PR #1, opened 2012-03-31
Some few fixes and enhancements

ptitjes:
I reintroduced ASTHelper class, fix the parent management (it was not propagating correctly and there was a lot of replicated code so I factored the parent setting code), and GenericProviderVisitor was not returning the return values correctly.

### DIFF
--- a/src/main/java/japa/parser/ASTHelper.java
+++ b/src/main/java/japa/parser/ASTHelper.java
@@ -1,0 +1,285 @@
+/*
+ * Copyright (C) 2008 Júlio Vilmar Gesser.
+ * 
+ * This file is part of Java 1.5 parser and Abstract Syntax Tree.
+ *
+ * Java 1.5 parser and Abstract Syntax Tree is free software: you can redistribute it and/or modify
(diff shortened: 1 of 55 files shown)


ptitjes:
Thank you :)

matozoid:
No, thank you :)

On Sun, Apr 1, 2012 at 1:46 AM, ptitjes
reply@reply.github.com
wrote:

ptitjes:
If you wish, I have a variant of DumpVisitor that respects blank lines from the input. It follows Node.getBeginLine, and getEndLine to determine if there were some blank lines between non-modified nodes that were in the original tree.

I also have some type-checking (expected and actual type) stuff, although it is not implemented for all kind of expressions yet. It could be optimized by introducing a "feature" attribute in addition to parentNode to know in which property and index (for lists) of its parent node it is located. (I would only have to add that to the setAsParentNodeOf() method).

Tell me what you think about it.

matozoid:
I've added you to the collaborator list. Add any features you like :-) I'll also add some issues to the issue tracker here.

ptitjes:
Great :)

PR #20, opened 2013-09-14
Java 7 support

dnadolny:
I've added the ability to parse Java 7 source, which includes: binary literals, underscores in numeric literals, the "diamond operator", try-with-resource blocks, and multi-catch clauses. The last one is the biggest change, because it involves adding a new method to the visitors.


matozoid:
Awesome! Thanks!

PR #26, opened 2013-10-24
Comments parser

ftomassetti:
Hi,
I am working on this patch to deal with all comments.

What I did was to remove from the grammar any attempt to "smartly" parse the comments.
They are instead parsed (optionally) in a separate step.

A separate hand-written parser parse the comments and associate some comments to the thing the comment and consider the other comments as "orphans".

For example:

class /\* An orphan comment */ A {
// An orphan comment

// A comment for a field;
int a;

/**
- A comment for a method
  _/
  void foo(){}
  } /_ Another orphan comment*/

Comments are dumped correctly and a few existing tests now gave the right output (some FIXME removed). I strived to add as many tests as possible for this part: I hope they can help both documenting and verifying the new functionalities.

I am available for any discussion and any correction other people could suggest.
Any advice to improve this patch is very welcome.


matozoid:
Merged it. Let's see what people think of it.

PR #34, opened 2014-05-22
Fix ArrayIndexOutOfBoundsException on DumpVisitor#printOrphanCommentsEnding

jahutton:
When there is a trailing // comment on a invocation line the everything list only has the one comment here.  do a index check before a get so the ArrayIndexOutOfBoundsException isn't through when it shouldn't.  This would happen if the everything list only contains comments, not sure how this comes to be.

Let me know if you want me to put together a unit test for this.


matozoid:
Hopefully someone can look into this later. Pulling it in.

PR #38, opened 2014-06-13
Correctly handle & visit the resources of a "try-with-resources" statement

sebastiankirsch:


matozoid:
Thanks!

arouel:
@sebastiankirsch Thanks for your contribution. It would be nice to have some tests too.

Issue #49, opened 2014-07-12
Document how comments attribution work


ftomassetti:
...and also how we would like it to work.

ftomassetti:
Initial documentation available here:
https://github.com/matozoid/javaparser/wiki/CommentsParsing

It would be nice to refer to the Javadoc documentation of the classes involved with comments handling. Does github help with producing Javadoc documentation?

ftomassetti:
I could add some tests to verify the parser works as written in the documentation.

Could someone take a look at the page and tell me if it makes sense?

SmiddyPence:
Made a couple of typo corrections.  Also added a couple more scenarios we have seen around "atypical" comment usage.  

Test would be very useful, especially given that a lot is open to interpretation unlike the base grammar.  As someone coming in cold its very hard to know if it is that way by intention or that particular case was just not considered.

ftomassetti:
Thank you!
Hey, the first atypical example is awful, we should fix that!

In the next days I will write those tests.

SmiddyPence:
I was thinking about it more and it is actually very difficult to do, because it is open to all kinds of interpretation.

``` java
// b is cool.
int b; 
```

That makes sense that it is attributed to the field b to both a human and parser.  However:

``` java
// Here be dragons
int b = 0;
int c = b;
// Here finisheth the dragons 
return c;
```

Here the comments have been use to describe a number of statements, which is obvious to a human; but not a parser.  

I wonder if it is trying to do too much? Perhaps a more broader brush could be easier to implement and still be useful to people? 

As before classes and methods could commentable, with a single comment by proximity.  Then other comments are just the "contained by" a method or a class and the api practitioner can use line number etc to work out what they are relevant for.

Or perhaps different modes different modes of parsing, one like now where is tries to go to the finest level of detail; where it is bias on developer interpretation - or the more broader approach is just a different implementation of the comment parser.

ftomassetti:
I agree there will be cases in which we cannot obtain the ideal results, we can just relay on good practices and conventions. In the case of "Here be dragons" I would expect a blank line between the comment and variable, we could use it to understand the comment is not directly related to the variable (we do not that at the moment).

I think we have different users interested in different things:
1) Many could simply not care about comments, that is why comments parsing should be optional or there should be the possibility to disable it (I do not remember what is the case, I will check)
2) Others could want could want to know which comments are present in a file, without knowing how they are attributed
[...]

Issue #45, opened 2014-07-03
Fix comment parsing


matozoid:
Make comment parsing work 100%, and write decent tests for it.

ftomassetti:
I think that we could close this issue when we merge #50 and close #49.

While more tests are always welcome we have now some documentation in place (and more to come as part of #49), with 35 tests related to comments out of 60 total tests.

SmiddyPence:
#50 and #49 are closed now. 

Others can create a new ticket for specific errors when found.

PR #50, opened 2014-07-12
Improvements to comments attribution

ftomassetti:
I create this pull request to discuss possible improvements to comments attributions. I added a couple of comments explaining why I expect certain results. If we agree that make sense I could document this logic.

### DIFF
--- a/src/main/java/japa/parser/JavaParser.java
+++ b/src/main/java/japa/parser/JavaParser.java
@@ -30,6 +30,7 @@ import japa.parser.ast.body.BodyDeclaration;
 import japa.parser.ast.comments.Comment;
 import japa.parser.ast.comments.CommentsCollection;
 import japa.parser.ast.comments.CommentsParser;
+import japa.parser.ast.comments.LineComment;
 import japa.parser.ast.expr.AnnotationExpr;
 import japa.parser.ast.expr.Expression;
 import japa.parser.ast.stmt.BlockStmt;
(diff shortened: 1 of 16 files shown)


ftomassetti:
Today I added some other commits to deal with issue #45 and issue #49

ftomassetti:
I will leave the merge request open until the 28 of July, then if I will not get comments I will merge it.

SmiddyPence:
Couldn't see any problems.

General gripe, not necessarily with this pull request is the untidiness of the tests; I think now we are adding more it is becoming more important.  

I think I prefer the convention of having the src used in tests in a separate file rather than inline Strings in the test code.  Especially for anything over a couple of line, potentially we can reuse some as well.  

The two files "TestCommentsParser" and "TestCommentsParsing" seem just to be two buckets with slightly different names. 

Some tests use Hamcrest matchers other don't.

What I assume is legacy test code, that uses main rather than JUnit.

Once this is merged I'm happy to take a pass over tidying it up. (Rather than just complaining about it!)

ftomassetti:
I agree with all you said. I think one set of tests was meant to check if the comments parser could get all the comments (with right content and position) while the other one should check that comments are attributed correctly. Probably now it is just a mess with no logical division... so it would be great if you can put some order, maybe when tests will be in order we will be encouraged to writ more of them.
Thanks!

ftomassetti:
Merged and opened issue #59 from the comment of @SmiddyPence

PR #65, opened 2014-08-08
JBehave Tests for Comment Parsing

SmiddyPence:
I made a demo video as a guide for how I tend to work with JBehave.  It doesn't go into JBehave in detail, more the process of writing tests and a few tips.

https://www.youtube.com/watch?v=IW5FsUGenNs

Also I wanted to practice my demoing.


ftomassetti:
Thank you very much, Nicholas!
Sorry if it took a while to get some comments.
IMHO the screencast is great and for me it is very useful that you describe IntelliJ integration (I am using it) but we should also consider casual contributors not using IntelliJ, so maybe we could write some basic instructions from them, possible based on a simple example.
I think you put in place all the infrastructure needed to address all the most common cases and so many tests can be written by copying & hacking a bit your tests.

Maybe we could have continuos integration using travis, to ensure all tests keep passing. I can take care of that (issue #67).

Also some tool for code coverage could be useful (issue #66).

Looking forward to have your work merged!

P.S. Do you mind if a tweet and refer this screencast?

talios:
Un-idling...

Loved the screencast - one thing tho, I also recommend looking at 
Thucydides, for reporting on top of JBehave - combines code coverage 
reporting per story, as well as timing.

http://www.wakaleo.com/blog/automated-acceptance-testing-with-jbehave-and-thucydides

My understanding of what Thucydides adds is an extra layer to the tests 
with a class annotated with @Steps, which give a finer granularity to 
the tests ( altho, thats more useful for acceptance testing, stepping 
thru a web page etc. and not so much how the parser works - maybe? ).

Mark

SmiddyPence:
I will definitely look into that.

It can be a real problem once you have a lot of this style of tests how you mange/maintain them, I have definitely noticed that.  I'm not sure it will be as much of a problem on a project of this size mind.  

It also ties in nicely with another thread about test reporting. 

Federico - sure it was made to share.

SmiddyPence:
Is there anymore feedback, shall I merge it?

ftomassetti:
I think the policy is "every pull request will be accepted". I am also as you feel a bit shy about accepting my own pull request (unless it is rather small), maybe we could ask @matozoid how we should do in this case.

From my point of view I have been looking forward to have your changes merged for a while already :D

matozoid:
If it doesn't break the build - go ahead :)

ftomassetti:
Seems a sensible policy :)

PR #75, opened 2014-12-12
Performance improvement

jmurciego:
Bubble sort algorithm over a double linked list has poor performance,
making the javaparser unusable in large projects.

This ugly refactor avoid the issue, reducing the complexity to O(n^2)


matozoid:
That's an ugly sort :-O Let's open an issue to fix this more thoroughly.

ftomassetti:
I think I could have been the culprit of that awful sort... thanks for deleting that shame!

Issue #74, opened 2014-11-28
Change package and maven group id


matozoid:
These two are left over from the project's life on Google Code.

Suggestions for improvements below:

matozoid:
Created a "javaparser" organization, maybe we should make it the official location of the project.

SmiddyPence:
Personally I think it is a good idea for an org in the long run.

Changing co-ordinates is not ideal, but I think makes more sense than maintaining the old ones now the project has moved hosting.

We can perhaps have a fuller gh-pages site with not just the JavaDoc, but information about the project past a present. As well as how to use etc.

sebastiankirsch:
+1

arouel:
+1

SmiddyPence:
So this branch: https://github.com/SmiddyPence/javaparser/tree/Milestone2

Has **Java 8**, **Maven Modules** and **Updated package name & Maven Co-ordinates**.

Do people want to review it there first or shall I create a Milestone2 branch here and push it over?

SmiddyPence:
Hi @matozoid

Can you transfer the repo to the organization?  

I can then update the `issueManagement` etc sections of the pom with the correct urls. Before the SonaType release.

matozoid:
Repository has been transfered!

SmiddyPence:
Cool I have raised the ticket.

Do you own the existing drone.io build?  Will need to move the repo url there as well... If you don't I can just add it to my drone account; no worries.

matozoid:
Better get that drone.io build off my account, so I can leave things to active users and committers like you! I'll delete mine. A good time to see if there's a nicer build site out there too.

sebastiankirsch:
I'd recommend [Travis CI](https://travis-ci.org).

ftomassetti:
+1 for Travis: I am using  it for different projects and I am happy with it (I have nothing against drone.io anyway)

SmiddyPence:
I left it as drone.io for the time being, just because it took about ~30 seconds.  If someone wants to set up Travis feel free, I didn't feel the need just to execute "mvn clean install".  Perhaps when #66 comes about the extra control will be useful.

In any case closing this now as the modules are in place.

Issue #71, opened 2014-09-24
Maven Modules


SmiddyPence:
I propose splitting the project into three modules:

**core** - Main module with all the libraries code
**grammar** - Contains grammar and generated sources
**acceptance-tests** - All the JBehave bases testing

Rational - Not plainly obvious what code is generated and that you have to run javacc:javacc initially from the command line.  The IDE can get pissy while referencing src files from the target directory.

Any other suggestions logical divisions, names there of...

SmiddyPence:
Ok so should have realised, that there would be a cyclic dependency between core and grammar.  So grammar is generating the src files directly to core atm.

Thoughts?

https://github.com/SmiddyPence/javaparser/tree/Milestone2

arouel:
It should basically work in same way as an ANTLR grammar generates sources before the compilation starts. So it should require only the right phase in the lifecycle to ensure everything works well.

The separation into **grammar** and **core** sounds not like a good idea to me. The grammar could be hidden for users and therefore I do not see any value in a separate module for it. It should be completely encapsulated and isolated to the outer world.

SmiddyPence:
The problem is not really a lifecycle issue.  It is that the generated code references non generated code and vice versa.  So ultimately you have to end up with the compilation happening in one module, where all the src resides.  Rather than having one module depend on the other.

I'm not clear on your second concern, you seem to imply hiding is bad; but then claim it should be encapsulated?  

From an end user point of view, there is no difference they add a jar and use it as normal, nothing changes.  The separating is mainly for maintenance to help people understand that some code is generated.  It will also make the generated sources first class citizens; making debugging and working with the IDE easier; rather than having to attach src files from the target directory.

arouel:
If they depend on each other it does not make sense to separate it into two JARs. In case we want to provide also OSGi meta data you get trouble because of that circular reference. I do not see any maintenance improvements by separating it. That an end user understand how it works together is questionable and can easily be solved with documentation on class-level. And that it brings easier working with in an IDE can be solved by packaging everything useful into the corresponding `*-sources.jar`.

SmiddyPence:
To a certain extent I agree while there are circular dependencies, it doesn't make sense to have only a grammar file in a module. However it does not always have to be this way, hopefully we can refactor this, as it is somewhat of a smells.  That is exactly my point to have a *-sources.jar", but it would be built from a separate module using vanilla maven rather than the assembly plugin or some such.

[...]

Issue #33, opened 2014-02-03
Make CuPrinter implement java.io.Closeable 


matozoid:
This allows Java 7 to auto close the resource. This would mean moving to java 7.

(imported from google code comments)

SmiddyPence:
Probably looking at putting that in a major release version.

i.e. v 2.0.0
With minimum Java 7 and support for Java 8 syntax.

P.S. Good work on the tidy up, was this the only relevant issue from google code? Pretty nice if that's all you're left with.

matozoid:
A new milestone was added :)

As far as I can see, there were about three issues that were solved (double check if you like,) the rest were complaints about lacking documentation and many samples that, in my opinion, didn't add much.

sebastiankirsch:
Hm, `java.io.Closeable` is around since 5, so we could implement it right away.

I'd assume that this enables auto-closing as well (even if javaparser is compiled with JDK < 7), but I don't know for sure. Any "Oracle Certified Associate, Java SE 7 Programmer" around to confirm this?

SmiddyPence:
That sounds correct, having swatted up. Only SCJP for 1.4 here :wink: 

Personally I was getting Closable and AutoClosable confused, had assumed there was only 1 for using the try-with-resource; which was only available >= 7.

Incidentally this CuPrinter has been removed anyway, (likely me)! Looks like it was just a cheeky little main method to dump a class out as a String.

Issue #27, opened 2013-10-31
Add Java 8 support


matozoid:

talios:
+1 on this - I just started updating my codebase to using JDK8 and I have a maven-enforcer-plugin extension that's using the javaparser ( +1 works AWESOME ) and would love to see JDK8 support soon.

shoito:
+1

ghost:
+1

matozoid:
Nobody is working on this, so pick it up if you like.

rpau:
Hi,

I have recently added Java 8 support to walkmod (check https://github.com/rpau/javalang). This open source project started from the Javaparser 1.6 grammar, so you feel free to add my grammar changes to Javaparser. 

I have tested my Java 8 parser with expressions extracted from the JSR 335. So, it would be nice to test the parser with your Java 8 code before adding my changes to this project and publish an official version to Maven. My parser also passes your parsing (Java 7) tests.

talios:
+1 will give it a shot on my JDK8 code and report back what I find...

rpau:
Hi,
Today I have read Java 8 allows to annotate create expressions, cast, implements and exceptions.
Check (http://docs.oracle.com/javase/tutorial/java/annotations/basics.html). I need to add the syntax into my Java8 grammar. I will notify you when the annotations are added (probably this weekend).

ptitjes:
@rpau In fact, this is any TypeReference that have to hold the annotations. (And not the expressions that contain TypeReferences.)

I'm quite sure the Checker Framework team (washington university) have already done that work on their fork of JavaParser for their initial experiments (almost one year and a half ago). Maybe you could get some stuff from them.

rpau:
Thanks @ptitjes for your help. Java 8 Annotations for any TypeReference added in my repo ( https://github.com/rpau/javalang). So, any comment is always welcome :)

ptitjes:
Well done! Hope this will go nicely into JavaParser!

ftomassetti:
Any news? We are looking forward to have Java 8 support :D

martoeng:
I've seen many Java 8 constructs in the code. Any information on when there will be a release that supports the Java 8 changes?

ftomassetti:
I suggest we start a separate branch, in that branch I would insert as many examples as possible of code written in Java 8. Do you think it is a good plan?
Does someone knows where to find good examples of code written in Java 8? Perhaps a common set of files used for testing some compiler or tool...

[...]

PR #78, opened 2015-01-02
Milestone2

SmiddyPence:
Pull request for:
#27 Add Java 8 Support
#71 Maven Modules
#74 Change package and group id

Due to the package name changes, this commit has **breaking** changes so I have up version-ed to 2.0.0.  I have kept the parser code to be JDK 6 compliant though, not refactoring for diamonds etc.

We still have **#33 Make CuPrinter implement java.io.Closeable** in the backlog.  Do we feel like this is a good time to make Java7 the minimum version and tackle this issue also?  Or keep the project using a wider range of JDK versions?

I will look to merge this is a week (09/01/15), please let me know feedback.  

Afterwards I'll do the Sonatype dance for new co-ordinates ;-)


sebastiankirsch:
+8
Thank you very much!

ftomassetti:
Should we update the Readme and all the other files to state that Javaparser is now a Java 1.8 parser?

SmiddyPence:
We will need to:

Move repo to the organization
Update urls in pom for sonatype release
Once on Sonatype at new co-ordinates update readme docs etc.

Issue #54, opened 2014-07-20
Lazy initialized lists should be used everywhere


ftomassetti:
We did that for comments in issue #3 but it applies everywhere.

I could do that in this way, if you agree:

``` java
    public List<Parameter> getParameters() {
        if (null == parameters) {
            parameters = new ArrayList<Parameter>();
        }
        return parameters;
    }
```

Thoughts?

sebastiankirsch:
Why not adapt all the setters/constructors and set either the specified non-null value or `Collections.emptyList()`? This simplifies the getter code and is probably faster & more memory-efficient then anything else.

The only downside is that the lists cannot be modified, but, alas, one needs to set your own list in such cases.

ftomassetti:
It makes sense: after all you are not supposed to edit the list (as you said).

I can do this change over the week-end if everyone agrees.

ftomassetti:
Just to be sure, before doing the same change on a ton of classes, would that work for you?

``` java
private List<AnnotationExpr> annotations = Collections.emptyList();

public List<AnnotationExpr> getAnnotations() {
    return annotations;
}

public void setAnnotations(List<AnnotationExpr> annotations) {
    if (null == annotations) {
        throw new NullPointerException();
    }
    this.annotations = annotations;
    setAsParentNodeOf(this.annotations);
}
```

sebastiankirsch:
Looking at the ASTParser line 3353 I'd assume that you cannot throw a NPE,
but should fall back to the empty list as well. Which is much more
user-friendly as well. Apart from that, looks great!

Federico Tomassetti notifications@github.com schrieb am Wed Feb 18 2015
at 4:40:21 PM:

ftomassetti:
Ok, let's do that for the sake of compatibility, maybe I would add a note to specify that in the future that could be changed

ptitjes:
It seems to me this deals with the same problems that PR #106 tries to fix. Do you @javaparser/owners think that we must take a definitive decision about this for 2.1 ?

ftomassetti:
Given this is swamped for the moment I will prepare another pull requests with the bug fixes present in this pull request. They should be less controversial

ftomassetti:
I think we have done this in the end, closing

JinyongPark:
without Maven, how i use your source? i use eclipse with Dynamic Web Project

ftomassetti:
please @JinyongPark do not ask questions unrelated this issue here. If you have questions open a new issue. Thank you!

PR #79, opened 2015-01-09
Fixing NullPointerException of getDeclarationAsString

chundongwang:
I was using getDeclarationAsString with my java library and hit this NullPointerException due to some methods not having parameters or throws. This change fixed the issue and my tool is happy now so hope it'd be helpful to others as well.

Thanks


ftomassetti:
To me it seems fine, but I would prefer to fix it by initializing the lists (throws, parameters, etc) using empty lists instead of null. I think the original author did that for performance reasons but I am not sure it is worthy... and we had already a few of these errors.

SmiddyPence:
I agree this part of larger problem, and related to #54.  The suggestion in #54 is ok, but it only caters when access lists as properties rather than fields.  In this case putting similar code around all field access is pretty grim.

Not sure what the lesser of two evils is here.

chundongwang:
To me I would rather keep the existing behavior since probably not a lot functions would have both parameters and throw expressions. Rather than this, having two identical/similar code in two different places bites me more.

ftomassetti:
Maybe we could do the lazy initialization thing in getParameters and getThrows and use them in foreach loop so we fix this problem with a smaller change AND we avoid returning null in getParameters and getThrows.

It is potentially true that users could expect getParameters and getThrows to return null, but imho this is bad and should be fixed eventually

chundongwang:
I tried to make it lazy initialization and found that DumpVisitor is checking null of getParameters and getThrows. And it'll put a "throw" there even if the method has no throw expression and caused several test failures.

chundongwang:
I fixed the DumpVisitor and looked up the whole source tree and that seems to be it. However I agree that it might break some existing projects out there. I personally would like to do the right thing here but it's up to you.

ftomassetti:
Thank you for your work, to me it seems great! @SmiddyPence do you think this pull request could be merged?

SmiddyPence:
Definitely looks cleaner to me, +1

ftomassetti:
then I think you should accept it, I do not think I have writing access on this project

matozoid:
Better give you access then.

ftomassetti:
Thank you, I immediately used this new acquired power to merge the branch :D
Thank you @chundongwang for your patch!

chundongwang:
My pleasure!

PR #86, opened 2015-01-16
Add missing functions to ModifierVisitorAdapter

chundongwang:
After moving to new version, my class hit a compiler issue complaining missing function. And it's due to three newly introduced methods of GenericVisitor. GenericVisitorAdapter has them but ModifierVisitorAdapter doesn't


SmiddyPence:
Nice one, good job.

PR #91, opened 2015-01-16
Fix #88

### LINKED ISSUE #88: com.github.javaparser.ast.expr.MethodReferenceExpr and com.github.javaparser.ast.expr.TypeExpr do not set themselves as parent of their child nodes
No surprise after #82, but I didn't catch it immediately. Will provide a fix.

sebastiankirsch:
Fixes #88 

Besides the comments, this seems to be all. If the comments should be reworked as well, I'll add it later on

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/LambdaExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/LambdaExpr.java
@@ -45,8 +45,8 @@ public class LambdaExpr extends Expression {
 	}
 
 	public void setParameters(List<Parameter> parameters) {
-		setAsParentNodeOf(parameters);
 		this.parameters = parameters;
+		setAsParentNodeOf(this.parameters);
 	}
 
 	public Statement getBody() {
@@ -54,8 +54,8 @@ public class LambdaExpr extends Expression {
 	}
 
 	public void setBody(Statement body) {
-		setAsParentNodeOf(body);
 		this.body = body;
+		setAsParentNodeOf(this.body);
 	}
 
 	@Override
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
@@ -50,14 +50,14 @@ public final class ImportDeclaration extends Node {
     }
 
     public ImportDeclaration(NameExpr name, boolean isStatic, boolean isAsterisk) {
-        this.name = name;
+        setName(name);
         this.static_ = isStatic;
         this.asterisk = isAsterisk;
     }
 
     public ImportDeclaration(int beginLine, int beginColumn, int endLine, int endColumn, NameExpr name, boolean isStatic, boolean isAsterisk) {
         super(beginLine, beginColumn, endLine, endColumn);
-        this.name = name;
+        setName(name);
         this.static_ = isStatic;
         this.asterisk = isAsterisk;
     }
@@ -119,6 +119,7 @@ public final class ImportDeclaration extends Node {
      */
     public void setName(NameExpr name) {
         this.name = name;
+        setAsParentNodeOf(this.name);
     }
 
     /**
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
@@ -50,16 +50,16 @@ public final class ImportDeclaration extends Node {
     }
 
     public ImportDeclaration(NameExpr name, boolean isStatic, boolean isAsterisk) {
+        setAsterisk(isAsterisk);
         setName(name);
-        this.static_ = isStatic;
-        this.asterisk = isAsterisk;
+        setStatic(isStatic);
     }
 
     public ImportDeclaration(int beginLine, int beginColumn, int endLine, int endColumn, NameExpr name, boolean isStatic, boolean isAsterisk) {
         super(beginLine, beginColumn, endLine, endColumn);
+        setAsterisk(isAsterisk);
         setName(name);
-        this.static_ = isStatic;
-        this.asterisk = isAsterisk;
+        setStatic(isStatic);
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/LambdaExpr.java
```diff
@@ -45,8 +45,8 @@ public LambdaExpr(int beginLine, int beginColumn, int endLine,
 	}
 
 	public void setParameters(List<Parameter> parameters) {
-		setAsParentNodeOf(parameters);
 		this.parameters = parameters;
```
ftomassetti, line 48: What happens if parameters is null?

ftomassetti:
+1 on the code (I have not read the tests)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/LambdaExpr.java
sebastiankirsch, line 48: This is implemented since long in the `Node` class and is null-safe ;)

SmiddyPence:
Would suggest moving the ParentVerifier visitor out into a separate class and put it in the visitor package alongside the existing `PositionTestVisitor`.  I would rename one or the other to make them consistent as well.

sebastiankirsch:
@SmiddyPence the reasoning behind having the visitor as a nested class was that no one outside of `ParsingSteps` will need access to `ParentVerifier` - in fact, if it wasn't for the gazillion methods that need to be implemented, I'd have defined an anonymous class.

SmiddyPence:
Just doesn't feel too pragmatic to me.  

Although the book definition of when to use a nested class. I think it will become unpleasant fast.  If we add just more visitor we will end up over a thousand lines for the class.  If we add 4 test orientated visitors well be over 2000 LoC in one file.

sebastiankirsch:
Would feel pragmatic to me if we split up once that critical size is reached. However, as one could argue that that size is reached already, I don't mind extracting it now.  
But I'm against moving it further away from the class by moving it into another package, let alone the visitor package - this class is a visitor by chance; its primary purpose is to verify that each of the nodes has a parent set.

Writing that I realize that we already have a rather "tech-focused" package naming schema in place. I'd suggest splitting up the `steps` package into `parsing`, `manipulation`, `comments`, etc. - keeping the `visitor` package as it covers parts of javaparser's service offering. Then the extracted `ParentVerifier` would just be located in the `parsing` section, as it is required only there.

SmiddyPence:
Sure, makes sense.

I'm principally against the inconsistency of having one test orientated visitor as a separate class and one nested.  Seeing as they are so large it just seems infeasible to treat them all as nested.

sebastiankirsch:
Yeah, I see that. It's just that the existing visitor is tested code, whereas the nested one is "merely" used to verify the output of the tested code. Should I make the adaptations in the context of this pull request or afterwards?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/LambdaExpr.java
ptitjes, line 48: My doing (a long time ago) as a matter of fact... :P

ptitjes:
@SmiddyPence @sebastiankirsch That seems to me both your visitor could be made in one: a catch-all visitor. I've got a generic one if you want.

It has the following shape:

```
public abstract class CatchAllVisitorAdapter<R, A> extends
        GenericVisitorAdapter<R, A> {

    protected R preVisit(final Node n, final A arg) {
        return null;
    }

    protected R postVisit(final Node n, final A arg) {
        return null;
    }

    @Override
    public R visit(CompilationUnit n, A arg) {
        R result = preVisit(n, arg);
        if (result != null) {
            return result;
        }
        result = super.visit(n, arg);
        if (result != null) {
            return result;
        }
        result = postVisit(n, arg);
        if (result != null) {
            return result;
        }
        return null;
    }

    // ...
}
```

With that, both your visitors could (and should imho) elegantly be inner classes.

ptitjes:
Appart from the visitor, I personally thinks that this should go to tree right away. (else someone will want to fix those :P)

sebastiankirsch:
@ptitjes yes, that thougt crossed my mind, too. As both visitors are of the void flavor, it's even easier than that.
But me as well, I wanted to get this through and do that afterwards...

ptitjes:
Guys, could we please go forward with this work and merge it ??
The factoring out of visitor will be done later as suggested by @sebastiankirsch.

ptitjes:
@sebastiankirsch Sorry but while rebasing, I've just seen you missed TypeParameter.
Here: https://github.com/ptitjes/javaparser/commit/d5a99541a3bd6a7f020e1282868ddd0bb4d1fef9#diff-d222855024502fe49b9d51b6dbbc94a5L71
And here: https://github.com/ptitjes/javaparser/commit/d5a99541a3bd6a7f020e1282868ddd0bb4d1fef9#diff-d222855024502fe49b9d51b6dbbc94a5L127

sebastiankirsch:
@ptitjes I'm sorry, I can't put together what I've been missing. The changes you are referring to are all on your fork, aren't they? Because it seems that neither of those changes have been merged into master.

ptitjes:
Hi @sebastiankirsch! Good to see you. Yeah was just mentioning that you missed `TypeParameter` in the clean to use a setter and the `setAsParentNodeOf()` call.

sebastiankirsch:
@ptitjes you mean in the `setAnnotations` method?

So, i did not "un-merge" some of your changes?

ptitjes:
Now that is me that don't understand :p

ptitjes:
@sebastiankirsch I meant:

Missing `setAnnotations()` call here: https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/TypeParameter.java#L73

Missing `setAsParentNodeOf()` call here: https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/TypeParameter.java#L132

sebastiankirsch:
Ah, I see. I just didn't catch that, true.

Opened up #269 for this...

PR #94, opened 2015-01-24
Issue76

ftomassetti:
Removing bubble sort and using Collections.sort (~mergesort)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
@@ -37,7 +37,11 @@ public class PositionUtils {
         return compare(a, b, ignoringAnnotations) <= 0;
     }
 
-    private static int signOf(int value) {
+    /**
+     * Signum function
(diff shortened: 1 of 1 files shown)


SmiddyPence:
In that case I would say go with `signum`.

ftomassetti:
Renaming signOf as signum and using @SmiddyPence implementation. Merging.

sebastiankirsch:
I know I'm late to the party, but you were so close, @ftomassetti: http://docs.oracle.com/javase/6/docs/api/java/lang/Integer.html#signum(int)
I will replace this quickly...

here we go: https://github.com/javaparser/javaparser/commit/4079fa3e9414232ed8a05ef04e045565ad850a67

ftomassetti:
never late for improving things, thanks I learned a new one. Sorry for missing it in the first place

sebastiankirsch:
No need to excuse ;)

ptitjes:
@ftomassetti @SmiddyPence @sebastiankirsch Textual-search of 'sort' in the project reveals no other sorting methods, so I believe there are no other hidden manual sorts.

I add this to 2.1.

PR #96, opened 2015-02-11
fixed encoding issue

mschieder:
fixes an encoding issue when the encoding parameter is not the same as the platform's default charset.


sebastiankirsch:
Hi @mschieder , thanks for the patch!
I'd suggest to get rid of the local variable in the `else` branch (as done in the `if` branch) before merging.
I know you just kept the original code, but this is more consistent than having "two ways" to compute the input stream.

PR #101, opened 2015-02-18
Adding the NamedNode interface

ftomassetti:
While this has not immediate benefits for our code, it could be useful for users, which could manipulate with the same method all nodes implementing this interface.


sebastiankirsch:
Missing
- TypeParameter
- MemberValuePair
- NameExpr (obviously :)
- ClassOrInterfaceType

I'm honestly a bit concerned that some nodes choose to store the name as a `String`, others as a `NameExpr`. I feel like using a `NameExpr` is "wrong" in many cases: this would not prohibit using a `QualifiedNameExpr`, which is not correct for most, if not all of all of the nodes.

ftomassetti:
Thanks, I added the missing classes. About the cases in which NameExpr is used... I am not sure what the best solution is, but I am open to suggestions :)

sebastiankirsch:
So, since we use gitter, let's discuss here:
https://gitter.im/javaparser/javaparser/issue-101

Please help remember to document the outcome of the discussion in the
issue, though.

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
```diff
@@ -29,7 +30,7 @@
 /**
  * @author Julio Vilmar Gesser
  */
-public final class ClassOrInterfaceType extends Type {
```
sebastiankirsch, line 31: that is not the name of the node, that is a "reference by name"... I think this shouldn't be a `NamedNode`

ftomassetti, line 31: Fixed

ftomassetti:
Is there anything else that we should fix for this patch?

matozoid:
Should this be in 2.1 or 3.0?

ftomassetti:
I would guess 2.1, it is adding an interface but it should not be a disruptive API change

matozoid:
I'll leave it at that then.

PR #102, opened 2015-02-18
Adding NodeWithModifiers interface

ftomassetti:


javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
```diff
@@ -65,34 +67,76 @@ public static AccessSpecifier getAccessSpecifier(int modifiers) {
...
+    @Deprecated
     public static boolean hasModifier(int modifiers, int modifier) {
         return (modifiers & modifier) != 0;
     }
 
+    public static boolean hasModifier(NodeWithModifiers node, int modifier) {
+        return (node.getModifiers() & modifier) != 0;
+    }
+
+    @Deprecated
```
sebastiankirsch, line 103: Can you please add a javadoc like

```
/** @deprecated use {@link #hasModifier(...)} instead */
```

everywhere?

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
```diff
@@ -53,6 +54,7 @@
 
     public static final int STRICTFP = Modifier.STRICT;
 
+    @Deprecated
```
sebastiankirsch, line 77: not sure if I want to suggest that or not - but according to DRY, this should be delegated to the new (or an extracted) method

Usually I would want that, but it is very unlikely that this code will ever be touched again...

sebastiankirsch:
Looking at the `ModifierSet` for the first time now, I wonder if it makes sense to introduce a class representing the modifiers, which can then be used as a field instead of the `int` value. This class would basically provide all the functionality of `ModifierSet`.

I'm usually a big fan of such encapsulated functionality - however I'm not sure if, for resource consumption's sake (similar to #54), this would be appreciated...

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
ftomassetti, line 103: Sure, I will

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
ftomassetti, line 77: mmm, I did not want the new methods to call the old methods because we are going to get rid of the old methods in the future. Doing the opposite would permit to be DRY but the problem is that the old method take an `int` while the new methods take a `NodeWithModifiers`, which I cannot build from an `int`. I mean, I could create an anonymous class to wrap the int and implement NodeWithModifiers but it seems overkilling to me

ftomassetti:
I will agree that `ModifierSet` is much better that using an `int`, however the nodes having modifiers, let's say MethodDeclaration to have the method `isPublic` should delegate the call to their fiels, something like:

``` java
class MethodDeclaration {
  private ModifiersSet modifiers;

  public boolean isPublic(){
    return modifiers.isPublic();
  }
}
```

Which has to be done for all the classes having modifiers and for each one has to be done for a few methods (isPublic, isPrivate, isStatic, isFinal, etc.).

As an alternative the classes could return an immutable instance of ModifiersSet on which the caller could call the methods isPublic, etc:

```
  bool isMyMethodPublic = myMethod.getModifiers().isPublic();
```

I think the former solution is more correct and elegant, but it requires a lot of repetitions (which could be removed when we will move to Java 8, using the default methods of interfaces).

How does it sound?

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
sebastiankirsch, line 77: I was unsure myself, just wanted to bring it up.
I'm fine with keeping that (simple) duplication.

sebastiankirsch:
Yeah, replacing the `int` with an immutable version of `ModifierSet` and exposing that was what I had in mind.

ftomassetti:
changes as suggested by @sebastiankirsch

sebastiankirsch:
Hm, I really think it is better to come up with an improved version of `ModifierSet` (or a new class) and use that instead of the `Set`. Sure, it's better than the plain `int`, but now I still have to do some "magic" in order to find out if it is `public static final`, namely call `Set.contains` several times. I'm completely fine having this technical solution in "the inside", but imho the better API would be to expose the `isPublic()` etc. methods explicitly...

ftomassetti:
Done as suggested by @sebastiankirsch. Now you can do:

```
node.getModifiersSet().isPublic()
```

ftomassetti:
As suggested, put ModifierSet back and renamed the new class Modifiers.
In addition to that Modifiers is made immutable.

javaparser-core/src/main/java/com/github/javaparser/ast/Modifiers.java
```diff
@@ -0,0 +1,131 @@
+package com.github.javaparser.ast;
+
+import com.github.javaparser.ast.AccessSpecifier;
+import com.github.javaparser.ast.NodeWithModifiers;
+
+import javax.lang.model.element.Modifier;
```
ptitjes: I disagree about that! I'm really not sure that we want to introduce a dependency on jx.l.e.
In fact, there is already a ref on j.l.reflect that is crazy and that we should remove it.

That may have seemed a good idea to the original author at that time but see:
https://github.com/ptitjes/javaparser/commit/cc6378936b0a072b80b1d9d420b66b8337cbc033#diff-af73e9494c4ad27e96149400188c9ea3R56

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
```diff
@@ -30,6 +33,21 @@
 public final class ModifierSet {
 
     /* Definitions of the bits in the modifiers field.  */
+    
+    private static Map<javax.lang.model.element.Modifier, Integer> ENUM_TO_INT = new HashMap<javax.lang.model.element.Modifier, Integer>();
```
ptitjes: Please no ref to jx.l.m in this. How do we support j8's 'default' if we build on j6 ?

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
```diff
@@ -19,8 +19,11 @@
 package com.github.javaparser.ast.body;
 
 import com.github.javaparser.ast.AccessSpecifier;
+import com.github.javaparser.ast.Modifiers;
+import com.github.javaparser.ast.NodeWithModifiers;
 
 import java.lang.reflect.Modifier;
```
ptitjes, line 28: We have to remove that ref to java.lang.reflect. We can't support j8 with reflection constants not up to date!

Here is what I propose:

```
public enum Modifier {
  ABSTRACT, FINAL, NATIVE, PRIVATE, /* ... */;

  // ...
  // Helpers tagada...
}

public interface NodeWithModifiers {
  int getModifiers();
  EnumSet<Modifier> getModifiersSet();
}

public class ModifierSet {

  // For compatibility
  public static final Modifier PUBLIC = Modifier.PUBLIC;

  // ...
  // Helpers tagada...
}
```

Else let's keep integers!

ftomassetti:
@sebastiankirsch I would prefer to avoid using "add" because the given modifier is not always added, it is added only if it is not already present. For this reason I think set/unset are more clear in this case

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
ftomassetti: Reflect 8 seems not do contain a reference to default among the modifiers, even in version 8
http://docs.oracle.com/javase/8/docs/api/java/lang/reflect/Modifier.html

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
ftomassetti, line 28: good point

ftomassetti, line 28: the problem is that for compatibility I have to specify and accept also integers, but I cannot find the value for the Default Modifier (it is not present in j.l.r.Modifier and looking on the JVM Specs I did not find it so far)

ftomassetti, line 28: "For the default keyword, it is even simpler: there is no modifier bit associated with the keyword. If a public, non-abstract, non-static method appears in an interface, it must be a default method. That’s how Method.isDefault() determines whether a method is a default method. Modifier.toString(…) has no chance of knowing whether the declaring class is an interface and therefore will never print default."

So the old interface based on ints cannot handle the default modifier.

ftomassetti, line 28: and in the current nodes is not treated as a modifier, but as a separate field of the method (see setDefault/isDefault).

It seems therefore that this solution works. Do we agree or do we still see problems with it?

ptitjes, line 28: @ftomassetti j.l.r is another view (reflection at runtime) from what we have to take (source at compile-time).
Indeed, if you look at the Java Language Specification (JLS8 Chapter 9.4 p. 286), you can see that 'default' is considered a modifier:

> InterfaceMethodModifier: one of
>    _Annotation_ **public** **abstract** **default** **static** **strictfp**

Thus I think we have to have our own constants. For the 'default', in my fix https://github.com/ptitjes/javaparser/commit/cc6378936b0a072b80b1d9d420b66b8337cbc033#diff-af73e9494c4ad27e96149400188c9ea3R56, I had to add a new constant not present in j.l.r.Modifier.

I understand that we need to keep the integer values for backward compatibility. And this is something enums are good for! So I restate what I had in mind, because I might not have been clear enough:

``` java
public interface NodeWithModifiers {

    @Deprecated
    int getModifiers();

    @Deprecated
    void setModifiers(int modifiers);

    EnumSet<Modifier> getModifiersSet();
}

public enum Modifier {

    // Exact values taken from j.l.r.Modifier
    // We can remove the integer field completely after deprecation
    PUBLIC(0x00000001, "public"),
    PRIVATE(0x00000002, "private"),
    PROTECTED(0x00000004, "protected"),
    STATIC(0x00000008, "static"),

    // ...

    STRICTFP(0x00000800, "strictfp"),
    // Note that the last comma after the last constant IS valid java
    // This way of defining enums is diff-proof when you add new values
    ;

    private final int integerValue;
    private final String label;

    Modifier(int integerValue, String label) {
        this.integerValue = integerValue;
        this.label = label;
    }

    @Deprecated
    public int integerValue() {
        return integerValue;
    }

    public String label() {
        return label;
    }

    // We can still add helpers
    public static boolean isPublic(NodeWithModifiers node) {
        return node.getModifiersSet().contains(PUBLIC);
    }

    public static void setPublic(NodeWithModifiers node) {
        node.getModifiersSet().add(PUBLIC);
    }

    // ...
}

@Deprecated
public /**/ class ModifierSet {

    // For compatibility
    @Deprecated
    public static final int PUBLIC = Modifier.PUBLIC.integerValue();

    @Deprecated
    public static final int PRIVATE = Modifier.PRIVATE.integerValue();

    // ...

    @Deprecated
    public static int addModifier(int modifiers, int mod) {
        return modifiers | mod;
    }

    @Deprecated
    public static int removeModifier(int modifiers, int mod) {
        return modifiers & ~mod;
    }

    @Deprecated
    public static boolean hasModifier(int modifiers, int modifier) {
        return (modifiers & modifier) != 0;
    }

    @Deprecated
    public static boolean isPublic(int modifiers) {
        return (modifiers & PUBLIC) != 0;
    }

    // ...
}
```

Note how the definition with the enum Modifier is way more compact, elegant and readable than the ModifierSet artillery of methods, and that for equivalent functionality!

Let's compare the now deprecated way and the new way:

``` java
NodeWithModifiers method = /* ... */;

// Querying modifiers
{
    // Deprecated way
    if (ModifierSet.isPublic(method.getModifiers())) { /* ... */ }
}
{
    // New way
    if (method.getModifiersSet().contains(Modifier.PUBLIC)) { /* ... */ }
    // Note nice completion of ^^^ in IDEs
    // Whereas completion is useless with integers
}

// Updating modifiers
{
    // Deprecated way
    int modifiers = method.getModifiers();
    int modifiers2 = ModifierSet.removeModifier(modifiers, ModifierSet.PRIVATE);
    int modifiers3 = ModifierSet.addModifier(modifiers2, ModifierSet.PUBLIC);
    method.setModifiers(modifiers3);
}
{
    // New way
    EnumSet<Modifier> modifiers = method.getModifiersSet();
    modifiers.add(Modifier.PRIVATE);
    modifiers.add(Modifier.PUBLIC);
    // Note nice completion of ^^^ in IDEs
    // Whereas completion is useless with integers
}
```

IMHO it is a complete win-win to go that way while still supporting the old way until end of deprecation time.

ftomassetti, line 28: In the current nodes it is not treated as a modifier, it is treated differently. In a method you treat it differently (you use the specific methods setDefault/getDefault). I think it makes sense to treat it like this. While in the code you use a keyword to indicate them, in practice it is not a modifier because a method with a body in an interface is a default method and a method without a body is not a default method.

sebastiankirsch:
Concerning the `default` modifier: PMD also treats it as a modifier and not explicitly. Just saying.

sebastiankirsch:
@ftomassetti I'm not completely convinced: [Set.add(E element)](http://docs.oracle.com/javase/6/docs/api/java/util/Set.html#add%28E%29) & [List.set(int i, E element)](http://docs.oracle.com/javase/6/docs/api/java/util/List.html#set%28int, E%29)

And the modifiers are some kind of list/set.

sebastiankirsch:
@ptitjes 
I like the `Modifier` enum as well. I just think it would even be better to define a `Modifiers` as @ftomassetti did, wrapping the `EnumSet`. In line with your old/new comparison, this would look like this:

```
// querying
if (method.getModifiers2().isPublic()) { /* ... */ }

// updating is a bit more cumbersome (at least if we strive for immutability)
method.setModifiers2(method.getModifiers2().add(Modifier.PUBLIC));
// BUT: it will even allows us to throw an exception if someone does this
method.setModifiers2(method.getModifiers2().add(Modifier.PRIVATE));
// because we cannot be private & public in the same instance...
```

javaparser-core/src/main/java/com/github/javaparser/ast/body/ModifierSet.java
ptitjes, line 28: Then I guess 'abstract' is not a modifier too... :P
I mean they could as well specify abstract methods on an abstract class as just not having a body but they chose to add a modifier keyword. They did the same with default methods on interfaces: they added a modifier keyword.
Well that is the purpose of a modifier.

Our role in parsing java sources is to model the syntax. Following your point of view, we could as well drop the isDefault boolean in MethodDeclaration and let the user deal with testing if there is a body or not. But then why not do the same for abstract methods in abstract classes.
Nah 'default' is definitely a modifier and does not deserve its own boolean flag.

I even think that InitializerDeclaration should use modifiers even if only one (namely 'static') is possible.

ptitjes:
@sebastiankirsch That would mean replacing an integer by two object instances (an EnumSet inside a Modifiers object). I don't think it is a good strategy for something like an AST.

And by the way if we need object-like accessors (not relying on helper classes) then I think the accessors should be on the Nodes with modifiers. So that we can do:

```
method.isStatic()
```

Because, when you think about it, the following sentence has not much any sense:

```
method.getModifiers().isStatic()
```

meaning what ? That the modifiers is static ? That should be the verb 'contains' right ? A set of modifiers **is not** _static_ or _public_. But it can **contain** the _static_ or the _public_ modifier.

Then I would agree to have the individual accessors on nodes directly. As for an implementation strategy, knowing that we don't have default methods, I don't mind having methods individually implemented on nodes, as that would permit to make evidence of what modifier has meaning on what node. (That is, for instance, the ConstructorDeclaration node would not feature isStatic() or isAbstract() methods)

ptitjes:
@sebastiankirsch About whether or not an EnumSet, I'd say we can have a Modifiers object but then put an (private) int inside. But IMHO that would be reprogramming EnumSet... but with specialized methods containsStatic(), containsPublic(), ... And that would also **mean to not take profit of switch { case ... }** and the fact that they are **compiler-checked for their exhaustivity**. That would be really a miss.

ftomassetti:
It would be nice to have the methods `isStatic`, `isPublic` and so on the node, but without the default implementation of method interface it means duplicating a lot of code. The alternative is the `getModifiers().isPublic`. Personally I do no think it is confusing. If we think it is we could later (i.e., in a next pull request) rename those methods from `isPublic` to `public`, etc.

I think that through good suggestions and nice discussions we have improved this patch a lot, let's reach an acceptable solution on the main points and commit it, we can always improve some details later :)

ptitjes:
@ftomassetti Ok but what about switch ??
Having accessors only is a pullback compared to int. You are forced to use if statements where you could use an efficient switch, which I guess I'm not the only one to use a lot. (Enums do permit that too!)
As for duplicating code, this is for some years until j8 becomes the common ground as j6 is for us now. If methods in nodes are just

``` java
public boolean isStatic() { return getModifiers().contains(Modifier.STATIC); }
```

we can deal with it. Ok it is duplicated but this is not like it is some complex processing duplicated that would lead to bugs not fixed in some part of the code.

ftomassetti:
I do not see why someone would like to do a switch. We are not talking about an homogeneous collections, modifiers indicate very different things one from another, therefore using an enum is kind of stretched

ptitjes:
I don't agree. In a refactoring that acts on methods, one could handle multiple **cases** depending on the modifiers. That is common ground. In fact, I even believe the only suited thing for that are integers.

``` java
switch (modifiers) {
case PUBLIC & STATIC:
  //...
  break;
case PUBLIC:
  // ...
  break;
}
```

In fact, you just made me realize that the only way to do that is integers.
I change my vote. Keep integers. Introduce how many helpers you want but please keep integers and don't deprecate them.

ftomassetti:
I think a refactoring would do different things with static or final. Private, public and protected could be somehow similar (and we have the AccessLevel enum or something similar) but this is not the case for all the modifiers. The only thing they have in common is that they are stored in the same byte in the bytecode, but this is not something that should drive our design imho.

So said, I am willing to discuss and I am sure my solution is not perfect (I have happily revised many times, according to the suggestion from @sebastiankirsch), but it seems that in any discussion is very difficult for you to make concessions :D

sebastiankirsch:
@ptitjes Hm, those resource constraints :(
Although, I'm not sure if that is something we have to consider - look at the fields of e.g. a `MethodDeclaration`:

```
private int modifiers;
private List<TypeParameter> typeParameters;
private Type type;
private NameExpr name;
private List<Parameter> parameters;
private int arrayCount;
private List<NameExpr> throws_;
private BlockStmt body;
private boolean isDefault = false;
```

and this is without the inherited members. Adding an object encapsulating an `int` would add one pointer per `MethodDeclaration` instance. Doesn't seem to be that much. On the other hand, an `EnumSet` would add those fields:

```
final Class<E> elementType;
final Enum<?>[] universe;
private long elements = 0L;
```

So if resources are indeed an issue, we should not use an `EnumSet`. Or we do it on the fly when `getModifiers()` is called - thus sacrificing speed.
I have to admit that I have a very limited view on using the API and never spent an idea on using javaparser to do refactoring; thus it's hard to argue for or against the `switch` scenario. My gut feeling is with Federico in this; it's hard to imagine that I need to cover every single value of a modifier in a switch statement (except for dumping/printing)...

Yes, the wording I chose was tailored towards the `Node`s. Seeing all the arguments I wonder if it would be that cumbersome to implement the `isPublic()` etc. methods directly at the nodes - this would allow us to offer only the "possible" modifiers (and make the interface useless - which may be an indication that it isn't required anyway, because who would be looking for all kinds of nodes with modifiers instead of specific types? But again, I only have my few use cases). And if I counted correctly, we have only 7 implementations.

@ftomassetti I understand you want to close this, but I find this discussion very productive - I don't feel that we are running in circles.
Also, I'm very hesitant to introduce a new API just to tear it down with the next pull request ;)

ptitjes:
Enums are now ruled out in my opinion anyway.

So we are back to something that may be your original proposal I guess.
Maybe we can first agree on part of the stuff, like:
- A Modifiers class that contains an integer with:
  - Getter/setter for the int. wording to be defined
  - Constant flags for that integer (values of which we define ourselves)
  - Individual flag accessors (boolean getters/setters) wording to be defined
- Replacing the integer in the nodes with modifiers by an instance of Modifiers
  - Appropriate delegation for the current integer getter/setter

We also have to find consensus on:
1. whether we deprecate access to the backing integer (through the nodes and/or through the Modifiers class)
2. wording for the int getter/setter (if access is not deprecated)
3. wording for the boolean getters/setters
4. whether we define an interface for nodes that have modifiers
5. whether we define additional specific boolean getters/setters on node with modifiers that appropriately delegate to the Modifiers instance

Did I miss something ??

My opinion on those points:
1. I think we let them be accessible through Modifiers, else it would be a show stopper for all those out there that use switches.
2. Maybe simply Modifiers.get() and Modifiers.set(int)
3. Yet an alternate wording proposition: Modifiers.hasStatic()/Modifiers.setStatic().
4. I don't mind. I think defining one has no cost.
5. I think it could be good there are individual getters/setters on nodes. I don't think it is that much to implement and maintain, and it would be nice to use.

ftomassetti:
Ok, let's try to summarize and see if we can agree and either drop this patch or complete it.

# Status of this patch

A few classes implement NodeWithModifiers. E.g.

```
abstract class BaseParameter extends Node implements NodeWithModifiers {
   private Modifiers modifiers = new Modifiers();
```

NodeWithModifiers looks like this:

```
public interface NodeWithModifiers {
    int getModifiers();
    Modifiers getModifiersSet();
}
```

while it would be call to have all the methods to read/write the flags on NodeWithModifiers that would mean having to add them to each single class implementing NodeWithModifiers (lot of redundant code).

Modifiers is where the most of the work is done. I report just the declaration of the methods:

```
public class Modifiers {
    boolean hasModifier(javax.lang.model.element.Modifier modifier);
    public boolean isAbstract();
    public boolean isFinal();
    public boolean isNative();
    public boolean isPrivate();
    public boolean isProtected();
    public boolean hasPackageLevelAccess();
    public boolean isPublic() ;
    public boolean isStatic() ;
    public boolean isStrictfp();
    public boolean isSynchronized();
    public boolean isTransient() ;
    public boolean isVolatile() ;
    public boolean has(Modifier modifier);
    public Modifiers set(Modifier modifier);
    public Modifiers unset(Modifier modifier) ;
    public AccessSpecifier getAccessSpecifier();    
```

# Questions
- Do we think this general schema (nodes implements NodeWithModifiers, NodeWithModifiers just return a Modifiers class) is acceptable or should it be changed?
- If we think the general schema is acceptable which methods should we add, remove, change?
- Once the general schema and the interface of Modifiers are defined we should discuss how to store the data in Modifiers

What do you think?

ptitjes:
Just a note to say that I intend to propose in 3.0 to have a node called `ModifiersAndAnnotations` to more closely follow the grammar defined by JLS and the lexical ordering of modifiers and annotations. That object would contain a list of interleaved `Modifier` and `Annotation` nodes. That might impact a lot the design decisions made here...

As for your proposal, @ftomassetti, I like it (and it is compatible with my above comment about `ModifiersAndAnnotations`). Just a few remarks/requests:
- I still think that the `isAbstract()`, ..., methods should be named `hasAbstract()`, ..., to be coherent with `has(Modifier)`. The names `containsAbstract()`, ..., and `contains(Modifier)` would fit me as well. But `Modifiers.isAbstract()` does not mean anything to me.
- I think we should remove `hasModifiers(javax.lang.model.element.Modifier)` and, if really needed, that should be moved to some utility class. Some class that make the two-way bridge between `javax.lang.model.element.Modifier` and our `Modifier` class.
- If we push this in 3.0, then maybe we could directly rename `Modifiers getModifiersSet()` to `Modifiers getModifiers()`. I'd rather not have that 'Set' part in the name.

ftomassetti:
- I am fine with using `hasAbtract` instead of `isAbstract`
- I agree on removing hasModifiers(javax.lang.model.element.Modifier)
- I would agree that getModifiers sound better but ideally I would like to get done with this before 3.0

About the changes in 3.0 I think this class could need some internal refactoring but it could keep having the same API. I see it as an abstraction above the pure list of tokens present.

ftomassetti:
Rebased, and updated with latest suggestions

ptitjes:
Well, I think having two Modifier names (Modifiers.Modifier and Modifier) and a ModifierSet makes the whole thing confusing. I advise to make something clean and not confusing for 3.0.

ftomassetti:
I am definitely ok with trying something clean, I assure you that I am not trying to do something confusing on purpose :)

Do you have specific suggestions?

ptitjes:
Well, my suggestion is to postpone to 3.0 as to not having to bother with not breaking API compatibility. Then we can make the correct renames.

matozoid:
Can you continue this discussion or close the request please, so this pull request doesn't stay here forever.

ftomassetti:
To simplify the discussion let's assume this patch will go in the 3.0 branch and we do not strictly need to be compatible with the existing code.

I would suggest to the following (just a sketch to get the basic idea):

``` java
interface NodeWithModifiers {
   ModifiersSet getModifiers();
  void setModifiers(ModifiersSet modifiers);
}
// Immutable collection of modifiers
class ModifiersSet {
    // Both these methods return a copy of the object
    ModifiersSet add(Modifier modifier);
    ModifiersSet remove(Modifier modifier);
    AccessLevel getAccessLevel();
    ModifiersSet setAccessLevel(AccessLevel accessLevel);
    // optional: add hasPublic, hasProtected, etc. 
}
```

Would that work for you?

ftomassetti:
Ok, given we are aiming at version 3.0 let's try to reconsider all the issues and go for a future-proof solution

## Problems with the existing solution
- it does not preserve order of modifiers
- it does not permit annotations to be interleaved with modifiers
- it does not offer a common interface for all nodes having modifiers

So there are a few problems to solve and I think the solution cannot be super-simple unfortunately.

## These new annotations among modifiers... what to about them?

Looking at the JLS (https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-MethodModifier) Modifiers and Annotations are mixed but it is not clear what these annotations are referring to:
- are they annotations to the method itself? Then why we can mix them with modifiers instead of putting them in one place?
- are they annotations to the return type? Then again why we can mix them with modifiers instead of putting them in one place?
- are they annotations for the modifiers themselves? That make more sense to me but what about the annotations beyond the last modifier? Or just when there are annotations and not modifiers? Do they belong to the return type in this case?

In my opinion we have two possibile solutions:
- represent an heterogeneous list of annotations and modifiers: this is closer to the specs but it is also cumbersome to use and slighlt less nice to use
- represent a list of modifiers which can have annotations. Annotations beyond the last modifier are assigned to the return type or the method itself.

## Solution A: the hetereogeneous list

We have a list (to preserve order) of Modifier and Annotation.

I think that if we have such list it would be very useful to add methods to add and remove modifiers, otherwise the user would have to iterate on this list to do so.

An interface coould permit easy querying and refactoring by exposing methods to add and remove modifiers. The add method could just create a Modifier at the end of the list (veryfing there is no duplication) and the delete method could find the Modifier and remove it. 

I am thinking something like:

``` java
interface NodeWithModifiers {

    // Add it at the end of the list
    void addAnnotationAmongModifiers(Annotation annotation);

    // Add annotation immediately before the modifier
    void addAnnotationToModifier(Modifier modifier, Annotation annotation);

    Iterable<Annotation> getAnnotationsBeforeModifier(Modifier modifier);

    Iterable<Annotation> getAllAnnotations();

    // Potentially we could also check if the modifier is legal for this node: we do not want to use
    // default on class methods, for example
    // If the modifier is already present it is a no-op and false is returned
    boolean addModifier(Modifier modifier);

    // If the modifier is not present it is a no-op and false is returned
    // Should we also remove preceeding annotations in this case?
    boolean removeModifier(Modifier modifier);

    // ...and isPrivate, isProtected, etc
     boolean isPublic();

    // ...and setPublic, setProtected, etc.
     void setPublic(boolean value);

    // This filter and return only the modifier, returning a list of Modifier instead of the internal
    // list of Nodes 
    Iterable<Modifier> getAllModifiers(); 
}
```

Given we do not have traits in Java I would use a proxy, so one class would implement all these methods, we would add this class as a field to all our classes needing to implement the interface and the implementation would be simply a proxy to the field.

## Solution B: the list of modifiers which can be annotated

Internally we could have a list of Modifiers where each Modifier is a proper node with has its own position (for possible space preserving operations) and annotations, so the annotations immediately preceeding a Modifier are associated to it. I think it is cleaner/simpler that having a flat list of Modifiers and Annotations (especially given we have not union type like in Ceylon :D).

Note that we could have to handle also the re-assignment of annotations (or decide if we want to remove them when we drop a Modifier).

``` java
interface NodeWithModifiers {

    // Potentially we could also check if the modifier is legal for this node: we do not want to use
    // default on class methods, for example
    // If the modifier is already present it is a no-op and false is returned
    boolean addModifier(Modifier modifier);

    // If the modifier is not present it is a no-op and false is returned
    // Should we also remove preceeding annotations in this case?
    boolean removeModifier(Modifier modifier);

    // ...and isPrivate, isProtected, etc
     boolean isPublic();

    // ...and setPublic, setProtected, etc.
     void setPublic(boolean value);

    // This filter and return only the modifier, returning a list of Modifier instead of the internal
    // list of Nodes 
    Iterable<Modifier> getAllModifiers(); 
}

class Modifier {
    void addAnnotation(Annotation annotation);
    Iterable<Annotation> getAnnotations();
}
```

ptitjes:
I'm in favour for something like A.

Some comments though:
- You'd also need an `boolean hasModifier(Modifier modifier);` method has there is no `contains(...)` method on the iterable returned by `getAllModifiers()`.
- Annotations can't annotate modifiers. So `void addAnnotationToModifier(Modifier modifier, Annotation annotation);` has no meaning. Maybe what you mean is `insertBefore(...)`. So you'd have:

```
void insertAnnotationBefore(Modifier before, Annotation annotation);
void insertAnnotationBefore(Annotation before, Annotation annotation);
void insertModifierBefore(Modifier before, Modifier modifier);
void insertModifierBefore(Annotation before, Modifier modifier);
void appendAnnotation(Annotation annotation)
void appendModifier(Modifier modifier)
```

All in all, what you propose could be greatly simplified, both in term of usability and implementability, with the introduction of an `ExtendedModifier` parent type for `Annotation` and `Modifier`:

```
interface NodeWithModifiers {

    boolean prependModifier(ExtendedModifier modifier);
    boolean insertModifier(ExtendedModifier before, ExtendedModifier modifier);
    boolean appendModifier(ExtendedModifier modifier);
    boolean removeModifier(ExtendedModifier modifier);

    Iterable<ExtendedModifier> getAllModifiers();
    Iterable<Modifier> getModifiers();
    Iterable<Annotation> getAnnotations();

    boolean hasModifier(Modifier modifier);
    boolean hasAnnotation(QualifiedNameExpr annotationName);
    Annotation getAnnotation(QualifiedNameExpr annotationName);

    // ...and isPrivate, isProtected, etc
    boolean isPublic();

    // ...and setPublic, setProtected, etc.
    boolean setPublic(boolean value);
}
```

ftomassetti:
@ptitjes I believe that the ExtendedModifier class would be rather confusing so I would avoid that. I could instead add the suggested methods.

Of course also voting to reject this proposal and perhaps submitting a better one is a valid option :)

vinitgaikwad0810:
Hi,

Does anyone of you have a visual example of AST constructed by javaparser?

I am beginner here and I wanted to multiple classes in one Java file and I
want to create member to class mapping.

Regards

ftomassetti:
Hi, we were glad to help but please ask the question on [the gitter channel](https://gitter.im/javaparser/javaparser?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge) or open an issue. This is not the right place for asking questions.

ptitjes:
@ftomassetti I won't reject your proposal. This was just a suggestion as I fear not having `ExtendedModifier` will give a very much complicated and obfuscated API.

ftomassetti:
Bot @ptitjes and @SmiddyPence are in favor of options A, so let's go for that one

ptitjes:
Just a note to make clear that I said that "I'm in favour for **something like A**.". I still think the proposed API needs more work.

matozoid:
Who wants to work on this?

ftomassetti:
This PR aged and @matozoid solved this issue. Time to close this one.

PR #104, opened 2015-02-19
First attempt at adopting jbehave-junit-runner

ftomassetti:
With this patch we use jbehave-junit-runner to better integrate jbehave with junit. In this way we will have one separate test for each scenario and we will have the stack trace of failing tests. It should help understanding which test fail and why.

We can also get rid off the FooEmbedder classes, simplifying a bit our tests machinery.


SmiddyPence:
+1

sebastiankirsch:
I love them stack traces ;)

Issue #100, opened 2015-02-18
Set language level to Java 8


ftomassetti:
Hi, what are the thoughts about that?
I would love to get started on Java 8 :)
Do we have any reason to stick with Java 6 (I think it is not supported anymore).
Java 8 has been released almost 1 year ago (March 2014).

matozoid:
Just one request from someone forced to use Java 1.5 at work (aaargh!!!): carefully document which version is compiled for which Java version!

sebastiankirsch:
I would immediately fork a java6-compatible version :/

Danny van Bruggen notifications@github.com schrieb am Wed Feb 18 2015 at
3:53:48 PM:

ftomassetti:
mmm ok, I see, probably times are not mature :)

sebastiankirsch:
Don't get me wrong, I really like java8. But for a library, I'd say this
cripples adoption...

Federico Tomassetti notifications@github.com schrieb am Wed Feb 18 2015
at 4:06:10 PM:

matozoid:
Suggestion: limit the src/main code to Java 6, but let the src/test code compile with 8. Besides that: investigate if there is anything that can be done with the Java 6 code that Java 8 users can take advantage of (making it easy to use functional stuff with it, for example)

Tip: use maven-enforcer-plugin and animal-sniffer-maven-plugin to enforce Java versions.

PR #105, opened 2015-02-21
Enforce compiling the library for a certain Java version

matozoid:
As discussed in https://github.com/javaparser/javaparser/issues/100

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -70,11 +70,11 @@
                             <rules>
                                 <requireJavaVersion>
                                     <!-- Make sure a compiler of this version is used -->
-                                    <version>1.6</version>
+                                    <version>${java.version}</version>
                                 </requireJavaVersion>
                                 <enforceBytecodeVersion>
                                     <!-- Make sure the dependencies are compiled for our Java version -->
-                                    <maxJdkVersion>1.6</maxJdkVersion>
+                                    <maxJdkVersion>${java.version}</maxJdkVersion>
                                 </enforceBytecodeVersion>
                             </rules>
                         </configuration>
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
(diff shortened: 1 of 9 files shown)


matozoid:
Aaaand I still need to fix something here.

javaparser-core/pom.xml
```diff
@@ -28,6 +33,61 @@
...
+                        <id>enforce-versions</id>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <requireJavaVersion>
+                                    <!-- Make sure a compiler of this version is used -->
+                                    <version>1.6</version>
```
ftomassetti: Can we refer to the property declared above?

javaparser-core/pom.xml
```diff
@@ -28,6 +33,61 @@
...
+                                </requireJavaVersion>
+                                <enforceBytecodeVersion>
+                                    <!-- Make sure the dependencies are compiled for our Java version -->
+                                    <maxJdkVersion>1.6</maxJdkVersion>
```
ftomassetti: As above, can we refer to the property java.version?

javaparser-testing/src/test/java/com/github/javaparser/bdd/ManipulationTest.java
```diff
@@ -15,7 +15,7 @@
 
     @Override
     public InjectableStepsFactory stepsFactory() {
-        Map<String, Object> state = new HashMap<String, Object>();
+        Map<String, Object> state = new HashMap<>();
```
ftomassetti, line 18: Isn't he diamond operator Java7+?

javaparser-testing/src/test/java/com/github/javaparser/bdd/ParsingTest.java
```diff
@@ -15,7 +15,7 @@
     
     @Override
     public InjectableStepsFactory stepsFactory() {
-        Map<String, Object> state = new HashMap<String, Object>();
+        Map<String, Object> state = new HashMap<>();
```
ftomassetti, line 18: ditto

pom.xml
```diff
@@ -42,7 +42,7 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<java.version>1.6</java.version>
+		<java.version>1.8</java.version>
```
ftomassetti, line 45: Java 8?

matozoid, line 45: That's the point: you can write everyting in Java 8, except for the core library, which is compiled with Java 6, and double checked to be compatible.

ftomassetti, line 45: I see, I missed the point :D

matozoid, line 45: Feel free to double check that everything compiles, and that the output library has the correct version. Maybe try adding a Java 8-only dependency to core, or add some Java 7 code, and see if the build fails as it should.

javaparser-testing/src/test/java/com/github/javaparser/bdd/visitors/PositionTestVisitor.java
```diff
@@ -10,6 +10,8 @@
 import com.github.javaparser.ast.type.*;
 import com.github.javaparser.ast.visitor.VoidVisitorAdapter;
 
+import java.util.HashSet;
```
sebastiankirsch: Is that intentional?

matozoid: No, it's accidental.

sebastiankirsch:
I'd give that a go, but I guess the travis.yml file causes a conflict?

matozoid:
It's mergable again, who wants to?

PR #108, opened 2015-02-24
Grammar work

ptitjes:
I completely reworked the way positions are handled in the grammar, which is as a result a lot simpler.

Now {begin|end}{Line|Column} are stored in Nodes as two Position objects (begin and end).

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/Position.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/Position.java
@@ -3,30 +3,54 @@ package com.github.javaparser;
 import com.github.javaparser.ast.Node;
 
 public class Position {
-    private int line;
-    private int column;
+    public final int line;
+    public final int column;
 
-    public static final Position ABSOLUTE_START = new Position(Node.ABSOLUTE_BEGIN_LINE,-1);
-    public static final Position ABSOLUTE_END = new Position(Node.ABSOLUTE_END_LINE,-1);
+    public static final Position ABSOLUTE_START = new Position(Node.ABSOLUTE_BEGIN_LINE, -1);
+    public static final Position ABSOLUTE_END = new Position(Node.ABSOLUTE_END_LINE, -1);
 
-    public static Position beginOf(Node node){
-        return new Position(node.getBeginColumn(),node.getBeginColumn());
+    /**
+     * @deprecated Use node.getBegin()
+     */
+    @Deprecated
+    public static Position beginOf(Node node) {
+        return node.getBegin();
     }
 
-    public static Position endOf(Node node){
-        return new Position(node.getEndColumn(),node.getEndColumn());
+    /**
+     * @deprecated Use node.getEnd()
+     */
+    @Deprecated
+    public static Position endOf(Node node) {
+        return node.getEnd();
     }
 
-    public Position(int line, int column){
+    public Position(int line, int column) {
         this.line = line;
         this.column = column;
     }
 
-    public int getLine(){
+    /**
+     * @deprecated Use Position.line
+     */
+    @Deprecated
+    public int getLine() {
         return this.line;
     }
 
-    public int getColumn(){
+    /**
+     * @deprecated Use Position.column
+     */
+    @Deprecated
+    public int getColumn() {
         return this.column;
+    }
+
+    public Position withLine(int line) {
+        return new Position(line, this.column);
+    }
+
+    public Position withColumn(int column) {
+        return new Position(this.line, column);
     }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
@@ -21,6 +21,7 @@
  */
 package com.github.javaparser.ast;
 
+import com.github.javaparser.Position;
 import com.github.javaparser.ast.body.AnnotationDeclaration;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.EmptyTypeDeclaration;
@@ -71,6 +72,13 @@ public final class CompilationUnit extends Node {
         setImports(imports);
         setTypes(types);
     }
+
+	public CompilationUnit(Position begin, Position end, PackageDeclaration pakage, List<ImportDeclaration> imports, List<TypeDeclaration> types) {
+		super(begin, end);
+		setPackage(pakage);
+		setImports(imports);
+		setTypes(types);
+	}
 
     @Override
     public <R, A> R accept(GenericVisitor<R, A> v, A arg) {
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -25,6 +25,7 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 
+import com.github.javaparser.Position;
 import com.github.javaparser.ast.comments.Comment;
 import com.github.javaparser.ast.visitor.DumpVisitor;
 import com.github.javaparser.ast.visitor.EqualsVisitor;
@@ -42,13 +43,9 @@ import com.github.javaparser.ast.visitor.VoidVisitor;
  */
 public abstract class Node {
 
-	private int beginLine;
+	private Position begin;
 
-	private int beginColumn;
-
-	private int endLine;
-
-	private int endColumn;
+	private Position end;
 	
 	private Node parentNode;
 
@@ -65,11 +62,13 @@ public abstract class Node {
 	public Node() {
 	}
 
+	public Node(Position begin, Position end) {
+		this.begin = begin;
+		this.end = end;
+	}
+
 	public Node(final int beginLine, final int beginColumn, final int endLine, final int endColumn) {
-		this.beginLine = beginLine;
-		this.beginColumn = beginColumn;
-		this.endLine = endLine;
-		this.endColumn = endColumn;
+		this(new Position(beginLine, beginColumn), new Position(endLine, endColumn));
 	}
 
 	/**
@@ -100,21 +99,63 @@ public abstract class Node {
 	public abstract <A> void accept(VoidVisitor<A> v, A arg);
 
 	/**
+	 * Returns the begin position of this node.
+	 *
+	 * @return the begin position, or <code>null</code> if the position has not been set
+	 */
+	public Position getBegin() {
+		return begin;
+	}
+
+	/**
+	 * Sets the begin position of this node.
+	 *
+	 * @param begin
+	 *              the end position
+	 */
+	public void setBegin(Position begin) {
+		this.begin = begin;
+	}
+
+	/**
+	 * Returns the end position of this node.
+	 *
+	 * @return the end position, or <code>null</code> if the position has not been set
+	 */
+	public Position getEnd() {
+		return end;
+	}
+
+	/**
+	 * Sets the end position of this node.
+	 *
+	 * @param end
+	 *            the end position
+	 */
+	public void setEnd(Position end) {
+		this.end = end;
+	}
+
+	/**
 	 * Return the begin column of this node.
 	 * 
 	 * @return the begin column of this node
+	 * @deprecated Use getBegin().column
 	 */
+	@Deprecated()
 	public final int getBeginColumn() {
-		return beginColumn;
+		return begin == null ? 0 : begin.column;
 	}
 
 	/**
 	 * Return the begin line of this node.
 	 * 
 	 * @return the begin line of this node
+	 * @deprecated Use getBegin().line
 	 */
+	@Deprecated()
 	public final int getBeginLine() {
-		return beginLine;
+		return begin == null ? 0 : begin.line;
 	}
 
 	/**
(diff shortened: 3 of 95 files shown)


javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -3,8 +3,8 @@
 import com.github.javaparser.ast.Node;
 
 public class Position {
-    private int line;
-    private int column;
+    public final int line;
```
ftomassetti, line 6: making the field public instead of using a getter we cannot later change the way it is stored
I guess it could be ok in this case

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -72,6 +73,13 @@ public CompilationUnit(int beginLine, int beginColumn, int endLine, int endColum
         setTypes(types);
     }
 
+	public CompilationUnit(Position begin, Position end, PackageDeclaration pakage, List<ImportDeclaration> imports, List<TypeDeclaration> types) {
```
ftomassetti, line 76: maybe package_ (in the code I have seen an underscore used after the identifier when it would correspond to a keyword)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -65,11 +62,13 @@
 	public Node() {
 	}
 
+	public Node(Position begin, Position end) {
+		this.begin = begin;
+		this.end = end;
+	}
+
 	public Node(final int beginLine, final int beginColumn, final int endLine, final int endColumn) {
```
ftomassetti, line 70: Deprecated?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -100,21 +99,63 @@ public Node(final int beginLine, final int beginColumn, final int endLine, final
...
+	 */
+	public Position getBegin() {
+		return begin;
+	}
+
+	/**
+	 * Sets the begin position of this node.
+	 *
+	 * @param begin
+	 *              the end position
```
ftomassetti, line 114: clean up

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -100,21 +99,63 @@ public Node(final int beginLine, final int beginColumn, final int endLine, final
...
+	/**
 	 * Return the begin column of this node.
 	 * 
 	 * @return the begin column of this node
+	 * @deprecated Use getBegin().column
 	 */
+	@Deprecated()
 	public final int getBeginColumn() {
-		return beginColumn;
+		return begin == null ? 0 : begin.column;
```
ftomassetti, line 147: I would throw an exception is the position is null

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -65,11 +62,13 @@
 	public Node() {
 	}
 
+	public Node(Position begin, Position end) {
+		this.begin = begin;
```
ftomassetti, line 66: I would check for nulls and prevent them

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -100,21 +99,63 @@ public Node(final int beginLine, final int beginColumn, final int endLine, final
 	public abstract <A> void accept(VoidVisitor<A> v, A arg);
 
 	/**
+	 * Returns the begin position of this node.
+	 *
+	 * @return the begin position, or <code>null</code> if the position has not been set
```
ftomassetti, line 104: are we sure null is acceptable?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -100,21 +99,63 @@ public Node(final int beginLine, final int beginColumn, final int endLine, final
...
+		return begin;
+	}
+
+	/**
+	 * Sets the begin position of this node.
+	 *
+	 * @param begin
+	 *              the end position
+	 */
+	public void setBegin(Position begin) {
```
ftomassetti, line 116: I would prevent nulls

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -158,9 +203,11 @@ public final int getEndLine() {
 	 * 
 	 * @param beginColumn
 	 *            the begin column of this node
+	 * @deprecated Use setBegin()
 	 */
+	@Deprecated()
 	public final void setBeginColumn(final int beginColumn) {
-		this.beginColumn = beginColumn;
+		begin = new Position(begin == null ? 0 : begin.line, beginColumn);
```
ftomassetti, line 210: I would add a method to position which takes the column a return an instance of Position with the same line and the given column:

```
begin = begin.withColumn(beginColumn);
```

javaparser-core/src/main/java/com/github/javaparser/Position.java
ptitjes, line 6: Well, I mean it is like a Point class. If we decide it is immutable then there is no real interest to have accessors.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -216,9 +267,11 @@ public final void setEndColumn(final int endColumn) {
 	 * 
 	 * @param endLine
 	 *            the end line of this node
+	 * @deprecated Use setEnd()
 	 */
+	@Deprecated()
```
ftomassetti, line 272: It could be not necessarily deprecated

javaparser-core/src/main/java/com/github/javaparser/parser/ASTParserBase.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * @author Didier Villevalois
+ */
+public abstract class ASTParserBase {
+
+
+	protected List add(List list, Object obj) {
+		if (list == null) {
+			list = new LinkedList();
```
ftomassetti: Maybe ArrayList could be faster

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -53,51 +54,15 @@ import com.github.javaparser.ast.type.*;
 /**
  * <p>This class was generated automatically by javacc, do not edit.</p>
  */
-final class ASTParser {
+public class ASTParser extends ASTParserBase {
+	// Make the parser public for now
```
ftomassetti: why you need that?

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
ptitjes, line 76: Copy-pasted names from the other constructors. I did not want to introduce inconsistancies and modify stuff outside of my current task (as too not confuse the diffs).

Well blame the previous coder!! :P
But I'd be happy to make the change in an other commit!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 66: Nah! Some people instanciate nodes manually without positions. The null checks are in the deprecated getter/setters.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
ftomassetti, line 76: make sense :)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 70: Well if everybody is ok with it, I'd like to deprecate all the constructors of all the nodes with integer positions.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 114: Arghhh... Copy/paste error!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 210: That wood be a good addition indeed! Added on my todo!

javaparser-core/src/main/java/com/github/javaparser/ast/TypeParameter.java
```diff
@@ -125,5 +141,6 @@ public void setTypeBound(final List<ClassOrInterfaceType> typeBound) {
 
     public void setAnnotations(List<AnnotationExpr> annotations) {
         this.annotations = annotations;
+	    setAsParentNodeOf(annotations);
```
ptitjes, line 144: Here I fixed a potential bug, while passing on that class...

javaparser-core/src/main/java/com/github/javaparser/ast/expr/TypeExpr.java
```diff
@@ -35,8 +43,6 @@ public Type getType() {
 
     public void setType(Type type) {
         this.type = type;
+	    setAsParentNodeOf(this.type);
```
ptitjes: And here too... That is code too quickly copied from rpau! Fixed now!

javaparser-core/src/main/java/com/github/javaparser/parser/ASTParserBase.java
ptitjes: Yep. I did not modify too much the original code, just extracted it yet. In fact, I plan to do more work on that stuff to simplify the grammar further and minimize the ammount of java code in the grammar.

javaparser-core/src/main/javacc/java_1_8.jj
ptitjes: I explained in the commit message that this is temporary (because of the package move and the need to be visible by the JavaParser class).

I plan to make a further front-end in beetween the JavaParser class. It will provide better error handling.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 210: Done in the additional commit:
https://github.com/ptitjes/javaparser/commit/70dab5734a87c73857e0204f10f4b1311b729ede

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
sebastiankirsch, line 70: Makes sense

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
sebastiankirsch, line 147: this would be a behaviour change: currently, it would return `-1`

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
sebastiankirsch, line 66: As the positions aren't mandatory, throwing an exception would indeed be mean...

javaparser-core/src/main/java/com/github/javaparser/ast/expr/TypeExpr.java
sebastiankirsch: actually, there's #91 waiting to be merged ;)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ptitjes, line 147: Actually it returns 0. But maybe that would be a good thing to return -1 in the new implementation. I wonder how many people rely on this to differentiate parsed node from manually created ones... (actually nobody I guess!)
What do you think ?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/TypeExpr.java
ptitjes: Uhuh, did not see that one... :)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
sebastiankirsch, line 147: Whoopsie. So why did I think an int has a default value of `-1`? I would definitively stick to `0` then!

ptitjes:
I did rebase upon the 2/6 step and add a new commit cc63789 just for the new 'default' modifier.

ptitjes:
@sebastiankirsch As for the two bug fixes, I'm willing to rebase my branch (btw github's pull requests seem to like it) on master once #91 is merged, so the bug fixes will disappear.

ptitjes:
@sebastiankirsch @ftomassetti @matozoid Suspend the pull request.

Hmmm. I just realized that I introduced a binary incompatibility because the ParseException moved in the com.github.javaparser.parser package.

If you have any suggestions as to how to solve that cleanly, I'm listening.

ptitjes:
Oki restored the ParseException as discussed on chat.
And rebase onto master.

ptitjes:
ParseException actually changes its API... so would need a major release!
Also, I wonder if this is a useful change as I'd prefer to directly pass tokens (as in lex-preserving) instead of positions.

ptitjes:
Indeed, it is superseded by https://github.com/ptitjes/javaparser/tree/lex-preserving

PR #116, opened 2015-03-05
Fix #84

### LINKED ISSUE #84: com.github.javaparser.ast.body.Parameter.getType() can be null since 2.0.0
I assume a parameter's type being `null` only occurs for lambda expressions, thus why it didn't pop up until now.

This causes `NullPointerException`s at least for
- `com.github.javaparser.ast.visitor.GenericVisitorAdapter`
- `com.github.javaparser.ast.visitor.ModifierVisitorAdapter`
- `com.github.javaparser.ast.visitor.VoidVisitorAdapter`

The other adapters seem to be unaffected, one of them was explicitly adapted to it. I'd appreciate a double-check of this.

Before introducing a mere `if (parameter.getType() != null)` everywhere, I wanted to discuss how this should be approached. Does it make sense to introduce a new class representing lambda parameters? Although lambda parameters can have a type too, it seems that it's the only instance a paramater can have no type. Or am I missing something?

ptitjes:
I'm willing to put the correct thingies in the JBehave test suite if someone tells me where to act.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2015 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with JavaParser.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.github.javaparser.ast.type;
+
+import com.github.javaparser.ast.visitor.GenericVisitor;
+import com.github.javaparser.ast.visitor.VoidVisitor;
+
+/**
+ * An unknown parameter type object. It plays the role of a null object for
+ * lambda parameters that have no explicit type declared. As such, it has no
+ * lexical representation and hence gets no comment attributed.
+ *
+ * @author Didier Villevalois
+ */
+public final class UnknownType extends Type {
+
+    public UnknownType() {
+    }
+
+    @Override
+    public <R, A> R accept(final GenericVisitor<R, A> v, final A arg) {
+        return v.visit(this, arg);
+    }
+
+    @Override
+    public <A> void accept(final VoidVisitor<A> v, final A arg) {
+        v.visit(this, arg);
+    }
+}
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java
@@ -430,6 +430,15 @@ public class CloneVisitor implements GenericVisitor<Node, Object> {
 	}
 
 	@Override
+	public Node visit(UnknownType _n, Object _arg) {
+		Comment comment = cloneNodes(_n.getComment(), _arg);
+
+		UnknownType r = new UnknownType();
+		r.setComment(comment);
+		return r;
+	}
+
+	@Override
 	public Node visit(ArrayAccessExpr _n, Object _arg) {
 		Expression name = cloneNodes(_n.getName(), _arg);
 		Expression index = cloneNodes(_n.getIndex(), _arg);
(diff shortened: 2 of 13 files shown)


ptitjes:
Ok that should do it!

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+package com.github.javaparser.ast.type;
+
+import com.github.javaparser.ast.visitor.GenericVisitor;
+import com.github.javaparser.ast.visitor.VoidVisitor;
+
+/**
+ * @author Didier Villevalois
+ */
+public final class UnknownType extends Type {
```
ftomassetti, line 33: I think a comment could be useful: it is not straightforward to understand what class represent

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
```diff
@@ -0,0 +1,45 @@
...
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with JavaParser.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/*
+ * Created on 05/10/2006
```
ftomassetti: You have been working a lot on this class :D

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java
```diff
@@ -429,6 +429,15 @@ public Node visit(WildcardType _n, Object _arg) {
 		return r;
 	}
 
+    @Override
+    public Node visit(UnknownType _n, Object _arg) {
+        Comment comment = cloneNodes(_n.getComment(), _arg);
```
ftomassetti: Can an UnknownType have a comment?

ftomassetti: If so it should dumped in the DumpVisitor, if not is should be prevent to have a comment associated

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitor.java
```diff
@@ -137,6 +133,8 @@
 
 	public R visit(WildcardType n, A arg);
 
+    public R visit(UnknownType n, A arg);
```
ftomassetti: should we use the tabs until we reformat the whole class?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitorAdapter.java
```diff
@@ -1488,6 +1483,11 @@ public R visit(final UnaryExpr n, final A arg) {
 		return null;
 	}
 
+    @Override
```
ftomassetti: as above, should we use tabs so that it is aligned to the other methods?

javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
```diff
@@ -200,6 +200,23 @@ Then lambda in statement 3 in method 1 in class 1 is parent of contained paramet
...
+Given a CompilationUnit
+When the following source is parsed:
+package com.github.javapasrser.bdd.parsing;
+import java.util.function.Function;
+public class MultiParameterizedLambdas {
+    public static void main(String[] args) {
+        BiFunction<Integer, Integer, String> f = (a, b) -> String.valueOf(a) + String.valueOf(b);
+    }
+}
+Then lambda in statement 1 in method 1 in class 1 has parameters with non-null type
```
ftomassetti, line 219: Should we also check that the type is the UnknownType?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java
ptitjes: I think not. This is a kind of null object and can't receive any comment as it has no lexical representation.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitor.java
ptitjes: I'll put tabs. (The problem is that is it is different from files to files)

javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
ptitjes, line 219: Well the problem was that the type was null. So I guess checking for more is over-testing.

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
ptitjes, line 33: Adding that.

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnknownType.java
ptitjes: Wow! Indeed I don't know where my IDE got that date...

ptitjes:
Fixed!

ftomassetti:
+1, thanks!

Issue #120, opened 2015-03-09
Tabs/spaces and indentation guidelines


ptitjes:
Yeah I know that this is a sensitive issue...
But there are still files with tabs, others with spaces, and even some that have mixed tabs and spaces! 

It is hard to do correct every time and can lead to surprising things with diffs. I'm sure we all agree on that. Personally, I don't really have a preference as this is just a preference in the IDE...

However, I noticed that IDEA by default adds spaces (in addition to tabs) when I line-break in the middle of a method declaration's argument list in order to align with the first argument.
That's the main problem I hit when trying to be like the current code.

So do we define guidelines ?

(and maybe apply a big code reformatting once and for all for the 2.1 release...)

ftomassetti:
+1 with the reformatting, I was thinking myself to do it. Maybe we could do one package at the time and consider the pull requests in the pipeline

ptitjes:
Yep, that's why I proposed to do it just before 2.1 release in order to have the less pull requests to rework. If everything goes well, every null-handling, new interfaces, fixes would have gone to master for the release. Hence, there would only be the work I started for 3.0 and I don't mind reworking them.

ptitjes:
So how do we proceed to get consensus ? Then I volunteer to make the reformatting, and there should be a second person that independently configures its IDE and do the reformatting again to check nothing has been missed.

ptitjes:
Also I noticed that we should agree on annotations on newline or not (and maybe about method order in visitors)

For instance, see difference in 
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java#L135
and in
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitorAdapter.java#L82

matozoid:
What I know: Eclipse loves messing around with line breaks, intellij respects them. As long as these two are used simultaneously, automatic formatting will lead to nonsense diffs.

What I do: I stopped caring, I only format the pieces of code I work on, leaving line breaks alone as much as possible, introducing as few "line too long: wrap" line breaks. I tweak default formatter settings as little as possible.

No problem with a full formatting of all code, but please resubmit all your pull requests afterwards :)

ptitjes:
[...]

PR #128, opened 2015-04-04
Bugs due to null lists

ftomassetti:
In some situations we do not handle correctly empty lists. In visitors we just check for null lists but empty lists have the same meaning so they should be treated in the same way (derived from #54 )

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/internal/Utils.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/internal/Utils.java
@@ -1,0 +1,20 @@
+package com.github.javaparser.ast.internal;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * @author Federico Tomassetti
+ * @since 2.0.1
+ */
+public class Utils {
+
+    public static <T> List<T> ensureNotNull(List<T> list) {
+        return list == null ? Collections.<T>emptyList() : list;
+    }
+
+    public static <E> boolean isNullOrEmpty(Collection<E> collection) {
+        return collection == null || collection.isEmpty();
+    }
+}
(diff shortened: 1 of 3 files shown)


ptitjes:
Would you mind keeping tabs because it obfuscates your changes ? I propose that we make a reformat code just before 2.1, once we agreed on #120.

Also could git rebase -i and do some commit squashing ?

ftomassetti:
Sure, I would just drop my last commit then, consider that the way the diff looks like depends on tab issues

ftomassetti:
And it should be done. About the commit squashing I am not sure, I think it is better to have more granular commits.

ptitjes:
Well in the second one you replaced tabs with spaces.
Also you can squash "Adding newline" and the first one.

ptitjes:
@ftomassetti Well, I did not mean squashing everything in one commit. I also do think granular commits are better. But when you add a commit to fix another one, you can 'git rebase -i', move the fix commit, squash it, and then 'git -f push'

ftomassetti:
You are right, I have squashed the commit about the newline

ptitjes:
Nice! I may be picky :P but in the second one you still tabs to space... You may want to turn off you auto-format on save. IIRC you use Idea, then there is a 'View > Active Editor > Show Whitespaces' command.

ptitjes:
I guess now you hate me for real :( Hopefully this will be easier when we will have all formated...

ftomassetti:
Ok, I fixed the ones you suggested and I put back the tabs in the method _printModifiers_

ptitjes:
That's good for me! +1

ftomassetti:
Thanks :D

ptitjes:
Thank you :P

ftomassetti:
Thanks for merging

Issue #95, opened 2015-02-11
CommentParser:  invalid comment within String found


quarkie:
The commentParser does recognize the following as a comment and  fails  which is not correct:
String failStr= ("/*/data");

ftomassetti:
Are you suggesting that "/*/" should be recognized as a valid comment?

According to http://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.7

I think it is not: the asterisk part of the start of the comment cannot be "reused" as part of the end of the comment. Am I wrong in interpreting your problem? Am I wrong in my interpretation of the language specs?

SmiddyPence:
Opposite way round.  (unverified yet)

It is not a valid comment; but the parser recognizes it as one.

ftomassetti:
My bad, I did not read correctly, ok, that is bad :)
Probably we should start by writing a failing test and then look in the comment parser.

SmiddyPence:
I'll write the test, you can fix it. Call it virtual pairing :wink:

ftomassetti:
Love that!

quarkie:
Yes it should not be recognized as a comment ^^ 
very fast reaction time  here  ;-)

ftomassetti:
We try to be fast, I think we always had the experience of reporting an issue and get an answer a long after it stopped to be useful :)
It is bad because users stop reporting problems, while we really need that: I would have never ran in this issue myself

quarkie:
i just saw that the version I am using is the old one from googlecode .-(. in the current trunk(github) the error might be fixed, as stringsare recognized.  sorry  for making you work on things that might be fixed :-(

ftomassetti:
No problem, it could be worthy to add that test anyway, please keep reporting issues, in the worst case we will just realize how awesome the project already is :D

ptitjes:
@SmiddyPence @ftomassetti Did one of you finally write a test ?

ftomassetti:
Just sent a pull request with the test (see #130). Should we close this issue?

PR #130, opened 2015-04-06
Adding test to verify that /*/ is not recognized as a comment

ftomassetti:
Following #95 we add a test to prevent possible future regressions.


ptitjes:
Appart from formating. +1

ftomassetti:
Formatting should be fixed now

ptitjes:
Indeed! +1

PR #132, opened 2015-04-15
Fix hashCode to differentiate between Nodes that have the same toString

nburles:
I don't think these Nodes compare as equal, and so they should not hash equally either.  I've added information such as line numbers to the hashCode method to fix this.

All the tests continue to pass.

If I'm wrong and they do compare as equal... maybe they shouldn't?  They're completely different Nodes in completely different places in the source...?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -250,8 +250,14 @@ public abstract class Node {
     hash = addToHashCode(hash, hashMultiplier, this.getBeginColumn());
     hash = addToHashCode(hash, hashMultiplier, this.getEndLine());
     hash = addToHashCode(hash, hashMultiplier, this.getEndColumn());
-    //hash = addToHashCode(hash, hashMultiplier, this.getData().hashCode());
-    hash = addToHashCode(hash, hashMultiplier, this.getComment().hashCode());
+    Object data = this.getData();
+    if (null != data) {
+      hash = addToHashCode(hash, hashMultiplier, data.hashCode());
+    }
+    Comment comment = this.getComment();
+    if (null != comment) {
+      hash = addToHashCode(hash, hashMultiplier, comment.hashCode());
+    }
 		return hash;
 	}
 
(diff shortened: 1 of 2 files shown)


ptitjes:
Actually we do not test for node's position in equality. See https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java#L123. Indeed, what we test for is structural equality. And thus structurally equal nodes coming from different compilation units and different locations are equals. In my opinion, it makes sense.

I guess we could make the base methods protected (instead of private) so that one can override them to implement more specific use cases such as yours. @nburles, would that fit your bill ?

For the record, as to hash code implementation, I'd feel better off with an approach similar to https://github.com/ptitjes/javaparser/commit/71ee31cdb2bdcb46ccaed8648393db0f548fa774. That particular visitor does not test for nodes' position but that would be easy to add by overriding HashCodeVisitor.nodeHashCode(Node) if it were protected (instead of private).

nburles:
Fine by me, I just needed it to work sufficiently for a paper I'm working
on - we were storing Nodes as the key of a HashMap... and getting confused
by a bug manifesting later on that shouldn't have been possible :-)

(And I hadn't seen ptitjes' patch - will that be applied at some point?)

ptitjes:
That patch will, if accepted, go in 3.0.

But we can make EqualVisitor methods protected or add an extension mechanism.

ftomassetti:
I am not sure if comparing for structural equality is enough to consider two nodes equals. If they are in different parts of the code they should be different (different line, column, parent). Also because nodes have different meaning in different positions, because of scope rules.

So, I agree with this approach (but I understand it should go in 3.0). I had to use workarounds when using JavaParser because of its concept of equality.

@nburles I hope your work on the paper goes well, I would be curious to read it and maybe we should have a section of papers which were written using JavaParser. I think I have used it in a few.

matozoid:
@ptitjes - would you like to open a pull request?

ptitjes:
Well, apart from the private methods that are hand-written, this is generated code from the AST and is a bit out-dated. I think you'd rather regenerate one based on the current AST (and keep your generator at hand like I do with https://github.com/ptitjes/jlato-bootstrap) for when the AST changes.
I don't mind at all the result looks similar as I did not invent anything in that commit. :p

ftomassetti:
Should we close this one?

matozoid:
Don't know. I can't really figure out what the resolution is :)

matozoid:
Will eventually be fixed by https://github.com/javaparser/javaparser/issues/353

Issue #134, opened 2015-05-05
Tailing semicolon cause ParseException


xkommando:
While parsing Android 19 source, I encountered ParseException caused by double semicolons in "import" statements.
The class is `com.android.mediadump.VideoDumpActivity` , and it can be found here https://android.googlesource.com/platform/frameworks/base/+/cd92588/media/tests/MediaDump/src/com/android/mediadump/VideoDumpActivity.java
the problem is caused by the double semicolons at line 22:

```
import android.content.Context;
import android.content.DialogInterface;; // line 22, two semicolons here
import android.content.Intent;
import android.content.SharedPreferences;
```

and the exception is:

```
Exception in thread "main" com.github.javaparser.ParseException: Encountered " "import" "import "" at line 23, column 1.
Was expecting one of:
    <EOF> 
    "abstract" ...
    "class" ...
    "enum" ...
...
```

Regards.

ftomassetti:
Thank you for reporting this. I think we should:
1. find out if the language specifications consider valid to use an orphan semicolon outside methods. After all it is an empty statement which probably does not make sense there
2. probably, whatever it is legit code or not  we should support it anyway

ftomassetti:
It took a while but it should have been fixed by #135 so I am going to close this one, please do not hesitate to reopen if the problem is still present

Issue #22, opened 2013-09-16
Change the AST nodes or create new ones to modify source code like refactoring.


matozoid:
This was Júlio's plan with the project.

ftomassetti:
Well, I think we could do that by dumping back the source code. It is already possible but original formatting is lost (we do not store whitespaces) while comments are not lost but they are re-arranged. I think we spenti energy on comments attribution also for this reason, to support refactoring.

Maybe we should think about additional methods that could help with refactoring and provide some examples in the documentation.

What do you think?

cciollaro:
Hi, I think preserving whitespace would be a great feature to have. Júlio stated back on the google page that the DumpVisitor needs to be updated to respect token positions. As I can see, ASTNodes already store row and column information.

ftomassetti:
Yes but if you do refactoring how that affects the whitespaces? I think it becomes quite tricky...

cciollaro:
That's a good point. If you change a node's length then you'll likely have to adjust the position of every node after it by that amount. e.g. if you add a line to a method, you need to update the position of every node after that node by adding 1 to the line positions.

I think another possibility is to give whitespace its own token.

Do either of those sound feasible?

ftomassetti:
yes but if you add one line it should have the indentation of the line above or below (what if they are different?). 

Now, suppose you refactor this code:

'''
int c = foo(bla bla bla,
                                     a+b);
'''

You extract the second parameter of foo with something else, while moving the second parameter outside, like that:

'''
int c = foo(bla bla bla, 10);
int d = a+b;
'''

What about the space to put in fron of 10? And the space to put in front of a+b?
I would not know how to manage the whitespace in this case.

ftomassetti:
To be clear: I think it would be VERY nice to have refactoring preserving indentation, I just think that in practice it is not as easy as it seems.

We can give it a try however.

SmiddyPence:
I think it will have to be template based.  So if you commit to having a class refactored, you are committing to the white-space conforming a determined standard - (and probably applied to the rest of the class as well not just the refactored section).

prior:
I'm trying to use javaparser to auto generate pull requests, so I'm trying to keep my diff's as tight as possible and not obliterate different peoples' projects' java styles -- unfortunately, as is-- I can't use.

[...]

PR #135, opened 2015-05-06
Fix issue 134: Tailing semicolon cause ParseException

ftomassetti:
See #134 for details.

The solution includes tests.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
@@ -27,7 +27,7 @@ import com.github.javaparser.ast.visitor.VoidVisitor;
 
 /**
  * <p>
- * This class represents a import declaration. Imports are optional for the
+ * This class represents a import declaration or an empty import declaration. Imports are optional for the
  * {@link CompilationUnit}.
  * </p>
  * The ImportDeclaration is constructed following the syntax:<br>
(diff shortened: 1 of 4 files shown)


ptitjes:
I would say that we need a node to represent the semi-colon, in the same way it is done for EmptyXXX nodes. Else the semi-colon is dropped on dump.

Also does your code handles semi-colons just before the first import statement (that is after package decl) ?

ftomassetti:
I am not sure we want a new node for that. I understand having `EmptyStatement` makes sense but `EmptyImport` probably not so much. While an `EmptyStatement` has a role depending on its position in the code, an `EmptyImport` would be always redundant.

I think we are talking about a pretty rare case (we have parsed tons of code and never encountered this problem so far). I do not know if we want to introduce in the AST a phantom node representing this spurious semicolon. I could not find a reference to it in the Language Specifics document.  

I think my code handles semi-colons also before the first import statement but I am not sure if it is what we want (so I did not write a test for that case: I am testing for extra semi-colons between two imports or after the last one).

ptitjes:
I was not speaking about EmptyStmt, but about EmptyTypeDecl and EmptyMemberDecl.

I did check this WE and indeed this is not specified in the JLS, whereas Empty(Type|Member)Decl are (explicitly stated as a legacy for coders coming from C and that should not be used for new code, btw).

However, despite that is not specified, the IDEs (and I guess the Java compiler) accepts (and emit a warning) dangling semi-colons both after package decl and import decl (as it is done in between and after type declarations). I think this is consistent and I can't find any rational reasons to introduce support for dangling semi-colons in between and after import declarations _and omitting after package decl_. Please, we need to stay consistent.

As of a representation of it in the AST, we are parsing Java code and not interpreting. If we accept a piece of code, that must have a representation in a way or another. If we do not, we will break hash, equality, and the (not explicitly stated) we-dump-what-we-parsed contracts.

ftomassetti:
`Please, we need to stay consistent.`

sure, the code does that already, I can add an additional test for that specific case

`As of a representation of it in the AST, we are parsing Java code and not interpreting. If we accept a piece of code, that must have a representation in a way or another.`

Well, it is not necessarily true the _abstract_ of AST tells us that we should represent meaningful information. This is way spaces or comments are normally not represented and a dangling semi-colon that has no effect whatsoever and it is basically irrelevant does not have necessarily to be included in the AST. I think the we-dump-what-we-parsed thing should not pollute the AST which should contain only information relevant and meaningful. I am not sure this is the case of the dangling semicolon because I do not understand what is representing in this context.

matozoid:
@ptitjes @ftomassetti - can you please continue the discussion or close this?

ptitjes:
I think there should be an AST node for any input element that are not white-spaces or comments, but I don't mind anymore ;)

ftomassetti:
I think this point is not so much technical, it is instead a matter of interpreting what should or should not go in the AST.

My point is that there are semicolons with no meaning, for example extra semicolons among the import declarations.

I do not think we can reach an agreement without extra opinions. I would suggest that we wait for comments from others and if we do not receive them we close this PR. The problem is that #134 needs a solution anyway.

@matozoid @SmiddyPence @sebastiankirsch and any other interested, any thought on this one?

matozoid:
As a user I would be annoyed with too much and too little information in the AST, but the "too much" option is the only one that I, as a user, could fix. I can't add information that JavaParser has thrown away. I vote keeping as much info as possible.... but I'm not a user ;)

ftomassetti:
Mmm, I am convinced from the explanation of @matozoid . Maybe we can add some sort of marker for nodes which are just syntactical so that users could skip them easily. I will update this PR.

matozoid:
This is quite connected to #22, I guess.

ftomassetti:
I added nodes for the empty import declarations. Instead of creating a new class I just adapted `ImportDeclaration` to represent also empty import declarations. I considered two alternatives:
- having a specific class for EmptyImportDeclaration but in that case we would need a separate collection for holding empty import declarations in the `CompilationUnit`. At that point it would be hard to have an ordered list of normal and empty import declarations as they appear in the code.
- I considered creating an `AbstractImportDeclaration` from which to derive the `EmptyImportDeclaration` and the `NormalImportDeclaration`. However at that point the collection in `CompilationUnit` should have been a list of `AbstractImportDeclaration` and each time the user wanted to access an element it should have casted to a `NormalImportDeclaration` to use it. Sure, we could add some method like `asNormalImportDeclaration` on `AbstractImportDeclaration` but that would be not great

To me this seems the lesser evil but I am absolutely open to suggestions for alternative ways to represent the empty import.

SmiddyPence:
Sorry got to this one late.  Personally I think ditching it as noise would be acceptable, but I understand the argument to keep it consistent with other empty elements.

+1 to ship in current incarnation.  I wouldn't want to see additional classes introduced to support such a narrow concept.

matozoid:
People seem reasonably positive. Rebase & pull :)

ftomassetti:
rebased, tests running

Issue #139, opened 2015-05-14
Document the meaning of arraysAnnotations in ReferenceType


ftomassetti:
[ReferenceType](https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/type/ReferenceType.java) has one field named `arrayAnnotations`. I think we should document it, because I cannot remember what it does without looking at the grammar. Is that the weird thing deprecated since Java 1.1?

jgesser:
​​
It is really weird. I didn't knew it was possible, I had to look at the JLS
to be sure. Actually it is new, since Java 1.8 it is possible to declare
Annotations between the dimensions of an array declaration, as can be seen
below.

public class Test {

```
@Target(ElementType.TYPE_USE)
@interface Ann {

}

// HERE
int @Ann [] @Ann [] array;
```

}

Maybe a new type of node could be created to carry the real positions of
these Annotations.
Instead of ReferenceType have a list of Annotations and the arrayCount
field, it would have just a list of this new node: Dimension.

R​egards​,

Julio Vilmar Gesser

2015-05-14 21:28 GMT+01:00 Federico Tomassetti notifications@github.com:

ftomassetti:
First of all it is an honour to get an answer directly from Julio :)

Then yes, I agree that we should probably create a new Node representing the array modifier (`[]`) and the associated annotation, if any.

In the meantime we should document the meaning of the existing field.

What do you think?

jgesser:
Sure. Document the entire AST was an open issue from the original project
that I could never solve alone.

Atenciosamente,

Júlio Vilmar Gesser

2015-05-15 9:28 GMT+01:00 Federico Tomassetti notifications@github.com:

ptitjes:
Guys, I have that as a proposition already: https://github.com/javaparser/javaparser/issues/112
:p

ftomassetti:
While I agree that your suggestion makes sense, what I am proposing is a smaller change: just create a new node representing the array modifier (`[]`) and the corresponding annotations.

For now I will send a patch for document the existing behavior.

ptitjes:
@ftomassetti Well this is exactly what I proposed: "ArrayType would denote only one level of array and hold the annotations for the (single-level) array type."

ftomassetti:
I do not think so: you are proposing to make ArrayType extends ReferenceType, which could make sense imho but it is a major change. I am instead proposing of creating a class, let's call it ArrayModifier:

```
[...]

PR #142, opened 2015-05-15
Document arraysAnnotations in ReferenceType

ftomassetti:
The field `arraysAnnotations` of `ReferenceType` is not obvious and to understand its meaning it is necessary to dig into the grammar.

It is used to collect annotations on the array modifiers, like the annotations used in this example:

`Foo @Ann1 [] [] @Ann2 [] foo;`

This patch is just about adding a Javadoc for the getter of the field and a reference in the setter.

See issue #139

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/type/ReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/type/ReferenceType.java
@@ -92,7 +92,7 @@ public final class ReferenceType extends Type {
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ptitjes:
We should also document the order the ArrayDim are met in arrayAnnotations.

ftomassetti:
Good point. Let's do that.

ptitjes:
+1 for me

ftomassetti:
Thanks!

ptitjes:
Thank you!

ftomassetti:
It is just documentation and it has been reviewed from @ptitjes . Merging

PR #146, opened 2015-05-29
Fixed creating Position using beginOf() or endOf() creator methods

lgawin:
Line was wrongly initialized


ftomassetti:
Seems fine to me, I will go on and merge it because it seems an obvious fix.
Thank you so much @lgawin for fixing this!

matozoid:
Does this need a fix in 2.1 as well?

ftomassetti:
Well, I think it would be useful to do so, do you want me to create a 2.1 branch and put this commit in that branch? There are another couple of small fixes in the pipeline/already committed that could all end up in a 2.1.1 release

matozoid:
If it's little work, go ahead :) If it becomes a lot of work, we have to consider if it's worth it, or if we just focus on 3.0

ftomassetti:
I created a 2.1 development branch, I will write on gitter also

Issue #149, opened 2015-06-14
Parsing Failed


akkinenirajesh:
Failed to parse any java file with typed lambda params. (code copied from java.time.chrono.AbstractChronology.java)

```
static final Comparator<ChronoLocalDate> DATE_ORDER =
    (Comparator<ChronoLocalDate> & Serializable) (date1, date2) -> {
        return Long.compare(date1.toEpochDay(), date2.toEpochDay());
    };
```

ftomassetti:
The problem is with the "combined cast": if you remove & Serializable it can be parsed. I started writing a test case and looking into the grammar for a solution.

akkinenirajesh:
Yes, I noticed. I am new to JavaCC so taking time for me to understand where to change. It seems to a new kind of Casting 
    static final Consumer<String> x=(Consumer<String> & Serializable)null; 
this seems to be valid in Java 8. Just tested in Eclipse. So changes should be in Casting Expression I believe.

ftomassetti:
Yes, I also think so. In my opinion the relevant piece of the grammar is:

```
Expression CastExpression():
{
    Expression ret;
    Type type;
    int line;
    int column;
    AnnotationExpr ann;
    List annotations = null;
}
{
  "(" {line=token.beginLine; column=token.beginColumn;}
   (ann = Annotation()   { annotations = add(annotations, ann);})*
  (
      LOOKAHEAD(2)
      type = PrimitiveType() ")" ret = UnaryExpression() { type.setAnnotations(annotations); ret = new CastExpr(line, column, token.endLine, token.endColumn,type, ret); }
    |
      type = ReferenceType() ")" ret = UnaryExpressionNotPlusMinus() { type.setAnnotations(annotations); ret = new CastExpr(line, column, token.endLine, token.endColumn,type, ret); }
  )
  { return ret; }
}
```

The problem is how to find a sensible representation of this new cast without changing the metamodel of the currently existing CastExpression.

akkinenirajesh:
How does "create a new CombinedCastExpr that will contain one or more CastExpr " sound ?
In this way we don't have to break any existing code.

akkinenirajesh:
Sorry, I think my idea is stupid. I did not think through...

ftomassetti:
With this change it parses the example given:

```
Expression CastExpression():
{
    Expression ret;
    Type type;
    int line;
    int column;
    AnnotationExpr ann;
    Type st;
    List annotations = null;
[...]

Issue #112, opened 2015-03-02
Remove ReferenceType.arrayCount in favor of ArrayType


ptitjes:
I don't think this can be done in a forward-compatible way.

ArrayType would denote only one level of array and hold the annotations for the (single-level) array type.

```
public final class ArrayType extends ReferenceType {

    private Type componentType;

    public ArrayType(int beginLine, int beginColumn, int endLine, int endColumn,
                     Type componentType, List<AnnotationExpr> annotations) {
        super(beginLine, beginColumn, endLine, endColumn, annotations);
        setComponentType(type);
    }
}
public abstract class ReferenceType extends Type {

    public ReferenceType(int beginLine, int beginColumn, int endLine, int endColumn,
                         List<AnnotationExpr> annotations) {
        super(beginLine, beginColumn, endLine, endColumn, annotations);
    }
}
```

instead of

```
public final class ReferenceType extends Type {

    private Type type;
    private int arrayCount;
    private List<List<AnnotationExpr>> arraysAnnotations;

    public ReferenceType(int beginLine, int beginColumn, int endLine, int endColumn,
                         Type type, int arrayCount, List<AnnotationExpr> annotations,
                         List<List<AnnotationExpr>> arraysAnnotations) {
        super(beginLine, beginColumn, endLine, endColumn, annotations);
        setType(type);
        setArrayCount(arrayCount);
        this.arraysAnnotations = arraysAnnotations;
    }
}
```

ptitjes:
OK, there are some more thoughts to put into this!

I just spotted a problem with my previous understanding of the specification and thus the implementation I did in #156.

So at the top of page 314 in section 9.7.4 ("Where annotations may appear") of the JLS, they give an example:

> `@C int @A [] @B [] f;`
> 
> @A applies to the array type int[][] , @B applies to its component type int[] , and @C
> applies to the element type int .

Thus the parsing I do for the array type encapsulation is wrong. What I implemented parses for the above expression:
`ArrayType(@B, ArrayType(@A, PrimitiveType(@C, Integer))`
where it should parse:
`ArrayType(@A, ArrayType(@B, PrimitiveType(@C, Integer))`

Also, note that square bracket pairs are also allowed (and annotated - see JLS chapter on Arrays) after the variable name in a variable declarator id.

[...]

Issue #111, opened 2015-03-02
Kill MultiTypeParameter and BaseParameter and introduce UnionType


ptitjes:
The idea is to make BaseParameter be Parameter again and introduce UnionType that wraps a list of Types.

The rational behind that is that it is more close to the semantic model. Also, having an object for the union type enables to consistently attribute (meaning put in the data of the AST node) a semantic type object to each syntactic type object. There is no way to associate a semantic type object to the type of a 'MultiTypeParameter' parameter as this object represents both the parameter and its type.

This can almost be done in forward-compatible way (thus deprecating the old form at first):

```
@Deprecated
class MultiTypeParameter extends Parameter {
  public MultiTypeParameter(int modifiers, List<AnnotationExpr> annotations, List<Type> types, VariableDeclaratorId id) {
    super(modifiers, annotations, new UnionType(types), id);
  }
}
class Parameter extends BaseParameter {
  public MultiTypeParameter(int modifiers, List<AnnotationExpr> annotations, Type type, VariableDeclaratorId id) {
    super(modifiers, annotations, type, id);
  }
}
@Deprecated
class BaseParameter {
  public MultiTypeParameter(int modifiers, List<AnnotationExpr> annotations, Type type, VariableDeclaratorId id) {
    // ...
  }
}

public abstract class VoidVisitorAdapter<A> implements VoidVisitor<A> {

    @Override public void visit(final MultiTypeParameter n, final A arg) {
        visit((Parameter)n, arg)
    }
}
```

ptitjes:
If everyone is OK with that, I add this to 3.0 milestone.

ftomassetti:
If we created the TypeBound, could we just remove the `MultiTypeParameter`?

ptitjes:
You mean UnionType right ? (search the jls for "UnionType") If yes, then indeed you can remove MultiTypeParameter.

matozoid:
So, considering #185 and #149, what is still to be done here?

matozoid:
Looks like all that was left to do was a clean up of unused nodes.

Issue #107, opened 2015-02-23
Annotations on types are not parsed and dumped correctly


ptitjes:
See failing test case:
https://github.com/ptitjes/javaparser/commit/beffe836060823820eed0c4e2fcdac99545a0486

```
@Target(ElementType.TYPE_USE)
@interface Ann1 {
}

@Target(ElementType.TYPE_USE)
@interface Ann2 {
}

class B {
    class C {
    }

    class A extends @Ann1 B.@Ann2 C {
    }
}
```

is valid Java 8 code.

It parses as :

```
class A extends @Ann1 B.C {
}
```

and if you instanciate the ast manually, when dumped it produces

```
class A extends @Ann2 @Ann1 B.C {
}
```

ptitjes:
https://github.com/ptitjes/javaparser/tree/zz-issue-107

ptitjes:
If everyone is OK with that, I add this to 3.0 milestone.

ftomassetti:
Hi @ptitjes . Do you have a patch for this? Do you want to still contribute one or do you prefer not to because of the license change?

ptitjes:
Well, I don't mind helping users and give some advises from time to time when I can, but I won't contribute code because of the license change and because I prefer to use my spare time for JLaTo...

The stuff in the '3.0 type hierarchy...' contains nothing spectacular and you can easily do something similar by yourself. And BTW it has a big performance problem because of the way the lookahead is done currently for cast expressions in JavaParser.

PR #150, opened 2015-06-14
Issue 149: parsing combined casts

ftomassetti:
In issue 149 has been reported one issue with a new form of casts (introduced in Java 8) that we were not supporting.

Example:

```
Object foo = (Cloneable & Serializable)bar;
```

I suggest one solution in which the existing node type `CastExpression` get an additional fields named `secondaryTypes` which collect all types but the first. We could of course add a method to get allTypes (so the first one and the secondary ones in one collection).

Otherwise we could add a new type of expression (CombinedCastExpression) and maybe rename the existing one to SimpleCastExpression. Maybe both types could inherited by an abstract type named CastExpression.

Suggestions?

### DIFF
--- a/javaparser-core/src/main/javacc/java_1_8.jj
+++ b/javaparser-core/src/main/javacc/java_1_8.jj
@@ -2256,7 +2256,9 @@ Expression CastExpression():
 	int line;
 	int column;
 	AnnotationExpr ann;
+	Type st;
 	List annotations = null;
+	List secondaryTypes = null;
 }
 {
   "(" {line=token.beginLine; column=token.beginColumn;}
@@ -2265,7 +2267,7 @@ Expression CastExpression():
   	  LOOKAHEAD(2)
   	  type = PrimitiveType() ")" ret = UnaryExpression() { type.setAnnotations(annotations); ret = new CastExpr(line, column, token.endLine, token.endColumn,type, ret); }
   	|
-  	  type = ReferenceType() ")" ret = UnaryExpressionNotPlusMinus() { type.setAnnotations(annotations); ret = new CastExpr(line, column, token.endLine, token.endColumn,type, ret); }
+  	  type = ReferenceType() ("&" st = ReferenceType() { secondaryTypes = add(secondaryTypes, st);} )* ")" ret = UnaryExpressionNotPlusMinus() { type.setAnnotations(annotations); ret = new CastExpr(line, column, token.endLine, token.endColumn,type, ret); }
   )
   { return ret; }
 }
--- a/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
+++ b/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
@@ -255,3 +255,16 @@ interface MyInterface {
 Then method 1 class 1 is a default method
 Then method 2 class 1 is not a default method
 Then all nodes refer to their parent
+
+
+Scenario: simple cast on lambda expression can be parsed
+
+Given a CompilationUnit
+When the following source is parsed:
+class A {
+    static final Comparator<ChronoLocalDate> DATE_ORDER =
+        (Comparator<ChronoLocalDate>) (date1, date2) -> {
+            return Long.compare(date1.toEpochDay(), date2.toEpochDay());
+        };
+}
+Then all nodes refer to their parent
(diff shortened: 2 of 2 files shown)


ptitjes:
Hi @ftomassetti. Some things that you might want to take in account:
- there is already a production that does something similar, the TypeBound production. In fact, the «& InterfaceType» part in the spec is in a shared production AdditionalBound shared by CastExpression (JLS page 551) and TypeBound (JLS page 57). I think we should not deviate much from the spec, so that might be a good thing to factor out that production too in our grammar. (JavaCC productions can take arguments - to pass the list of types and list of annotations - see below)
- according to the spec, you should parse annotations for each of the additional bounds (like in TypeBound - which you really might want to factor out)
- as for the AST, i'd prefer you introduce a IntersectionType ast node to put together the main type and the additional bounds. That is the name of it according to the spec (JLS page 551). Another possibility is to make the type field be a list of types - also suggested by the spec (JLS page 551). But I really don't like Simple/CombinedCastExpr or additional 'secondaryTypes' field.

rpau:
Hi @ftomassetti and @ptitjes ,

I would create the "IntersectionType", called preferably "BoundType" (sounds better). If the CastExpr interface changes (having a list of types), the users of this library would have so much things to change to make their code compile. If we just add a new Type, nothing changes for them if they are working with code written in Java 6 o Java 7.

The approach of having a super class "AbstractCastExpr" and thow children types ("CastExpr") and "MultipleCastExpr" could also work, but it seems less natural.

ftomassetti:
So, would you have:
- `(Foo)` to be parsed as it is now (CastExpression with a field name type of type ReferenceType)
- `(Foo & Bar)` to be parsed as a CastExpression with a field name type of type IntersectionType?

In this case the problem would be where to put the annotations: currently the annotations stay in the CastExpression, while we would need to have IntersectionType contain the annotations to all types but the first. At that point we should probably make it contain also the annotation on the first type but that would be changing the current behavior.

@rpau this initial solution I proposed does not change CastExpr: we just add a field named secondaryTypes and I would suggest to add another list as a projection of type + secondaryTypes but none of the existing methods of CastExpr would change.

rpau:
Hi @ftomassetti 

You are right that the part of the annotations becomes ambiguous. What I would suggest you is accumulating the annotations that receives the included types in the cast expression. What do you think?

@ftomassetti sorry If I understood different your initial solution. It can also work, but sounds more artificial for me (but it is just an opinion :) ), because in fact, when you are using multiple types in the cast expression, you are referencing all those types which are subclasses of all the referenced types (the same than having just once), but the AST structure is different. I guess that if someone wants to make some semantic analysis with the AST, the people would like to attach information in a kind of type node instead of the cast expression.

ptitjes:
@rpau @ftomassetti I'm strongly against putting the annotations in the expression. The annotations are relative to a type. Also "accumulating" the annotations will give later trouble as we have for ReferenceType (see #112) and for annotations in inheritance lists (see #107). Let's not do again what we already have to fix.

As for evolution, we have a 3.0 in the process and that is what it is for. I think we can cook an upgrade guide along the way. So I favor:
A composite IntersectionType (that is the name of it in the spec, sorry to insist, see JLS page 70 - and also that is the name given to that thing in type-theory) that inherits from Type and use the Type.annotations field of each of the children types to store their relative annotations.

The Type hierarchy needs reworking anyway and we will also introduce UnionType and ArrayType along with IntersectionType (that is one important point of 3.0, see #112, #107 and #111). If you look at the spec all those three are explicitly mentioned as possible types in the type system. Also you can look at the javax.lang.model.type which actually models that.

rpau:
@ptitjes +1 for removing annotations from the CastExpression. My project, javalang does not stores annotations in the CastExpression.

@ptitjes  +1 for IntersectionType name. I have checked the specification.

ftomassetti:
@rpau thank you for your comments, I see the point why having a sort of combined type would help several type of analysis.

The point is that the current 2.1 parser cannot parse Java 8 code correctly and I think we should fix it being as back compatible as possible. While we can look for more elegant solutions in 3.0 (for which we do not have the strict constraint of backwards compatibility) in 2.1. the first rule is to not break existing usages while supporting this kind of cast.

This is my opinion that said I will drop this PR and let someone one fix it if he/she wants: every time I started a discussion with @ptitjes my PR ended up being stalled so I will just give up :)

I would leave this PR open just to keep the conversation and let others fix this if they want too, please feel free to close this PR if no one is interested in picking it up.

ptitjes:
I don't mind that you fix it for 2.1. I just think 2.1 is a bug fix branch. If bug fixing requires introducing new things then let's do it in a clean way in 3.0.

matozoid:
Can someone open an issue for this, and close this PR?

ptitjes:
The issue is #149.

ftomassetti:
I will close this issue: at this point we can go for a compatibility-breaking solution in 3.0

PR #151, opened 2015-06-14
issue 148: add additional test to verify we dump correctly orphan com…

ftomassetti:
…ments in a for loop

The issue reported a problem with has been solved with recent releases of JavaParser. This PR is just about adding an additional test case to assure we dump correctly orphan comments inside for statements.


SmiddyPence:
+1

ftomassetti:
This had just a test and it has been approved by our test expert, so I merge it

Issue #152, opened 2015-06-15
Parse Failed due to ConditionalExpr in lambda


akkinenirajesh:
Example code (java.util.function.Pridicate)

```
static <T> Predicate<T> isEqual(Object targetRef) {
    return (null == targetRef)? Objects::isNull : object -> targetRef.equals(object);
}
```

Above code fails to parse as generateLambda method does not consider the case where ret is ConditionalExpr

rpau:
Solved in https://github.com/javaparser/javaparser/pull/153

ftomassetti:
A PR has been merged for this, I would close the issue. Feel free to reopen it if you still experience this problem.

PR #153, opened 2015-06-15
issue #152: Parse Failed due to ConditionalExpr in lambda

rpau:
Solved. It was a bug in the Expression() rule


matozoid:
Thank you!

PR #154, opened 2015-06-22
Javaparse2jctree

pcgomes:
I have added the files that convert a Java Parser AST into a JCTree AST into the fork javaparser2jctree.

The files and it utility are the following:

com/github/javaparser/jctree/Main.java: Entry class that glue all parts. Shows how to make translation, until the code generation.
com/github/javaparser/jctree/ast/*.java: Node types for the JCTree AST, extended to store comments (I call then AJCTree).
com/github/javaparser/jctree/ast/visitor/APretty.java: Visitor that prints an AJCTree AST, including the comments.
com/github/javaparser/jctree/ast/visitor/PrintAstVisitor.java: Visitor that dumps an AJCTree AST, printing the node type before its content.
com/github/javaparser/ast/visitor/ASTDumpVisitor.java: Visitor that dumps a Java Parser AST, printing the node type before its content.
com/github/javaparser/ast/visitor/ComplyToJCVisitor.java: Visitor that formats the Java Parser AST, to comply to the JCTree one.
com/github/javaparser/ast/visitorJavaParser2JCTree.java: Visitor that translated the Java Parser AST to the AJCTree AST.


javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ASTDumpVisitor.java
```diff
@@ -0,0 +1,1741 @@
...
+import java.util.List;
+
+import static com.github.javaparser.ast.internal.Utils.isNullOrEmpty;
+
+/**
+ * Dumps the AST to formatted Java source code.
+ *
+ * @author Julio Vilmar Gesser
+ */
+public class ASTDumpVisitor implements VoidVisitor<Object> {
```
ftomassetti, line 90: I am not sure why this appear to be a new file: you did not add this file right?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/JavaParser2JCTree.java
```diff
@@ -0,0 +1,1763 @@
...
+   private Name.Table names;
+
+   public JavaParser2JCTree(Context context) {
+      make = TreeMaker.instance(context);
+      names = Names.instance(context).table;
+   }
+
+   @Override
+      public JCTree visit(final AnnotationDeclaration n, final Object arg) {
+         /* TODO - Annotations are not currently supported
```
ftomassetti, line 109: We should probably fix this before accepting the pull request

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/AJCAnnotation.java
```diff
@@ -0,0 +1,38 @@
+package com.github.javaparser.jctree.ast;
+
+import com.sun.tools.javac.tree.JCTree.*;
+import com.sun.tools.javac.tree.JCTree;
+import com.sun.tools.javac.util.List;
+//import com.sun.tools.javac.util.Name;
```
ftomassetti, line 6: Please remove

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/AJCAnnotation.java
```diff
@@ -0,0 +1,38 @@
...
+//import com.sun.tools.javac.code.Symbol;
+//import com.sun.tools.javac.code.Scope.ImportScope;
+//import com.sun.tools.javac.code.BoundKind;
+
+public class AJCAnnotation extends JCAnnotation implements JavaParserComments {
```
ftomassetti, line 11: Can you rename the class or add a comment to make clearer its role?

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/AJCArrayAccess.java
```diff
@@ -0,0 +1,38 @@
+package com.github.javaparser.jctree.ast;
+
+import com.sun.tools.javac.tree.JCTree.*;
+//import import com.sun.tools.javac.tree.JCTree;
```
ftomassetti, line 4: please remove

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/visitor/APretty.java
```diff
@@ -0,0 +1,1172 @@
...
+public class APretty extends Pretty {
+   public APretty (Writer out, boolean sourceOutput) {
+      super(out,sourceOutput);
+   }
+   public void visitTopLevel(JCCompilationUnit that) {
+      if (((AJCCompilationUnit) that).hasComment()) {
+         try {
+            println();
+            print("/*" + ((AJCCompilationUnit) that).getComment() + "*/");
+         } catch (IOException e) {
```
ftomassetti, line 22: Why are we treating exceptions in this way?

pom.xml
```diff
@@ -25,6 +25,32 @@
...
+        <property>
+          <name>java.vendor</name>
+          <value>Oracle Corporation</value>
+        </property>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>com.sun</groupId>
+          <artifactId>tools</artifactId>
+          <version>1.7</version>
```
ftomassetti, line 41: I think we are still supporting JDK 6 and I am not sure if we can insert this dependency

ftomassetti:
Thank you very much for sending this PR, in my opinion it is a very interesting addition. What do you think @matozoid  @sebastiankirsch @SmiddyPence and others?

I think there is some refinement work to do before merging this pull request. For example it would be absolutely great to have some tests.

Another aspect that we should pay attention to is the dependencies: so far JavaParser is compatible with every JDK 6, so we should not introduce dependencies that break this contract. If this is the case and it cannot be avoided maybe we should add this as a separate module (we can help in organizing it).

Another thing to check is the fact that the build is currently not completing successfully (see https://travis-ci.org/javaparser/javaparser/builds/67854479 for details).

For any doubts, questions and help we are here so don't be afraid of pestering us here or on gitter.

Thank you again for your work!

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ASTDumpVisitor.java
pcgomes, line 90: This file was copied from DumpVisitor, and altered. The different between DumpVisitor and ASTDumpVisitor is that in the later I have added a print of the AST node in the first line of the visitors' declaration. (E.g. printer.printLn("NameExpr"); )

Maybe this file has no use for Java Parser. You tell me.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/JavaParser2JCTree.java
pcgomes, line 109: Java annotations are not currently supported, as I have mentioned before. It was not a priority, since in my project I preserve the "per-node" comments.

I may implement the support for this soon, though.

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/AJCAnnotation.java
pcgomes, line 6: Do you mean to remove the commented "imports"?

No problem, I kept them so that, if I had to make massive changes into the files, all of them had a header with same number of lines.

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/AJCAnnotation.java
pcgomes, line 11: Sure, I can do both. The classes extend the JCTree AST nodes, so it preserve the comments per-node.

Do you have suggestions for the new name?

javaparser-core/src/main/java/com/github/javaparser/jctree/ast/visitor/APretty.java
pcgomes, line 22: If I remember correctly, the methods println() and print() from Pretty raise IOException.

But the catch blocks, with System.err.println have no specific reason. I may remove them.

pom.xml
pcgomes, line 41: This dependency was inserted just because I have adopted JDK7. If I have to change the AJC files to comply to JDK 6, this will also be reflected here.

Moreover, I have edited the pom.xml file poorly, given my lack of experience with maven. Probably there's a better way to add tools.jar to the javac's classpath.

pcgomes:
Thank you @ftomassetti for reviewing this. I make take care of most of the fixes, if assists me.

I have checked the compile errors, and I guess that most (if not all) are caused by different JDK versions. As I said above, I have been using JDK 7. I'll install JDK 6 and check if it's the case.

ftomassetti:
does it work for you compiling with maven?
You should just run

```
mvn clean test
```

I think that also if you have a more recent JDK it compiles to Java 6 because of the maven configuration. Still there is the fact you are accessing tools.jar which I am not sure if it is portable

pcgomes:
Hi, @ftomassetti 

It does work in my environment when compiling with maven. I may give you a shell in my machine, so you check the building yourself. Just send me your ssh-rsa key.

I understood the issue of adding the code for JDK7. I have seen in your personal page that you're already planning the JavaParser 3.0. Will it support JDK7? If so, I could add the code in the new codebase.

I don't understand what you mean about it not being portable. It's in present in any jdk installation. The only real concern I see is the versioning. Especially because the inner JCTree classes are not supported, and may change at any point. From the header of one of the files:

"This is NOT part of any supported API.
-  If you write code that depends on this, you do so at your own risk.
-  This code and its internal interfaces are subject to change or
-  deletion without notice."

Here, we might "whilelist" the OpenJDK 7 versions that are supported, if necessary. Though I don't think many things change within the same major release. I'll ask this with a friend working in Oracle with javac.

matozoid:
Alright, having this path magic for the tools.jar is brittle. It does not work on my machine, it does not work on the build server, it does not work at @ftomassetti 's machine, so it will not work at quite a few machines for people who want to work on the project, and that's a no go. Yes, we know how to fix it, but people should not have to fix things specifically for this project.

A nice approach would be to put this code in a separate, optional module or so, give that a normal dependency to tools.jar (which does not exist), then ask people to manually upload their tools.jar in their local maven repo. Or keep using path magic. I guess people who want to use this know what they are doing anyway :)

ptitjes:
@matozoid +1

ftomassetti:
hi @pcgomes I think what you have done could be useful but (as @matozoid said and @ptitjes agreed on) it is not 100% reliable for all users on all platforms. I think JavaParser has to be a bit conservative because many users rely on it and it could be puzzling if all of a sudden a piece of functionality is not available for everyone.

What do you think to release your code as an additional module to be used together with JavaParser? If you want I can help you with the setup process (pom.xml & the boring bits).

pcgomes:
Hi @ftomassetti , @matozoid and @ptitjes,

First of all, thanks for reviewing the suggestion. I agree that JavaParser has to be conservative, and the failure to find tools.jar in other platforms is definitely a no-go to go for the main tree.

Thus, I think that the suggestion for the additional module is good, and I will go for it. Your help with the boring bits is highly appreciated. Let's get started?

ftomassetti:
hi @pcgomes, sorry for the late reply, I moved country in these days and life get in the way...
first thing first: I think you should create a new repository. Once you have done it I can help you creating the initial pom.xml. Then we will just copy the code there and start writing some tests. I could send you Pull Requests that you could review before accepting. Does it work for you?

pcgomes:
Hi @ftomassetti , it's my turn to say I am sorry for the long time. I have been busy with PhD-related things, and could not focus on this.  I will create the new repository. Once I am done, I come back and tell you, so you can help me with other things.

ftomassetti:
perfect!

pcgomes:
Hi @ftomassetti ,

I have created the repository and have added the initial files. It contains the visitors for the javaparser AST, the JCTree's nodes, JCTree's visitors (to verify the transformation), and a Main class, to be the reference how to glue everything together.

They're available at: https://github.com/pcgomes/javaparser2jctree

Let me know the next steps!

ftomassetti:
sure, I will open issues in the project itself!

matozoid:
Has this moved to its own project? So, can this be closed?

ftomassetti:
Yes, there is now a separate project here: https://github.com/pcgomes/javaparser2jctree
We can close this one.

Thanks @pcgomes !

pcgomes:
@ftomassetti  Thank you for the support!

Issue #147, opened 2015-06-08
Add a clone method


ftomassetti:
Reading this question (http://stackoverflow.com/questions/30633535/copy-nodes-in-ast) it seems that it is not obvious how to clone methods. I think we could easily add a clone method wrapping the CloneVisitor.
What do you think?

ptitjes:
It makes sense. In addition to implement Object.clone(), Node could even implement the Cloneable interface. +1 for me.

ftomassetti:
This has been done in #161

PR #161, opened 2015-07-13
Issue 147: add clone method to Node

ftomassetti:
This is a proposal: let me know if it makes sense to you. The rationale is to make cloning easier to use. ~~While this solution means few lines of code it uses reflection.~~

~~\* We could pre-calculate the method of CloneVisitor to call at construction time~~
~~\* We could change approach and have different implementations of the clone method (each one calling a different method of CloneVisitor)~~

Reference: #147


ftomassetti:
Thanks to @ptitjes for fixing my broken proposal!

ftomassetti:
Any comment on this one?

SmiddyPence:
+1

Issue #144, opened 2015-05-19
License under an Apache license


glaforge:
GPL and LGPL licensed libraries is often problematic when integrating in other projects.
Could it be possible to also license the project under the Apache license, or even MIT / BSD?

ftomassetti:
Personally I am in favor of using the most liberal license but I think this would be a major decision and it would require every contributor to agree, but I will confess that I am completely ignorant on the possible legal requirements for changing license.
@matozoid @SmiddyPence @sebastiankirsch what do you think?

glaforge:
IANAL, but you just need the major contributors to agree, so just a handful of persons being okay with that.

ftomassetti:
@glaforge could you help us understand in which cases would help to have a different license? Do you have in mind anything specific?

melix:
See http://www.apache.org/legal/resolved.html (LGPL section) : LGPL cannot be included in any Apache licensed product.

glaforge:
As an example, if the Groovy project wanted to use this parser, it couldn't, because it's not allowed :-(

ftomassetti:
Thank you guys, it helps, let's see what the other guys involved in the project think.

glaforge:
Okie dokie.
The point is that if you want the project to be used widely, the current licenses are too restrictive :-(
The Groovy project might have a use for that parser, but unfortunately, with the license, we're not allowed to use it.

matozoid:
I <3 BSD. I'm not making any money off this, and even if someone copies everything, erases my name and sells it, well, it's better than this stuff being used less because of the license.

glaforge:
So that's a +1 vote :-)

matozoid:
Aye, and there's a discussion on gitter going on about licenses. Not sure if you can see it, but we're wondering if you could explain what the problem with the license is.

ftomassetti:
I agree 100% with @matozoid

glaforge:
I went on gitter but didn't see any discussion about licenses.
Basically, the link given by @melix gives the point of view of the lawyers from the Apache Software Foundation, which explains that Apache licensed projects can't use GPL-ed / LGPL-ed code:
http://www.apache.org/legal/resolved.html#category-x
I'm definitely not a lawyer though, but if you want the project to be as open as possible, letting anybody use it however they see fit, etc, a more business-friendly license is needed.
GPL/LGPL are quite viral and would force ASL licensed project to use the same license, which complicate things, especially for projects hosted at the Apache Foundation.

[...]

PR #166, opened 2015-07-25
Dual license

ftomassetti:
Applying the changes suggested by @SmiddyPence in issue #144.

This is intended as a proposal, please revise the wording!


ftomassetti:
Using this script we should be able to update the comment at the top of all source files, so if there is some rewording to do it will be easy. So please correct me!

https://gist.github.com/ftomassetti/400a2450e716e400184f

ftomassetti:
@ptitjes suggested that the GPL license should be included too because the LGPL refers to it. I added it as LICENSE.GPL

SmiddyPence:
I can't see the update to the grammar file. 

We also need a reference to the javalang or Walkmod project with respect to what @rpau requested.  I'm not sure what form this would take, perhaps @rpau can suggest?

ftomassetti:
- corrected license as suggested by @matozoid 
- added reference to Walkmod

ftomassetti:
I updated the comment in the grammar file, as suggested by @SmiddyPence

Issue #168, opened 2015-08-01
Wrong indentation for lambdas with BlockStmt as the body


markzyu:
I utilize this library to generate source files according to other java files.

Here is part of my generator related to the generation of lambda expressions:

``` Java
    LambdaExpr lambda = new LambdaExpr();
    lambda.setParametersEnclosed(true);
    lambda.setParameters(Arrays.asList(singleParam));
    lambda.setBody(lambdaBody);
    ASTHelper.addArgument(netThenCall, lambda);
    ASTHelper.addStmt(body, netThenCall);
```

Here is part of the generated java file (variables' and functions' names are obfuscated):

``` Java
    public XXX connect(boolean b, YYY argsProvider) {
        if (argsProvider != null)
            Func1(b, argsProvider);
        Func2((_XXX_type args)->{
    var1.connect(args.var2);
    return null;
});
        return this;
    }
```

ptitjes:
The problem comes from the `body.toString()` call here: http://git.io/vOG3T

You can easily fix this by replacing lines 1568 to 1573 by:

```
if (body instanceof ExpressionStmt) {
    // Print the expression directly
    ((ExpressionStmt) body).getExpression().accept(this, arg);
} else {
    body.accept(this, arg);
}
```

And voilà!

markzyu:
Thanks! I will fix this issue locally, but it would be much appreciated if the issue can be fixed in a later release.

ftomassetti:
@C4Phone let us know if we can help with your new PR

ptitjes:
@ftomassetti It was #170, and @matozoid already merged it.

matozoid:
Close this then?

ptitjes:
Closing you can :p

markzyu:
I am okay with closing this issue.

PR #169, opened 2015-08-02
Fix issue #168 + Tests + boxedType()/unboxedType() convenience methods

markzyu:
1. Fixed Issue #168, thanks to the comment by ptitjes
2. Added function `public ClassOrInterfaceType toClassType();` in `PrimitiveType` class. 
   - Although this is useful for me, maybe some other people is looking for a conversion in the oppsite direction...

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
@@ -29,7 +29,24 @@ import com.github.javaparser.ast.visitor.VoidVisitor;
 public final class PrimitiveType extends Type {
 
 	public enum Primitive {
-		Boolean, Char, Byte, Short, Int, Long, Float, Double
+		Boolean ("Boolean"),
+		Char    ("Character"),
+		Byte    ("Byte"),
+		Short   ("Short"),
+		Int     ("Integer"),
+		Long    ("Long"),
+		Float   ("Float"),
+		Double  ("Double");
+
+		private final ClassOrInterfaceType correspondingClassType;
+
+		public ClassOrInterfaceType toClassType() {
+			return correspondingClassType;
+		}
+
+		private Primitive(String nameOfCorrespondingClassType) {
+			correspondingClassType = new ClassOrInterfaceType(nameOfCorrespondingClassType);
+		}
 	}
 
 	private Primitive type;
@@ -57,6 +74,10 @@ public final class PrimitiveType extends Type {
 
 	public Primitive getType() {
 		return type;
+	}
+
+	public ClassOrInterfaceType toClassType() {
+		return type.toClassType();
 	}
 
 	public void setType(final Primitive type) {
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
@@ -20,6 +20,8 @@
 
 package com.github.javaparser.ast.type;
 
+import java.util.HashMap;
+
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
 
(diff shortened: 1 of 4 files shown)


ptitjes:
@C4Phone Nice! Good work!
Some comments though:
- I guess the issue fix would require a test in the JBehave test suite.
- Nice to have added spaces before and after the arrow.
- Would you mind renaming your `toClassType()` method to `toBoxedType()` ?
  Also, maybe you could add the converse as to be complete. I'd see two methods in ClassOrInterfaceType:

```
public boolean isBoxedType() { /* ... */ }
public PrimitiveType toUnboxedType() throws UnsupportedOperationException { /* ... */ }
```

What's your opinion ?

markzyu:
I agree with your opinion. Maybe I should close this pull request and open a new one when I have implemented the `toUnboxedtype()` function.

While I think I can implement the conversion between boxed type and unboxed type, I haven't used JBehave before, so adding a test will probably take me some time. But I will try :)

markzyu:
Hi, I just implemented the `toUnboxedType()` function and included some lambda expressions in dumping_scenarios.story.

javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
```diff
@@ -29,7 +31,32 @@
...
+		Boolean ("Boolean"),
+		Char    ("Character"),
+		Byte    ("Byte"),
+		Short   ("Short"),
+		Int     ("Integer"),
+		Long    ("Long"),
+		Float   ("Float"),
+		Double  ("Double");
+
+		static final HashMap<String, PrimitiveType.Primitive> unboxMap = new HashMap<String, PrimitiveType.Primitive>();
```
ptitjes: Maybe this Map should go directly inside `PrimitiveType`.

javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
```diff
@@ -29,7 +31,32 @@
...
+		static {
+			for(PrimitiveType.Primitive unboxedType : PrimitiveType.Primitive.values()) {
+				String boxedTypeName = unboxedType.toBoxedType().getName();
+				unboxMap.put(boxedTypeName, unboxedType);
+			}
+		}
+
+		private final ClassOrInterfaceType boxedType;
```
ptitjes, line 51: You should directly store the name `String` here instead of the `ClassOrInterfaceType`. Indeed consider the following scenario:

```
ClassOrInterfaceType boxedInt = Primitive.Int.toBoxedType();
// Add a NonNull annotation
boxedInt.setAnnotations(/*...*/);
```

This would add annotations directly in the `ClassOrInterfaceType` that you store in `Primitive`. This is certainly not what a user would expect :p

javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
```diff
@@ -29,7 +31,32 @@
...
+
+		public ClassOrInterfaceType toBoxedType() {
+			return boxedType;
```
ptitjes, line 54: So here you have to instantiate a new `ClassOrInterfaceType`...

javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java
```diff
@@ -29,7 +31,32 @@
...
+		}
+
+		private Primitive(String nameOfBoxedType) {
+			boxedType = new ClassOrInterfaceType(nameOfBoxedType);
```
ptitjes, line 58: And here assign the name directly...

javaparser-testing/src/test/resources/com/github/javaparser/bdd/dumping_scenarios.story
```diff
@@ -3,10 +3,19 @@ Scenario: When printing the instantiation we should use the right amount of spac
 Given the class:
 public class A {
     Object foo = new Object();
+	Runnable runnable = ()-> System.out.println("running");
```
ptitjes, line 6: That's perfect but I guess it would be even better if you copy this scenario and create a new one with your lambda in place of the `Object foo = new Object();`.

ptitjes:
I made some comments in-line. I hope you don't mind giving it a tweak again!
Sorry the `ClassOrInterfaceType` field in `Primitive` did pass my first review...

ptitjes:
At this point, I advise you to:
- `git reset --mixed 1e280d271aaa0a4e673dbf65bd3bec32f39a4e55`, to reset git to the commit before yours
- `git checkout -B fix-issue-168`, to create a branch for your pull request
- do your changes
- and finally use `git add --patch` to re-create your commits.

I just connected to gitter (https://gitter.im/javaparser/javaparser) to assist you if ever you need help in doing that.

You'll have to recreate a pull request however, because GitHub tracks pull requests with branches.

markzyu:
Thanks for the replies! They can help me improve, too. Sorry that I did not follow up immediately with any new commit, because I went to sleep right after the fourth commit.

I am closing this pull request and will follow your instruction to move my commits to another branch before making another request.

javaparser-testing/src/test/resources/com/github/javaparser/bdd/dumping_scenarios.story
markzyu, line 6: Sure... These flaws in my code should have been really obvious to me...

PR #172, opened 2015-08-05
Issue Fix 171: Spaces in the file path cause a failure in the Visitor…

maxstreichert:
…Test


ftomassetti:
It seems good to me, thank you for taking the time to submit this.

I will merge it

Issue #167, opened 2015-07-28
Dump tests don't work because of false delimiter support on Windows


maxstreichert:
I run junit Tests on eclipse on Windows and some tests fail because of false delimiter support.
The failing tests are all with the dump scenarios and assertions seem always like this: 

Expected: is "public class A {\r\n\r\n    Object foo = new Object();\r\n}"
     but: was "public class A {\n\n    Object foo = new Object();\n}"

The reason is that jBehave provides with @Given a string with OS-specific delimiters. In case of Windows the delimiter of the newline is "\r\n". But the javaparser provides after parsing a string the Unix-specific newline-delimiter, in this case "\n".

What I tried to do: I changed the delimiters of the story-files to the Unix-like, and checked via notepad++ that there are really Unix-like delimiters, but I didn't work :(

ftomassetti:
I suspect most contributors run test under Linux so does also our CI so we did not notice such errors so thank you for reporting them.

Can you tell us which tests exactly are failing? Can you report the error messages?

Thanks!

SmiddyPence:
Investigating.

Ok I have reproduced on Windows.  

Thanks for reporting.

SmiddyPence:
Will double check fix on linux later and file pull request.

PR #173, opened 2015-08-07
#167 Use Platform Independant Line Separator in DumpVisitor

SmiddyPence:
& remove unused imports.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
@@ -132,7 +132,7 @@ public final class DumpVisitor implements VoidVisitor<Object> {
 		}
 
 		public void printLn() {
-			buf.append("\n");
+			buf.append(System.getProperty("line.separator"));
 			indented = false;
 		}
 
--- a/javaparser-testing/src/test/java/com/github/javaparser/bdd/steps/DumpingSteps.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/bdd/steps/DumpingSteps.java
@@ -22,32 +22,18 @@ package com.github.javaparser.bdd.steps;
 
 import com.github.javaparser.JavaParser;
 import com.github.javaparser.ParseException;
-import com.github.javaparser.TokenMgrError;
 import com.github.javaparser.ast.CompilationUnit;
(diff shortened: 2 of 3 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
```diff
@@ -132,7 +132,7 @@ public void printLn(final String arg) {
 		}
 
 		public void printLn() {
-			buf.append("\n");
+			buf.append(System.getProperty("line.separator"));
```
ftomassetti, line 135: Looks good, Java 7 has System.lineSeparator but I think it is the best we can do with our JDK 6 constraint

javaparser-testing/src/test/java/com/github/javaparser/bdd/steps/DumpingSteps.java
```diff
@@ -22,32 +22,18 @@
...
 import com.github.javaparser.ast.comments.*;
-import com.github.javaparser.ast.expr.Expression;
-import com.github.javaparser.ast.expr.IntegerLiteralExpr;
-import com.github.javaparser.ast.stmt.BlockStmt;
-import com.github.javaparser.ast.stmt.ExpressionStmt;
-import com.github.javaparser.ast.type.PrimitiveType;
 import com.github.javaparser.ast.visitor.DumpVisitor;
-import com.github.javaparser.bdd.TestUtils;
-import org.jbehave.core.annotations.Alias;
+
```
ftomassetti, line 28: Definitely a good idea :D

pom.xml
```diff
@@ -193,4 +193,15 @@
 			</plugins>
 		</pluginManagement>
 	</build>
+	<profiles>
```
ftomassetti, line 196: I read around that this is a solution to avoid too-strict checks from the Javadoc utility that comes with JDK 8. Is that correct?

It seems good to me, but my understanding of this is limited

ftomassetti:
It seems good to go to me

pom.xml
SmiddyPence, line 196: Yes.  Currently with Java8 if we generate the javadoc compilation will fail.  With problems like return & param types not documented.  Some of these problems we can (should) fix as they are hand written classes, but the majority are from the generated classes.  So there is little we can do.

Issue #175, opened 2015-08-11
Update the license section in the POM


ftomassetti:
As discussed on gitter

SmiddyPence:
resloved

PR #179, opened 2015-08-12
#175 Updated license in poms

SmiddyPence:
Review and pull please

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -1,22 +1,22 @@
-<?xml version="1.0" encoding="UTF-8"?>
 <!--
   ~ Copyright (C) 2007-2010 Júlio Vilmar Gesser.
   ~ Copyright (C) 2011, 2013-2015 The JavaParser Team.
   ~
   ~ This file is part of JavaParser.
   ~
-  ~ JavaParser is free software: you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


javaparser-core/pom.xml
```diff
@@ -1,22 +1,22 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
ftomassetti, line 1: Why has been this line removed here and in the other pom.xml?

ftomassetti:
It looks good to me, I would just leave the line specifying the encoding

javaparser-core/pom.xml
SmiddyPence, line 1: Convention is not to include it, was only present in one of our current poms.

Typically license doesn't appear either though.

ftomassetti, line 1: My bad, I thought it was suggested to include it

ftomassetti:
There is also the `licenses` tag in the main pom to be updated

SmiddyPence:
@ftomassetti I have changed this to use the `<license>` block instead.  I couldn't find examples where verbatim is the pom like we have.

ftomassetti:
Makes sense to me, it should be easier to maintain also and harder to get unnoticed because it appears on Maven central

Issue #180, opened 2015-08-12
Wrong endLine/endColumn for primitive ArrayCreationExpr


Legioth:
ArrayCreationExpr created for primitive types has its endLine and endColumn set according to the primitive type token instead of the terminating ] token. There's a different projection for non-primitive arrays that seems to get the expected terminating token.

SmiddyPence:
Thanks for reporting.

I agree the way it works for object types sounds correct.

ftomassetti:
PR #182 prepared

SmiddyPence:
merged

PR #182, opened 2015-08-14
Issue 180: the end position of ArrayCreationExpr for primitive types is wrong

ftomassetti:
Correcting issue #180 as nicely suggested by @Legioth
Adding test to prevent regressions

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ASTHelper.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ASTHelper.java
@@ -22,6 +22,7 @@
 package com.github.javaparser;
 
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.BodyDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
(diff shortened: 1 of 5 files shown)


ftomassetti:
Any comment on this one?

SmiddyPence:
Perhaps the name is not the best, but the intention of "SharedSteps" was to include more of the common aspects of loading classes etc.  Open to something more descript.

The `@Then` s I would suggest are more to do with ensuring the correctness of parsing scenarios. So ParsingSteps.

Also I would tend to group whens and thens together, we have a when right after some thens in this case.

Two of the `@Then` have contradicting parameter names `$line` and `column`.

ftomassetti:
Thank you for your review. 
- I moved all but the first new step to ParsingSteps. I left `whenTheFollowingSourceIsParsedTrimmingSpace` in SharedSteps because it is a variant of `whenTheFollowingSourceIsParsed` which was already in SharedSteps. If you prefer that I move both to ParsingSteps I will do that.
- I moved the steps at the end of ParsingSteps so the only 'when' step is at the end of the file, all the others are 'then' steps. Do you prefer that I move the 'when' step at the top of ParsingSteps?
- I corrected the contradicting $line params

SmiddyPence:
Sure makes, sense.

Personal preference would be the given-when-then, but we're nit picking here though ;-)

ftomassetti:
Nothing agains nit picking :)
I reorder the steps and added comments to identify the blocks

SmiddyPence:
Cool, closing.

Issue #181, opened 2015-08-14
Parsing doesn't retain comments if the method has only comments


kashishm:
#### Original method :

```
public void helloWorld(String greeting, String name) {
        //sdfsdfsdf
            //sdfds
        /*
                            dgfdgfdgfdgfdgfd
         */
}
```
#### After parsing :

```
public void helloWorld(String greeting, String name) {
}
```

ftomassetti:
Thanks for reporting this, I will look into that in the next days (everyone else feel free to jump in if you can look into it sooner)

kashishm:
In the following example, parsing removes half comments.

#### Original method :

```
public void helloWorld(String greeting, String name) {
        //sdfsdfsdf
        //sdfds
        /*
                            dgfdgfdgfdgfdgfd
         */
        System.out.println(greeting + ", " + name);
        //sdfsdfsdf
        //sdfds
        /*
                            dgfdgfdgfdgfdgfd
         */
}
```

#### After parsing :

```
public void helloWord(String greeting, String name) {
        //sdfsdfsdf
        //sdfds
        /*
                            dgfdgfdgfdgfdgfd
         */
        System.out.println(greeting + ", " + name);
}
```

ftomassetti:
I think that when dumping we do not print the orphan methods following the last statement of a body declaration.

ftomassetti:
PR proposed #183

kashishm:
Thanks for the quick response and fix.

ftomassetti:
#183 should have fixed this, please re-open or create a new issue if you experience other problems

PR #183, opened 2015-08-14
Issue 181

ftomassetti:
To solve #181 we need to print orphan comments at the end of a method.
Note that the alignment of comments could be improved but I suggest to start by fixing this issue first, and later address that.


SmiddyPence:
+1 

Seems sensible in lieu of a wider whitespace preservation debate.

ftomassetti:
The patch seems limited, I got approval by @SmiddyPence and no further comments so I would merge this (if you think this is inappropriate let me know and I will revert the patch and avoid doing that in the future)

PR #185, opened 2015-08-17
Issue 149: new solution introducing the MultiBoundType

ftomassetti:
With this PR we are able to parse the examples reported in issue #149 
We introduce a new kind of node representing multiple bounds as they could appear in cast expression:

``` java
(Iterable & Serializable) foo;
```

If the general approach is acceptable for you all we can add more tests for the visitors which had to be changed, so we should checked that all the visitor handle the new node type correctly.


ftomassetti:
Update the patch to introduce UnionType and rename IntersectionType

SmiddyPence:
+1 for the immediate solution.

Then in conjunction with #111 for 3.0

ptitjes:
@ftomassetti, I have a feeling that there is a mismatch in your PR with respect to `MultiTypeParameter` and `CatchClause`: You make `CatchClause.param` be a `Parameter` (and not a `MultiTypeParameter`) but you let `MultiTypeParameter` extends `BaseParameter` (and not `Parameter`). I think there is a problem here.

Also if your wish is to not break API for now, you should let `CatchClause.param` be named `CatchClause.except` and be a `MultiTypeParameter`. I would advise you to not touch `CatchClause` at all and add temporary (already deprecated) methods on `MultiTypeParameter` that delegates to `UnionType` to fake the old `get/setBounds` methods.

ftomassetti:
Well, MultiTypeParameter does not matter now, it could be actually removed because it was only used in CatchClause.

If you prefer we can remove the commits touching MultiTypeParameter and introduce just the IntersectionType here. We would need to touch all the visitors and test their behavior with the new type.
How does it sound?

ptitjes:
Oh sorry! No no, I thought you wanted to have something non-API-breaking before 3.0. Sorry, I'm a bit lost. Ignore if not relevant.

ftomassetti:
No need to be sorry and thank you for your review

matozoid:
Can be pulled when rebased.

ftomassetti:
Rebased, tests running

PR #187, opened 2015-08-17
Parse annotations around reference types

ftomassetti:
In both throws and implements clause we should parse annotations preceding the elements.

Fixing this would permit to parse two files present in the OpenJDK which are currently not parsed.

If you agree with this approach I could just add a few extra tests to protect from regressions.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
@@ -29,6 +29,7 @@ import com.github.javaparser.ast.comments.JavadocComment;
 import com.github.javaparser.ast.expr.AnnotationExpr;
 import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.type.ReferenceType;
 import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.visitor.GenericVisitor;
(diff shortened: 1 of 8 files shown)


ptitjes:
Do not forget the extends clause for both class declarations and interface declarations... (as per the JLS as usual...)

SmiddyPence:
Defer to you wisdom.

matozoid:
Is this done, or is it still a proposal?

ftomassetti:
This is done, but there are other places where the same change could be applied (permitting to use annotations in more places). However we could fix that later.

ptitjes:
Why not fix that correctly and slowly now ? Is there something for which this has to be commited immediately ?

Fixing that for classes' and interfaces' extends clauses is just a matter of two copy pastes and some more test cases. You've already made the main work of understanding the how and when. So either you apply it consistently or you forget it and remake the intellectual work in some weeks.

FYI, JLaTo parses that from the beginning...

ftomassetti:
Rebased

ftomassetti:
Note that we are using `ReferenceType` to host references to type instead of using NameExpr. This gives us the possibility to add annotations to that particular usage of the type. However ReferenceType permits also to specify the array modifier, which is not legal in most places. 

I would open an issue to refactor ReferenceType to do one thing (contains annotations) instead of two (containg annotations and array modifiers). What do you think?

matozoid:
Please rebase this, pull it, and add the new issue to keep things moving.

ftomassetti:
rebased, tests running

PR #195, opened 2015-09-04
Issue192: add extra tests to prevent regressions

ftomassetti:


ftomassetti:
This is just about adding extra tests and it is has been open for almost 2 months so I will merge it, please let me know if it is an issue and I will revert it.

PR #196, opened 2015-09-17
Adds IntelJ files to gitignore

JLLeitschuh:


ftomassetti:
Thanks! Why it was not working before? Which files were not ignored and should have been? Why removing the `.idea` item?

.gitignore
```diff
@@ -6,5 +6,53 @@ target
 /target
 .project
 .classpath
-.idea/
```
JLLeitschuh, line 9: This was moved

.gitignore
```diff
@@ -6,5 +6,53 @@ target
...
+
+# Created by https://www.gitignore.io/api/intellij
+
+### Intellij ###
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio
+
 *.iml
+
+## Directory-based project format:
+.idea/
```
JLLeitschuh, line 18: Down to here

ptitjes:
All those comments make a lot of noise. You should keep the comments to the bare minimum and rework what IDEA has generated. Also that will help git diff recognize the '.idea/' line!

JLLeitschuh:
You may want to look at this and implement a gitignore that covers all of the different tools that this project uses:
https://www.gitignore.io

SmiddyPence:
I'm at a loss at to what benefit this brings over the catchall? 

Usually there is some issue that gets raised that needs addressing.

JLLeitschuh:
Perhaps this won't help you.
I added it because when I imported the project into Intellj I got files that I had to manually ignore.
I'm just looking to make it easier for future devs that want to jump in.

https://www.gitignore.io/api/maven,gradle,windows,osx,eclipse,intellij,linux,java

ptitjes:
The build is done by maven. If you correctly import the project in IDEA, IIRC only `.idea/` and `*.iml` are needed to be ignored.

We do not use gradle. The windows, osx and linux sections are useless. Also the java stuff is useless.

Could you be specific on what files you had to ignore ?

SmiddyPence:
Thank you for you efforts in trying to make JavaParser better, however we have decided we will not be incorporating this request.

We do still need your help in making the parser better for everyone.  If you would like to contribute to the project, please look at the existing issues.  Feel free to join the public channel on gitter and discuss any ideas you may have. [Chat](https://gitter.im/javaparser/javaparser)

Looking forward to your next PR!

PR #197, opened 2015-09-17
Adds documentation to various confusing places

JLLeitschuh:
There were various locations that I was very confused about.
I added documentation as accurate as I could understand it.


ftomassetti:
+1
We definitely need better comments, thanks a lot for this PR!

JLLeitschuh:
I figure if I'm using the library and figuring out how it works I may as well add the documentation for the next person who needs to figure it out.

Issue #157, opened 2015-06-27
Enhance parser frontend


ptitjes:
This bug is to discuss enhancements to the parser front-end as highlighted in #115, that is:
- Make JavaParser a state-full object
  - Give access to the (exception-free) parsing mechanism
  - Keep static helpers that throw ParseException
- Add problem reporting mechanism
  - Introduce a Location object (source path + Position, comparable)
  - Introduce a Problem object (comparable based on Location)
  - Introduce a ParseResult` object that contains the parsed node and reported problems
  - For the static helpers, make a `ParseException` that contains a list of problems

I'd rather not introduce a `Reporter` interface as I previously noted in #115. Indeed, lexical preservation might enable us to later do continuous parsing à-la Papa Carlo, so I propose the creation of a `ParseResult` object which might later provide a method like `reparse(Parser, ParseHandler)` with `ParseHandler` being a reactive sink for problems and node addition/removal/replacement events. 

Here are, in a glance, the APIs that I propose for discussion:

``` java
public class Location {
  public java.nio.Path path(); // May be null if not parsed from a file
  public Position position();
}

public class Problem {
  public enum Severity { FATAL, ERROR, WARNING, INFO }
  public Severity severity();
  public Location location();
  public String message();
}

public class ParseResult<N extends Node> {
  public List<Problem> problems();
  public N node();
}

public class ParserConfiguration {
  private boolean parseComments; // + getter/setter
  // other configuration...
}

public abstract class ParseContext<N extends Node> {
  // Required package-protected boilerplate code to call ad-hoc parser production...
}

public class JavaParser {
  public JavaParser(ParserConfiguration configuration) { /* ... */ }

  public static ParseContext<CompilationUnit> COMPILATION_UNIT = ...;
  public static ParseContext<BodyDeclaration> BODY_DECLARATION = ...;
  public static ParseContext<Type> TYPE = ...;
  public static ParseContext<Statement> STATEMENT = ...;
  public static ParseContext<Expression> EXPRESSION = ...;
  // ...

  public <N extends Node> ParseResult<N> parse(ParseContext<N> context, InputStream input, Charset encoding);
  public <N extends Node> ParseResult<N> parse(ParseContext<N> context, String input, Charset encoding);

  public ParseResult<CompilationUnit> parse(Path path, InputStream input, Charset encoding);
  // And some other variations

  // + current static helpers throwing ParseException
}

public class ParseException extends Exception {
  private List<Problem> problems;
  // ...
}
```

[...]

PR #207, opened 2015-10-06
Enabled access to the tokens after parsing

sebkur:
Enabled the COMMON_TOKEN_ACTION=true option in the java_1_8.jj file.

Implemented a CommonTokenAction(Token token) method in the
TOKEN_MGR_DECLS that collects encountered tokens in a list.

Added a getter for this list of tokens in the ASTParser.

Made ASTParser public so that it may be used directly, not just using
the JavaParser wrapper class.

Made JavaParser.insertComments(...) public so that it may be used to
attach comments to a CompilationUnit that has been parsed using the
ASTParser directly.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -440,7 +440,17 @@ public final class JavaParser {
         return b.getBeginLine()>(a.getEndLine()+1);
     }
 
-    private static void insertComments(CompilationUnit cu, String code) throws IOException {
+    /**
+     * Parses the comments in the Java code contained in the String and attaches
+     * them to nodes of the CompilationUnit.
+     *
+     * @param cu
+     *            the CompilationUnit to attach the comments to
+     * @param code
+     *            Java source code
+     * @throws IOException
+     */
+    public static void insertComments(CompilationUnit cu, String code) throws IOException {
         CommentsParser commentsParser = new CommentsParser();
         CommentsCollection allComments = commentsParser.parse(code);
 
--- a/javaparser-core/src/main/javacc/java_1_8.jj
+++ b/javaparser-core/src/main/javacc/java_1_8.jj
@@ -23,7 +23,7 @@ options {
   LOOKAHEAD=1;
   STATIC=false;
   JAVA_UNICODE_ESCAPE=true;
-  //COMMON_TOKEN_ACTION=true; // Will not use CommonTokenAction to handle JavaDoc and comments
+  COMMON_TOKEN_ACTION=true; // Using the CommonTokenAction callback to collect tokens for later usage
   //SUPPORT_CLASS_VISIBILITY_PUBLIC=false;
   JDK_VERSION = "1.6";
   TOKEN_FACTORY = "ASTParser.GTToken";
@@ -56,7 +56,7 @@ import com.github.javaparser.ast.type.*;
 /**
  * <p>This class was generated automatically by javacc, do not edit.</p>
  */
-final class ASTParser {
+public final class ASTParser {
 
     void reset(InputStream in, String encoding) {
         ReInit(in, encoding);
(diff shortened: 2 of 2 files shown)


sebkur:
Using these changes we can use the tokens of the parsed source code like this:

```
Reader reader = new StringReader(code);
ASTParser astParser = new ASTParser(reader);
CompilationUnit unit = astParser.CompilationUnit();
JavaParser.insertComments(unit, code);

List<Token> tokens = astParser.getTokens();
for (Token token : tokens) {
    // do something with each token
}
```

ftomassetti:
I am not familiar with that feature of JavaCC but this addition seems very useful to me. Thanks a lot for sending this patch!
I am not sure it is a good idea to make ASTParser visible. Could we perhaps create a new class wrapping ASTParser and exposing just the methods to collect tokens? Could that be clearer for user?

sebkur:
I can see that the ASTParser exposes a lot of methods that users might be overwhelmed with. It is not directly clear which part of a source file theses methods will parse.

In my specific use case of syntax highlighting, it makes sense however to have access to some of those methods. The typical situation might be parsing a whole CompilationUnit, but it also makes sense being able to parse just valid fragments of a java source file. For example if I'm parsing just a method declaration or maybe just a few statements, it is useful being able to parse them independently.

To achieve this I would start parsing the greatest possible match and fall back to smaller units should exceptions occur during parsing. I could try CompilationUnit() first, then TypeDeclaration() and then probably ClassOrInterfaceBodyDeclaration().

The JavaParser already accounts for some of these situation as it provides parseImport(), parseStatement(), parseBlock() and some others. We could of course add more of these for everything that makes sense to be parsed independently, but I guess we would end up repeating many of the methods from the ASTParser.

The parse() method is also overloaded for various types of input and some parameters while the other parse\* methods work only for strings. I think it would make sense to provide overloaded variants for the other methods as well. However this becomes pretty verbose quickly depending on the number of methods from the ASTParser that would be made available through the JavaParser.

Not sure which is the best approach here.

ptitjes:
@sebkur Could be neatly done with `ParseContext` as proposed here https://github.com/javaparser/javaparser/issues/157. However, it seems to me unrelated to the topic of this ticket which was about access to tokens.

ftomassetti:
@sebkur I think we can expose just the few methods of ASTParser which makes sense. We could want to parse a class declaration, a statement, an expression and of course a CompilationUnit. I was thinking about 4-6 methods or so to expose while ASTParser has over 100 public methods and several internal classes. While the subset of the language which makes sense to parse (class declaration, statement, expression, etc.) is very stable the other 100 methods could change over time as we refactor internally the parser.

Perhaps we could limit ourselves to just parsing strings: if someone is interested in getting the tokens of an expression he can just take his InputStream and convert it to a String. We could go as far as provide an utility method to do so but probably it would be too much clutter to have several variants of the same method, repeated for several methods.

@ptitjes it seems out of scope to me: this is a nice patch to which we can discuss minor adjustments and having it approved while #157 seems a complete refactoring which a large impact

More comments would be welcome (@matozoid @SmiddyPence @sebastiankirsch)

ptitjes:
@ftomassetti Introducing `ParseContext` is not a "large refactoring" nor a refactoring at all... It requires one new method in the `JavaParser` class, a `ParseContext` abstract class and some implementation classes of `ParseContext`.

ftomassetti:
@ptitjes in my humble opinion it seems out of the scope of this patch which to me seems very nicely focused and useful as it is. I think it would be great if you wanted to propose that other change in a separate PR.

ptitjes:
I kinda repeat myself, https://github.com/javaparser/javaparser/pull/207#issuecomment-147354916

> However, it seems to me unrelated to the topic of this ticket which was about access to tokens.

ftomassetti:
@sebkur what do you think about my comments? Would it make sense to expose a simpler interface instead of making ASTParser public?

sebkur:
Sorry for not answering that long. Yes, for me that makes perfect sense. I agree that making the ASTParser public is not a good idea. I'll try to come up with a decent interface as soon as I find some time.

Although finding a proper solution for https://github.com/javaparser/javaparser/issues/157 might make the things I'm interested in possible along the way, I think I'm too unfamiliar with the requirements to come up with an implementation of that.

ftomassetti:
No problem. Please disregard #157 , your patch it is very close to what we need to deliver value and I think that with a few changes it is ready to go.

sebkur:
Superseded by https://github.com/javaparser/javaparser/pull/229

PR #213, opened 2015-10-14
Fix TryStmt NullPointer if Resources not provided

### LINKED ISSUE #212: TryStmt throws null pointer if setResources is not called
If you create a `TryStmt` without calling `setResources(Collections.emptyList())` the to string dump visitor throws a null pointer exception.
This line:
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java#L1420
Should check for null as well as if empty.

JLLeitschuh:
Checks if the resources is null prior to checking if resources is empty.

Closes #212


ftomassetti:
Thank you for submitting this patch.
I think this issue has been corrected in master (but not yet released). If you look at TryStmt.getResources as it is on master (https://github.com/javaparser/javaparser/blob/defc804a38374b48c9d004299627a0b07812d326/javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java) it should never ever returns null.

Therefore I would close this PR. Hope to see more coming in the future!

PR #215, opened 2015-10-16
Prepare next release (2.3.0? 3.0.0?)

ftomassetti:
- updated version number
- added a changelog

About the version number:
I would go for 2.3.0: I think we introduce compatibility breaking changes but we do not have enough change to move to 3.0.0 in my opinion. What do you think?

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,0 +1,7 @@
+Version 2.3.0
+-------------
+
+* ClassOrInterfaceType implements NamedNode
+* DumpVisitor can now be extended
+* Improved documentation
+* AST: lists are now lazy initialized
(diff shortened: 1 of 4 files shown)


SmiddyPence:
ironically up versioning it probably the one thing the release plugin does reliably for you ;-)

ftomassetti:
right but the version was 2.2.3 while I think we should release 2.3.0. I was not sure if the plugin would ask me which version I wanted to release or just remove the "-SNAPSHOT" suffix and use the current version

SmiddyPence:
It will, but depending on how many times you have to do it this may be quicker.  You can keep bashing enter your way :)

JanBessai:
I've just built it locally and it worked fine for me :+1:

ftomassetti:
This is been released: merging this PR because it contains the changes necessary to do the release and the changes from the maven release plugin

PR #218, opened 2015-10-21
Report whole package name

olivergondza:
It seems that the parser reports only the last package component. Attaching reproducing test, I would welcome some guidance on where to fix this.

`Then the package name is com.github.javaparser.bdd(com.github.javaparser.bdd.steps.ParsingSteps): expected:<[com.github.javaparser.]bdd> but was:<[]bdd>`

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/PackageDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/PackageDeclaration.java
@@ -90,12 +90,19 @@ public final class PackageDeclaration extends Node {
     }
 
     /**
-     * Return the name of the package.
-     * 
+     * Return the name expression of the package.
+     *
      * @return the name of the package
      */
     public NameExpr getName() {
         return name;
+    }
+
+    /**
+     * Get full package name.
+     */
+    public String getPackageName() {
+        return name.toString();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ftomassetti:
Yes, as @ptitjes it works as intended but the API could be confusing. Perhaps we should improve it. Anyway thank you for taking the time to send a PR. We hope to see more of them in the future. 
If you have questions or doubts or do you want to tell us which awesome things are you doing with JavaParser feel free to join our channel on gitter (you can find the link in the readme).

ptitjes:
@ftomassetti said:

> the API could be confusing. Perhaps we should improve it.

Indeed! I'd even say the design is not correct as having `MethodDeclaration.name` being a `NameExpr` enables one to a have a `QualifiedNameExpr` as a method name. Quite fun but not Java...

@ftomassetti You can look at [JLaTo](https://github.com/ptitjes/jlato)'s [Name](http://static.javadoc.io/org.jlato/jlato/0.0.5/org/jlato/tree/name/Name.html) and [QualifiedName](http://static.javadoc.io/org.jlato/jlato/0.0.5/org/jlato/tree/name/QualifiedName.html) for an alternative design.

olivergondza:
Confirming that `node.getName().toString()` does what I need. I need it to generate fully-qualified symbol name (perhaps there is a way to do so and I missed that, too). Something like:

```
java.io.File
java.io.File#separatorChar
java.io.File#getName()
java.io.File#renameTo(java.io.File)
java.io.File.TempDirectory#tmpdir
```

olivergondza:
I have fixed the test and proposed more convenient api. Merge it if you want.

matozoid:
Looks reasonable to me.

ftomassetti:
@olivergondza can you please rebase this one? We will merge it if you can do that. Thank you!

olivergondza:
Rebased on top of master.

ftomassetti:
Thank you @olivergondza !

Issue #217, opened 2015-10-21
TypeDeclaration is not a DocumentableNode


olivergondza:
What I am trying to do is following:

``` java
for (TypeDeclaration t: compilationUnit.getTypes()) {
    t.getJavaDoc();
}
```

ftomassetti:
TypeDeclaration is not documentable but all of its subclasses are DocumentableNode. It could make sense to make it implement DocumentableNode directly for this specific usage scenario.

Thanks for reporting this!

PR #219, opened 2015-10-21
Make TypeDeclaration implements DocumentableNode directly

ftomassetti:
All the subclasses of TypeDeclaration are implementing DocumentableNode, however TypeDeclaration (which is abstract) it is not.
This prevent the reasonable usage presented in #217 .
The solution is to declare TypeDeclaration to implement DocumentableNode directly.


olivergondza:
Fix looks good to me.

SmiddyPence:
+1 Ship it

ftomassetti:
thanks both of you guys, merging

Issue #160, opened 2015-07-07
Double backslash in strings and lost comments


Irioth:
Can't retrive comments after double backslash in strings.

```
public class b {

    public void m1() {
        String s = "\\";
    }

    /**
     * Comment
    */  
    public void m2() {
        return;
    }
}
```

Parser will losing comment in this example. If I change value of string `s` to something without backslash everything works fine.

version: 2.1.0

ftomassetti:
that is interesting, thanks for catching this. It seems an issue with the CommentsParser

ftomassetti:
Maybe is this:
"https://github.com/javaparser/javaparser/blob/109e43888bc1b771780ce155d732e870f63bfb7a/javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsParser.java"

```
                case IN_STRING:
                    if (!prevTwoChars.peekLast().equals('\\') && c == '"') {
                        state = State.CODE;
                    }
                    break;
```

We did not consider the double backslash. I think it was me introducing this bug while fixing #37

ftomassetti:
I am sorry it took so long but we have now a PR to fix this

matozoid:
Can this be closed now?

ftomassetti:
I think it was fixed in #224 so I would close this one

PR #224, opened 2015-10-28
Issue 160: sequences of slashes in Strings confuse the CommentsParser

ftomassetti:
This should solve #160 and the tests added should protect against regressions.


ftomassetti:
It should be less hacky now. Thanks for the feedback!

PR #227, opened 2015-10-29
Issue 211: consider literals with newlines

ftomassetti:
- a test verifies that string literals with newlines are parsed and dumped correctly
- we prevent StringLiteralExpr with newlines or tabs from being instantiated

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
@@ -32,9 +32,13 @@ public class StringLiteralExpr extends LiteralExpr {
 	protected String value;
 
 	public StringLiteralExpr() {
+        this.value = "";
 	}
 
 	public StringLiteralExpr(final String value) {
+        if (value.contains("\n") || value.contains("\t")) {
+            throw new IllegalArgumentException("Illegal literal expression: some characters have to be escaped");
+        }
 		this.value = value;
 	}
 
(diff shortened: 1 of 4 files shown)


SmiddyPence:
Are we better off with:

`System.getProperty("line.separator"));`

Considering https://github.com/javaparser/javaparser/issues/167 ?

ftomassetti:
I really do not know about that. I think you cannot use `\n` neither `\r` on any platform. It does not really depend if that character is used to form newline on that platform or not. Perhaps we should prevent all the characters for which there is an escaped form to be used. Can you use a real tab character in a string or do you have to replace it with `\t`? 

Here I cannot read that you should not use newlines or tabs in string: https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.10.5

I am probably missing the correct place in the documentation

ptitjes:
Just to help...

From JLS 3.10.5:

> StringLiteral := " {StringCharacter} "
> StringCharacter := (InputCharacter but not " or )  || EscapeSequence

From JLS 3.4:

> InputCharacter := UnicodeInputCharacter but not CR or LF

See 3.10.6 for EscapeSequence.

So no newlines in strings. Also beware of Unicode unescaping during Lexical Translations (3.2).

JLLeitschuh:
I would argue that escaping the characters for the user would be more intuitive. However, this is a valid alternative.

matozoid:
This could use a negative test and a more informative error message.

ftomassetti:
Added negative test, revised error message and rebased

matozoid:
Can we get a parse location into the error message? If not: pull.

ftomassetti:
@matozoid we generate the error in the constructor of StringLiteralExpr, when a position is not available. That makes sense because the StringLiteralExpr could be used to create new nodes (which do not have a position). 

However an error during parsing will generate a TokenMgrError which will contain the current position, so it should be fine. I am going to pull this. Thanks for the review!

PR #229, opened 2015-11-07
JavaParser refactoring

sebkur:
This PR is the result of the discussion on this one: https://github.com/javaparser/javaparser/pull/207

The main parsing methods of the JavaParser now do not return a Node
subclass anymore, but instead an instance of the new ParseResult class.
This class wraps a Node plus some extra information about the parsing
process. At the moment the additional information is the list of tokens
that the parser has built while parsing the source code. It could
however be extended in the future to provide more information about the
parsing process.

Changes in detail:

Enabled the COMMON_TOKEN_ACTION=true option in the java_1_8.jj file.

Implemented a CommonTokenAction(Token token) method in the
TOKEN_MGR_DECLS that collects encountered tokens in a list.

Added a getter for this list of tokens in the ASTParser.

Added the ParseResult class as a wrapper for Node and List<Token>.

Changed parsing methods of JavaParser to return ParseResults instead of
Node subclasses.

Fixed testing source code to work with the new JavaParser API.

Cleaned up the JavaParser's parse() methods that parse a
CompilationUnit:
- Consistent overloaded versions for all input types (String,
  InputStream, File and Reader)
- Only the parse() method that accepts a String as input and mandatory
  'encoding' and 'considerComments' arguments contains the logic of how
  to actually parse and create a result. All other version just forward
  to this method by building a String and passing it to this method.
- Added JavaDoc for some undocumented public methods.
- Fixed JavaDoc: new return types.


sebkur:
As opposed to the changes proposed in https://github.com/javaparser/javaparser/pull/207, this PR actually modifies the public API of the core JavaParser class and will break code depending on the current API. Don't know if this is desirable. Existing code is pretty easy to fix though, since it is just required to append a `.getNode()` in places where the results of parsing are being processed. See the changes in the testing source code for examples of the necessary updates for existing code.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -86,72 +83,145 @@ public static void setDoNotAssignCommentsPreceedingEmptyLines(boolean doNotAssig
...
     /**
+     * Parses the Java code contained in the {@link String} and returns a
+     * {@link CompilationUnit} wrapped in a {@link ParseResult} that represents
+     * it.
+     *
+     * @param code
+     *            {@link String} containing Java source code
+     * @param encoding
+     *            encoding of the source code
+     * @return CompilationUnit representing the Java source code
```
ftomassetti: this should probably be updated

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -86,72 +83,145 @@ public static void setDoNotAssignCommentsPreceedingEmptyLines(boolean doNotAssig
...
      *            encoding of the source code
+     * @param considerComments
+     *            whether to parse comments and attach to nodes
      * @return CompilationUnit representing the Java source code
      * @throws ParseException
      *             if the source code has parser errors
      */
-    public static CompilationUnit parse(final InputStream in,
-                                        final String encoding, boolean considerComments) throws ParseException {
+    public static ParseResult<CompilationUnit> parse(final String code,
```
ftomassetti: To maintain compatibility we could have both one method with the same name and the same return type as before and a new method named "parseWithTokens" or something similar. The new method would return the ParseResult and the method with the old name ("parse") would just call the new method, take the node and throw away the tokens. What do you think? I know it could be a bit repetitive to write but in this way we would stay compatible and offer a clean interface for many users which are not interested in tokens, while giving all the information to users interested in both nodes and tokens

sebkur:
Well I think compatibility is important. I now copied all parse() methods and created parseToResult() counterparts.

Of course all in all there's much redundancy in the class now. In the previous discussion you mentioned to maybe just support parsing from String and moving conversion from File, Reader and InputStream to a utility class. Internally the current implementation converts to String anyway. I think that would be useful, also because this will provide more flexibility with the parseStatement, parseBlock, etc methods as well. 

However, removing the existing methods with File, Reader or InputStream arguments would break compatibilty as well, so this is probably not an option. Maybe we should create such a utility class and mark the non-String methods as deprecated and then remove them after a while.

An alternative might be to create a fresh parser class, designed with more flexibility in mind and keep the JavaParser the way it was. I'm thinking about moving away from static parsing methods and instead having a Parser class that wraps an ASTParser object and offers some high level instance methods for parsing code to CompilationUnits, Blocks, Statements etc. Having a new class would give the freedom to introduce a cleaner API without maintaining a set of legacy methods.

ptitjes:
Let me note that the copy from streams/readers to string inside the parse methods was made when the actual CommentsParser was added. I personally think that this is a bad choice performance and memory wise. Indeed using streams and readers you can have direct buffers without doing in memory copies to strings.

Moving to accepting only strings as a parse interface would force to accept this design for the time being. I'd rather suggest to keep the parse methods taking streams/readers and work later on enhancing the CommentsParser to work with streams as this was done before and is done in some other projects (:p).

sebkur:
I agree that performance-wise converting to String is not a good idea. Also the ASTParser supports Reader and InputStream so there's basically no need for the detour over String. With a new Parser class that would not use static methods but instance methods instead, this should be possible without the redundant methods for each input type.

ftomassetti:
It looks good to me, I will ping other contributors on the gitter channel to see if there are other comments.

We can of course improves things in successive PRs and having the CommentsParser works with streams would be nice.

sebkur:
I have created a draft for the idea with the parser with instance methods as opposed to the current approach with static parsing methods in the `JavaParser` class. The benefit with this approach is on the one hand that it's then easy to access additional information like the tokens after parsing and on the other hand that it is straightforward to allow parsing methods for all kinds of parts of a `CompilationUnit` like `Statment`s or `Block`s without writing repetitive code for different input sources. Here's the link to the branch:

https://github.com/sebkur/javaparser/commits/instance_parser

Here's how parsing from an `InputStream input` works:

```
InstanceParser parser = new InstanceParser(input);
CompilationUnit unit = parser.parse();
```

And here's how to insert comments into a parsed `CompilationUnit`

```
InputStream input = new FileInputStream(filename);
String text = streamToString(input);

InstanceParser parser = new InstanceParser(new File(filename));
CompilationUnit unit = parser.parse();

CommentsInserter commentsInserter = new CommentsInserter();
commentsInserter.insertComments(unit, text);
```

What do you think about this?

ftomassetti:
Sorry for taking so long to answer. I personally think that having one single method to call could be simpler, however I see the implementation advantage of this alternative approach. Could be possible to implement JavaParser using InstanceParser? People could keep using the existing interface or use directly InstanceParser, if they prefer to.

This is just my opinion of course. I am not sure what the other contributors think about it.

sebkur:
Yes, implementing JavaParser using InstanceParser is straightforward. JavaParser would then essentially be a class that offers helper methods for one-line parsing calls.

I think the name 'InstanceParser' is a bit confusing though. I'm not sure what would be a better name for it.

ftomassetti:
It is difficult to come up with this name :) I think InstanceParser could be called just Parser or perhaps JavaParser should be renamed JavaParserFacade (but I guess we would have some resistance on doing that)

matozoid:
Is anybody still working on this?

sebkur:
Not really at the moment. I would still very much like this feature to be available on the official API, so I'd be willing to pick it up. I will try to make some progess in the next few days. I'll rebase on the current master and try to come up with useful names for the new classes.

sebkur:
As mentioned before (https://github.com/javaparser/javaparser/pull/229#issuecomment-156077537) the work on this continues in a different branch: https://github.com/sebkur/javaparser/commits/instance_parser

I have merged into that branch the changes from last months without problems.

The `InstanceParser` has been renamed to `InstanceJavaParser`, I think that name is better than `InstanceParser`, but still suggestions are welcome for better names.

To sum things up, there are two new classes when compared to the master branch:
- `InstanceJavaParser` (https://github.com/sebkur/javaparser/blob/instance_parser/javaparser-core/src/main/java/com/github/javaparser/InstanceJavaParser.java)
- `CommentsInserter` (https://github.com/sebkur/javaparser/blob/instance_parser/javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java)

Which can be used more flexibly to parse source code and also have access to the tokens after parsing for example.

As suggested before (https://github.com/javaparser/javaparser/pull/229#issuecomment-157046987) I have refactored the `JavaParser` to be based on the new `InstanceJavaParser`. As a result the class is a lot more lightweight, also because the code related to comment parsing / insertion is no encapsulated in the `CommentsInserter` class.

You can review the changes in the mentioned branch. If you think it is going in the right direction, I will create another branch that contains the squashed changes of that branch and open yet another PR based on that.

ftomassetti:
I think we are most definitely going in the right direction. Thank you for your efforts and patience.
- In CommentsInserter fields start with an underscore. In JavaParser I do not think that convention is used anywhere (it sounds a bit like C++ :D)
- It could make sense to add yourself to the contributors (in pom.xml)

sebkur:
Sure, field names with underscores is definitely not Java convention. Changed that and added myself to the contributors list. I also created a new branch and PR: https://github.com/javaparser/javaparser/pull/297

sebkur:
Superseded by https://github.com/javaparser/javaparser/pull/297

PR #230, opened 2015-11-09
Update DumpVisitor.java

cdietrich:
Make SourcePrinter better accessable


ftomassetti:
In which cases do you think it would made sense to have SourcePrinter more visible? Isn't intended to be used only in subclasses of the DumpVisitor?

trew:
It makes sense if you don't want to create a subclass but instead something like this:

``` java
new DumpVisitor()
{
  @Override
  protected SourcePrinter createSourcePrinter()
  {
    return new SourcePrinter("  ");
  }
}
```

matozoid:
I guess that makes sense.

PR #232, opened 2015-11-10
issue231: fix how empty Enums are dumped

### LINKED ISSUE #231: DumpVisitor wrong for Enums
public enum ABC {} is serialized as public enum ABC implements {}

ftomassetti:
This should fix #231


matozoid:
Looks fine to me.

PR #241, opened 2015-12-02
issue233: consider in DumpVisitor(ConstructorDeclaration, Object) tha…

### LINKED ISSUE #233: Extra whitespace in DumpVisitor for ConstructorDeclaration
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java#L969

`n.getTypeParameters()` will never return null. It should be `if (!n.getTypeParameters().isEmpty())`.

ftomassetti:
…t now lists are lazy initialized

Fix #233


SmiddyPence:
+1 Ship it

Issue #242, opened 2015-12-14
@FunctionalInterface annotation should be used to flag Single Abstract Method interfaces


matozoid:
A Single Abstract Method (SAM) interface is a Java interface containing only one method. The Java API is full of SAM interfaces, such as java.lang.Runnable, java.awt.event.ActionListener, java.util.Comparator and java.util.concurrent.Callable. SAM interfaces have a special place in Java 8 because they can be implemented using Lambda expressions or Method references.

Using @FunctionalInterface forces a compile break when an additional, non-overriding abstract method is added to a SAM, which would break the use of Lambda implementations.

ftomassetti:
Currently JavaParser supports Java 6 so we cannot use [FunctionalInterface](https://docs.oracle.com/javase/8/docs/api/java/lang/FunctionalInterface.html) because it was introduced in Java 8.

We should consider that for the future, I would mark this issue with the Label "Java8" for future references.

ptitjes:
As I explained in https://github.com/javaparser/javaparser/pull/248#issuecomment-164832149, IMHO, this should be resolved for the two mentioned interfaces by configuring Sonar to ignore the two warnings.

ftomassetti:
I think we could close his one

matozoid:
Won't fix until JavaParser's core is Java 8.

matozoid:
We can check this one again.

DeepSnowNeeL:
With my modifications for #363 and related to  #372 , those 2 interfaces have more than 1 method now

matozoid:
Here it goes again!

PR #248, opened 2015-12-15
Code Quality Improvement - @FunctionalInterface annotation should be used to flag Single Abstract Method interfaces

civanyp:
This pull request is focused on resolving occurrences of Sonar rule squid:S1609 - “@FunctionalInterface annotation should be used to flag Single Abstract Method interfaces”. You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1609.

Please let me know if you have any questions.

Christian

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NamedNode.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NamedNode.java
@@ -28,6 +28,7 @@ package com.github.javaparser.ast;
  * 
  * @since 2.0.1 
  */
+@FunctionalInterface
 public interface NamedNode {
     String getName();
 }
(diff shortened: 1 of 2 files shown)


ptitjes:
Apart from the Java8 dependency, IMHO, there is not much meaning to annotate those two interfaces with FunctionalInterface.

Indeed, those are not SAMs in the sense that they are not intended to be implemented as a lambda.

The fact that those two interfaces have only one method is just purely a coincidence, so to speak.

I consider this Sonar warning should be ignored. I don't know how this tool works but I am quite sure there should be a way to configure it to ignore those warnings.

matozoid:
This was already discussed in https://github.com/javaparser/javaparser/issues/242

ftomassetti:
I would reject this PR as discussed in #242. I agree with @ptitjes that it does not necessarily make sense to use this annotation in this context and anyway it would require to use Java 8 while we currently use Java 6

Issue #243, opened 2015-12-14
@Override annotation should be used on any method overriding


matozoid:
Using the @Override annotation is useful for two reasons : 

It elicits a warning from the compiler if the annotated method doesn't actually override anything, as in the case of a misspelling. 

It improves the readability of the source code by making it obvious that methods are overridden.

ftomassetti:
Agreed. How can we easily find all the places where @Override should be used?

ptitjes:
With JavaParser :p

matozoid:
+1 for @ptitjes :D

matozoid:
But: the last five issues were suggested through private email, and I registered them here. They'll be picked up.

ftomassetti:
It will be very nice to do so with JavaParser, however it would mean to find ancestors of classes and that would not be trivial. I would be in awe of whoever would do that and extremely thankful :)

SmiddyPence:
According to IntelliJ that is all of them.  Any further offers?

PR #250, opened 2015-12-15
#243 Missing @Override

SmiddyPence:
According to IntelliJ


ftomassetti:
After a careful review I can say that it makes sense to me :) +1

PR #253, opened 2015-12-21
Issue 252: VoidVisitorAdapter skips some NameExprs

clause:
VoidVisitorAdaper was updated to visit child NameExprs.

In the process of making this change, I found that in several cases nodes were not properly set as the parent of NameExprs thy contain.  This has also been addressed.

Changes have resulted in 4 tests failing.  However, I think that the tests are now incorrect and should be updated.  For example, because ClassOrInterfaceDeclaration now correctly contains it's name, orphan comments are no longer assigned to it.  Similarly, the total number of visited nodes has now increased.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/ConstructorDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/ConstructorDeclaration.java
@@ -143,11 +143,12 @@ public final class ConstructorDeclaration extends BodyDeclaration implements Doc
     }
 
     public void setName(String name) {
-        this.name = new NameExpr(name);
+        setNameExpr(new NameExpr(name));
     }
 
     public void setNameExpr(NameExpr name) {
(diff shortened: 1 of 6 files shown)


ftomassetti:
Thank you for this great patch! Indeed we need to fix those tests to merge this PR. Either you fix it or we will take your PR and add the patches for the tests before merging it.

clause:
I don't know how to correct some of the tests because I'm not sure what behavior would be considered correct.  If you can update, that would be a big help.

ftomassetti:
Sure, I will take a look. Indeed having a new node change how orphan comments are attributed.

matozoid:
@ftomassetti this looks very interesting - can we get the tests fixed?

ftomassetti:
Done in #296

matozoid:
So this one can be closed?

ftomassetti:
I think so

PR #256, opened 2015-12-28
Expose diamond operator

### LINKED ISSUE #255: Diamond operator information is lost
Looks like there's no way to know if an object allocation was done using Java 7's Diamond Operator, this information is simply lost in the parsing.

Would be useful to get this information exposed.

tanob:
Fixes #255 

@ftomassetti please provide any feedback

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/TypeArguments.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/TypeArguments.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright (C) 2015 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.ast;
+
+import com.github.javaparser.ast.type.Type;
+
+import java.util.Collections;
+import java.util.List;
+
+import static com.github.javaparser.ast.internal.Utils.ensureNotNull;
+import static java.util.Collections.unmodifiableList;
+
+public class TypeArguments {
+    public static final TypeArguments EMPTY = withArguments(Collections.<Type>emptyList());
+
+    private final List<Type> typeArguments;
+    private final boolean usesDiamondOperator;
+
+    private TypeArguments(List<Type> typeArguments, boolean usesDiamondOperator) {
+        this.typeArguments = ensureNotNull(typeArguments);
+        this.usesDiamondOperator = usesDiamondOperator;
+    }
+
+    public List<Type> getTypeArguments() {
+        return unmodifiableList(typeArguments);
+    }
+
+    public boolean isUsingDiamondOperator() {
+        return usesDiamondOperator;
+    }
+
+    public static TypeArguments withDiamondOperator() {
+        return new TypeArguments(Collections.<Type>emptyList(), true);
+    }
+
+    public static TypeArguments withArguments(List<Type> typeArguments) {
+        return new TypeArguments(typeArguments, false);
+    }
+}
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
@@ -22,12 +22,11 @@
 package com.github.javaparser.ast.type;
 
 import com.github.javaparser.ast.NamedNode;
+import com.github.javaparser.ast.TypeArguments;
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
 
 import java.util.List;
-
-import static com.github.javaparser.ast.internal.Utils.*;
 
 /**
  * @author Julio Vilmar Gesser
@@ -38,7 +37,7 @@ public final class ClassOrInterfaceType extends Type implements NamedNode {
 
     private String name;
 
-    private List<Type> typeArgs;
+    private TypeArguments typeArguments = TypeArguments.EMPTY;
 
     public ClassOrInterfaceType() {
     }
@@ -52,12 +51,22 @@ public final class ClassOrInterfaceType extends Type implements NamedNode {
         setName(name);
     }
 
+    /**
+     *
+     * @deprecated use the other constructor that takes {@link TypeArguments}
+     */
+    @Deprecated
     public ClassOrInterfaceType(final int beginLine, final int beginColumn, final int endLine, final int endColumn,
-            final ClassOrInterfaceType scope, final String name, final List<Type> typeArgs) {
+                                final ClassOrInterfaceType scope, final String name, final List<Type> typeArgs) {
+        this(beginLine, beginColumn, endLine, endColumn, scope, name, TypeArguments.withArguments(typeArgs));
+    }
+
+    public ClassOrInterfaceType(final int beginLine, final int beginColumn, final int endLine, final int endColumn,
+                                final ClassOrInterfaceType scope, final String name, final TypeArguments typeArguments) {
         super(beginLine, beginColumn, endLine, endColumn);
         setScope(scope);
         setName(name);
-        setTypeArgs(typeArgs);
+        setTypeArguments(typeArguments);
     }
 
     @Override public <R, A> R accept(final GenericVisitor<R, A> v, final A arg) {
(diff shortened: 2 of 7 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
```diff
@@ -53,8 +55,14 @@ public ClassOrInterfaceType(final ClassOrInterfaceType scope, final String name)
     }
 
     public ClassOrInterfaceType(final int beginLine, final int beginColumn, final int endLine, final int endColumn,
-            final ClassOrInterfaceType scope, final String name, final List<Type> typeArgs) {
+                                final ClassOrInterfaceType scope, final String name, final List<Type> typeArgs) {
```
ftomassetti, line 60: Perhaps instead of using a boolean parameter we could either:
- provide an enum with three values (TypeParamsNotSpecified, TypeParamsDerived, TypeParamsNotSpecified)
- make this constructor private and expose two static factory methods

ftomassetti:
Thank you for providing this patch! It looks good and it is great that you took the time to write also a test.
However it needs some refinements because it does not allow to use whitespace in the diamon operator. For example, this is legal Java code:

```
class A {
    List<String> args = new ArrayList<  /*hello*/  >();
}
```

but it cannot be parsed with this patch.

I could parse this case with these changes:

```
diff --git a/javaparser-core/src/main/javacc/java_1_8.jj b/javaparser-core/src/main/javacc/java_1_8.jj
index b711f3a..b0d93f0 100644
--- a/javaparser-core/src/main/javacc/java_1_8.jj
+++ b/javaparser-core/src/main/javacc/java_1_8.jj
@@ -1106,7 +1106,6 @@ TOKEN :
 | < ELLIPSIS: "..." >
 | < ARROW: "->" >
 | < DOUBLECOLON: "::" >
-| < DIAMOND: "<>" >
 }

 /* >'s need special attention due to generics syntax. */
@@ -1816,26 +1815,16 @@ ClassOrInterfaceType ClassOrInterfaceType():
 }
 {
   <IDENTIFIER> {line=token.beginLine; column=token.beginColumn;} { name = token.image; }
-  [
-    LOOKAHEAD(2)
-    typeArgs = TypeArguments()
-    {
-        Object firstTypeArg = typeArgs.remove(0);
-        usesDiamondOperator = firstTypeArg instanceof String && "<>".equals(firstTypeArg);
-    }
-  ]
-  { ret = new ClassOrInterfaceType(line, column, token.endLine, token.endColumn,null, name, typeArgs, usesDiamondOperator); }
+  [ LOOKAHEAD(2) typeArgs = TypeArguments() {typeArgs.remove(0);} ]
+  {
+      usesDiamondOperator = typeArgs != null && typeArgs.isEmpty();
+      ret = new ClassOrInterfaceType(line, column, token.endLine, token.endColumn,null, name, typeArgs, usesDiamondOperator);
+  }
   (
      LOOKAHEAD(2) "." (ann = Annotation()   { annotations = add(annotations, ann);})*  <IDENTIFIER> { name = token.image; }
-     [
-       LOOKAHEAD(2)
-       typeArgs = TypeArguments()
-       {
-            Object firstTypeArg = typeArgs.remove(0);
-            usesDiamondOperator = firstTypeArg instanceof String && "<>".equals(firstTypeArg);
-       }
-     ]
+     [ LOOKAHEAD(2) typeArgs = TypeArguments() {typeArgs.remove(0);} ]
      {
+       usesDiamondOperator = typeArgs != null && typeArgs.isEmpty();
        ret = new ClassOrInterfaceType(line, column, token.endLine, token.endColumn,ret, name, typeArgs, usesDiamondOperator);
        ret.setAnnotations(annotations);
        annotations = null;
@@ -1870,15 +1859,10 @@ List TypeArguments():
 {
    (
      "<" { ret.add(new int[]{token.beginLine, token.beginColumn}); }
-     type = TypeArgument() { ret.add(type); } ( "," type = TypeArgument() { ret.add(type); } )*
+     (type = TypeArgument() { ret.add(type); } ( "," type = TypeArgument() { ret.add(type); } )*)?
      ">"
    )
    { return ret; }
- |
-   (
-     <DIAMOND> { ret.add("<>"); } /* Mark that TypeArguments uses Diamond Operator */
-   )
-   { return ret; }
 }

 Type TypeArgument():
diff --git a/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story b/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
index 71ae94a..189a45c 100644
--- a/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
+++ b/javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
@@ -367,4 +367,14 @@ class A {
     List<String> args = new ArrayList<>();
 }
 When I take the ObjectCreationExpr
+Then the type's uses diamond operator flag should be true
+
+Scenario: Diamond Operator can be parsed also with space and comments
+
+Given a CompilationUnit
+When the following source is parsed:
+class A {
+    List<String> args = new ArrayList<  /*hello*/  >();
+}
+When I take the ObjectCreationExpr
 Then the type's uses diamond operator flag should be true
\ No newline at end of file
```

I sent you this patch as a PR to your patch. Feel free to reject it and implement an alternative solution if you wish.

Thank you again for your contribution!

tanob:
Thanks for the feedback! I've incorporated your changes to the grammar and the scenario in the PR.

One thing I've noticed in the scenario you brought up (`new ArrayList</*hello*/>`) is that with the current implementation the comment lives as an orphan node inside the `ClassOrInterfaceType`. If we move the comment to another position inside the object creation, e.g. `new ArrayList /* hello */<>`, there's no difference in the object tree. I'm assuming this is not an issue so let me know if you think otherwise. (By the way, where is this "comments allowed everywhere" specified in the JLS?)

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
tanob, line 60: I also dislike the boolean because that information should actually be inside some class that defines stuff related to the Type Arguments. I'd prefer to encapsulate the `List<Type> typeArgs` inside a class called `TypeArguments` and have the information about the usage of the Diamond Operator inside it. I don't see a problem having a boolean inside that class.

I didn't go that route because it requires more changes and I wanted to get some feedback.

Should I go this route? Any other ideas?

ftomassetti, line 60: Yes, that would be a nice idea. The only issue is that it would break compatibility and this is something to consider. That said I agree it would be a nicer design.

I would say that I could merge myself a solution which maintains compatibility while for a solution which breaks compatibilityI would ask for a second opinion from another committer. I am not against that but it is just something that we should be more careful with.

ptitjes:
If I may, I'd like to put some perspective here. Before the lazy lists feature, it was possible to have either a null list of type arguments or an empty list of type arguments and this was two semantically correct configurations. This was also correctly reflected by both the grammar and `DumpVisitor`.

This was lost in this commit: https://github.com/tanob/javaparser/commit/78b667cadac59412be67d870c6e5684f8908f909#diff-f31b869d419e1ca7d610eee86be362cfR222

I'd like to advocate the fact that having two fields `typeArgs` and `usesDiamondOperator` could lead a user to believe all the configurations implied are valid (i.e. that [typeArgs = {a, b}; usesDiamondOperator=true] is also a valid configuration).

I'd rather allow `typeArgs` to be `null` again, as this has a valid meaning, or use something like an `Option` of list as this is done in JLaTo.

ftomassetti:
I would advocate against the usage of null having a special meaning, and currently we cannot use Optional because we are Java 6-compatible.

Invalid combinations could be detected in the constructor or prevented using static factory methods.

tanob:
@ftomassetti I've introduced a `TypeArguments` class to remove the boolean and took care to be backwards compatible. Also, I've marked one of `ClassOrInterfaceType`'s constructor as `@Deprecated`.

Let me know what you think.

javaparser-core/src/main/java/com/github/javaparser/ast/TypeArguments.java
```diff
@@ -0,0 +1,57 @@
...
+    private TypeArguments(List<Type> typeArguments, boolean usesDiamondOperator) {
+        this.typeArguments = ensureNotNull(typeArguments);
+        this.usesDiamondOperator = usesDiamondOperator;
+    }
+
+    public List<Type> getTypeArguments() {
+        return unmodifiableList(typeArguments);
+    }
+
+    public boolean isUsesDiamondOperator() {
```
ftomassetti: I would rename this methis _isUsingDiamondOperator_ or _usesDiamondOperator_

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
```diff
@@ -78,8 +87,15 @@ public ClassOrInterfaceType getScope() {
...
-        typeArgs = ensureNotNull(typeArgs);
-        return typeArgs;
+        return typeArguments.getTypeArguments();
+    }
+
+    public TypeArguments getTypeArguments() {
+        return typeArguments;
+    }
+
+    public boolean isUsesDiamondOperator() {
```
ftomassetti: Same here

javaparser-testing/src/test/resources/com/github/javaparser/bdd/parsing_scenarios.story
```diff
@@ -357,4 +357,24 @@ class A {
...
+
+Scenario: Diamond Operator can be parsed also with space and comments
+
+Given a CompilationUnit
+When the following source is parsed:
+class A {
+    List<String> args = new ArrayList<  /*hello*/  >();
+}
+When I take the ObjectCreationExpr
+Then the type's uses diamond operator flag should be true
```
ftomassetti: Perhaps we should add a newline at the end of the file.

It could also make sense to add a test which verify that the diamond operator flag is set to false for:
- the case in which type arguments are not specified (e.g., List)
- the case in which type arguments are specified (e.g., List<String>)

ftomassetti:
To me it looks very good. I added a couple of minor comments but for the rest it seems good to go.

Thank you for the great work!

tanob:
@ftomassetti thanks for the feedback, did the suggested changes and squashed the commits :+1:

ftomassetti:
Great stuff, we hope to see more PRs coming from you :)

PR #257, opened 2015-12-29
issue247: parsing slash asterisk sequence in block comments

### LINKED ISSUE #247: A block comment containing '/*' is not parsed correctly
Hi,
Parsing this file is not working. The parser duplicates some methods:
[BeanConfig.txt](https://github.com/javaparser/javaparser/files/62645/BeanConfig.txt)

ftomassetti:
The issue is that this is invalid:

```
/*/
```

But this is valid:

```
/*      /*/
```

A slash followed by an asterisk can be part of a block comment but a slash, an asterisk and a slash are not a valid block comment.

Fix #247


SmiddyPence:
:+1: 

:+1: for keeping the most used Deque in use on the internet ;-)

ftomassetti:
Thanks @SmiddyPence 

Let's keep Deque alive!

PR #268, opened 2016-02-07
Implement AnnotableNode interface in BaseParameter

fjudd:
I use a method in a project of mine to extract a specific annotation from an _AnnotableNode_. When I passed a _Parameter_, I've noticed that its superclass _BaseParameter_ doesn't implement the _AnnotableNode_ interface even though it provides a _getAnnotations()_ method.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/BaseParameter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/BaseParameter.java
@@ -28,7 +28,7 @@ import java.util.List;
 
 import static com.github.javaparser.ast.internal.Utils.*;
 
-public abstract class BaseParameter extends Node {
+public abstract class BaseParameter extends Node implements AnnotableNode {
     private int modifiers;
 
     private List<AnnotationExpr> annotations;


ftomassetti:
BaseParameter is extended both by Parameter and MultiTypeParameter (which could benefit from some documentation). Would make sense to make those subclasses Annotable instead of the base class? Just in case later we have something extending BaseParameter which is not Annotable. Also, it is really MultiTypeParameter annotable?

ftomassetti:
P.S. Thank you for this PR!

fjudd:
That would definitely make sense. I only implemented it in `BaseParameter` because it already provided a getter and setter for annotations.

matozoid:
@ftomassetti - do you suggest moving the getAnnotable method to the subclasses too?

ftomassetti:
I think it is fine and we should accept the PR. Both kinds of parameters can be annotated so it makes sense to have the interface on the base class. If in the next edition there will be a new type of parameter which can not be annotated we will refactor it :)

matozoid:
Why are pull requests marked invalid when coverage decreases nowadays? :-/ Well, whatever. Accepted :)

ftomassetti:
Accepted? Did you perhaps close it instead of merging it?

matozoid:
Oops!

Issue #271, opened 2016-02-12
TypedNode Interface with getType/setType


LunNova:
Could we get a TypedNode interface?

It would be implemented by all nodes which have getType/setType methods.

```
void setType(Type t);
Type getType();
```

I'm trying to remap types in some code, currently this requires either a large number of instanceof checks, casts and code duplication or costly/slow reflection as there is no interface implemented by everything which has a Type.

If this feature request makes sense/is approved I can make a PR for it.

ftomassetti:
I am in favor of this kind of interfaces: we introduced a few liked the NamedNode and we had a shot at the NodeWithModifiers interface. I agree they helped a lot in processing code.

matozoid:
Sounds like something that should have been there already :) If more methods-that-should-have-been-part-of-a-type are known, please open issues for those too.

Go ahead and submit a PR.

LunNova:
Changed the name to TypedNode as that seemed more consistent with NamedNode which @ftomassetti mentioned.

thejohnfreeman:
Can this be closed?

ftomassetti:
Yes, it makes sense to close this one. Thanks @thejohnfreeman

PR #272, opened 2016-02-13
Add TypedNode interface

LunNova:
Implements #271 by adding a TypedNode interface with the following methods:

```
void setType(Type t);
Type getType();
```

I'm seeing a single test failure in com.github.javaparser.bdd.DumpingTest: https://gist.github.com/nallar/8703500c7516f09cbd76

It's failing before this commit and looks like a line-endings issue, so may just be an issue with my local environment.

I haven't added any new tests for TypedNode as I'm not sure how to add/write tests with JBehave.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
@@ -22,6 +22,7 @@
 package com.github.javaparser;
 
 import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.TypedNode;
 import com.github.javaparser.ast.body.AnnotableNode;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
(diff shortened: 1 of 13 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
```diff
@@ -21,10 +21,7 @@
  
 package com.github.javaparser.ast.body;
 
-import com.github.javaparser.ast.AccessSpecifier;
-import com.github.javaparser.ast.DocumentableNode;
-import com.github.javaparser.ast.NamedNode;
-import com.github.javaparser.ast.TypeParameter;
+import com.github.javaparser.ast.*;
```
LunNova, line 24: Hopefully this was correct. I'm using the JavaParser code style for idea from dev-files.

Might be a good idea to add a .editorconfig? Some files have tabs, some spaces.

javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
```diff
@@ -105,11 +106,8 @@ private static int beginColumnWithoutConsideringAnnotation(Node node) {
     }
 
     private static Node beginNodeWithoutConsideringAnnotations(Node node) {
-        if (node instanceof MethodDeclaration) {
-            MethodDeclaration casted = (MethodDeclaration) node;
-            return casted.getType();
-        } else if (node instanceof FieldDeclaration) {
-            FieldDeclaration casted = (FieldDeclaration) node;
+        if (node instanceof MethodDeclaration || node instanceof FieldDeclaration) {
```
ftomassetti, line 109: What about using `node instanceof TypedNode` instead?

ftomassetti:
It looks good to me and all tests are passing on the server.

Sorry for the late reply and thank you so much for sending this PR!

Let's see if others want to comment on it, if not I will merge it in a few days.

javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
LunNova, line 109: Is that correct? It won't be equivalent to the old code as there are more things implementing `TypedNode` than `Method`/`FieldDeclaration`.

ftomassetti, line 109: You are right, sorry my bad

ftomassetti:
No other comments, let's merge this

PR #279, opened 2016-02-26
Update comment_attribution_scenarios.story same comments problem

waqas19921:
Update comment_attribution_scenarios.story same comments are considered only once so consider some same comments also #278


matozoid:
According to the original issue, a line of code should be changed. This pull request only changes a line in a test. Can you explain what we have here please?

matozoid:
Closing because there is no response. Please create a new PR when required.

PR #280, opened 2016-02-26
JavaDoc Comment Begin Line Equality Checking is missing

waqas19921:
JavaDoc Comment Begin Line Equality Checking is missing that is why if same java doc comments are added at different lines. JavaParser.java only adds a single javadoc comment in the whole compilation unit


ftomassetti:
I think we started discussing if equality should include also the position of the node or not. For example in #132

waqas19921:
this should be included otherwise for same comment at different locations is considered only once in whole compilation unit.

matozoid:
This particular visitor does not check positions, so should not suddenly start doing so for one type of node. You can create your own visitor that does this (by extending this one, for example.)

matozoid:
(continued in #365, where I'm running into exactly the problem you're describing here)

Issue #270, opened 2016-02-10
Don't let TokenMgrError extend Error anymore


sebastiankirsch:
It seems rather awkward that TokenMgrError extends `java.lang.Error`, as it is not indicating "serious problems that a reasonable application should not try to catch" to state the javadoc. Instead, it should probably extend `RuntimeException` (to be backward-compatible).

asheiduk:
Get caught on this too!

PR #285, opened 2016-03-01
Upgrade to JavaCC 6.1.3 and use new TokenMgrException.

asheiduk:
Solves #270 

JavaCC has new option `JAVA_TEMPLATE_TYPE = "modern"`. This option also changes TokenMgrError to TokenMgrException, a RuntimeException.

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -44,6 +44,13 @@
                         </configuration>
                     </execution>
                 </executions>
+				<dependencies>
+					<dependency>
+						<groupId>net.java.dev.javacc</groupId>
+						<artifactId>javacc</artifactId>
+						<version>6.1.2</version>
+					</dependency>
+				</dependencies>
             </plugin>
             <plugin>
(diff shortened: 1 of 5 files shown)


matozoid:
Thanks for your contribution - could you check why one test is failing and update the request?

ftomassetti:
Now the tests seem to pass, but we have a conflict

matozoid:
There is one test failing when compiling on Windows, something is wrong with end of line characters. I'm not sure if it is this PR that is causing the problem though.

ftomassetti:
Oh, I see, sorry about that, I thought it was corrected. Perhaps we should have tests running also on a Windows machine

matozoid:
No, I'm sorry, I was very unclear. I just tested it on our master branch and the bug is there as well. This PR is not causing it (sorry @asheiduk ) and can be merged. I'll make a bug report.

Okay, it cannot be merged due to InstanceJavaParser having been introduced. Will do this tonight.

matozoid:
Redid this change in https://github.com/javaparser/javaparser/pull/328

PR #294, opened 2016-04-07
Update VoidVisitorAdapter.java

JaviRpo:
The validation isn't necessary


ftomassetti:
This seems sound to me. Any comments?

PR #301, opened 2016-05-03
Add missing overrides for ModifierVisitorAdapter

### LINKED ISSUE #300: ModifierVisitorAdapter does not provide default overrides for UnionType and IntersectionType
I'll try to volunteer a patch later this week.

thejohnfreeman:
Closes #300

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java
@@ -788,6 +788,30 @@ public abstract class ModifierVisitorAdapter<A> implements GenericVisitor<Node, 
 		return n;
 	}
 
+    @Override
+    public Node visit(final IntersectionType n, final A arg) {
+        final List<ReferenceType> elements = n.getElements();
+        if (elements != null) {
+            for (int i = 0; i < elements.size(); i++) {
+                elements.set(i, (ReferenceType) elements.get(i).accept(this, arg));
+            }
+            removeNulls(elements);
+        }
+        return n;
+    }
+
+    @Override
+    public Node visit(final UnionType n, final A arg) {
+        final List<ReferenceType> elements = n.getElements();
+        if (elements != null) {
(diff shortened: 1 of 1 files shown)


ftomassetti:
It seems to not compile:

```
[ERROR] /home/travis/build/javaparser/javaparser/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java:[804,12] cannot find symbol
  symbol:   class R
  location: class com.github.javaparser.ast.visitor.ModifierVisitorAdapter<A>
```

ftomassetti:
I think it makes sense but we should fix the return type of the second method

thejohnfreeman:
Yeah, sorry about that. I fixed that and tried to add a test. Let me know if it doesn't meet the standards or is in the wrong place. This is my first time using JBehave.

ftomassetti:
No problem, tests are there to help us catch these things :)

javaparser-testing/src/test/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapterTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+package com.github.javaparser.ast.visitor;
+
+import org.jbehave.core.annotations.Then;
+
+import static org.junit.Assert.assertNotNull;
+
+public class ModifierVisitorAdapterTest {
+
+  @Then("ModifierVisitorAdapter requires no overrides")
```
ftomassetti: I do not understand what this test is doing. It instantiate an object and then it verifies it is not null?

thejohnfreeman:
Yes. The real test is that the construction compiles. If ModifierVisitorAdapter is missing any overrides like it was before this patch, then the construction will not compile. I added the assertion just so that the test has at least one. I'll make this test look however you want, just tell me.

ftomassetti:
I would suggest we just remove the test and merge the change as it is.

Thanks for your help on this one!

thejohnfreeman:
As you wish.

ftomassetti:
Looks perfect. Thanks again!

Issue #314, opened 2016-06-01
ModifierVisitorAdapter tries to modify unmodifiable collection typeArgs


svlassiev:
ModifierVisitorAdapter.visit(final ClassOrInterfaceType n, final A arg) method gets typeArgs from node n and tries to modify it, but ClassOrInterfaceType.getTypeArgs() returns unmodifiable collection.
That causes UnsupportedOperationException.

ModifierVisitorAdapter.java

```
...
312:        final List<Type> typeArgs = n.getTypeArgs();
313:            if (typeArgs != null) {
314:                for (int i = 0; i < typeArgs.size(); i++) {
315:                    typeArgs.set(i, (Type) typeArgs.get(i).accept(this, arg));
316:                }
...
```

PR #315, opened 2016-06-01
Fix for "ModifierVisitorAdapter tries to modify unmodifiable collection typeArgs #314"

svlassiev:
Return type arguments collection without wrapper.


matozoid:
Thanks!

Issue #321, opened 2016-06-15
When visit ArrayCreationExpression initializer is missed in some cases.


svlassiev:
In ModifierVisitorAdapter, GenericVisitorAdapter, VoidVisitorAdapter and DumpVisitor in case of non-null dimensions of ArrayCreationExpression node an initializer of the node is not visited.
This means in this valid code
String[] array = new String[]{"one", "two" , "three"};
visitors miss to visit nodes that represent literals "one", "two" and "three".

matozoid:
Can you open a pull request?

svlassiev:
Actually, bug is still open due to lack of new tests.
And I don't understand how I can reopen pull request if it's already merged.

matozoid:
Oh, I get it now, this is the one I accidentally merged. Let's just close it then.

PR #322, opened 2016-06-15
Fix for "When visit ArrayCreationExpression initializer is missed in some cases. #321".

svlassiev:
Visit array initializer even if array creation expression had dimensions.


ftomassetti:
It would be very nice to have a test for this also

svlassiev:
OK. I'll take a look and add tests.

matozoid:
Wait, I merged this a little early :) Tests are still appreciated!

svlassiev:
Tests are still pending.

PR #334, opened 2016-07-05
issue333: configuring Travis to run on JDK 6, 7 and 8

ftomassetti:
As said in the subject

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,9 @@ language: java
 install: mvn install -DskipTests=true
 script: mvn test
 jdk:
+  - openjdk6
+  - openjdk7
+  - oraclejdk7
   - oraclejdk8
 notifications:
   webhooks:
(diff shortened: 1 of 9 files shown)


ftomassetti:
Apparently this fails because of this:

`<java.version>1.8</java.version>`

Should we change it to:

`<java.version>1.6</java.version>`?

ftomassetti:
Ok, in the code under testing there were some Java 8 features (the diamond operator). It should now compile also on Java 6

ftomassetti:
Please ignore the failure from AppVeyor: in this branch we did not have the configuration for AppVeyor so it fails

matozoid:
A while ago I configured the code to be all Java 8, except the core, including very strict checks that the core is 1.6 compatible, so I vote against these changes :)

ftomassetti:
Oh, I see, sorry for the confusion then!
Would it make sense to compile just the core with the JDK 6?
If not I will just drop this PR

matozoid:
If you can configure it that way, go ahead!

matozoid:
#371 will help with this.

matozoid:
Alright, @ftomassetti , how do we continue here? Everything is Java 8 now.

matozoid:
@ftomassetti , you're cluttering the PR area! :)

ftomassetti:
That is an old habit of mine :)
I am back from holidays. I guess we can just close this one if we moved to Java 8

Issue #141, opened 2015-05-15
methodDeclaration.getJavaDoc() returns null 


kronar:
CompilationUnit cu = JavaParser.parse(f);
List<TypeDeclaration> types = cu.getTypes();
List<ImportDeclaration> imports = cu.getImports();
Assert.assertNotNull(imports);

ftomassetti:
Yes, this is known: lists are initialized with null and nothing prevent to call setters for lists with null.
I wanted to correct this behavior (see #54 and #106) but the PR has been stuck for a while.

Until it gets fixed (if we agree on fix it) you should expect every getters returning a list to treat null or empty lists as equivalent.

kronar:
@ftomassetti I've mean that after parsing class with imports  this list is still empty

ftomassetti:
I think it is the expected behavior, if the class has no imports because instead of assigning empty lists we leave them uninitialized (so with value null)

ftomassetti:
if instead the file contains imports than it is a new bug, in that case please post the source code

kronar:
I've created project that reproduces this issue
https://github.com/kronar/javaparser-issue-141

mvn clean test

ftomassetti:
Thank you for creating the example: best bug report ever!

About the issue: for me the test pass, and the test verifies that the imports are not null and not empty. I tried parsing the given class (https://github.com/kronar/javaparser-issue-141/blob/master/src/main/java/ClassToParse.java) and I can see that the `getImports` return a list of one element, as expected.

Modifying your test to print the imports:

```
    @Test
    public void testIssue141() throws IOException, ParseException {
        CompilationUnit parse = JavaParser.parse(new File("./src/main/java/ClassToParse.java"));
        List<ImportDeclaration> imports = parse.getImports();
        System.out.println("IMPORTS "+imports);
        Assert.assertNotNull(imports);
        Assert.assertFalse(imports.isEmpty());
    }
```

I get:

```
IMPORTS [import com.google.common.base.Optional;
]
```

Am I not understanding the issue or we have a bug that happens just in your environement? Let me know if I can help in figuring out this one, I am curious.

kronar:
Seems it's my fault.
I've fixed test case. 
Javadoc is null. See latest commits

ftomassetti:
I updated the title of the issue. Yes, I can reproduce the issue. You can get the comment using the method getComment() instead of getJavaDoc().

I will look more into this to understand why this happens.

Parth:
[...]

PR #336, opened 2016-07-05
Fix issue 141, JavaDoc getter

matozoid:
This makes getJavaDoc work, as reported in https://github.com/javaparser/javaparser/issues/141

This could use some test, I guess.


matozoid:
And sorry for including an improvement for my previous PR in here ;)

ftomassetti:
+1

Issue #281, opened 2016-02-28
Character encoding in Comments Parser is broken 


danielbobbert:
Fix #41 sought to handle the case where source files are not encoded in UTF-8. Unfortunately the author of the fix chose to use the default Charset for encoding/decoding. Now, if you run the parser on UTF-8 encoded source code that contains special characters, and the platform you are running the parser on happens to have a default characterset that cannot display those special characters, then the comment parser will fail to parse the comment or simply replace the special characters by question marks.

The comments parser should simply use UTF-8 for both encoding and (re-)decoding the source code. This way, the intent of the original fix for #41 is maintained while still making sure that all special characters can be represented and parsed.

```
public CommentsCollection parse(final String source) throws IOException, UnsupportedEncodingException {
    // use UTF-8 to ensure that we can represent all characters
    String encoding = "UTF-8";
    InputStream in = new ByteArrayInputStream(source.getBytes(Charset.forName(encoding)));
    return parse(in, encoding);
}
```

ftomassetti:
I think you can use the method of CommentsParser which let you specify the encoding. Would that help in your case?

danielbobbert:
I tried to. That helps for parsing, but then there is no easy way of inserting the parsed comments into the CompilationUnit because JavaParser.insertCommentsInCu() is private.

danielbobbert:
BTW: I was wondering why you keep encoding strings to streams back and forth anyways. CommentsParser.parse(String) builds an InputStream (with an arbitrary encoding) just for CommentsParser.parse(InputStream, String) to turn that into a Reader again (using the same encoding).

An easier (and more efficient) way would be to offer a method that takes a Reader right way:

```
public CommentsCollection parse(final String source) throws IOException, UnsupportedEncodingException {
    // simply use the Reader-variant to omit all re-encoding
    return parse(new StringReader(source));
}

// existing method, now forwards to the Reader-variant
public CommentsCollection parse(final InputStream in, final String charsetName) throws IOException, UnsupportedEncodingException {
    return parse(new InputStreamReader(in, charsetName));
}

// new method that takes a Reader instead of an InputStream
public CommentsCollection parse(final Reader in) throws IOException {
    boolean lastWasASlashR = false;
    BufferedReader br = new BufferedReader(in);
    ...
```

matozoid:
[...]

PR #337, opened 2016-07-06
Issue 281: simplify stream reading

matozoid:
The second suggestion from: https://github.com/javaparser/javaparser/issues/281 and some small stuff


javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -21,32 +21,34 @@
...
+class CommentsInserter {
     private boolean doNotAssignCommentsPreceedingEmptyLines = true;
     private boolean doNotConsiderAnnotationsAsNodeStartForCodeAttribution = false;
+    
+    CommentsInserter() {
+    }
 
-    public void insertComments(CompilationUnit cu, String code)
-            throws IOException {
+    public void insertComments(CompilationUnit cu, String comments) throws IOException {
```
ftomassetti: It is not 100% clear to me what comments should contain: just the text of the comments? All the original code?

javaparser-core/src/main/java/com/github/javaparser/InstanceJavaParser.java
```diff
@@ -80,7 +88,19 @@ public InstanceJavaParser(Reader reader) {
      *             if the source code has parser errors
      */
     public CompilationUnit parse() throws ParseException {
-        return astParser.CompilationUnit();
+        try {
+            return astParser.CompilationUnit();
+        }finally {
```
ftomassetti: Nitpicking, level master: we could add a space before finally

javaparser-core/src/main/java/com/github/javaparser/InstanceJavaParser.java
```diff
@@ -91,7 +111,11 @@ public CompilationUnit parse() throws ParseException {
      *             if the source code has parser errors
      */
     public BlockStmt parseBlock() throws ParseException {
-        return astParser.Block();
+        try{
+            return astParser.Block();
+        }finally {
```
ftomassetti: Same here

javaparser-core/src/main/java/com/github/javaparser/InstanceJavaParser.java
```diff
@@ -103,7 +127,11 @@ public BlockStmt parseBlock() throws ParseException {
      *             if the source code has parser errors
      */
     public List<?> parseStatements() throws ParseException {
-        return astParser.Statements();
+        try {
+            return astParser.Statements();
+        }finally {
```
ftomassetti: here

javaparser-core/src/main/java/com/github/javaparser/InstanceJavaParser.java
```diff
@@ -115,7 +143,11 @@ public BlockStmt parseBlock() throws ParseException {
      *             if the source code has parser errors
      */
     public Statement parseStatement() throws ParseException {
-        return astParser.Statement();
+        try {
+            return astParser.Statement();
+        }finally {
```
ftomassetti: and there

ftomassetti:
This seems a great patch! It is removing a lot of code. Good job!

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: I'm adding some javadoc here :)

PR #338, opened 2016-07-06
Fix for issue 263: lost orphan comments before initializer block

matozoid:
This does fix the issue, but I'm really unhappy with it. Adding that line feels completely random, and the result causes the comments to move one line down. Comments?


matozoid:
This is not the correct solution.

PR #340, opened 2016-07-07
Upgrade dependencies in poms to latest versions

matozoid:
See title.


ftomassetti:
Are we sure the dependencies used in core are compatible with Java 6? Is there an easy way to test it?

matozoid:
Core has no dependencies :)

ftomassetti:
Very good observation :)

matozoid:
+1?

ftomassetti:
Sure +1!

PR #342, opened 2016-07-08
Propagate removal in ModifierVisitorAdapter

### LINKED ISSUE #331: A more intelligent `ModifierVisitorAdapter`?
The manual includes [a lengthy example on how to remove "dead" context nodes surrounding a removed node](https://github.com/javaparser/javaparser/wiki/Manual#remove--delete-node-from-ast). This isn't an isolated case. There are many nodes that have required children: if the child is removed, the parent should be removed as well. Should we have that behavior by default?
- [ ] `ExpressionStmt`: should be removed if the expression is removed
- [ ] `VariableDeclarationExpr`: should be removed if all children are removed
- [ ] `VariableDeclarator`: should be removed if the id is removed

Going out on a limb, I can think of others:
- [ ] `IfStmt`: should be removed if the "then" clause is removed
- [ ] `DoStmt`: should collapse to a `WhileStmt` if the body is removed
- [ ] `AssignExpr`: should be removed if either side is removed
- [ ] `BinaryExpr`: should collapse to the other side if either side is removed, or be removed if both sides are removed
- [ ] `FieldAccessExpr`: should be removed if the scope is removed
- [ ] `CastExpr`: should collapse to the expression if the type is removed

What do we think? I'll work on a PR unless I hear some objections.

thejohnfreeman:
Closes #331 

Includes import optimizations thanks to IntelliJ. I can revert them if desired.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitorAdapter.java
@@ -21,10 +21,10 @@
  
 package com.github.javaparser.ast.visitor;
 
-import com.github.javaparser.ast.comments.BlockComment;
+import java.util.List;
+
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.ImportDeclaration;
-import com.github.javaparser.ast.comments.LineComment;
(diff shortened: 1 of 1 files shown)


matozoid:
I can't imagine why people would want a broken AST, so thank you for this :)

Issue #245, opened 2015-12-14
Redundant Field Initializers


matozoid:
Java will initialize fields with known default values so any explicit initialization of those same defaults is redundant and         results in a larger class file (approximately three additional bytecode instructions per field).

Issue #246, opened 2015-12-14
Objects should be compared with "equals()"


matozoid:
Using the equality (==) and inequality (!=) operators to compare two objects does not check to see if they have the same values. Rather it checks to see if both object references point to exactly the same object in memory. The vast majority of the time, this is not what you want to do. Use the .equals() method to compare the values of two objects or to compare a string object to a string literal.

PR #351, opened 2016-07-12
Big bunch of minor cleanups

matozoid:
Lots of small things that were annoying me for a while now. Most of it refactored with tools.

Also closes this: https://github.com/javaparser/javaparser/issues/245

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
@@ -132,8 +132,7 @@ public final class PositionUtils {
             if (bl>contained.getBeginLine()) return false;
             if (bl==contained.getBeginLine() && bc>contained.getBeginColumn()) return false;
             if (container.getEndLine()<contained.getEndLine()) return false;
-            if (container.getEndLine()==contained.getEndLine() && container.getEndColumn()<contained.getEndColumn()) return false;
-            return true;
+            return !(container.getEndLine() == contained.getEndLine() && container.getEndColumn() < contained.getEndColumn());
         }
         return true;
     }
--- a/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
@@ -132,6 +132,7 @@ public final class PositionUtils {
             if (bl>contained.getBeginLine()) return false;
             if (bl==contained.getBeginLine() && bc>contained.getBeginColumn()) return false;
             if (container.getEndLine()<contained.getEndLine()) return false;
+            // TODO < or <= ?
             return !(container.getEndLine() == contained.getEndLine() && container.getEndColumn() < contained.getEndColumn());
         }
         return true;
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -21,7 +21,6 @@
 
 package com.github.javaparser.ast;
 
-import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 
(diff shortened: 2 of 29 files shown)


matozoid:
https://github.com/javaparser/javaparser/issues/246 could be closed because none of that type of error are reported.

javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
```diff
@@ -132,8 +132,7 @@ public static boolean nodeContains(Node container, Node contained, boolean ignor
             if (bl>contained.getBeginLine()) return false;
             if (bl==contained.getBeginLine() && bc>contained.getBeginColumn()) return false;
             if (container.getEndLine()<contained.getEndLine()) return false;
-            if (container.getEndLine()==contained.getEndLine() && container.getEndColumn()<contained.getEndColumn()) return false;
-            return true;
+            return !(container.getEndLine() == contained.getEndLine() && container.getEndColumn() < contained.getEndColumn());
```
ftomassetti, line 136: Should it be <=?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -330,9 +328,7 @@ public void setParentNode(Node parentNode) {
 
     protected void setAsParentNodeOf(List<? extends Node> childNodes) {
         if (childNodes != null) {
-            Iterator<? extends Node> it = childNodes.iterator();
-            while (it.hasNext()) {
-                Node current = it.next();
+            for (Node current : childNodes) {
```
ftomassetti, line 331: thank you, my eyes hurt at looking at the old way of doing iterations :)

javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
```diff
@@ -91,7 +91,7 @@ public void setCommentedNode(Node commentedNode)
     {
         if (commentedNode==null)
         {
-            this.commentedNode = commentedNode;
```
ftomassetti, line 94: We should also fix the brackets in this file but this change is good

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
```diff
@@ -1587,8 +1587,7 @@ public void visit(LambdaExpr n, Object arg) {
         printJavaComment(n.getComment(), arg);
 
         List<Parameter> parameters = n.getParameters();
-		boolean printPar = false;
-		printPar = n.isParametersEnclosed();
+		boolean printPar = n.isParametersEnclosed();
```
ftomassetti: we could even make printPar final

ftomassetti:
It looks good to me (I did not look at the JavaCC file)

javaparser-core/src/main/java/com/github/javaparser/PositionUtils.java
matozoid, line 136: Tomorrow I'll write a bunch of tests for Range and Position, I'll pick this up too.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
matozoid: If I start making everything final that can be final, I will submit another huge patch :)

javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
matozoid, line 94: Fixed.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid, line 331: How did we ever live without generics and foreach?

PR #361, opened 2016-07-12
Issue 341: Use Position and Range

matozoid:
A rather huge commit deprecating the use of four integers to indicate a source file position. It introduces Range to replace it, and a Range is two Positions.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/Position.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/Position.java
@@ -18,36 +18,78 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  */
- 
+
 package com.github.javaparser;
 
 import com.github.javaparser.ast.Node;
 
-public class Position {
-    private int line;
-    private int column;
+/**
+ * A position in a source file. Lines and columns start counting at 1.
+ */
+public class Position implements Comparable<Position> {
+    private final int line;
+    private final int column;
 
-    public static final Position ABSOLUTE_START = new Position(Node.ABSOLUTE_BEGIN_LINE,-1);
-    public static final Position ABSOLUTE_END = new Position(Node.ABSOLUTE_END_LINE,-1);
+    public static final Position ABSOLUTE_START = new Position(Node.ABSOLUTE_BEGIN_LINE, -1);
+    public static final Position ABSOLUTE_END = new Position(Node.ABSOLUTE_END_LINE, -1);
 
-    public static Position beginOf(Node node){
-        return new Position(node.getBeginLine(),node.getBeginColumn());
-    }
-
-    public static Position endOf(Node node){
-        return new Position(node.getEndLine(),node.getEndColumn());
-    }
-
-    public Position(int line, int column){
+    public Position(int line, int column) {
         this.line = line;
         this.column = column;
     }
 
-    public int getLine(){
+    /**
+     * Convenient factory method.
+     */
+    public static Position pos(int line, int column) {
+        return new Position(line, column);
+    }
+
+    public int getLine() {
         return this.line;
     }
 
-    public int getColumn(){
+    public int getColumn() {
         return this.column;
+    }
+
+    public Position withColumn(int column) {
+        return new Position(line, column);
+    }
+
+    public Position withLine(int line) {
+        return new Position(line, column);
+    }
+
+    public boolean isAfter(Position position) {
+        if (position.line == Node.ABSOLUTE_BEGIN_LINE) return true;
+        if (line > position.line) {
+            return true;
+        } else if (line == position.line) {
+            return column > position.column;
+        }
+        return false;
+
+    }
+
+    public boolean isBefore(Position position) {
+        if (position.line == Node.ABSOLUTE_END_LINE) return true;
+        if (line < position.line) {
+            return true;
+        } else if (line == position.line) {
+            return column < position.column;
+        }
+        return false;
+    }
+
+    @Override
+    public int compareTo(Position o) {
+        if (isBefore(o)) {
+            return -1;
+        }
+        if (isAfter(o)) {
+            return 1;
+        }
+        return 0;
     }
 }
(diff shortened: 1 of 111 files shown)


javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
+    public final int column;
 
-    public static Position endOf(Node node){
-        return new Position(node.getEndLine(),node.getEndColumn());
-    }
+    public static final Position ABSOLUTE_START = new Position(Node.ABSOLUTE_BEGIN_LINE, -1);
+    public static final Position ABSOLUTE_END = new Position(Node.ABSOLUTE_END_LINE, -1);
+    
+    /** The first position in the file */
+    public static final Position HOME = new Position(1, 1);
```
ftomassetti: If I have an empty file the EOF is at position 1,1?
The newline on an empty line is on column 1?

javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
+    public static final Position UNKNOWN= new Position(0, 0);
```
ftomassetti: Imho using -1 would make clearer that the value is not valid

javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
 
-    public Position(int line, int column){
+    public Position(int line, int column) {
```
ftomassetti: Should we check that the value is either positive or equal to UNKNOWN?

javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
         return this.line;
     }
 
-    public int getColumn(){
+    public int getColumn() {
         return this.column;
     }
+
+    public Position withColumn(int column) {
+        return new Position(line, column);
```
ftomassetti: should we use "this" on line to make more evident what we are doing

javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
+        if (line < position.line) {
+            return true;
+        } else if (line == position.line) {
+            return column < position.column;
+        }
+        return false;
+    }
+
+    @Override
+    public int compareTo(Position o) {
```
ftomassetti: I think it would be simpler to implement isAfter and isBefore referring to compareTo with something like:

```
int partial = Math.signum(this.line - other.line);
return partial == 0 ? Math.signum(this.column - other.colunn) : partial;
```

javaparser-core/src/main/java/com/github/javaparser/Range.java
```diff
@@ -0,0 +1,70 @@
...
+    public Range withBegin(Position begin) {
+        return range(begin, end);
+    }
+
+    public Range withEnd(Position end) {
+        return range(begin, end);
+    }
+
+    public boolean contains(Range other) {
+        return begin.isBefore(other.begin) && end.isAfter(other.end);
```
ftomassetti, line 58: what if the are the same range?

javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -18,36 +18,82 @@
...
+        if (isBefore(o)) {
+            return -1;
+        }
+        if (isAfter(o)) {
+            return 1;
+        }
+        return 0;
+    }
```
ftomassetti: we probably need hashCode and equals

javaparser-core/src/main/java/com/github/javaparser/Range.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public boolean isBefore(Position position) {
+        return end.isBefore(position);
+    }
+
+    public boolean isAfter(Position position) {
+        return begin.isAfter(position);
+    }
```
ftomassetti, line 67: we probably need hashCode and equals

ftomassetti:
Good change! Thank you for all the work you are doing!

javaparser-core/src/main/java/com/github/javaparser/Position.java
matozoid: It's what I reverse engineered. Look at "Scenario: A Class With Block Comments is processed by the Comments Parser" for example, the comments start at character 1, and the top line is also 1. Or am I missing the point?

javaparser-core/src/main/java/com/github/javaparser/Position.java
matozoid: I agree, however these are the values that the code used to set if you set no position. 0,0,0,0. It's more about not wanting to break stuff than about having a good value. I can try setting them to -1,-1 and check what happens if you like?

matozoid: I'm not sure what to do with ABSOLUTE_START/END, by the way.

javaparser-core/src/main/java/com/github/javaparser/Position.java
matozoid: Yes, I thought I was being really smart here ;)

javaparser-core/src/main/java/com/github/javaparser/Position.java
matozoid: I'll see if it can be done without breaking anything.

javaparser-core/src/main/java/com/github/javaparser/Position.java
matozoid: Here I'm also carefully transplanting existing code :-) It would feel better with a bunch of tests in place.

javaparser-core/src/main/java/com/github/javaparser/Range.java
matozoid, line 58: When?

matozoid:
I'm going to assume a +1 from @ftomassetti here :-)

ftomassetti:
Good assumption :)

PR #394, opened 2016-07-20
Generate an enhanced manifest for OSGi support

### LINKED ISSUE #390: Add OSGi metadata to javaparser-core
It would be great if you could add OSGi metadata to javaparser-core to make it simpler to reuse the code in OSGi environments.

A good start would be exporting everything except `com.github.javaparser.ast.internal` with version 0.0.0 - assuming you don't want to do fine-grained OSGi versioning for now.

rombert:
Fixes #390

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -111,7 +111,7 @@
             <plugin>
                 <groupId>biz.aQute.bnd</groupId>
                 <artifactId>bnd-maven-plugin</artifactId>
-                <version>3.2.0</version>
+                <version>${bnd-maven-plugin.version}</version>
                 <executions>
                     <execution>
                         <goals>
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -107,6 +107,30 @@
                     </dependency>
                 </dependencies>
             </plugin>
+            <!-- Generate an OSGi-enabled MANIFEST during the build -->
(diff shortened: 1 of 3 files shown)


matozoid:
All I can say is: it does not seem to break anything, so I like it.
@ftomassetti , @DeepSnowNeeL - have anything to add?

javaparser-core/pom.xml
```diff
@@ -107,7 +107,31 @@
                     </dependency>
                 </dependencies>
             </plugin>
+            <!-- Generate an OSGi-enabled MANIFEST during the build -->
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-maven-plugin</artifactId>
+                <version>3.2.0</version>
```
DeepSnowNeeL: Maybe put the version in a variable like other plugins ?

matozoid: Hm, good for consistency.

rombert: OK, I'll add that. Do you prefer me to push a second commit with the fix or squash the existing one?

rombert:
I've added the plugin version to parent pom, feel free to squash when merging.

matozoid:
Thanks, will piggy back this on version 2.5.1!

rombert:
Great, thank you!

Issue #381, opened 2016-07-18
Change bitfields into EnumSets


matozoid:
Modifiers are giving some bugreports thanks to their lack of typing. An EnumSet would force a contract that avoids these problems.

DeepSnowNeeL:
It could be integrated to my work if you want & released with 3.0 since
it's breaking changes.

Le lun. 18 juil. 2016 12:00, Danny van Bruggen notifications@github.com a
écrit :

matozoid:
I prefer keeping increments as small and focused as possible.

PR #400, opened 2016-07-20
Issue363 builder style

DeepSnowNeeL:
There is still work left around expressions (don't know yet how I will add builder methods to this)

I need to add a lot of tests to cover every method I created.
The firsts commits aren't really usefull since I was since searching how I'll implement this.

Finally, everything is taking advantage of java8 capabilities :)

If you have ideas about new cool methods, feel free to suggest :D

(Main classes to look for modifications : CompilationUnit, every child of TypeDeclaration, every interface starting with "NodeWith" in nodeTypes package)

PS: It includes my other PR #381


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithMembers.java
```diff
@@ -53,10 +54,22 @@ public default FieldDeclaration addField(Class<?> typeClass, EnumSet<Modifier> m
      * @return the {@link FieldDeclaration} created
      */
     public default FieldDeclaration addField(String type, EnumSet<Modifier> modifiers, String name) {
+        return addField(new ClassOrInterfaceType(type), modifiers, name);
```
matozoid: Maybe use "Modifier... modifiers" here? Or maybe that's better for another issue. This PR is big enough as it is.

matozoid:
+1 huge and very useful addition! I say we merge it and fine tune it later.

matozoid:
Thank you very much!

matozoid:
Ahem...

Issue #416, opened 2016-08-08
Array declaration is parsed wrong


Miglecz:
Compiler gives an error on the following (stating the the method is already defined):

```
public void m1(boolean[] boolArray) {}
public void m1(boolean boolArray[]) {}
```

But the parser creates different objects out of it.

`boolean boolArray[]` is not recognised the same way as the other one. It is treated as a simple `PrimitiveType`.

matozoid:
Okay, detecting duplicate methods is indeed the task of the compiler. Syntactically it is correct, so JavaParser parses it.

The same can be said about those two not being the same.

``` java
        CompilationUnit parse = JavaParser.parse(new StringReader("public class Foo {\n" +
                "    boolean[] boolArray;\n" +
                "    boolean boolArray[];\n" +
                "}\n"));

        List<FieldDeclaration> fields = parse.getClassByName("Foo").getFields();

        assertEquals(1, ((ReferenceType)fields.get(0).getType()).getArrayCount());
        assertEquals(1, fields.get(1).getVariables().get(0).getId().getArrayCount());
```

As you can see there was no information lost, it is just stored in another place, attached to either the type or the id (name) of the field.

Is that a good design? I'm not sure.

matozoid:
No more comments?

matozoid:
I guess it's fine then.

matozoid:
This is actually being addressed in #237 now.

PR #420, opened 2016-08-17
Various niceties

matozoid:
Cleaned up the code and added some things while inspecting issue #416


matozoid:
Looks like I accidentally branched off from issue 417. If there are problems with that PR, this one should be changed too.

DeepSnowNeeL:
You should note in the javadoc that getFields and getMethod are cloned
lists ? Maybe use Collection.unmodifiableList to enforce this ?

matozoid:
Hmmm, good point. To be consistent with the rest of the code they should be modifiable, on the other hand all these modifiable lists are giving us a lot of trouble with people adding nodes without setting the parent correctly. I'll take a look.

DeepSnowNeeL:
We could make the getMembers() return a unmodifiableList and add a generic
method addMember(BodyDeclaration<?> member) which will do the link job
correctly.(And do this for every other mutable list accessible today)

It would be breaking changes but goes on with the modifications for 3.0

matozoid:
That's... ugly, but it would avoid the problem :-) I'll open an issue for it.

PR #435, opened 2016-08-24
Some other builder methods and new nodes types for 3.0

DeepSnowNeeL:
I'm working a bit more on code generation so I create new handy methods :)

This time, more focused on statements


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Add an extends to this and automatically add the import
+     * 
+     * @param clazz the class to extand from
+     * @return this
+     */
+    public default T addExtends(Class<?> clazz) {
+        ((Node) this).tryAddImportToParentCompilationUnit(clazz);
+        return addExtends(clazz.getSimpleName());
```
matozoid, line 21: I think you can get rid of the cast by declaring the interface as `NodeWithExtends<T extends Node>`

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/CatchClause.java
```diff
@@ -68,6 +69,7 @@ public CatchClause(final Range range,
 		v.visit(this, arg);
 	}
 
+    // TODO Should we deprecate those ?
 	public BlockStmt getCatchBlock() {
```
matozoid, line 78: Why would we deprecate them?

DeepSnowNeeL: Because of getBody() setBody() inherited by NodeWithBody ? Or we let them as duplicates doing the same thing (strange)?

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
DeepSnowNeeL: Huh ? I don't understand how ?

matozoid: Excuse me, github hid part of my comment, please take another look.

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/CatchClause.java
matozoid: Sure, throw them out, it's violating DRY anyway (we're using the word "catch" twice in catchClause.getCatchBlock() )

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
chmuche: Hello, to the idea to provide new Node Type, why to not pass Node as an Interface ?
I certainly too young on this project to see the side effect of this refractor.
In that way, you can do this, 
`public interface NodeWithExtends<T> implements Node {`

I say this because, if I create a new class like that 

```
public class Foo implements NodeWithExtends<EnumDeclaration> {
  public List<ClassOrInterfaceType> getExtends(){/*implement*/}
public T setExtends(final List<ClassOrInterfaceType> extendsList){/*implement*/}
}
```

Your default method throw a ClassCastException

DeepSnowNeeL: Actually no : 
The interface will have a template type which needs to be a subtype of `Node`(so every T needs to extends `Node`)
But 
This interface can be applied to any type (and his template type can't apply any restriction)

:(

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/CatchClause.java
DeepSnowNeeL: I prefer to keep them deprecated untill we reach like another major version, I don't like when methods just doesn't exist anymore without providing information, this way we let them fix it on the long run.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
```diff
@@ -0,0 +1,37 @@
...
+     * 
+     * @param name the name of the type to extends from
+     * @return this
+     */
+    @SuppressWarnings("unchecked")
+    public default T addExtends(String name) {
+        ClassOrInterfaceType classOrInterfaceType = new ClassOrInterfaceType(name);
+        getExtends().add(classOrInterfaceType);
+        classOrInterfaceType.setParentNode((Node) this);
+        return (T) this;
```
chmuche, line 35: `return (T) this;` will  work only if **this** is the same type of **T**, your interface do not impose this.
and `(Node) this` will work only if **this** extend Node or you can't be sure.

DeepSnowNeeL, line 35: Thoses interfaces are for internal use only.
They were created to avoid code duplication by taking advantage of default
methods.
There is no point of using them outside of the library or using them on a
non-subtype of Node.
Unfortunately package private exists for types but not library private :p

matozoid: Actually, user code could instanceof for one of these interfaces on a node if it is for example only interested in annotations. But yes, there is no other use for them. Unless some people want to implement their own nodes, which I wouldn't recommend.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
matozoid: But why would you want to apply it to something other than a Node?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/VariableDeclarationExpr.java
```diff
@@ -47,106 +49,140 @@
...
+        List<VariableDeclarator> vars = new ArrayList<>();
+        vars.add(new VariableDeclarator(variableName));
+        setVars(vars);
+    }
+
+    public VariableDeclarationExpr(final Type type, VariableDeclarator var) {
+        setType(type);
+        List<VariableDeclarator> vars = new ArrayList<>();
+        vars.add(var);
+        setVars(vars);
```
chmuche: Why not using Arrays.asList(var) ? or asList(var) with Arrays.asList in a static import

DeepSnowNeeL: Good point, I tend not to use this as it's a bit of an obscure method, being able to use it with an array or values isn't right to me.
I'll rework this

matozoid:
Okay, there are still one or two remarks open. Nothing dramatic, nothing that cannot go in an alpha release, so I'm pulling this.

Issue #353, opened 2016-07-12
Generate visitors from the AST classes


matozoid:
Instead of manually patching the visitors every time we make some changes (and regularly forgetting to...) we should generate the visitors from code.

Work in progress over here, where I use @ftomassetti 's symbol solver to build the AST model, then use my own code generator to generate the visitors: https://github.com/matozoid/javaparser/tree/generate-visitors

matozoid:
Will fix #220

matozoid:
Might fix #236

matozoid:
Will prevent #252 from happening again

matozoid:
After some request for specific visitors, it seems like a good idea to make this a public tool.

A user might specify:
- the visitor's class name
- the visitor's argument that is passed into every method
- if it's a void visitor
- if it's not a void visitor, what the return type is
- what happens to the visitor before generating the methods (adding custom methods)
- what happens to the visitor after generating the methods (who knows?)

then a callback gets called for every node type, with
- the node class
- the names and type of the fields of this node
- the setters and getters for the fields of this node
- the visit method body, for the user to fill
- a map with information about all node types in case the user needs it

This will generate a CompilationUnit that can be written to disk and compiled.

matozoid:
Also, since we're getting more and more repetitive code in the nodes, we could generate parts of those classes. Maybe with annotations, like:
@GenerateDefaultConstructor

@GenerateProperty(nullable, ...)

So the generator can read from the javaparser-core source directory, find all annotated methods, replace them and write them back.

ftomassetti:
I am really looking forward to this one!

> After some request for specific visitors, it seems like a good idea to make this a public tool.

Were you thinking about having it as a class in JavaParser or as the third project under the JavaParser umbrella?

matozoid:
Okay, @DeepSnowNeeL is actually working on this but I'll give you my personal thoughts: something that generates part of JavaParser seems good to have away from the source of JavaParser. I was thinking of a separate module so that at some time in the future a single build will build the generator, generate the visitors and other code, build the parser, and build the core.

matozoid:
Here we go again: https://github.com/matozoid/javaparser/commits/issue_353_generate_visitors

ftomassetti:
But if you use it JavaSymbolSolver does it mean you would java JavaParser depending on JavaSymbolSolver? Or would this be a separate module/project?

matozoid:
This time I'm not using JSS.

[...]

PR #440, opened 2016-08-29
Issue 430 clean up annotation handling

### LINKED ISSUE #430: The Annotation inner TypeReference are not visited in Adapter
Java8 introduce a new ElementType in the Annotation Target (`ElementType.TYPE_PARAMETER`)
[JSL 9.6.4.1. @Target  ](https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.4.1)

In VoidAdapter<A> the method :

```
@Override public void visit(final ReferenceType n, final A arg) {
        visitComment(n.getComment(), arg);
        n.getType().accept(this, arg);
    }
```

Don't visit the annotations

A little workaround is to override this method like this :

```
@Override
    public void visit(final ReferenceType n, final A arg) {
        for (AnnotationExpr annotation : n.getAnnotations()) {
            annotation.accept(this, arg);
        }
        super.visit(n, arg);
    }
```

this Issue is in tree Adapter Void, Generic, Modifier but not in DumpVisitor

matozoid:
Fixes #430

I've double checked the visiting of annotations for all Types. Type has eight subclasses: ReferenceType, ClassOrInterfaceType, UnionType, UnknownType, PrimitiveType, WildcardType, VoidType and IntersectionType. With some research I came to the conclusion that all of these types except UnknownType can be annotated, so I made sure their annotations were visited in all visitors.

I did not look at other nodes with annotations, and I think I should continue work on #353 to make sure there are no errors in these visitors.

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -22,12 +22,6 @@
             <comments>A business-friendly OSS license</comments>
         </license>
     </licenses>
-    <dependencies>
-        <dependency>
-            <groupId>org.jetbrains</groupId>
-            <artifactId>annotations</artifactId>
-        </dependency>
-    </dependencies>
 
     <properties>
         <java.version>1.8</java.version>
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -22,6 +22,12 @@
             <comments>A business-friendly OSS license</comments>
         </license>
     </licenses>
+    <dependencies>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+        </dependency>
+    </dependencies>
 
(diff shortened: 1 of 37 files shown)


ptitjes:
Hey, while you're at that you might want to support annotations on ellipsis like this:

https://github.com/ptitjes/jlato/blob/d82d1/src/test/resources/org/jlato/integration/TypeDecls/method-with-annotated-ellipsis/formatted.txt

Caught that while parsing some files of the openjdk test suite some twenty day ago... (cf. JLS 8.4.1)

matozoid:
Cripes! Makes sense though. Thanks!

matozoid:
Appveyor fails for an odd reason - I cannot check this right now.

Issue #136, opened 2015-05-10
Continuous parsing.


artur-shaik:
Does we have continious parsing or is it in the plan, that will give ability to parse files, even with errors?

ftomassetti:
I think that writing an error-tolerant grammar is in general quite difficult, especially for a grammar quite ambiguous as the Java one. However we could introduce some functionality in this sense. We use JavaCC for the parser and here there is some information on error recovery: https://javacc.java.net/doc/errorrecovery.html

talios:
I've not actually played with it too much, but I wonder how some form of RxJava integration could work here - maybe not in exactly the way @artur-shaik maybe thinking - either as a direct part of javaparser or a separate artifact/project ( so as to not tie an RxJava dependency to the core library ).

One could have say an `Obversable<CompilationUnit> observeCompilation(File file)` , which whenever the `file` changes ( file watchers, polling, ... ) the parser gets executed and a new `CompilationUnit` instance is emitted.

Rather than refer to it as _continuous parsing_ you'd call this _reactive parsing_ - subscribe to compilation events and process the results down a reactive flow.

That could be quite powerful.

ftomassetti:
Well, if you are interested in that I wrote an incremental Java parser (https://github.com/ftomassetti/JavaIncrementalParser). I have not played with it much but it should be able to parse a reasonable amount of Java 7 code. It is written in Scala using the papa-carlo library for incremental parsing.

artur-shaik:
Ok, thanks for link to javacc docs. I played with those error skip codes, and I see that there is ability to catch ParseException and skip some tokens until semicolon or close braces. This method allow skip erroneous lines or may be blocks. Withal we could collect all catched exceptions to report it later.

What about @talios said. Sounds powerful and complex :smile: . For now I wish to just have ability to   parse whole class as it has no errors.

wsdjeg:
mark
EDIT:https://github.com/artur-shaik/vim-javacomplete2/issues/185

matozoid:
A start has been made on this.

wsdjeg:
Nice! And thanks for your work.

matozoid:
You can track it under this label: https://github.com/javaparser/javaparser/labels/error%20recovery

artur-shaik:
@matozoid , thank you very much for mention it. I'm quite busy now, so I can not track contributions related to this field.
I will try to check it later.

matozoid:
[...]

PR #447, opened 2016-08-31
Issue 157 enhance parser frontend

matozoid:
#157 suggests a new interface for JavaParser. This PR contains a reduced version of the suggestion.

There are two interfaces:
1. the static methods on JavaParser form the "quick & dirty" interface
2. the methods on an instance of JavaParser are the "power user" interface.

Interface 1 has not been changed much. There is a different exception, and that exception can hold multiple parsing problems.

For interface 2:
- what we're parsing (compilation unit, expression, block...) is now specified as a parameter instead of using a different parameter. This happens to be a single-method interface, so now any production of ASTParser can be used.
- where the source code is coming from is now determined by JavaCC's Provider interface. The Providers factory class has several methods for creating Providers.
- parsing doesn't always fail on the first parsing problem anymore. It still does in many cases, but this can be improved. This way we can collect the problems and return them, instead of returning only one. Additionally the parse result, the node, is returned if parsing was completed. Both can be returned from a single parse call: a list of problems and a result node, which lays the ground work for continuous parsing (#136)

Things that are significantly different from #157:
- There is no reference to the origin of the source code (the path() suggestion.) This can be done, but I don't know how useful it is (we can indicate a file, but how do we say that the source code came from a string or an input stream?) It can always be added on later.
- Problems don't have a severity. I can't imagine needing it right now since everything seems to be an error. Maybe when this feature is used extensively there will be a need.

Things to do:
- Use error recovery to turn more fatal problems into recoverable errors, see #136 
- Write tests...

Thanks to @ptitjes for thinking this up!

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
@@ -5,7 +5,7 @@ import java.util.List;
 /**
  * Thrown when parsing problems occur during parsing with the static methods on JavaParser.
  */
-public class ParseProblemException extends Exception {
+public class ParseProblemException extends RuntimeException {
     /** The problems that were encountered during parsing */
     public final List<Problem> problems;
     
--- a/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
@@ -2,10 +2,23 @@ package com.github.javaparser;
 
 import java.util.List;
 
+/**
+ * Thrown when parsing problems occur during parsing with the static methods on JavaParser.
+ */
 public class ParseProblemException extends Exception {
-  public final List<Problem> problems;
-
-  public ParseProblemException(List<Problem> problems) {
-    this.problems = problems;
-  }
+    /** The problems that were encountered during parsing */
+    public final List<Problem> problems;
+    
+    ParseProblemException(List<Problem> problems) {
+        super(createMessage(problems));
+        this.problems = problems;
+    }
+    
+    private static String createMessage(List<Problem> problems){
+        StringBuilder message = new StringBuilder();
+        for(Problem problem: problems){
+            message.append(problem.toString()).append("\n");
+        }
+        return message.toString();
+    }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
@@ -1,0 +1,11 @@
+package com.github.javaparser;
+
+import java.util.List;
+
+public class ParseProblemException extends Exception {
+  public final List<Problem> problems;
+
+  public ParseProblemException(List<Problem> problems) {
+    this.problems = problems;
+  }
+}
--- a/javaparser-core/src/main/java/com/github/javaparser/Providers.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/Providers.java
@@ -10,7 +10,7 @@ import java.nio.file.Path;
  * Factory for providers of source code for JavaParser.
  * Providers that have no parameter for encoding but need it will use UTF-8.
  */
-public abstract class Providers {
+public final class Providers {
 	public static final Charset UTF8 = Charset.forName("utf-8");
 
 	private Providers() {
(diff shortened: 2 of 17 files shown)


ptitjes:
Wow!!!
I'm so mixed fellings: glad to finally see thing coming up together, so sad to not have been given the opportunity to make those happen...
Thanks for the kudos anyway!
(Shit, we are on par on the error reporting front now! I feel like I have to finish that error recovery very quickly...)

matozoid:
@ptitjes it's funny how you suddenly started working on jlato again when I picked up development here :)

matozoid:
I just realized that if we want self-describing and rereadable providers, that the Provider interface is too bare.

matozoid:
Stuck issue #448 in here too.

chmuche:
Hello, the feature look great, but I have a question, You always check if on of the method `setSource` is called before parsing with the method `checkSourceSet` so today this code break the compatibility.
So I think we can use `setSource` as factory method, in that case no need to check.
I think somethings like that :

```
/**
     * Instantiate the parser. Note that parsing can also be done with the static methods on this class.
     * Creating an instance will reduce setup time between parsing files.
     */
    private JavaParser(Provider provider) {
        this.provider= provider;
        if(astParser==null) {
            astParser = new ASTParser(provider);
        }else{
            astParser.ReInit(provider);
        }
    }

    public static JavaParser setSource(Provider provider) {
        Objects.requireNonNull(provider, "The provider can't be null");
        return new JavaParser(provider);
    }

    public static JavaParser setSource(Reader reader) {
        Objects.requireNonNull(reader, "The reader can't be null");
        return setSource(new StreamProvider(reader));
    }

    public static JavaParser setSource(InputStream input) throws IOException {
        Objects.requireNonNull(input, "The InputStream can't be null");
        return setSource(new StreamProvider(input));
    }

    public static JavaParser setSource(InputStream input, Charset encoding) throws IOException {
        if (input == null){
            throw new IOException("The InputStream can't be null");
        }
        Objects.requireNonNull(encoding, "The encoding can't be null");
        return setSource(new StreamProvider(input, encoding.name()));
    }

    public static JavaParser setSource(File file) throws IOException {
        Objects.requireNonNull(file, "The file can't be null");
        return setSource(new FileInputStream(file));
    }

    public static JavaParser setSource(File file, Charset encoding) throws IOException {
        Objects.requireNonNull(file, "The file can't be null");
        Objects.requireNonNull(encoding, "The encoding can't be null");
        return setSource(new FileInputStream(file), encoding);
    }

    public static JavaParser setSource(String source) {
        Objects.requireNonNull(source, "The source can't be null");
        return setSource(new StringReader(source));
    }
```

javaparser-core/src/main/java/com/github/javaparser/Providers.java
```diff
@@ -0,0 +1,57 @@
...
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.OpenOption;
+import java.nio.file.Path;
+
+/**
+ * Factory for providers of source code for JavaParser.
+ * Providers that have no parameter for encoding but need it will use UTF-8.
+ */
+public abstract class Providers {
```
chmuche: Why **abstract** and not **final** ?

matozoid: Oh woops...

javaparser-core/src/main/java/com/github/javaparser/ParseProblemException.java
```diff
@@ -0,0 +1,24 @@
+package com.github.javaparser;
+
+import java.util.List;
+
+/**
+ * Thrown when parsing problems occur during parsing with the static methods on JavaParser.
+ */
+public class ParseProblemException extends Exception {
```
chmuche: Me when I code I prefere to create RuntimeException, Or using an another Pattern because with exception we can't do composition.If you want to talk about, PM me on gitter :smile: otherwise it's fine :smile_cat:

matozoid: Nah, I think I agree, it's the Java 8 world and we want to use the functional programming features, and checked exceptions are hated there.

DeepSnowNeeL: Yep I agree, checked exceptions are annoying :p

matozoid:
@chmuche - I don't really get your other comment. Did you notice that setSource is gone? It was only there in alpha-1 and wasn't a great design, introducing state where it shouldn't have been.

I like Objects.requireNotNull, will make an issue for discussing nullity stuff.

matozoid:
Needs more work on exception handling, but this gets merged right now because it is causing merge problems.

Issue #109, opened 2015-02-26
Conditional expression should not be accepted as left-hand side of assignment


ptitjes:
Per JLS8 §15.26:

```
AssignmentExpression:
    ConditionalExpression
    Assignment
Assignment:
    LeftHandSide AssignmentOperator Expression
LeftHandSide:
    ExpressionName
    FieldAccess
    ArrayAccess
```

Indeed, the grammar has been copied originally from the PMD one which does not build a full parse tree. Hence the comment in the grammar:

```
/*
 * This expansion has been written this way instead of:
 *   Assignment() | ConditionalExpression()
 * for performance reasons.
 * However, it is a weakening of the grammar for it allows the LHS of
 * assignments to be any conditional expression whereas it can only be
 * a primary expression.  Consider adding a semantic predicate to work
 * around this.
 */
```

ptitjes:
A test case on that branch
https://github.com/ptitjes/javaparser/commits/zz-issue-109

Commit:
https://github.com/ptitjes/javaparser/commit/744a4a30f81e8612878e86f1cfebb7d3a35ae322

ptitjes:
If everyone is OK with that, I add this to 3.0 milestone.

matozoid:
@ptitjes - can we grab your fix, or would you like us to redo it?

ptitjes:
@matozoid hmm... what fix ? :)

matozoid:
Oh... Well, can we grab your test? ;)

ptitjes:
Sure :)

matozoid:
I wonder if this is best fixed by introducing a "PrimaryExpression" node that has the allowed node types as children.

matozoid:
Thanks for reporting this!

PR #450, opened 2016-09-03
Fix for #109

matozoid:
#109 do not accept non-primary expressions on the left hand side of assignments.

The fix is to check the type of the expression going in.

I'm thinking that this could be caught in the AST instead.

Sorry for the unclear commits thanks to reformatting: what is added is a type check in the grammar.


matozoid:
The fix here is incorrect and ugly, so this doesn't get pulled.

Will reopen another time.

Issue #457, opened 2016-09-10
ArrayCreationExpr.arrayCount always stays 0


matozoid:
Since we already have a list with all initializers, we can just as well delete this field.

See #237

ptitjes:
You can have both dimension-expressions and dimensions in a ArrayCreationExpr (see JLS8 section 15.10.1).
For instance, `new int[10][10][][]` is a valid array creation expression.

matozoid:
Yes, new int[10][10][][] would have arrayCount=2. I misunderstood this functionality.

PR #459, opened 2016-09-10
Issue 237 arraytype

### LINKED ISSUE #237: Refactor ReferenceType
Refactor ReferenceType to do one thing (contains annotations) instead of two (containg annotations and array modifiers).

This is a follow-up of #187

matozoid:
Fixes #237 part 1: ReferenceType no longer holds array information, this is now the task of a wrapping of ArrayTypes.


javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -153,6 +153,16 @@ final class ASTParser {
 	    return ret;
     }
 
+    private ReferenceType juggleArrayTypes(Range range, Type type, int arrayCount, List<List<AnnotationExpr>> arrayAnnotations) {
+        if (arrayCount == 0) {
+            return new ReferenceType(range, type);
+        }
+        for (List<AnnotationExpr> annotations: arrayAnnotations) {
+            type = new ArrayType(range, type, annotations);
```
ptitjes: Annotations are parsed from left to right, thus instantiations of ArrayTypes should happen from right to left. (c.f. example in JLS section 9.7.4)

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
```diff
@@ -526,7 +527,13 @@ public void visit(final ReferenceType n, final Object arg) {
 			}
 		}
 		n.getType().accept(this, arg);
-		printArrayBraces(n, arg);
+	}
+
+	@Override
+	public void visit(ArrayType n, Object arg) {
```
ptitjes: Unfortunately, if the ArrayType wrapping order is fixed (see comment bellow in the grammar), this can't be handled this easily. This must be handled with a loop (or a collecting inner-visitor) unwrapping all inner ArrayTypes up to a (non-Array)Type.

For instance, for a type 
`ArrayType(annotations=@B, ArrayType(annotations=@C, PrimitiveType(annotations=@A, primitive=int)))`
should be printed as
`@A int @B [] @C []`

matozoid:
@ptitjes - I will write out exactly what the code is doing - I checked that it is correct but it's good to double-check.

matozoid:
Fixes #457 now, because things are completely refactored :)

matozoid:
This should fix everything in #237 and #112 except for array braces on methods.

ptitjes:
@matozoid I still believed that it is incorrect. You should had some tests that check the structure of some parsed array types.

Because, it seems to me that you currently parse `@A int @B [] @C []`
as

```
ArrayType(annotations=@C, ArrayType(annotations=@B, PrimitiveType(annotations=@A, primitive=int)))
```

where you should parse it as

```
ArrayType(annotations=@B, ArrayType(annotations=@C, PrimitiveType(annotations=@A, primitive=int)))
```

A q&d fix would be to **prepend** to the annotations array (instead of appending) when you parse the array dimensions.

Also, on another topic you should read the different sections of the JLS about array creation expressions. Indeed, I find your wording poorly chosen for `DimensionedArrayType` as it is not a type. It implies a type contribution (the brackets) but the integer expression inside the brackets are not in the type.

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -153,6 +153,23 @@ final class ASTParser {
 	    return ret;
     }
 
+    private ReferenceType juggleArrayTypes(Range range, Type type, int arrayCount, List<List<AnnotationExpr>> arrayAnnotations) {
+        if (arrayCount == 0) {
+            return new ReferenceType(range, type);
```
ptitjes: `type` can only be a `ClassOrInterfaceType`. You should cast it to `ReferenceType` rather than wrapping it and both `ClassOrInterfaceType` and `ArrayType` extends `ReferenceType`.

javaparser-core/src/main/java/com/github/javaparser/ast/type/ArrayType.java
```diff
@@ -0,0 +1,26 @@
...
+import com.github.javaparser.ast.visitor.GenericVisitor;
+import com.github.javaparser.ast.visitor.VoidVisitor;
+
+import java.util.List;
+
+/**
+ * Every type followed by [] gets wrapped in an ArrayType for each [].
+ */
+public class ArrayType extends ReferenceType {
+    public ArrayType(Range range, Type type, List<AnnotationExpr> annotations) {
```
ptitjes: `ArrayType` can't have annotations on its own. The `annotations` field of `ReferenceType` should be moved down to `ClassOrInterfaceType` (which should extend `ReferenceType`).

The JLS states:

> There are four kinds of reference types: class types (§8.1), interface types (§9.1),
> type variables (§4.4), and array types (§10.1).

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: Double-checked this and you're correct. I blame late night checkins. The funny thing is that I had it from right to left the first time around, but decided to change it for some reason.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/DumpVisitor.java
matozoid: Fixed!

javaparser-core/src/main/java/com/github/javaparser/ast/type/ArrayType.java
matozoid: Okay, this is going over my head: do you mean that ReferenceType can be abstract, and ClassOrInterfaceType, ArrayType and TypeParameter should extend it? This means that ReferenceType does not need its type field anymore? That actually seems to make a lot of sense.

Also: I thought that introducing ArrayType was a good thing because we could have annotations everywhere? I think I don't understand.

javaparser-testing/src/test/resources/com/github/javaparser/bdd/visitor_scenarios.story
```diff
@@ -73,4 +73,4 @@ Given a CompilationUnit
 Given a VoidVisitorAdapter with a visit method that asserts sensible line positions
 When file "JavaConcepts.java" is parsed
 When the CompilationUnit is visited by the PositionTestVisitor
-Then the total number of nodes visited is 1427
```
ftomassetti, line 76: This test was intended as a sort of big alarm telling how if we are dropping tons of nodes without realizing but we end up having to touch it at every change. Should we either make it more flexible (like checking the nodes are 1500 > x > 1300) or remove it?

matozoid:
@ptitjes about DimensionedArrayType: what would you suggest? Leave off "Type" and call it ArrayDimension or ArrayCreation or so?

javaparser-testing/src/test/resources/com/github/javaparser/bdd/visitor_scenarios.story
matozoid: Although it is freaking annoying, it did make me notice today that the node counter was broken.

Wait, that's not important :) So yeah, I guess we can remove it.

(edit: it's in the latest commit)

javaparser-core/src/main/java/com/github/javaparser/ast/type/ArrayType.java
matozoid: Added a commit that does step 1: make ReferenceType "just a base class" so the cast works. Saves a huge amount of nodes in the AST!

ptitjes:
@matozoid  Huh! I thought I had open the `TypeParameter` package and inheritance issue on JLaTo... :P

matozoid:
@ptitjes - I read through JLS 10.3, and from that I deduce that ArrayCreationExpr didn't need to be changed at all. So I'll roll that stuff back.

ptitjes:
@matozoid I don't know about ArrayCreationExpr (didn't look to be precise) but the things you did yesterday for ArrayType seem correct to me. But tests cases to test the wrapping are mandatory in my opinion as this is intricate.

matozoid:
Yes, I'll add one until I finish my fancy serializer thingy.

About ArrayCreationExpr: I added DimensionedArrayType which you didn't like at all. This was an attempt to have array creation also use a nested structure like ArrayType and ArrayAccessExpr. However, the JLS says nothing about nesting ArrayCreationExpr's, so I guess I was a little too enthusiastic there, and that code was fine as it was. (Although I still don't like the combination of an array of dimensions and an integer arrayCount.)

matozoid:
I like how ArrayCreationExpr now looks, but I don't like the package for ArrayCreationLevel.

ptitjes:
@matozoid In response to https://github.com/javaparser/javaparser/pull/459#issuecomment-246446116

We get back to the same debate. Should JavaParser's AST structure represent the syntax as closely as possible (1) or (2) should it provide an abstract view over the code ?

I tend for (1) and think (2) is to be provided by an upper layer (already semantics IMHO).

In the context of ArrayCreationExpr, (1) means having separate informations (for dimension expressions and dimensions) in the AST for the user to be able to manipulate precisely the read code. (2) means having some methods and facades that aggregates the various informations to provide semantic data (namely the type of the instantiated object). The semantic part could go as far as javaparser-model and java-symbol-solver go.

For comparison:
- Javac, through the com.sun.tree API, provides a view like (2). Indeed, it is interested in compilation only.
- Eclipse/IDEA (and other IDE) provide both (1) and (2) by an additional layer of functionality (namely the type model representation). Indeed, they have to spit back the code as it were.

matozoid:
The idea in my mind is to, yes, follow the Java syntax as described in the JLS closely, and offer a bunch of helper methods - like the builder methods added by @DeepSnowNeeL, or the ones I added issues for yesterday. It's really a question of what the users want to do with it, or what goals we have with the library.

But what triggered this reply? The ArrayCreationExpr remark? That was simply me complicating stuff where nobody asked for it, so I half-reverted half-simplified it.

matozoid:
Right, so 
`@A int @B [] @C [] a;` parses fine now, but the rather obvious case of 
`int a @A [] @B[];` does not! Annotations are not supported here. Am I missing something here?

matozoid:
Updated more or less like what I was saying in #237 , meaning that we can now deal with array braces in the strangest places.

I think there's more thinking to be done here. Currently the grammar has some trouble passing around a PrimitiveType as a ReferenceType, since the grammar itself does not use ArrayType. Maybe it should.

matozoid:
See #237 for discussion.

If there are no serious issues, I'll pull this next sunday, october 2nd.

matozoid:
@DeepSnowNeeL - I've changed some getType()s into getElementType()s so I could get ahead with this feature. This is not the same thing. We probably need to look at that, but my head hurts :(((

matozoid:
Is anybody still awake? :D

matozoid:
#465 and #466 are indirectly implemented. There are methods on ArrayType that convert between (an element Type and a List<ArrayBracketPair>) and (a Type optionally wrapped in ArrayType(s) )

Issue #454, opened 2016-09-04
Inspect current Windows compile problem


matozoid:
It compiles fine on Windows, but not on our appveyor build tool.

matozoid:
It's something with M2_HOME, @ftomassetti - do you feel like looking at it?

ftomassetti:
There are 2 ways to fix this:
- fix the M2_HOME path
- stop installing our own Maven version

I went for option 1. There is one test failing but it seems legit

matozoid:
Thanks!

PR #462, opened 2016-09-11
issue454: appveyor: setting M2_HOME

ftomassetti:
See #454 

Fixing the path to M2_HOME.

Notes: we install our own maven and one test is still failing


matozoid:
Thanks!

Issue #424, opened 2016-08-19
Reformat all source code according to the formatting rules.


matozoid:
MAKE SURE nobody is working on anything, or merging will be hell!

rosrossito:
Matozoid, hello,
may I join this project?
Can you assign this task to me? I would like to try to make it.

matozoid:
Sure, this weekend I'll pull all the pull requests, so then would be a good time to do this.

matozoid:
@DeepSnowNeeL can you finalize #435 before Sunday?

DeepSnowNeeL:
Well you can include it like this, I'll continue to propose PR every time I use javaparser and find a way to do better with code building :)
Could you check that the eclipse formatting is updated and does the right thing ? I think I always break formatting... :(

Thanks !

matozoid:
@DeepSnowNeeL - good point. I personally only format the lines I touch, which is a pain.
@rosrossito - could you first investigate the current formatters and see if you can get them to format stuff identically?

rosrossito:
Ok. I will try.

2016-09-01 13:49 GMT+03:00 Danny van Bruggen notifications@github.com:

rosrossito:
Danny,

could you clarify the task more precisely?

I mean in what modules current formatters are located?

Do you have already this formatting rules?

2016-09-01 15:24 GMT+03:00 Ruslan Omelchenko <
ruslan.omelchenko.irtcits@gmail.com>:

matozoid:
Check https://github.com/javaparser/javaparser/wiki/Coding-Guidelines

matozoid:
I think Intellij should have the option to "import *" turned off.

rosrossito:
I followed the instruction given in guiedlines, but I am not shure that I
made it properly.
Also cloned project was built with error.
May be you can give me some hint concerning that.
Nevertheless I pulled request.

2016-09-11 1:10 GMT+03:00 Danny van Bruggen notifications@github.com:

matozoid:
There was a comment that maybe the files in dev-files shouldn't be in git, but in the wiki, next to the instructions for formatter settings. What do you think?

DeepSnowNeeL:
I think this is handy and does not impact the final jar so I would keep it
like this

matozoid:
In that case we could also add scripts to put those files in the correct place.

matozoid:
So the tasks are:
[...]

PR #469, opened 2016-09-14
Reformat all source code according to the formatting rules. #424

rosrossito:


.IdeaIC14/config/codestyles/JavaParser-idea.xml
```diff
@@ -0,0 +1,38 @@
+<code_scheme name="JavaParser">
```
matozoid: Can you move this file to /dev-files/ ?

matozoid:
So if we format files with Eclipse and Intellij with these configuration files, do they differ a lot?

ptitjes:
Having that in `dev-files/` would indeed permit to have a diff with the previous version...

ptitjes:
And if it is not clear that those files are in the repo as explained at the bottom of the wiki page, maybe it should be quickly mentioned at its top.

I always have been cold about having ide-specific hidden directory in projects but that dates from when maven support in eclipse was not very good. So I may be wrong...

rosrossito:
I moved file to folder /dev-files/.
Although on the wiki page it is mentioned that it must be in
.IdeaC14/config/codestyles...

matozoid:
So, how are the results? Is formatting in intellij and eclipse identical now?

matozoid:
This is not going anywhere.

Issue #468, opened 2016-09-12
Make TypeArguments a Node


matozoid:
I see no reason why it isn't?

PR #470, opened 2016-09-18
Helper methods

DeepSnowNeeL:
Awful formatting problem I know.. :/


DeepSnowNeeL:
I think we should create NodeWithTypeArguments, I saw a lot of expressions which could use some helper methods like addTypeArgumet(Class) or Type or String :)

matozoid:
Maybe do that when you pick up #468 ?

matozoid:
Do you want this pulled now? It would be convenient, because I would have to merge it into my array branch.

DeepSnowNeeL:
If you think it's better now go on, i'll continue to PR when I find new utility methods :)

matozoid:
Here we go :)

Issue #463, opened 2016-09-11
Removing nodes


ftomassetti:
This issue derives from a comment to a [post introducing Javaparser](http://tomassetti.me/getting-started-with-javaparser-analyzing-java-code-programmatically).

It is not trivial to remove nodes from the AST, right? You need to know their parents and the collections they are part of to remove them, there is not a generic way to do so. Should we fix that?

matozoid:
Well, I guess at least one user wants it :)

About collections: our use of plain Lists is quite a pain, as you can see in #421. I'm open to discuss creating our own collection type.

ptitjes:
I remember there was a lengthy discussion for the specific collection when null-management was discussed and I was disappointed it resolved to a no-go !

matozoid:
Hey @ftomassetti - maybe you want to steer this discussion a bit more. I feel it may die otherwise.

DeepSnowNeeL:
Can we provide a generic .remove() method on Node which will be overriden
when specific actions are required ? We will need to check the type of the
parent and correctly remove it from the collection.

If you guys have a specific idea about how to do it, tell me. I'll give it
a try and do a PR soon

matozoid:
Well, what would you have it do when it is contained in a collection? The parent points to the node that has the collection. So in which collection is it?

ftomassetti:
We can do a generic remove implementation being using dirty tricks based on reflections and possibly overriding this behavior where necessary: ugly but doable quickly.

We could use our own collections but: 
1. what about single children not contained in collections? 
2. it seems a rather significant change and I would like to understand first where we are on #354 because it could make more sense to halt implementing major changes and push the rewriting instead.

DeepSnowNeeL:
I'll make you a poc soon,
I intend to provide a remove() method and on it call
parent.tryToRemove(child)
The parent will be responsible to try to remove it from all its possible
fields / containers and both methods will return true if they succeed (in
the spirit of List.remove(x))

ftomassetti:
Ok, but how do you want to implement it? Using a generic, reflection-based approach or implementing remove specifically for each single type of node?

DeepSnowNeeL:
I think i'll go with a specific remove for each node type, it won't be that
complicated. But you are right if we just loop over fields and inherited
fields checking their types (lists/other nodes) and checking
references/contains to do the remove it might be better, i'm ok to do both,
what you guys think about this ?

ftomassetti:
[...]

PR #471, opened 2016-09-20
Remove method

DeepSnowNeeL:
as discussed in #463
We will need to do some tests, if someone want's to help :D :D


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -369,4 +372,37 @@ public void tryAddImportToParentCompilationUnit(Class<?> clazz) {
...
+        Class<?> parentClass = parentNode.getClass();
+        while (parentClass != Object.class) {
+            for (Field f : parentClass.getDeclaredFields()) {
+                f.setAccessible(true);
+                try {
+                    Object object = f.get(parentNode);
+                    if (object == null)
+                        continue;
+                    if (List.class.isAssignableFrom(object.getClass())) {
+                        List<?> l = (List<?>) object;
```
matozoid: Can you use Collection here?

matozoid:
Here, I'll help with the tests. We need...
- a test when parent is null.
- a test when the node is in a collection.
- a test when the node is a normal child.

:-P

DeepSnowNeeL:
Haha :P I mean write thoses ! To be correct we would need to write a test
per node type inside every possible parent node type
#imgoingtodie #nope #plzhelp #hashtag :P

But allright for 3basic tests, I guess we will have issues with "remove
does not work" > "Check the return value and provide a minimal test case
plz" > "oh allright this case wasn't handled !"

It would be great to check at least all the fields that we have in nodes to
confirm that it only contains Collection subtypes or
reference-equals-ok-to-check types

matozoid:
My energy is going towards 3.0, so keep looking for someone to do the boring part of your work ;)

matozoid:
Here, maybe someone from hacktoberfest will help you out :)

matozoid:
@DeepSnowNeeL - this PR is about to miss its second release chance!

DeepSnowNeeL:
What does it need to get into the release ? Just a few tests ? I might be
able to write some basic tests sunday :)

Le sam. 8 oct. 2016 à 19:52, Danny van Bruggen notifications@github.com a
écrit :

matozoid:
Yep, cool!

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithStatements.java
```diff
@@ -43,4 +43,25 @@ public default T addStatement(int index, final Expression expr) {
         return addStatement(index, stmt);
     }
 
+    public default <A extends Statement> A addAndGetStatement(A statement) {
```
matozoid: Intellij tells me that "public" is not necessary here.

DeepSnowNeeL: As in every method in an interface :p I don't know what's the right/standard way, i'm going to google a bit

DeepSnowNeeL: Allright http://stackoverflow.com/questions/161633/ let's remove them :D

matozoid:
I'm going to assume this is done and merge it :)

Issue #456, opened 2016-09-10
Remove "data" field from Node


matozoid:
This field is meant as a place where the user can put additional stuff in a node for own use.

I have never heard of anybody using it, and I would suggest keeping track of data like this in a separate Map<Node, UserObject> so it becomes well typed.

If you feel that this field must remain, say so below.

ftomassetti:
+1 for removing it

ptitjes:
It is often useful (and it is common practice) to attach data directly to the AST while type-checking it, etc...
If you want to attach typed information instead of just an object, there are other means like:

```
interface TypedDataKey<T>{}
```

Have your data field like:

```
private Map<TypedDataKey<?>, Object> data;
```

Document that users should use constants to implement the TypedDataKey and use an IdentityHashMap, and also enforce typing and casting in the accessors. (See Wicket for an example of this pattern.)

PR #472, opened 2016-09-25
Copy Wickets way of dealing with metadata. Also copy their JavaDoc :-)

matozoid:
Fixes issue #456 

I'll ask what the Wicket guys think of this :D


javaparser-testing/src/test/java/com/github/javaparser/ast/MetaDataKeyTest.java
```diff
@@ -0,0 +1,32 @@
...
+        node.setMetaData(ABC, "Hurray!");
+        node.setMetaData(LISTY, Arrays.asList("a", "b"));
+        node.setMetaData(ABC, "w00t");
+
+        assertThat(node.getMetaData(ABC)).contains("w00t");
+        assertThat(node.getMetaData(LISTY)).containsExactly("a", "b");
+        assertThat(node.getMetaData(DING)).isNull();
+    }
+
+}
```
ftomassetti: Newline at the end of the file?

matozoid: Newline at the end of the file...

matozoid:
I think it looks very pretty. Thanks for the hint, @ptitjes !

martin-g:
Hi,

I've been notified that you needed to copy some code from Apache Wicket for your needs.
Since Apache Wicket is ASL2 licenced you can do whatever you want with the source code without asking anyone!
If you want you can mention somewhere that your project has copied code from other projects. But this is not mandatory at all!

Good luck!

ptitjes:
« Licenced to steal » ? :)

And BTW, I wasn't suggesting to copy but rather get inspired... The pattern in itself was well-known before Wicket makes use of it, as using an anonymous inner class is the only way to retrieve the type parameter at runtime. (Web-search for that.)

That said, IMHO the term 'data' would be a better fit than 'mata-data' here...

matozoid:
I understand, the decision to copy (most) was completely mine :) I especially like the Javadoc. Maybe I should take Jonathan Locke's name out though!

Will change to data.

ptitjes:
Or 'userData', or something else that makes it clear it is intended for the only use of the user, without reading the docs...

Issue #437, opened 2016-08-24
CloneVisitor has many unused variables


matozoid:
... meaning that things aren't getting cloned.

It is also largely untested, it needs a decent test suite.

PR #473, opened 2016-09-25
Issue # 437 CloneVisitor has many unused variables

KyleYue:
I removed the getJavaDoc statements in the CloneVisitor which were not used. I also replaced two deprecated methods.  Added one new test for CloneVisitor class to test clone javadoc.


matozoid:
Good work! ( #437 )

Issue #355, opened 2016-07-12
Add breadth first travel to the TreeVisitor


matozoid:
And add bdd tests for the class.

zero33542:
Hello, I am now adding breadth first travel and I want to ask how do I add bdd tests?

matozoid:
Thank you for helping!

The bdd tests are in the javaparser-testing module, they use http://jbehave.org/ . They are quite useful for testing parsing+dumping. If you don't think it is a good match, make a junit test.

PR #474, opened 2016-09-25
Add breadth first travel to the TreeVisitor #355

zero33542:
I add visitBreadthFirst method in TreeVisirtor.java and add junit test in BreadthFirst directory.


matozoid:
Oh hey, you put your tests in the package with the source code on the wiki :-) Do you want this stuff on the wiki as a sample, or do you want to move the test?

zero33542:
Hi, I'll move the test out of wiki_samples. After doing that, do I need to PR again? Or I don't need to?

ftomassetti:
We may want to do a squash after merging this

matozoid:
@zero33542 if you push to the branch of the PR, github automatically notices, so everything is fine.

matozoid:
Hm, I don't think the unit test actually does a test, just a System.out.println. Can you put something together that asserts it does?

matozoid:
I've rewritten the tests and committed them through another PR.

Issue #476, opened 2016-09-26
Make all setters return this.


matozoid:
Since we have our nice "mutable builder" pattern where setters return this, we should apply this to every setter in every node.

@DeepSnowNeeL - interested? :)

PR #479, opened 2016-09-28
Made Node setters return this

ghost:
As described in #476, made Node setters return this

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
@@ -21,25 +21,21 @@
 
 package com.github.javaparser.ast;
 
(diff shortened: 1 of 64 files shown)


ghost:
Sorry for the import changes, that's IntelliJ doing it automatically 😟

matozoid:
Good work!

Did you join through Hacktoberfest?

ghost:
Actually yes :p I thought I could help since it didn't seem that hard and java is kinda my thing so. IntelliJ was complaining about the code style though but I just ignored that and tried to use yours (tabs instead of spaces etc)

matozoid:
Cool! So Hacktoberfest actually works :) If you want to do more, you can check the editor settings files and their instructions here: https://github.com/javaparser/javaparser/wiki/Coding-Guidelines

PR #480, opened 2016-09-29
Remove "download as zip/tar.gz" from http://javaparser.org/

UlhasDeshmukh:
Removed "download as zip/tar.gz" from http://javaparser.org/


SmiddyPence:
What is the motivation?

The html is generated from the README.md and the gh-pages theme.  So next time we regenerate the site these changes will be lost.  I don't think people will want to manually tweak this each time.

matozoid:
The motivation is in the linked issue :)

But if this is a generated file, then it is no use to fix it this way. @UlhasDeshmukh - can you take another look at how this works?

matozoid:
No response from @UlhasDeshmukh , closing.

PR #481, opened 2016-09-30
Make all Node setters return this

d-glenz:
All setters in Node and derived AST classes have their setter extended, so they return this with the appropriate return type.


matozoid:
Thank you for your contribution! Can you explain how this PR is an improvement over #479 ?

matozoid:
#479 was merged.

PR #486, opened 2016-10-03
bug of empty ImportDeclaration toString

niaoge:
see code when create an  empty ImportDeclaration meens like eclipse organize then imports add blank line


matozoid:
An empty import declaration is cause by finding a single ";" in the imports. The output must also be that ";" so this is not a bug.

If you like to change this behaviour, "extend" the DumpVisitor and override this method with your functionality.

What are you having trouble with? Eclipse's organize imports? There is a blank line added somewhere?

matozoid:
No response from @niaoge , closing.

PR #491, opened 2016-10-05
change import handling

### LINKED ISSUE #488: Visitors visit ImportDeclaration.name even when it is empty
ImportDeclaration.getName may throw an exception when the import declaration is empty. At least the Void and GenericVisitorAdapter don't take this into account and always visit it. This may lead to unexpected exceptions.

Some solutions I can think of:
1. fix the visitors
2. let an empty import declaration return null (or Optional.empty()) as the name
3. let an empty import declaration return "" as the name

I think 2 is the most "JavaParser-like" of the three.

matozoid:
This changes "import ...;" handling from using a single class (ImportDeclaration) to using one for every type (see [the JLS](https://docs.oracle.com/javase/specs/jls/se8/html/jls-7.html#jls-7.5)) and an additional type for stray comma's.

I would like some feedback from users if this is needlessly complicated or not. I personally think the stricter typing and the acces to a ClassOrInterfaceType for most types is pretty neat.

Fixes #488


javaparser-testing/src/test/java/com/github/javaparser/bdd/steps/ExistenceOfParentNodeVerifier.java
```diff
@@ -461,6 +503,42 @@ public void visit(TypeExpr n, Void arg) {
         }
 
         @Override
+        public void visit(ArrayBracketPair n, Void arg) {
```
ftomassetti, line 506: Why is this new?

javaparser-core/src/main/java/com/github/javaparser/ast/imports/ImportDeclaration.java
```diff
@@ -0,0 +1,86 @@
...
+    public ImportDeclaration() {
+        super();
+    }
+
+    public ImportDeclaration(Range range) {
+        super(range);
+    }
+
+    /**
+     * Factory method for the subclasses of this class.
```
ftomassetti: This is not instantiating EmptyImportDeclarations, right? So it is not all subclasses

javaparser-testing/src/test/java/com/github/javaparser/bdd/steps/ExistenceOfParentNodeVerifier.java
matozoid: I assumed this visitor should check all parents, so when I noticed a bunch of visits were missing I added them all.

matozoid:
@ftomassetti but is this more or less useful than what it was?

javaparser-core/src/main/java/com/github/javaparser/ast/imports/ImportDeclaration.java
matozoid: Rephrased it.

PR #493, opened 2016-10-06
Issue 492 replace commentsparser

### LINKED ISSUE #492: Replace CommentsParser
JavaCC is already parsing all comments, making CommentsParser redundant. CommentsParser causes two-pass parsing which creates complexity in JavaParser.java.

It would be a good preparation for #484

matozoid:
Fixes #492 

Look at how much garbage this removes from the JavaParser class :) :) :)


javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -46,59 +43,46 @@
...
+    private void insertComments(CompilationUnit cu, TreeSet<Comment> comments) {
+        if (comments.isEmpty())
             return;
 
-        // I should sort all the direct children and the comments, if a comment
-        // is the first thing then it
-        // a comment to the CompilationUnit
-        // FIXME if there is no package it could be also a comment to the
-        // following class...
+        /* I should sort all the direct children and the comments, if a comment
```
ftomassetti, line 55: Why a block comment?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -46,59 +43,46 @@
...
     /**
-     * Comments are attributed to the thing the comment and are removed from
-     * allComments.
+     * Comments are attributed to the thing they comment and are removed from
+     * the commentsCollection.
      */
-    private void insertCommentsInCu(CompilationUnit cu,
-            CommentsCollection commentsCollection) {
-        if (commentsCollection.size() == 0)
+    private void insertComments(CompilationUnit cu, TreeSet<Comment> comments) {
```
ftomassetti, line 51: should _comments_ be renamed in _commentsCollection_ or the JavaDoc be changed?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -46,59 +43,46 @@
...
-        // I should sort all the direct children and the comments, if a comment
-        // is the first thing then it
-        // a comment to the CompilationUnit
-        // FIXME if there is no package it could be also a comment to the
-        // following class...
+        /* I should sort all the direct children and the comments, if a comment
+         is the first thing then it
+         a comment to the CompilationUnit */
+
+        // FIXME if there is no package it could be also a comment to the following class...
```
ftomassetti, line 59: You see, you should never touch a line containing a _FIXME_ because now it is your responsibility :)

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -139,11 +122,11 @@ private void insertCommentsInNode(Node node,
             }
         }
 
-        // at this point I create an ordered list of all remaining comments and
-        // children
+        /* at this point I create an ordered list of all remaining comments and
```
ftomassetti, line 125: Line comment? Capital letter for the comment?

(I must have found my nitpick-hat)

javaparser-core/src/main/java/com/github/javaparser/ParseStart.java
```diff
@@ -8,21 +8,18 @@
...
  * COMPILATION_UNIT indicates a complete Java file,
  * and CLASS_BODY would indicate the part of a class that is within { and }.
- * @see JavaParser#parseStructure(ParseStart, Provider)
+ * @see JavaParser#parse(ParseStart, Provider)
  */
 @FunctionalInterface
 public interface ParseStart<R> {
 	ParseStart<CompilationUnit> COMPILATION_UNIT = ASTParser::CompilationUnit;
 	ParseStart<BlockStmt> BLOCK = ASTParser::Block;
-	ParseStart<List<Statement>> STATEMENTS = ASTParser::Statements;
```
ftomassetti, line 25: Why this disappeared?

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
```diff
@@ -21,84 +21,83 @@
...
-        for (Comment c : getAll()){
+    public boolean contains(Comment comment) {
+        for (Comment c : getComments()) {
             // we tolerate a difference of one element in the end column:
             // it depends how \r and \n are calculated...
-            if ( c.getBegin().line==comment.getBegin().line &&
-                 c.getBegin().column==comment.getBegin().column &&
-                 c.getEnd().line==comment.getEnd().line &&
-                 Math.abs(c.getEnd().column-comment.getEnd().column)<2 ){
+            if (c.getBegin().line == comment.getBegin().line &&
```
ftomassetti, line 73: nice cleanup :)

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -256,13 +261,39 @@ SKIP :
...
     }
+    
+    CommentsCollection getCommentsCollection() {
+        return commentsCollection;
+    }
 
     private void CommonTokenAction(Token token) {
         tokens.add(token);
+
+        while (token.specialToken != null) {
```
ftomassetti, line 277: we could improve the formatting here. Did you write this code or took it from CommentsParser?

javaparser-testing/src/test/resources/com/github/javaparser/bdd/comment_parsing_scenarios.story
```diff
@@ -54,11 +54,11 @@ Then block comment 4 is " comment put randomly in class:    another orphan.    I
 Then block comment 5 is " a comment lost inside a compilation unit. It is orphan, I am sure you got this one  "
 Then the block comments have the following positions:
 |beginLine|beginColumn|endLine|endColumn|
-|3|1|3|82|
-|4|1|4|25|
-|7|5|7|30|
-|10|5|13|32|
-|17|1|17|89|
+|3|1|3|81|
```
ftomassetti, line 57: Do we want to introduce this change in the way the endColumn is calculated? Would this cause problems to users?

ftomassetti (commented):
Looks good!
Adieu CommentsParser, we loved you!

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: I still have time to revert this! ;)

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: Because the comment spans multiple lines and was getting wrecked in reformatting. A block comment should keep the content of the comment together better.

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: I asked for compliments, not for nit picking :)

javaparser-core/src/main/java/com/github/javaparser/ParseStart.java
matozoid: Because JavaParser now always inserts comments in whatever it has parsed. The only requirement is that it has to be a Node. A List is not a Node. So, just for this I would have had to create a few separate methods that avoid adding comments, or maybe iterate the list and insert comments into each of them... Lots of trouble, and easily avoided by using BLOCK, for example.

This does make a case for a class that is a Node _and_ a List. This would also help a lot with our mutable list parenting issues.

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: Everything seems well indented to me, can you point out the problem?
It's newly written, by the way.

javaparser-testing/src/test/resources/com/github/javaparser/bdd/comment_parsing_scenarios.story
matozoid: This is also a nice one. I could have changed the code to match the tests, but I checked it by hand and the end of the range was pointing one character after the end. The old parser was buggy. So I would say it would cause users to not have to work around this.

javaparser-core/src/main/java/com/github/javaparser/ParseStart.java
ftomassetti: Makes sense

javaparser-core/src/main/javacc/java_1_8.jj
ftomassetti: A few lines seems long (at least in the split view I am using for this review). For example line 282 or 292.

There a few places where a space is missing around operators (e.g., lines 282 and 284)

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: Renamed!

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: I shortened those lines and reformatted the method.

Issue #421, opened 2016-08-18
Make returned collections unmodifiable


matozoid:
See #420 and #316. Make sure that internal collections of nodes cannot be modified by getting that collection, then adding to it. The main issue here is that JavaParser doesn't notice the addition, and cannot set the parent of the added nodes correctly.

To still support adding nodes, we will have to add an "addXxx(Node node)" method next to the getter of the collection.

DeepSnowNeeL:
It might be convenient to provide removeXXX() too ?

matozoid:
Two other options:
- subclass ArrayList and override mutator method to set the parent correctly. This doesn't prevent users from creating a completely new plain ArrayList and passing that to "setXxxs(List)" We would have to intercept lists there and turn them into our own subclassed list. That seems like a lot of work for something that should be simple.
- use a library like javaslang or Guava to give us immutable collections. We still have to add "add" and "remove" and it gives us a dependency.

I still like the unmodifiableList best, even though hiding a mutable collection behind something that throws an exception when you call its mutator methods is :hankey:

matozoid:
Lists are now a first class citizen of the AST. It stopped using Java collections for lists of nodes. There is now a NodeList type that is primarily a Node, but also support some of Java List's methods. The advantage is that lists now have a parent, and the children of the list have the list as the parent. Before, the children had their parent set to the owner of the list.

The most important advantage is that users have to make a lot more effort to confuse the parenting of nodes, since NodeLists will take care of their children's parent nodes.

So previously:

``` java
a.getAbcs() -> List
a.getAbcs().get(0) -> x
x.getParentNode() -> a
```

but now:

``` java
a.getAbcs() -> NodeList
a.getAbcs().get(0) -> x
x.getParentNode() -> NodeList
x.getParentNode().getParentNode() -> a
```

Meaning that travel up and down the parent chain now takes an identical amount of steps.

What users might notice:
- `getChildrenNodes` returns only direct children. If you have logic that works on children and their order, it will probably need work. For easier transitioning, there is `getBackwardsCompatibleChildrenNodes()`
- `getParentNode` returns a list if the node is inside a list. If you have logic that expect a parent to be a specific kind of node, you may now be getting a list instead. (But the parent of that list is the node you're looking for.)

This change went hand in hand with stricter null checking. It is no longer allowed to pass "null" for "empty list". Use `new NodeList<>()` now.

If requested in the future, NodeList could be made to implement the Java List interface as well.

[...]

PR #497, opened 2016-10-10
Issue 421 parenting of collections

matozoid:
Here's a demo for #421 of using a special collection to store nodes.

It fixes that lists of nodes are no longer an alien thing in the AST, now they are just another node.

It also fixes that nodes that are added always get their parent set, no matter what happens.

An improvement would be to completely implement the java.util.List interface by delegating everything to innerList.

Note that I couldn't get the visitors to work with a NodeList with generics, so I used a raw type there.


matozoid:
The more I'm thinking about it, the more this seems the most solid way to go.

matozoid:
The more I think about this, the more this seems to be the most solid solution.

matozoid:
Visitors are really cleaning up, because we can simply visit the nodelist, instead of having to loop through every list in the visitor.

PR #501, opened 2016-10-15
Issue 421 parenting of collections

matozoid:
See #421


javaparser-testing/src/test/resources/com/github/javaparser/bdd/manipulation_scenarios.story
```diff
@@ -27,13 +27,6 @@ When the List of VariableDeclarations are set as the resources on TryStmt
 Then all the VariableDeclarations parent is the TryStmt
 
 
-Scenario: Adding null to a TryStmt it handles the outcome gracefully
```
matozoid, line 30: In case anybody is bothered by removing a test: I thought this was a good idea because you can't set null on TryStmt anymore, and TryStmt will never have zero child nodes anymore, so the whole idea behind this test is now invalid.

matozoid:
Immediately fixing most of #451 here.

Will make range optional once I start rewriting comment support.

Issue #503, opened 2016-10-17
Get back the isStatic and isAsterisk on ImportDeclaration


ftomassetti:
They were removed because they make no sense for EmptyImportDeclaration however we could put them back and:
1. return `Optional<Boolean>` or
2. throw an `UnsupportedOperationException` if called on `EmptyImportDeclaration`. In this case the user should call first `isEmpty` on the import to verify if it makes sense to call `isStatic` or `isAsterisk`

matozoid:
Well, I did the whole ImportDeclaration refactoring because someone ran into those exceptions in the first place, so I prefer 1 :)

1 was actually a bit of a joke from my side, to show how complicated the EmptyImportDeclaration is making everything!

Here's an option 3: create "NonEmptyImportDeclaration" as a subclass to the four importdeclarations that are not EmptyImportDeclaration, and put the methods on there.

ftomassetti:
Damn EmptyImportDeclaration! :D 

Well, your 3rd option would make sense. After all in 99% of cases we will just throw away the EmptyImportDeclaration.

PR #509, opened 2016-10-19
issue503: introduce NonEmptyImportDeclaration

ftomassetti:
Following discussion on #503 I introduced NonEmptyImportDeclaration as a common ancestor for all useful imports so that we can put back methods which makes sense for all imports but for the EmptyImportDeclaration


matozoid:
Thanks!

Issue #504, opened 2016-10-17
Redo NodeList


ftomassetti:
Something like `getParentNode`.

If the parent is a Node it would return the result of getParent, otherwise it would return getParent.getParentNode().

matozoid:
Hmmm... I can see the use, but we need another name since a NodeList is... a Node :)

DeepSnowNeeL:
getEnclosingNode ?

matozoid:
Well, "Node" is the problem word.

`getNonListParentNode` is ugly :)

matozoid:
If you think the introduction of NodeList was a mistake, that would be good feedback too 😊

DeepSnowNeeL:
Well if it interefere with getParent it will be annoying, i'll migrate soon
my code to the last alpha and let you know if it's really a pain to upgrade
and what feel annoying :)

ftomassetti:
I think that NodeList is useful but maybe it could not be a node.

Unfortunately I did not follow all the discussions (shame on me) but for me NodeList is not a Node, it is a container of nodes.

We could take EMF as an example (EMF: the meta-modelling framework used in several products including [Xtext](http://xtext.org)): a node is always contained in a containment reference (which could be a List or thing with just one element). The containmente reference would be ~ our NodeList.

Maybe we could have:

```
interface NodeContainer { 
   NodeContainer getContainer();
   Node getContainingNode(); // as getContainer but skips NodeLists
}

// alternatively only some nodes could be NodesContainer and we could exclude 
// the nodes which cannot have children
class Node implements NodeContainer {}

class NodeList implements NodeContainer {}
```

Would this make sense? We would have still NodeList as part of the hierarchy (necessary to make easy insertions and deletions) but for certain usages we could easily skip it.

matozoid:
_crunch_ _crunch_ that seems to make sense. So NodeList is no longer a parent to its children. NodeList will simply set the parent to its own parent. Uhuh, this could work. Also: I don't have to undo a few days of work, so that makes me happy ;)

matozoid:
And: it immediately solves the problem of having two lists with more or less the same content in NodeList.

ftomassetti:
Do you want me to send a PR for this or do you prefer to do it yourself?

DeepSnowNeeL:
Take a look at w3c dom elements & nodes on the standard java libs to see
how they did it, it uses a NodeList when you do getChildNodes() (btw, it
doesn't implements iterable and is awful to use :p)

[...]

Issue #516, opened 2016-10-20
Null part 3


matozoid:
First, roll back #451.

This leaves us with extensive runtime null checks, so not everything is lost, but it would be nice to have compile time feedback. I only know about @Nullable annotations for this. There are two problems: they require a library dependency, and they aren't standardized (there are many different sets.)

If nobody has a better idea, I can see if I can make a `javax.annotation.Nonnull` in javaparser-core and get intellij idea to recognize it.

matozoid:
@DeepSnowNeeL please discuss here. I'm not going to start working on JavaParser again before we have a few people agreeing on what we want here.

DeepSnowNeeL:
I suggest : 
For example taking the class : `AnnotationMemberDeclaration` with its actual field : `Optional<Expression> defaultValue`
- Going back to `Expression defaultValue`
- On the setter, take an Expression that can be null, with the javadoc clearly saying that it can be null 
- On the getter return an `Optional<Expression>` using `Optional.ofNullable(defaultValue)`

This way the API clearly suggest that this value might be optional when getting it, and states in javadoc when it can be null.

The setters for non-null fields will use Objects.requireNonNull() to runtime-check that it doesn't take nulls, with the javadoc clearly stating that it does not accept nulls with bold text + `@Throws` . (We can't rely on `@NonNull` annotations, since it's a dead JSR (305), and it only works with an external tool like intelliJ or findbugs, it won't prevent users to compile with javac) so javadoc is our best bet.

Who is for, who is against ? Express yourself below.

matozoid:
For fields and setters I agree with @DeepSnowNeeL, but...
- I find it odd to have setters use nulls, and their getters use Optional. I'd rather pick one way of dealing with optionality, and since the internet seems to agree that you should only use Optional only as a return value, I'd prefer to not use it at all.
- JavaDoc should be well written, but is not "in your face" and often ignored.
- The big two IDE's have support for all kinds of nullability annotations. It's not perfect, but it is not hard to configure and will happily mark all your possible null dereferences in yellow :)

But for me it's mostly the first point.

@ftomassetti @SmiddyPence @mlangkabel ? Any idea how to get out of this impasse?

matozoid:
(reverted the optional stuff on master in preparation of the outcome of this)

matozoid:
@DeepSnowNeeL looks like nobody is all that interested in this subject. If you submit a PR it will go in.

DeepSnowNeeL:
[...]

Issue #484, opened 2016-10-02
Rewrite comment support


matozoid:
Rewrite comment support to store comments in a logical way.
- make sure the new way can be explained in a few lines.
- make sure that it is clear where a comment is located in the source, so it can reliably be reproduced.

The suggested approach is to store for each node the comments that come directly before it.

Since some comments will now be lost (they appear inside what is logically the node, but do not appear before any subnode) these will be stored in the node in a separate list.

``` java
// nice
/** Javadoc */
class X{
    /*wow*/
    int a;
    /*such doge*/
}
```

"nice" and "javadoc" are before class X, and are the comments for the node that represents class X.
"wow" is inside class X and has a node after it: the field declaration. It is the comment for this field declaration.
"such doge" is inside class X but has no node after it which is inside X. Therefore it goes into the "trailingComments" list of class X's node.

NEED MORE INPUT:
- what are the use cases people have for comments?
- are there any comments about this approach?

ftomassetti:
I think sometimes node refer to what precedes them:

```
int a; // about variable a
```

matozoid:
In my opinion this turns the algorithm from parsing into attribution - which seems more complex and subjective to me. I was thinking this logic could be in a second pass.

ftomassetti:
Agreed that would make sense to have it in a second step, but then the
comments should probably be part of this list, not assigned to elements
following them because it is attribution

DeepSnowNeeL:
Can't we store all comments in the Class body as a list of objects
containing : Comment,NodeBefore,NodeAfter
and providing some methods on nodes to get the comment before or after it
like field.getCommentAfter() field.getCommentBefore()

matozoid:
@ftomassetti what do you mean by "this list"?

@DeepSnowNeeL do you mean store all comments in the CompilationUnit? So instead of having the lists on the nodes themselves, store them in a single long list? We can do that, but what does it improve?

matozoid:
Note that I've got the proposed system working in a small POC by simply reading the "special tokens" that javacc reads. Because obviously, if javacc is parsing the java code, it also needs to parse the comments in and around it. So at least it improves that we don't need a secondary parsing pass anymore.

DeepSnowNeeL:
No I mean store the comments on the enclosing node, a comment for a field
would be stored on the classorinterfacetype, a comment in a method would be
stored on the method node. This way we could localize in which node they
were and before and after which node.
I didn't thought about faisability with javacc tho.

[...]

PR #522, opened 2016-10-29
Make NodeList NOT a Node.

matozoid:
Fix issue #504.

NodeList is now a stand alone class that shares functionality from Node only by using two (new) interfaces: HasParentNode to integrate into the general parent node setting mechanism, and Visitable to integrate into visitors.

Since NodeList is no longer a Node, it will no longer show up as a child or parent node.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
@@ -60,7 +60,7 @@ class CommentsInserter {
         // so I could use some heuristics in these cases to distinguish the two
         // cases
 
-        List<Node> children = cu.getChildrenNodes();
+        List<Node> children = cu.getChildNodes();
 
         Comment firstComment = comments.iterator().next();
         if (cu.getPackage() != null
@@ -91,7 +91,7 @@ class CommentsInserter {
         // if they preceed a child they are assigned to it, otherweise they
         // remain "orphans"
 
-        List<Node> children = node.getChildrenNodes();
+        List<Node> children = node.getChildNodes();
 
         for (Node child : children) {
             TreeSet<Comment> commentsInsideChild = new TreeSet<>(NODE_BY_BEGIN_POSITION);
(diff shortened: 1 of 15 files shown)


matozoid:
@ftomassetti @DeepSnowNeeL - since you supplied the feedback that led to this change, can you give this a thumbs up or down?

DeepSnowNeeL:
looks good, looking forward to the next alpha (if possible with #516 ) to switch to it and test everything

matozoid:
I'm sorry, you can't wish a feature into a release ;) #516 needs some kind of agreement among a few users. I've pulled the NodeList change forward because I was having issues with NodeList in #484.

matozoid:
I'll just assume @ftomassetti loves this PR so I can make a release. Feel free to open issues if I'm wrong.

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -60,7 +60,7 @@ private void insertComments(CompilationUnit cu, TreeSet<Comment> comments) {
         // so I could use some heuristics in these cases to distinguish the two
         // cases
 
-        List<Node> children = cu.getBackwardsCompatibleChildrenNodes();
+        List<Node> children = cu.getChildNodes();
```
ftomassetti, line 63: This arguably looks nicer :D

javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
```diff
@@ -0,0 +1,49 @@
...
+     * "this" for everything except NodeLists. NodeLists use their parent as their childrens parent.
+     */
+    Node getParentNodeForChildren();
+
+    @SuppressWarnings("unchecked")
+    default <N> N getParentNodeOfType(Class<N> classType) {
+        Node parent = getParentNode();
+        while (parent != null) {
+            if (classType.isAssignableFrom(parent.getClass()))
+                return (N) parent;
```
ftomassetti, line 27: I think you can use `classType.cast()` to avoid the warning. Maybe I will send a PR for that

matozoid: Can you scan the codebase for these casts? They happen quite often.

ftomassetti:
Sure, I will look for SuppressWarnings

matozoid:
I don't think they're all suppressed, @ftomassetti

PR #532, opened 2016-11-03
Cosmetic changes

ftomassetti:
I started with the intention of changing some casts to use `Class.cast` but I found just one of those. In the process I changed a few minor things.

The biggest thing is that all node types (but NodeWithDeclaration) has as type parameter `N extends Node` while before it was `T`


matozoid:
Nice

PR #536, opened 2016-11-05
#516 some optionals

DeepSnowNeeL:
WIP

@matozoid is this okay to you ? if yes i'll continue in this way for the rest of "// TODO nullable" I can find


javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java
```diff
@@ -162,7 +162,7 @@ public Visitable visit(AnnotationDeclaration _n, Object _arg) {
 	public Visitable visit(AnnotationMemberDeclaration _n, Object _arg) {
 		NodeList<AnnotationExpr> annotations = cloneList(_n.getAnnotations(), _arg);
 		Type<?> type_ = cloneNode(_n.getType(), _arg);
-		Expression defaultValue = cloneNode(_n.getDefaultValue(), _arg);
+		Expression defaultValue = cloneNode(_n.getDefaultValue().orElse(null), _arg);
```
matozoid: Consider making an optional-specific "cloneNode" for this

matozoid:
Yep, that looks like it :) Good luck, I know how much fun it is to update those visitors every time.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithStatements.java
```diff
@@ -71,8 +70,20 @@ default ExpressionStmt addAndGetStatement(Expression expr) {
...
-    
+
     default boolean isEmpty() {
         return getStmts().isEmpty();
     }
+
+    @SuppressWarnings("unchecked")
+    default N copyStatements(NodeList<Statement> nodeList) {
+        for (Statement n : nodeList) {
+            addStatement((Statement) n.clone());
```
DeepSnowNeeL, line 81: It actually make a shallow copy, do we have something to deep copy statements ? (I tried using CloneVisitor but without success)

matozoid: Why was clonevisitor not a success? It's in use behind n.clone()

DeepSnowNeeL: Oh yeah ! I thought it was using java standard clone() for a moment ! so it is a deep copy, great

matozoid:
Are you also going to do the Javadoc? ;)

Actually, it would be good to at least document the setters that can take a null, with a little explanation of what null means in that case. "Null means no initializer" or so. In other cases it would be "Null is not allowed"

DeepSnowNeeL:
Well I did the javadoc for the setter, but yep i'll try to be precise in
what null means

Le dim. 6 nov. 2016 à 16:37, Danny van Bruggen notifications@github.com a
écrit :

DeepSnowNeeL:
I'm almost done but @matozoid can you check the 1 last test, I can't get what's the difference, maybe a CR LF windows problem ?

matozoid:
Okay, I'll see.

matozoid:
It's working fine for me. Don't look at appveyor if that's what you're worried about. It hasn't been working for months now :(

matozoid:
I see you also did a lot of the reformatting job :D

Double check the Javadoc for NodeWithTypeArguments please.

DeepSnowNeeL:
Well my eclipse always reformat / re-import and there is no way to disable that unless I ctrl+Z after each save (it undo the organize import)

What about the NodeWithTypeArguments ?

matozoid:
NodeWithTypeArguments: it mentions null in the Javadoc.

I used to use Eclipse, and you can turn that off :) But never mind, you improved the formatting.

DeepSnowNeeL:
done :)

matozoid:
Merge time?

matozoid:
Thank you very much for your contribution :)

Issue #403, opened 2016-07-27
NameExpr for ClassOrInterfaceType


mlangkabel:
I'd like to request storing the name of a `ClassOrInterfaceType` as `NameExpr` instead of using a simple String. This would allow us to query the name's location (which is not always consistent with the type node's location -> e.g. when the type has type arguments).

matozoid:
This is actually already stored as a NameExpr, it just has strings in the API for ease of use. getNameExpr and setNameExpr will get it.

Is that helpful?

mlangkabel:
Thanks for getting back to me. I just re-checked the implementation and I really cannot find where those methods are implemented. The constructor of the `ClassOrInterfaceType` only requires a String and no NameExpr as argument. Maybe I missed something. I hope I don't make a fool of my here ;)

matozoid:
No, I'm the fool. I have no idea where I was looking but today I see the same as you :-) Will fix.

PR #549, opened 2016-11-09
Fix naming

### LINKED ISSUE #358: Node naming
After #189 and #252 , it seems we should refactor the naming of nodes.

Please discuss.

matozoid:
Fixes #358 and #403.

Impact: There is no more "String name" in any node. We have either Name (which can be a qualified name) or SimpleName (which is certainly not qualified.)

This means that all names are now visitable.


matozoid:
I'll just merge this to prevent merge conflicts with the other changes that are being done.

If there are comments, please open issues.

Issue #550, opened 2016-11-10
Make NodeList implement java.util.List


matozoid:
NodeList is our collection for nodes. Its main task is making sure every node in the collection has a parent. Currently it is a stand alone object, but we would like it to behave just like a native collection, so we want it to implement java.util.List.

When doing this task, every method can be delegated to the internal list, EXCEPT for the ones that add a node to the collection. This should take care to set the parent node before delegating to the internal list.

Alternatively, if it is possible to directly inherit from ArrayList, that might make things more efficient. At least it will remove the delegation, and therefore remove an additional object to click open while debugging.

PR #556, opened 2016-11-10
make NodeList implement List

DeepSnowNeeL:
Here we go

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
@@ -113,7 +113,10 @@ public class NodeList<N extends Node> implements List<N>, Iterable<N>, HasParent
 
     @Override
     public N remove(int index) {
-        return innerList.remove(index);
+        N remove = innerList.remove(index);
+        if (remove != null)
+            remove.setParentNode(null);
+        return remove;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -161,4 +171,222 @@ public Node getParentNodeForChildren() {
...
+
+    /**
+     * @param o
+     * @return
+     * @see java.util.List#remove(java.lang.Object)
+     */
+    @Override
+    public boolean remove(Object o) {
+        return innerList.remove(o);
+    }
```
matozoid, line 226: Hmmm, does this need de-parenting?

javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -161,4 +171,222 @@ public Node getParentNodeForChildren() {
...
+
+    /**
+     * @param c
+     * @return
+     * @see java.util.List#removeAll(java.util.Collection)
+     */
+    @Override
+    public boolean removeAll(Collection<?> c) {
+        return innerList.removeAll(c);
```
matozoid: And here, etc...

matozoid: Actually, I don't think we do that in other nodes as well, right? Hmmm, that means that the childNodes list will accumulate all kinds of crap when you start modifying the AST...

matozoid:
Ready to merge?

matozoid:
Fixes #550

DeepSnowNeeL:
I think so

matozoid:
Note that I squashed this one too @DeepSnowNeeL

PR #558, opened 2016-11-11
Add get(int i) shortcuts for list getters.

### LINKED ISSUE #524: Make shortcuts for list.get(i)
So many `getTypes().get(0)` in the tests. Why not have helper methods for every list: `getType(int i) { return getTypes().get(i); }`

(this needs the equivalent setType(i, ...) too)

qnnguyen:
Fixes #524 
Changed usages of old getters, found by grep -r "get[^)][^)]\*()\.get([^)][^)]\*)" \*


qnnguyen:
Pretty big commit, but all basically the same change repeatedly.

ftomassetti:
It makes life easier to the user but it also means more code to mantain :)

DeepSnowNeeL:
Please do it with default methods in interfaces !! It will be a lot easier

qnnguyen:
Good point! Sorry, bit rusty with Java :( I'll make the changes

matozoid:
Thank you very much for the contribution!

This does the getters of #524, but there was a remark about setters too, therefore #524 stays open.

matozoid:
Congratulations by the way, @qnnguyen, you are contributor number 40! :)

Issue #559, opened 2016-11-11
Observing nodes


ftomassetti:
If we want to promote JavaParser as a tool to change code I think it would be useful to have the possibility to react to changes. To do that we could use the observer pattern. In this way I could register to get notifications for all changes and possibly react by doing something (update some caches, try to compile the new code, dump code on the disk, verify some property is respected, etc.)

Thoughts?

matozoid:
I've been thinking how to generate Problems from the grammar, from the code while parsing, and on demand after a user made changes. Would this help with that?

ftomassetti:
I am not familiar with the Problems thing but the "after a user made changes" sounds something we could use: we could add observers to nodes and on any change we could mark nodes as "dirty" and then regenerate problems when needed (for example the first time someone ask them)

matozoid:
--> https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/Problem.java
A problem is essentially a single piece of feedback to the user of JavaParser. From "could not open file" to "annotations are not supported at this source level" to "opening bracket at position 10,10 was not closed"

A TryStmt can be invalid, you can set every optional part to null and it doesn't compile anymore. I don't want to throw an exception immediately when that happens, but I would like to be able to give that feedback somehow.

Sounds useful, but let's move this particular Problem problem into the future :)

matozoid:
Callbacks are all the hype nowadays. Is there some library we should look at for inspiration?

DeepSnowNeeL:
Just make sure it's functional interfaces so it can be converted to lambdas
:)

matozoid:
I played with it a little, and here is my feedback:

``` java
package com.github.javaparser;

import com.github.javaparser.ast.CompilationUnit;
import com.github.javaparser.ast.Node;
import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
import com.github.javaparser.ast.observing.NodeObserver;
import com.github.javaparser.ast.type.ClassOrInterfaceType;

import static com.github.javaparser.ast.NodeList.nodeList;

public class A {
    public static void main(String[] args) {
        CompilationUnit compilationUnit = JavaParser.parse("class X {}");

        ClassOrInterfaceDeclaration x = compilationUnit.getClassByName("X");
        /* Is this the best scope for a callback? It seems to be attached to exactly one node, 
         and it will receive only events generated by this node. 
[...]

PR #560, opened 2016-11-11
Observer pattern for nodes and nodelists

ftomassetti:
This is a prototype of the proposal in #559


matozoid:
How would you like us to respond to a prototype? Are you looking for comments? Are you actively working on this? Do you want this in a release?

ftomassetti:
I just added it in case we wanted to discuss the details. I basically would apply the same pattern to all the remaining node classes if this seems fine

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
```diff
@@ -137,25 +137,29 @@ public SimpleName getName() {
      * @return this, the AnnotationMemberDeclaration
      */
     public AnnotationMemberDeclaration setDefaultValue(Expression defaultValue) {
+        notifyPropertyChange("defaultValue", this.defaultValue, defaultValue);
```
matozoid: Oh my, identifiers in strings! This is a maintenance nightmare!

matozoid: I really think we should start generating parts of the AST. Will look into that after generating visitors.

matozoid:
The pattern seems pretty normal to me. +1

ftomassetti:
Ok, this should be ready to be considered:
- the pattern has been applied to all the setters
- Node has now a method to get the NodeLists it owns
- there is an Observer which propagates itself, by adding itself to all new nodes added to the element it is observing
- Node has both a method to register the observer just on itself and one to register an observer on itself and all descendants nodes and node lists
- added a couple of tests to show the usage and do a basic sanity check

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnionType.java
```diff
@@ -32,6 +30,7 @@ public UnionType(NodeList<ReferenceType<?>> elements) {
     }
 
     public UnionType setElements(NodeList<ReferenceType<?>> elements) {
+        notifyPropertyChange("elements", this.elements, elements);
```
matozoid: Do you think that maybe we should have an enum for the property names? That way we don't get typo's.

ftomassetti: Well, I think an enum or constants or something similar would make sense if we were using the value in multiple places. Maybe for those properties which are present in multiple nodes (name, type, etc) it could make sense, but for the ones used just once we would just have the same risk of typos, because the name would be typed exactly once anyway.

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
ftomassetti: Yes, I also think we should look into generating those classes

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnionType.java
matozoid: It is used once in JavaParser, but I expect that users will also use this text. I would personally feel a lot safer having this statically checked instead of equals'ed at runtime.

ftomassetti: right, I forgot to consider that users will use this. Ok, I can go with the enum.

If this is the only concern I will change every call to use the enum right now, if there are more I will apply all the changes at the same time.

matozoid: I have nothing more.

ftomassetti:
Ok, I have refactored to used the enum. It gives us two additional benefits:
- it is easier now to verify name consistency: for example, we have both ID and IDENTIFIER
- it is easier for user to see which nodes use which properties. This is useful because the same observer can be used for different types of nodes

matozoid:
+1 for pull, but check @DeepSnowNeeL 's comment first: #559

ftomassetti:
@matozoid I addressed the comment from @DeepSnowNeeL (answer on #559 and new code committed)

matozoid:
Thank you for a new feature and being very cooperative!

ftomassetti:
Thank you and @DeepSnowNeeL for helping improving it!

PR #562, opened 2016-11-14
correct a bug introduced with optionals

DeepSnowNeeL:
(I would love an alpha10 maven central with this fixed :D thx)


matozoid:
Is it a problem if that alpha waits until coming Sunday?

DeepSnowNeeL:
np :D

matozoid:
Okay, this one really needs documentation to indicate what the difference is between "none" and "empty list"!

matozoid:
Thanks!

PR #566, opened 2016-11-14
removed duplicate and old stuff on remove method

### LINKED ISSUE #565: Node.remove should invoke the setter
`Node.remove` directly set the field of the parent while it should call the setter. The setter would notify the observers, otherwise we have changes which cannot be observed.

DeepSnowNeeL:
fix #565 too like #568 but with @ftomassetti 's junit test + removed some old stuff in Node.remove + left some TODO's to finish the job about notifying things in NodeList


ftomassetti:
This test should pass:

```
    @Test
    public void deleteAParameterTriggerNotifications() {
        String code = "class A { void foo(int p) { }}";
        CompilationUnit cu = JavaParser.parse(code);
        List<String> changes = new ArrayList<>();
        AstObserver observer = new AstObserverAdapter() {

            @Override
            public void listChange(NodeList observedNode, ListChangeType type, int index, Node nodeAddedOrRemoved) {
                changes.add("removing [" + nodeAddedOrRemoved + "] from index " + index);
            }
        };
        cu.register(observer, Node.ObserverRegistrationMode.SELF_PROPAGATING);

        cu.getClassByName("A").getMethodsByName("foo").get(0).getParameter(0).remove();
        assertEquals(Arrays.asList("removing [int p] from index 0"), changes);
    }
```

DeepSnowNeeL:
It was caused by the methods you didn't add the notify to, like remove(Object) which was used by the remove() method.
I left some TODO's where we should notify, @ftomassetti is it ok if you finish this ? (you can push to this branch I think)

ftomassetti:
Mmm, there is quite some work to do. I think we need first of all to write some tests for these:
1. Tests to cover all cases of `Node.remove()`
2. Tests to cover addAll
3. Tests to cover clear
4. Tests to cover set
5. Tests to cover removeAll
6. Tests to cover retainAll
7. Tests to cover replaceAll
8. Tests to cover removeIf

I created separate issues: one for the point 1 (#565) and 1 for the other points, which are all related (#569). While working on this I noticed also #567 which is small and can be treated separately (see PR #568 ).

matozoid:
Closing because it shouldn't be pulled.

PR #570, opened 2016-11-14
issue565: invoke setters to remove arbitrary Node by reflection

ftomassetti:
The PR contains 3 tests:

* removing from a NodeList
* removing from a single property when allowed (method body)
* removing from a single property when not allowed (type name)


javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -48,9 +48,12 @@ private void own(N node) {
     }
 
     public boolean remove(Node node) {
-        notifyElementRemoved(innerList.indexOf(node), node);
+        int index = innerList.indexOf(node);
```
ftomassetti, line 51: This is a separate issue, I will remove this change from this draft

Issue #290, opened 2016-04-01
When there is a String constant "\\" compilationUnit ignores all further comments


ndimitry:
Here is code that reproduces this issue. This line ruins comments processing:  String str = "\";

```
package org.test;

import com.github.javaparser.JavaParser;
import com.github.javaparser.ast.CompilationUnit;
import com.github.javaparser.ast.body.MethodDeclaration;
import com.github.javaparser.ast.visitor.VoidVisitorAdapter;

import java.io.File;
import java.io.FileInputStream;

public class TestComments {
    String str = "\\";

    /**
     * Comment that is ignored
     */
    public void someTestMethod() {}

    public static void main(String[] args) throws Exception {
        // Correct path so that it will point to the file with this code
        try (FileInputStream in = new FileInputStream(
                new File(".\\src\\main\\java\\org\\test\\TestComments.java"))) {
            CompilationUnit compilationUnit = JavaParser.parse(in);

            new VoidVisitorAdapter<Object>() {
                public void visit(MethodDeclaration method, Object arg) {
                    System.out.println("Method: " + method.getName());
                    System.out.println("Comment: " + method.getComment());
                    System.out.println("JavaDoc: " + method.getJavaDoc() + "\n");
                }
            }.visit(compilationUnit, null);
        }
    }
}
```

ftomassetti:
We should first of all verify if this is still happening with the current master version or 2.4.0 (which is going to be released soon)

ftomassetti:
I can confirm that this works in the current version (3.0.0-alpha.9). I also opened a PR to add a test to prevent regressions.

Given the issue is solved I am closing this task.

Thanks for reporting it!

PR #577, opened 2016-11-19
issue290: add test against regression

ftomassetti:
Simple test to verify #290 is corrected and it will not happen again


matozoid:
@ftomassetti - can you make it use the Providers class so that character encoding is handled?

ftomassetti:
Sure, just tried doing this:

```
    private ParseResult parseSample(String sampleName) {
        InputStream is = this.getClass().getResourceAsStream(
                "/com/github/javaparser/issue_samples/" + sampleName + ".java.txt");
        Provider p = Providers.provider(is, Charsets.UTF_8);
        return new JavaParser().parse(ParseStart.COMPILATION_UNIT, p);
    }
```

Is that what you had in mind?

matozoid:
Yes, or leave the UTF_8 off, because that's the default. Do you think we should have a method on Providers that makes this case a bit more readable?

PR #579, opened 2016-11-20
Issue 578: set range of compilationunit to cover the whole file

### LINKED ISSUE #573: Need Help With Visitor
I've reviewed the examples but am unclear how to achieve the following:

How would one implement a visitor to identify all occurrences of instantiation of object type "Bar" with a parameter of object type "Foo"?

i.e.

Foo o = new Foo();
Bar b = new Bar(o);

I'd like to identify the source file and line number.

Thank you.

### LINKED ISSUE #578: Range of the compilation unit not including whole file
Consider this case:
```
// This is my class, with my comment
class A {
    static int a;
}
```

The comment is considered outside the range of the compilation unit. While I understand it is outside the class I think the comment should be considered part of the compilation unit. The compilation unit should represent the whole file, from the first character to the last one.

What do you think?

matozoid:
Fixes #578

### DIFF
--- a/javaparser-core/src/main/javacc/java_1_8.jj
+++ b/javaparser-core/src/main/javacc/java_1_8.jj
@@ -1308,7 +1308,7 @@ CompilationUnit CompilationUnit():
   ( in = ImportDeclaration() { imports = add(imports, in); } )*
   ( tn = TypeDeclaration() { types = add(types, tn); } )*
   (<EOF> | "\u001A" /** ctrl+z char **/)
-  { return new CompilationUnit(range(pos(1, 1), tokenEnd()), pakage, imports, types); }
+  { return new CompilationUnit(range(Position.HOME, tokenEnd()), pakage, imports, types); }
 }
 
 PackageDeclaration PackageDeclaration():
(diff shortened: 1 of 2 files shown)


matozoid:
Add a test, decrease coverage... :( :(

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -1302,14 +1302,13 @@ CompilationUnit CompilationUnit():
...
 {
-  [ LOOKAHEAD(PackageDeclaration()) pakage = PackageDeclaration() {begin = begin.orIfInvalid(pakage.getBegin());} ]
-  ( in = ImportDeclaration()      { begin = begin.orIfInvalid(in.getBegin()); imports = add(imports, in); } )*
-  ( tn = TypeDeclaration() { begin = begin.orIfInvalid(tn.getBegin()); types = add(types, tn); } )*
+  [ LOOKAHEAD(PackageDeclaration()) pakage = PackageDeclaration() ]
+  ( in = ImportDeclaration() { imports = add(imports, in); } )*
+  ( tn = TypeDeclaration() { types = add(types, tn); } )*
   (<EOF> | "\u001A" /** ctrl+z char **/)
-  { return new CompilationUnit(begin.invalid() ? Range.UNKNOWN : range(begin, tokenEnd()), pakage, imports, types); }
+  { return new CompilationUnit(range(pos(1, 1), tokenEnd()), pakage, imports, types); }
```
ftomassetti: Maybe you can use Postion.HOME here

PR #585, opened 2016-11-21
Issue 548 import parsing in grammar

### LINKED ISSUE #548: Move import parsing back to grammar
`ImportDeclaration.create` is currently interpreting some data that comes from the parser and turning it into the correct nodes. That is a task for the grammar itself.

This will impact `CompilationUnit.addImport` too.

matozoid:
Fixes #548


matozoid:
Nothing really changes for users, it's just moving parsing code from Java to the grammar.

PR #595, opened 2016-11-23
Issue 547 null range

### LINKED ISSUE #547: Use Optional<Range> instead of Range.UNKNOWN
The notion that a node has no location information is currently hidden by using "Range.UNKNOWN" which is a range from 0,0 to 0,0. It would be much more in-your-face if it was null (or Optional.empty())

matozoid:
Fixes #547

This makes range nullable instead of using Range.UNKNOWN.

@DeepSnowNeel - I can't get Node.remove() to work correctly here, I just don't get what it is trying to do so one test keeps failing. Help!


ftomassetti:
Just remove casts to `Node` in one place. Line 474 of Node:

`Object result = isOptionalAssignableFrom(method.getGenericReturnType(), this.getClass()) ? ((Optional)resultRaw).get(): resultRaw;`

ftomassetti:
In practice what is trying to do is call getters and if the getters return the node itself then he uses the corresponding setter to remove the node.

I am trying to delete a Node from its parent but I do not know where it is so I call `myNode.getParent().getFoo()` and notice that it return `myNode` then I can call `myNode.getParent().setFoo(null)` and remove `myNode` from the AST.

Now, if `getFoo` returns something that is not a Node we are sure is not what we are interested into so we can skip it.

ftomassetti:
We could refine `Node.remove` a little bit later but now all tests should pass

matozoid:
Thanks, it's working again - but the remove code doesn't seem very well-tested :(

ftomassetti:
We could take several Java files, build the AST and try to remove all nodes (starting from the leaves). This would permit to battle-test the remove method. What do you think?

matozoid:
I think we need a new issue with that suggestion for milestone 3.0.0 :)

Is this PR okay?

ftomassetti:
We have it now #597.

Looking at the PR. 110 files changed :D

.gitignore
```diff
@@ -1,3 +1,4 @@
+lib
```
ftomassetti, line 1: why?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -112,12 +114,20 @@ void insertComments(Node node, TreeSet<Comment> commentsToAttribute) {
...
-            if (comment.isLineComment()) {
-                for (Node child : children) {
-                    if (child.getEnd().line == comment.getBegin().line
-                        && attributeLineCommentToNodeOrChild(child,
-                                comment.asLineComment())) {
-                            attributedComments.add(comment);
+            if (comment.getRange().isPresent()) {
+                Range commentRange = comment.getRange().get();
+                if (comment.isLineComment()) {
+                    for (Node child : children) {
```
ftomassetti: could we use a filter here, to make the code more concise?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -112,12 +114,20 @@ void insertComments(Node node, TreeSet<Comment> commentsToAttribute) {
          there is something contained in their line */
         List<Comment> attributedComments = new LinkedList<>();
         for (Comment comment : commentsToAttribute) {
-            if (comment.isLineComment()) {
-                for (Node child : children) {
-                    if (child.getEnd().line == comment.getBegin().line
-                        && attributeLineCommentToNodeOrChild(child,
-                                comment.asLineComment())) {
-                            attributedComments.add(comment);
+            if (comment.getRange().isPresent()) {
```
ftomassetti: could we use a filter here, to make the code more concise?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -167,9 +177,10 @@ void insertComments(Node node, TreeSet<Comment> commentsToAttribute) {
     private boolean attributeLineCommentToNodeOrChild(Node node, LineComment lineComment) {
         // The node start and end at the same line as the comment,
         // let's give to it the comment
-        if (node.getBegin().line == lineComment.getBegin().line
+
```
ftomassetti: why the extra white line?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
```diff
@@ -192,11 +203,14 @@ private boolean attributeLineCommentToNodeOrChild(Node node, LineComment lineCom
...
     private boolean thereAreLinesBetween(Node a, Node b) {
+        if (!a.getRange().isPresent() || !b.getRange().isPresent()) {
+            return true;
+        }
         if (!PositionUtils.areInOrder(a, b)) {
             return thereAreLinesBetween(b, a);
         }
-        int endOfA = a.getEnd().line;
-        return b.getBegin().line > (endOfA + 1);
+        int endOfA = a.getRange().get().end.line;
```
ftomassetti: Maybe this is a little bit too long to write. Should we create & use a `getEndLine` method?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -113,21 +130,30 @@ public final Comment getComment() {
...
+        }
+        return Optional.of(range.end);
     }
 
     /**
      * Sets the begin position of this node in the source file.
      */
     public Node setBegin(Position begin) {
+        if (range == null) {
+            range = Range.range(0, 0, 0, 0);
```
ftomassetti: Mmm, should we let range be null? Or throw an error?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -431,7 +473,18 @@ public boolean remove() {
                     if (optSetter.isPresent()) {
                         try {
                             Object resultRaw = method.invoke(parentNode);
-                            Node result = isOptionalAssignableFrom(method.getGenericReturnType(), this.getClass()) ? (Node)((Optional)resultRaw).get(): (Node) resultRaw;
+                            Node result;
+                            if (isOptionalAssignableFrom(method.getGenericReturnType(), this.getClass())) {
+                                Optional optionalResultRaw = (Optional) resultRaw;
+                                if (optionalResultRaw.isPresent()) {
+                                    Object o = optionalResultRaw.get();
+                                    if (Node.class.isAssignableFrom(o.getClass())) {
```
ftomassetti, line 455: we do not need this check because anyway we are comparing by identity later, so no need to cast

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
```diff
@@ -67,14 +69,27 @@ public void addComment(Comment comment) {
...
-            if (c.getBegin().line == comment.getBegin().line &&
-                    c.getBegin().column == comment.getBegin().column &&
-                    c.getEnd().line == comment.getEnd().line &&
-                    Math.abs(c.getEnd().column - comment.getEnd().column) < 2) {
-                return true;
+            if(!c.getRange().isPresent()){
+                return false;
+            }
+            Range cRange = c.getRange().get();
+            if(cRange!=null) {
```
ftomassetti: a space after the if

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
```diff
@@ -67,14 +69,27 @@ public void addComment(Comment comment) {
...
+        }
         for (Comment c : getComments()) {
-            // we tolerate a difference of one element in the end column:
-            // it depends how \r and \n are calculated...
-            if (c.getBegin().line == comment.getBegin().line &&
-                    c.getBegin().column == comment.getBegin().column &&
-                    c.getEnd().line == comment.getEnd().line &&
-                    Math.abs(c.getEnd().column - comment.getEnd().column) < 2) {
-                return true;
+            if(!c.getRange().isPresent()){
```
ftomassetti: a space after the if

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
```diff
@@ -67,14 +69,27 @@ public void addComment(Comment comment) {
     }
 
     public boolean contains(Comment comment) {
+        if(!comment.getRange().isPresent()){
```
ftomassetti: a space after the if

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
```diff
@@ -67,14 +69,27 @@ public void addComment(Comment comment) {
...
-                    Math.abs(c.getEnd().column - comment.getEnd().column) < 2) {
-                return true;
+            if(!c.getRange().isPresent()){
+                return false;
+            }
+            Range cRange = c.getRange().get();
+            if(cRange!=null) {
+                // we tolerate a difference of one element in the end column:
+                // it depends how \r and \n are calculated...
+                if (cRange.begin.line == commentRange.begin.line &&
```
ftomassetti: could we replace the first two conditions with `cRange.begin.equals(commentRange.begin)` ?

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -209,7 +209,7 @@ final class ASTParser {
     }
 
     private static class RangedList<T extends Node> {
-        Range range = Range.UNKNOWN;
+        Range range = Range.range(0, 0, 0, 0);
```
ftomassetti: I do not like that because it is not true: the element has not that position. If this value represent `Range.UNKNOWN` then we should keep `Range.UNKNOWN`

ftomassetti:
Basically it seems a Node could either have or not have a Range: if we parse it it has a Range (which could become invalid as we modify the Node) while if it is created it could have no Range. Maybe we could add a method like `isPlaced()` to verify which is the case

matozoid:
Thanks! That's very thorough. I'll get to work in the weekend.

I considered removing getBegin/End/setBegin/End from Node, because being able to set half a range makes it easy to put the range in an invalid state. What do you think?

ftomassetti:
mmm, yes, I understand it makes sense to remove the setters but I think removing the getters lead to more awkward code. Maybe being asymmetrical is not ideal but anyway most people would be interested just in the getters, never in setting the range.

.gitignore
matozoid: Because I'm using intellij's OSGI plugin and that insists on adding some jars there. Harmless, I think?

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: Using getEnd and getBegin now

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: ok

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: ok

javaparser-core/src/main/java/com/github/javaparser/CommentsInserter.java
matozoid: ok

.gitignore
matozoid: Intellij's OSGI plugin wanted to put some jars here. Do you want it gone?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: deleted setBegin/End

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: I'm going to leave that for another issue :)

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
matozoid: Reformatted the file

javaparser-core/src/main/java/com/github/javaparser/ast/comments/CommentsCollection.java
matozoid: Yes

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: Hmmm, that's a temporary position that should never get outside the parser... But maybe UNKNOWN is still of some use to people so I'll restore it.

matozoid: No wait. We just changed the semantics to unknown = Optional.none(). Can't have two things claiming the same.

matozoid:
Okay, here's another go, I think I processed all your feedback.

.gitignore
ftomassetti: sure, I was just curious

Issue #499, opened 2016-10-14
Make naming consistent


matozoid:
- Make fieldnames the same as getter/setter names
- Use "...List" for nodes that are a list
- Only use abbreviations like "Expr" when it's not the only word in a name, so no more "getArgs()"

matozoid:
- ArrayCreationExpr calls the element type just "type" where it should be "elementType" (add a getType here that adds ArrayType around it.)

rosrossito:
Hi, I would like to try to fix this issue.

matozoid:
Cool, the ArrayCreationExpr thing is already done.

rosrossito:
Could you clarify the task?  I looked folder "body" and it seems that everything is fine there. Could you give me some example?

matozoid:
How about ClassOrInterfaceDeclaration.extendsList and ConstructorDeclaration.throws_? EnumConstantDeclaration.args or arguments? InitializerDeclaration.isStatic and ClassOrInterfaceDeclaration.interface_? MethodDeclaration.typeParameters or typeParams?

rosrossito:
I see.
Thank you.

2016-11-05 16:49 GMT+02:00 Danny van Bruggen notifications@github.com:

rosrossito:
Is this necessary to make these changes in folder javaparser-testing also?

2016-11-05 20:20 GMT+02:00 Ruslan Omelchenko <
ruslan.omelchenko.irtcits@gmail.com>:

matozoid:
That would be consistent. Do you have something we can look at already?

rosrossito:
I changed all code except that folder.

2016-11-12 14:46 GMT+02:00 Danny van Bruggen notifications@github.com:

matozoid:
Yes, that is not the question. This task is one of the last things to finish before the big 3.0 release and I have exactly in my mind how to do it and would like to get it done and I'm really eager to do it. Please show what you have been doing in the last three weeks or I'll take over and do it in a few hours.

rosrossito:
Sorry I am not so fast. 
I pulled two request.
I still have problems with building project because of "maven-resources-plugin.version" in pom.xml of javaparser-testing.

ftomassetti:
Hi @rosrossito , I do not think you created any pull request. You created your own fork and committed your changes in your fork.

[...]

PR #601, opened 2016-11-28
Consistent naming

matozoid:
Handles #499.

Decided not to append "List" everywhere because that gets ugly.

Attempted to get field names to match getters and setters. Creatively worked around fields that are named after keywords.

Use "is" in the field names of booleans.

When expr, stmt, id, etc. are used as field name, expand them to expression, statement, identifier, etc.

Add some other improvements because I was changing things anyway.

This is only the ast.body package, I'd like to hear if this is going the right way.


javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
```diff
@@ -3,29 +3,29 @@
...
+    SYNCHRONIZED("synchronized"),
+    NATIVE("native"),
+    STRICTFP("strictfp");
 
-    String lib;
+    String asString;
 
-    private Modifier(String lib) {
-        this.lib = lib;
+    Modifier(String asString) {
```
ftomassetti: do we need this string? We can calculate it simply as `name().toLowerCase()`

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -119,41 +119,41 @@ public ClassOrInterfaceDeclaration(Range range, final EnumSet<Modifier> modifier
...
     public NodeList<ClassOrInterfaceType> getImplements() {
-        return implementsList;
+        return implementedTypes;
     }
 
     public NodeList<TypeParameter> getTypeParameters() {
         return typeParameters;
     }
 
     public boolean isInterface() {
```
ftomassetti, line 134: should we add `isClass`?

javaparser-core/src/main/java/com/github/javaparser/ast/body/ConstructorDeclaration.java
```diff
@@ -63,7 +63,7 @@
 
     private NodeList<Parameter> parameters;
 
-    private NodeList<ReferenceType<?>> throws_;
+    private NodeList<ReferenceType<?>> thrownTypes;
```
ftomassetti: what about `thrownExceptions`?

javaparser-core/src/main/java/com/github/javaparser/ast/body/Parameter.java
```diff
@@ -172,24 +173,25 @@ public Parameter setVarArgs(boolean isVarArgs) {
...
     }
 
-    public VariableDeclaratorId getId() {
-        return id;
+    @Override
+    public VariableDeclaratorId getIdentifier() {
+        return identifier;
     }
-
+
```
ftomassetti: nitpicking black belt: do we want space on an empty line?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/CastExpr.java
```diff
@@ -51,7 +54,7 @@ public CastExpr(Type<?> type, Expression expr) {
     public CastExpr(Range range, Type<?> type, Expression expr) {
         super(range);
         setType(type);
-    	setExpr(expr);
+        setExpression(expr);
```
ftomassetti: should be `setValue` instead of expression? After all also the CastExpr itself is an expression

javaparser-core/src/main/java/com/github/javaparser/ast/expr/CastExpr.java
```diff
@@ -51,7 +54,7 @@ public CastExpr(Type<?> type, Expression expr) {
     public CastExpr(Range range, Type<?> type, Expression expr) {
```
ftomassetti: Or CastExpression?

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExpression.java
```diff
@@ -0,0 +1,10 @@
+package com.github.javaparser.ast.nodeTypes;
+
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.Expression;
+
+public interface NodeWithExpression<N extends Node> {
```
ftomassetti, line 6: What are those nodes with Expressions? Are Expressions wrapping other expressions? Any sort of node containing one single expression?

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithSimpleName.java
```diff
@@ -21,22 +21,23 @@
...
+import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.expr.SimpleName;
 
 /**
  * A node having a name.
  *  
  * The main reason for this interface is to permit users to manipulate homogeneously all nodes with a getName method.
  */
-public interface NodeWithSimpleName<T> {
+public interface NodeWithSimpleName<N extends Node> {
```
ftomassetti, line 32: Good :D I wanted to suggest this one and I forgot

ftomassetti:
Thanks for this one: it is maybe not exciting code to write but I think it is pretty useful.
None of my comments are strong, just thoughts I had going over the code and all of them can be ignored

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
matozoid: ¯\_(ツ)_/¯  Don't know. Maybe a slight preference to keep it like this to keep enum-name disconnected from keyword.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ConstructorDeclaration.java
matozoid: Oh yeah, good.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
matozoid: Should we split COID into ClassDeclaration and InterfaceDeclaration?

javaparser-core/src/main/java/com/github/javaparser/ast/body/Parameter.java
matozoid: Absolutely!
(I'll remove it ;) )

javaparser-core/src/main/java/com/github/javaparser/ast/expr/CastExpr.java
matozoid: I was planning to keep the abbreviations for names that are not just the word itself: class Expression, class CastExpr.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExpression.java
matozoid: They are nodes containing an expression. They can be expressions themselves. The interface is really general as you can see, but it is easy for refactoring ;)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/CastExpr.java
matozoid: Should it? I'll go check the JLS for naming... https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.16

The right side is definitely called "...expression" here, so I vote for keeping it this way.

matozoid:
No problem, my main goal is to get all the fieldname/getter/setters in sync so it's easier to make a metamodel :)

matozoid:
I think the change that may annoy people is that the AST has become more verbose. getIdentifier() instead of getId(), getExpression() instead of getExpr(). The sooner I know if this is seen as positive or negative, the easier it will be to change.

Issue #596, opened 2016-11-23
Fix remaining nullable fields to use Optional


matozoid:
Some were missed in the previous PR

DeepSnowNeeL:
Will do it ! Can we list them ?

matozoid:
Just look at all TODO's, there are around six that mention null left.

PR #604, opened 2016-11-30
More optionals

DeepSnowNeeL:
Handles #596 (+include @matozoid PR #601)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
```diff
@@ -105,13 +107,13 @@ public static Name parse(String qualifiedName) {
      */
     public String asString() {
         if (qualifier != null) {
```
matozoid, line 109: Oopsie! Please fix :)

PR #605, opened 2016-11-30
Remove MinValue nodes

### LINKED ISSUE #602: Remove IntegerLiteralMinValueExpr and LongLiteralMinValueExpr
I have no idea why there are special AST nodes for the minimum value of an int or long.

matozoid:
Fixes #602


ftomassetti:
+1 even if I have still to understand why these classes existed in the first place...

matozoid:
@jgesser should know...

jgesser:
@matozoid @ftomassetti 
Unfortunately I don't remember. It probably made sense 10 years ago...
I raised my old CVS from the grave trying to find some clue in the commit history from the original sources but it seem to be there since the beginning.
So, probably it can be safely removed.

jgesser:
I thing maybe I have remembered why these classes exists:
In the beginning, there was only one `LiteralExpr` class, in which the value field was of type Object. This field was used to store every kind of literals, but they were transformed in the corresponding object type (Byte, Integer, Long, Float, etc.)
As the `LiteralExpr` does not maintain the minus sign (the minus sign is represented by an `UnaryExpr` encapsulating the `WhateverLiteralExpr`), it was generating an error when parsing a source with a literal like:
`int i = -2147483648;`
The javacc was like this:
`<INTEGER_LITERAL> { ret = new LiteralExpr(Integer.parseInt(token.image)); }` // error here

When I refactored the code to store the literal value as String I forgot to remove this ugly solution.

matozoid:
@jgesser cool, thanks for the feedback! That means this PR is a go, and that we will release 3.0-RC.1 after months of work :)

Issue #591, opened 2016-11-23
Make pattern/tree matching functionality like ANTLR


matozoid:
[information](https://github.com/antlr/antlr4/blob/master/doc/tree-matching.md)

Extend the grammar to allow placeholders for certain types of nodes (expressions, statements, types, names...) and put these placeholders in the AST. Add functionality to either find this pattern in the AST, or to fill the placeholders.

ftomassetti:
Incidentally we could use something like this also for lexical preservation, maybe.
So we could have the children as placeholders

matozoid:
I'm building a toy project that translates Java into Commodore 64 BASIC. I immediately missed a feature in JavaParser, and I'm building it now. It's a minimal implementation of what [http://www.rascal-mpl.org/](Rascal) does, mixed with the regular expression approach:

```java
        // A pattern is simply an AST fragment.
        // The only thing it supports is named placeholders for some types of nodes.
        Expression pattern = parseExpression("6+6");
        // An "Iterable" interface for matching seems a decent approach:
        for(AstMatch match: pattern.matcher(expression)) {
            // So we matched a node that is a BinaryExpr with two IntegerLiteralExpr's containing "6".
            // Replace it with integer literal 36.
            match.replace(parseExpression("36"));
        }

        // Here are some placeholders.
        // They work because the grammar is extended to parse the things between `` into PlaceholderStmt or PlaceholderExpr nodes.
        Expression pattern = parseExpression("`a`+`b`");
        for(AstMatch match: pattern.matcher(expression)) {
            // Now we matched a node that is a BinaryExpr with two nodes in the position of a and b.
            // Replace it with a method call to "add"
            match.replace(parseExpression("Utils.add(`a`, `b`)"));
            // Or maybe do something else with the placeholders:
            Node a = match.get("a");
            Node b = match.get("b");
        }
        
        // Shortcuts:
        someNode.replace(Node pattern, Node replacement); 
        someNode.findAll(Node pattern, Consumer<AstMatch> consumer);
```

I'm not sure how to match something like "a method with any combination of modifiers" when I can also match "a public static method".

ftomassetti:
Nice! I was trying to build something like this when we were working on the patterns library. I should have some code somewhere...

matozoid:
Patterns library? The matcher library? I think it would complement this feature very well!

ftomassetti:
Yes, I meant the matcher library. It could become part of this project

Issue #582, opened 2016-11-20
Make parse errors in blocks recoverable


matozoid:
These should look for the next "}"

matozoid:
We could insert a "BrokenStatement" in the AST with the literal text of the statement in it.

Issue #581, opened 2016-11-20
Make parse errors in statements recoverable


matozoid:
These should recover by looking for the next ";"

matozoid:
We could insert a "BrokenStatement" in the AST with the literal text of the statement in it.

martinlippert:
I am very interested in this whole "make javaparser recoverable from errors" story. Any news on this?

matozoid:
It's not currently being worked on, but the steps to take are more or less clear.
1. We need to play around with the information on this page: https://javacc.java.net/doc/errorrecovery.html to make sure the parser doesn't throw parse errors, but only inserts a "bad node" and adds a Problem. 
2. We need to move the semantic checks (meaning: the check in Java code) in the grammar to a validation framework that we yet have to build.

There are issues for both of these, waiting for someone to pick them up :-)

martinlippert:
Thanks for the details. At the moment we don't have the cycles to work on this on our own. If that changes (and this isn't already done), I will let you know for sure.

ftomassetti:
_Here is the marketing department speaking_ I could not help but notice that your GitHub profile says "Pivotal". It would be could for us to brag about Pivotal being using JavaParser. We have to start creating our testimonials page. Do you think would be possible to bother some of you with a couple of questions on JavaParser? It would not take long, you should just say "it is great, it changed my life and my dog loves me more since I use JavaParser". That would mean a lot to us. _End of marketing comunication, back to business_

martinlippert:
Hey _marketing department_, I understand and support your request, but it is way too early for us to provide a quote. We need a gather a lot more experience using this before we can be part of a testimonials page and haven't even decided yet which exact way to go. I would prefer to use this project for parsing code, but since we are working on code editing capabilities, the error recovery is a very essential part. So this missing piece is preventing us from diving deeper into this at the moment, but again, we are at the very beginning here. But we can for sure come back to the _marketing department_ questions then.

ftomassetti:
thanks @martinlippert and for anything feel free to reach us

matozoid:
@jeandersonbc is now working on this.

martinlippert:
great news and let me know if there is something to try

jeandersonbc:
glad to hear you are interested @martinlippert! we are thinking on make something based on the so-called "panic mode" recovery, which basically skips the input stream until it finds a "synchronization token" (like a semicolon or closing brackets) to continue the parsing:

```{java}
import; // first parsing error

[...]

PR #609, opened 2016-12-05
Remove EmptyImportDeclaration.

matozoid:
Note: we were using EmptyImportDeclaration as a placeholder for broken import statements. Something like "import static x;" can be parsed, but doesn't fit in the AST, so EID was used there.

Since this is a pattern we can expect when we add more recovery into the parser (#136), we need a way to deal with this. My current "good enough" solution is to introduce BadImportDeclaration which stores the literal text and nonsense for other fields. More Bad nodes are needed for #581 and #582.

A variant on this pattern will appear when we introduce placeholders in the AST - #591. We need to be able to store an Expression which is a placeholder, so a PlaceholderExpression. We might run into trouble here with having to define a Placeholder variant of every type of Node we have!


matozoid:
Note that I killed a test too, since parsing ";" as an import is no longer valid.

PR #610, opened 2016-12-05
Setting position of package protected constructors

ftomassetti:
Hi guys, I received this patch by e-mail from Leon Poyyayil

> First of all: a big thank you for the outstanding project of the Java
parser! Works like a charm.
> 
> Pushing some existing code through it, I encountered a small issue:
> 
> When parsing a class which contains constructors with package protected
access (e.g. no modifiers at all), then the begin position of the
resulting ConstructorDeclaration is zero - invalid.
> 
> This might not be a big issue in most cases, but for the purpose where I
ran into it, it was crucial (attempting to count lines of methods).
> 
> => Good news: I found a working solution/fix.
> 
> I have attached a patch, which contains the fix as well as a new unit
test with associated test data.
> 
> Based on the checkout of version 3.0.0-RC.1, you can apply the patch in
its main directory as follows:
> 
>   patch -p1 < PackageProtectedCtorNodePosition.patch
> 
> Hope that this is acceptable to you - I just wanted to share it, to give
something back.
> 
> The patch also contains a somewhat more elaborate description of the
issue in the test data source code.
The patch further contains a missing dependency for the slf4j-api,
necessary to run the unit tests in maven.
> 
> Note: I have compared the unit test output results with and without my
patch and they are equivalent. Unfortunately, the current release has
failures in the unit tests, but the patch doesn't change the behavior.
The new test case is also verified to work as expected (fails without
patch and passes with it).


matozoid:
Cool, that's exactly the correct fix.

matozoid:
Can we make the tests more "unit" ? Five years ago we started with "parse this huge Java file" as the only test, so let's not introduce more chunks of Java when we want to test only a single thing. I think this simply tests that `JavaParser.parse("class X {}")` gives X position 1,1 - not none().

matozoid:
Another thing: can you quickly go through all places where we use modifiers and check that the logic is correct there?

ftomassetti:
Instead of parsing two files I moved to parse short snippet of online code covering a few more cases which are correct now but that we may want to protect from regressions

matozoid:
+1 from me :)

ftomassetti:
To me all places seem ok but to be sure I have added a few more tests

Issue #592, opened 2016-11-23
Redo array support


matozoid:
- do not store a `List<ArrayBracketPair>` anymore
- turn the currently generated `getType()` methods into getters, and do the generating at parse time

because:
- exact bracket location is of little interest to users analysing the AST
- having a generated type is awkward since it is not really part of the AST
- for refactoring and reprinting source code this information is not absolutely necessary, depending on the algorithm used for reprinting

result:
- all types are part of the AST again
- `int[] a[]` will pretty print as `int a[][]`
- it will be possible to invalidate the AST by taking `int a, b` and setting the type of b to `String`

ftomassetti:
For fields and variables we could have:
```
int[] a[], b;
```
Would we be able to figure out the right types?

While for paramters not being able to distinguish between `int[] a[]` and `int[][] a` or `int a[][]` would not be that important.

matozoid:
`int[] a[], b` would give:
```
a = ArrayType(ArrayType(int))
b = ArrayType(int)
```
and the whole expression doesn't mention type, so no more elementType. Where do you suspect a problem?

ftomassetti:
I am happy if they give me the "calculated type" (so `ArrayType(ArrayType(int))` in the case of `a`).
The important thing is that the type is part of the AST (has a valid parent). Otherwise I would prefer to get the type of the field and Array brackets of `a` to build the type myself.

matozoid:
That's exactly the problem I want to solve - make the type part of the AST :)

matozoid:
Here's the newest plan (compare to #237 )

``` java

int[] a, b[], c;

- VariableDeclarationExpr
    - List<VariableDeclarator> vars
        - Type type
        - VariableDeclaratorId id

class X {
    int[] a, b[], c;
}

- FieldDeclaration
    - List<VariableDeclarator> variables
        - Type type
        - VariableDeclaratorId id

class X {
    int[] a()[] {
         ...
    }
}

- MethodDeclaration
    - Type type

void a(int[] b[]) {
    ...
}

- Parameter
    - Type type
    - VariableDeclaratorId id
```

Issue #612, opened 2016-12-06
Remove VariableDeclaratorId


matozoid:
Without bracket information, VariableDeclaratorId is identical to SimpleName. Removing it would simplify the AST

PR #615, opened 2016-12-07
Issue 592 redo array support

matozoid:
Fixes issue #592. Makes type a first class citizen Node everywhere. Calculates the correct ArrayType() nesting while parsing. No longer supplies information about where array brackets are located. No longer has element type as a first class citizen, but if you have a type, you can always ask it for its element type.


matozoid:
Also fixes #612

matozoid:
@ftomassetti some observer tests keep failing and I don't know why. Would you like to help? (If not, I'll investigate deeper)

matozoid:
@ftomassetti never mind, found it.

matozoid:
@ftomassetti but maybe you still want to double check what I did to the tests.

Issue #580, opened 2016-11-20
Make a method on Providers for resources.


matozoid:
So that we don't have to go through `getClass().getResourceAsStream("/blah.java"), UTF_8`

tiaped:
Hi I would like to work on this issue.
But I've found only three uses of getResourceAsStream and all in tests.

Do you have a use case for this issue?

If the use is limited to tests why not use a TestUtils in package com.github.javaparser of the test project.

Thanks in advance for answers

matozoid:
Hi, thanks for offering some help.

The core library is not made only to be used in the test module. It is made to be used by users. The Providers class is there to make it easy to pass any kind of input to JavaParser. Since people often use resources, it seems logical to support those in Providers.

tiaped:
Thanks for the answer.
OK, so let's go.
If I do not mistake, the aim is to use something like this 
Charset charset = Charset.forName("US-ASCII");
Providers.provider(Providers.getResource("path/to/resource",charset),charset);

I could also add a providerFromResource(String pathToResource, Charset charset) method, but I do not think it matches current method names

I'll work on getClass().getClassLoader().getResourceAsStream("/blah.java") if you don't think it's silly. :)
Though I could try both.

matozoid:
Okay, cool.

The API should be one call, like `Providers.provider("path", charset)` - if there is an overloading problem, make it `Providers.resourceProvider` or so.
Also have an alternative that defaults charset to UTF_8.

Try it out by writing a test and calling JavaParser with it :)

Oh, make sure to test if everything is closed nicely when parsing is done!

If you could add some Javadoc with a line or two telling the user if the path needs a "/" at the start, and where it looks, that would be great.

tiaped:
Hi again,

I nearly finished. But I have one last question. Sorry in advance for the poor english.
As resources are related to class loaders I added a third alternative that allow to provide the class loader.
Typically when you call a resource, it is in a jar, Java files are rarely present in jars. So I still so not see the use case unless a source files are provided by a specific class loader.

I might have misunderstood the need of this issue :)

matozoid:
My main use case is: "I've put some source code on /src/main/resources/ and I want to analyse it", so `provider("/template/Abc.java")` would get input from there.

It would be possible that someone puts the source jar of some artifact on the classpath, but I agree that seems far fetched. You choose if you want to support that or not :)

DeepSnowNeeL:
I actually do this to generate code from my annotation processor :)

[...]

PR #616, opened 2016-12-08
Issue #580 : Make a method on Providers for resources.

tiaped:


javaparser-core/src/main/java/com/github/javaparser/Providers.java
```diff
@@ -57,4 +57,27 @@ public static Provider provider(String source) {
...
+     * @param pathToResource path to resource in provided class loader. As working relatively to a class loader, no "/"
+     * at the beginning of the path
+     * @param encoding encoding
+     * @return the provider associated to pathToResource
+     */
+    public static Provider resourceProvider(ClassLoader classLoader, String pathToResource, Charset encoding) {
+        return provider(classLoader.getResourceAsStream(pathToResource), encoding);
+    }
+
+    public static Provider resourceProvider(String pathToResource, Charset encoding) {
```
matozoid, line 74: Does this and the following one need a "/"? I'm alway confused...

matozoid:
This is a fine PR :) 

Implements #580

javaparser-core/src/main/java/com/github/javaparser/Providers.java
tiaped: I'm far to be a master of clarity :) 
I will add javadoc to each of the methods. In order to add the precision for each one of them.
Is the first javadoc OK for you?

matozoid: Oh, I'm all for documenting only things that are not obvious, so you can leave out "@param encoding encoding @return the provider associated to pathToResource"

(Just assume that people only look at the Javadoc if they are confused, so it should contain information on things that may be confusing :) )

matozoid:
Thanks @tiaped - you're welcome to pick up more issues!

Issue #622, opened 2016-12-13
Make assigning comments optional again


matozoid:
... with a flag in ParserConfiguration. The default should be to assign them.

PR #625, opened 2016-12-13
JavaParser.java cleanups

matozoid:
* reuse javaparser instance when using static methods
* let ParserConfiguration follow the usual JP style


javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -47,7 +47,10 @@
  * @author Júlio Vilmar Gesser
  */
 public final class JavaParser {
+    private static final JavaParser staticJavaParser = new JavaParser();
+
     private final CommentsInserter commentsInserter;
+    private final ParserConfiguration configuration;
```
ftomassetti, line 53: Is this field used somewhere?

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -47,7 +47,10 @@
  * @author Júlio Vilmar Gesser
  */
 public final class JavaParser {
+    private static final JavaParser staticJavaParser = new JavaParser();
```
ftomassetti: what about `defaultJavaParser` or `defaultInstance`?

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
matozoid: Yes, for the new "attributeComments" flag. I forgot to refer to the issue: #622

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
matozoid: Well, it's the one for the static methods. To be honest, I had an "instanceParser" and "staticParser" first, so that made more sense.

I don't care much - what would you like its name to be?

ftomassetti: defaultInstance, if it makes sense for you

PR #626, opened 2016-12-13
Better problem stacktrace output

DeepSnowNeeL:


matozoid:
Choose: either use Utils.EOL everywhere, or use System.lineSeparator() everywhere :)

(Didn't know that last one existed!)

PR #631, opened 2016-12-13
Add getElementType and getCommonType to field and variable declarations

### LINKED ISSUE #621: Helper for fieldDeclaration.getVariable(0).getType()
Previously we could do getElementType() directly on a FieldDeclaration,
Now that's been reworked, it would be great to have a helper to directly get the type of the first variable declared by this field.

`getFirstVariableType()` can be cool
Maybe we could have `getFirstVariableIdentifierAsString()` too ?

Thanks

matozoid:
Fixes #621


matozoid:
Travis died randomly :(

Issue #632, opened 2016-12-13
Pretty printing variables with array types


ftomassetti:
Currently this `byte[] rawData` is pretty printed as `byte rawData[]`.
I would suggest we pretty print it as `byte[] rawData`. We could check the common type between all the fields and print it in front of the variables, while putting extra brackets behind the variables.

For example:
`int a[], b[][]` -> `int[] a, b[]`
`int[] a[]` -> `int[][] a`
`int[] a, b[]` -> `int[] a, b[]`

What do you think?

ftomassetti:
If you agree I can send a PR

matozoid:
Please do send a PR, I know I've taken the lazy solution here :)

ftomassetti:
I can easily find the common type between different fields but to preserve annotations correctly probably we should remember where the brackets were...

matozoid:
Well, that certainly makes it more complex, but if you do it right you should be able to add brackets to the type on the left as long as all variables have the same annotations on those brackets...

Or you say "if there are annotations on the brackets I'll just stick all brackets behind the identifiers, as it is now. It's not like I've ever seen an annotated array bracket.

PR #637, opened 2016-12-14
Issue632: pretty printing array brackets using common type

ftomassetti:
See #632


matozoid:
Work in progress?

ftomassetti:
Well, I guess I just need to fix the tests, not the code :) working on it

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
```diff
@@ -36,4 +39,12 @@ default VariableDeclarator getVariable(int i) {
     }
 
     N setVariables(NodeList<VariableDeclarator> variables);
+
+    public default Type getCommonType() {
+        return getVariables().stream().map(v -> v.getType())
```
matozoid: You're conflicting with #631 :-O Shall I go pull some PR's?

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
```diff
@@ -36,4 +39,12 @@ default VariableDeclarator getVariable(int i) {
...
+                .sorted((Comparator<Type<?>>) (a, b) -> a.getArrayLevel() - b.getArrayLevel())
+                .findFirst()
```
matozoid: I guess you're going to add some code that checks the annotations?

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
ftomassetti: I will wait for #631 to be merged and adapt

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
ftomassetti: Damned annotations :D
Well, this seems an interesting problem that I will fix :D

matozoid: Oh look, I can edit eveybody's comments! :D

ftomassetti: Ah ah ah, with great power comes great responsibility :D 
Ok, I will do it, after #631 is merged

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
matozoid: Okay, it's merged. Note that ".junit." has disappeared from the package name of many unit tests.

matozoid: Oh hey, the merged getCommonType and yours are not the same. Mine only returns the type of the variables if it is the same for each one. What shall we do about naming?

The thing you are describing has no specific name in the JLS, as far as I know. Neither does mine.

ftomassetti:
Ok, considering annotations was tricky...

matozoid:
This is awesome!

*sits and cries in a corner* ;)

ftomassetti:
Maybe not the clearest code ever written... :)
This is the day I deserved the hell of developers

matozoid:
Oooh, I just thought of something. getMaximumCommonType() is only used for pretty printing, I would say. So why not move it to the pretty printer?

ftomassetti:
I think it could be potentially be more useful, but given it is not a clear method it could make sense to move to a less visible/general place. I will do that

matozoid:
It can always be moved around again.

matozoid:
Thanks for doing this feature!

ftomassetti:
thanks for the feedback

Issue #639, opened 2016-12-14
Make TAB character size used by JavaCC configurable


matozoid:
It is set to 8, and all positions are calculated using this number. This may very well be wrong for some cases. I think it is mostly 4 nowadays, but should be configurable.

matozoid:
This was found after reading @larsac07's thesis: http://bora.uib.no/bitstream/handle/1956/12756/144804725.pdf

@larsac07 - if you have more suggestions, please add :)

larsac07:
Yay! Someone read my thesis!

matozoid:
@larsac07 turns out you weren't the only one bothered by the TAB thing, by the way - in the new version it was changed to 1 already :)

PR #640, opened 2016-12-14
Make tab size configurable

matozoid:
Issue #639. Also makes public fields in configuration private.


DeepSnowNeeL:
Shall it be a tabStr instead of an int ? Imagine I want my tab to be real tabs "\t" ? :P

matozoid:
This is the input side: when JavaCC sees a tab character, it will set the position forward to the next tab stop of this size.

DeepSnowNeeL:
Oh allright ! I tought it was for the pretty printer :)

PR #647, opened 2016-12-16
Minor changes to JavaToken

ftomassetti:
Correct a typo and rename a field


ftomassetti:
Rebased

matozoid:
"Image" is the name javacc uses, but I like this as well.

PR #649, opened 2016-12-17
Correct Range contains to consider equal ranges as contained

ftomassetti:
Introduce `strictlyContains` with the same logic `contains` had before.

Add tests for several cases, considering both `contains` and `strictlyContains`.


matozoid:
Isn't normal terminology "inclusive" and "exclusive" or so?

ftomassetti:
I was thinking about "strict subsets" and "subsets" where A is a subset of A but it is not a "strict subset" of A.

matozoid:
I'm out of ideas ¯\\\_(ツ)\_/¯

+1

PR #651, opened 2016-12-18
Add field after building it

ftomassetti:
Why? Because in this way observers on the class get the field when it is built, instead of getting an half-built Field that they could not process properly.


matozoid:
Do you think we should do this in more places?

ftomassetti:
Maybe yes, but these things will become evident while working on Lexical Preservation (because it is based on observers)

PR #658, opened 2016-12-18
Issue 718 create metamodel

matozoid:


javaparser-bootstrap/pom.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>javaparser-parent</artifactId>
+        <groupId>com.github.javaparser</groupId>
+        <version>3.0.1-RC.2-SNAPSHOT</version>
```
ftomassetti: Is there any way we could avoid having to specify the version here?
Also, is 3.0.1, not 3.0.0?

javaparser-bootstrap/src/main/java/com/github/javaparser/bootstrap/VisitorGenerator.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+
+public class VisitorGenerator {
+    private static JavaParserMetaModel javaParserMetaModel = new JavaParserMetaModel();
+
+    public static void main(String[] args) throws IOException {
+        final Path root = new File(VisitorGenerator.class.getProtectionDomain().getCodeSource().getLocation().getPath() + "/../../../javaparser-core/src/main/java/").toPath();
```
ftomassetti: does this work on Windows? Should we use File.separator?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/VoidVisitor.java
```diff
@@ -41,189 +35,180 @@
  */
 public interface VoidVisitor<A> {
 
-    //- Compilation Unit ----------------------------------
-
-    void visit(CompilationUnit n, A arg);
+    void visit(AnnotationDeclaration n, A arg);
 
-    void visit(PackageDeclaration n, A arg);
+    void visit(AnnotationMemberDeclaration n, A arg);
```
ftomassetti, line 43: would it make sense to order the methods by grouping related nodes? Like, all the literals near each other

javaparser-bootstrap/pom.xml
matozoid: This is how it is in all poms. The release plugin changes that around at release time.

matozoid:
Don't you think it's cool that I've built a "SourceRoot" thing where you can parse arbitrary java files by specifying package and name, and SourceRoot keeps track of it so you can write all of this files out by just calling save()? I'm using that to do in-place generation of code: the actual Java files in the core module are being adjusted. So in theory we can write code here that modifies any file in core (like the copyright thing at the top)

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/VoidVisitor.java
matozoid: I would consider adding some meta-data that indicates this grouping first - but it's not a priority since I want this to actually generate the visitors first :)

ftomassetti:
Yes, it is cool :D

javaparser-bootstrap/pom.xml
ftomassetti: right!

javaparser-bootstrap/src/main/java/com/github/javaparser/bootstrap/SourceRoot.java
```diff
@@ -0,0 +1,84 @@
...
+import com.github.javaparser.ast.DataKey;
+import com.github.javaparser.printer.PrettyPrinter;
+
+import java.io.*;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.*;
+
+/**
+ * A collection of files located in one directory on the file system.
```
ftomassetti: And also subdirectories, right?

javaparser-bootstrap/src/main/java/com/github/javaparser/bootstrap/SourceRoot.java
```diff
@@ -0,0 +1,84 @@
...
+        startPackage = startPackage.replace(".", File.separator);
+        return Paths.get(root.toString(), startPackage, file).normalize();
+    }
+
+    private Path packagePath(String startPackage) {
+        startPackage = startPackage.replace(".", File.separator);
+        return Paths.get(root.toString(), startPackage).normalize();
+    }
+
+    public void save() throws FileNotFoundException, UnsupportedEncodingException {
```
ftomassetti: should these be `saveAll`?

matozoid: Okay

javaparser-bootstrap/src/main/java/com/github/javaparser/bootstrap/SourceRoot.java
matozoid: Okay

javaparser-bootstrap/src/main/java/com/github/javaparser/bootstrap/VisitorGenerator.java
matozoid: Okay

matozoid:
I'm brainstorming different levels of equality now...
* "strict" which check on token level: `1+1` is not equal to `1+    1`
* "normal" "default" or "(no special name)" which checks on AST level except comments: `1/*hello*/ + 1` equals `1+1`
* "structural" which is like normal, but ignores differences in literals: `1+1` equals `2+3` but not `1*1`
* "withComments" which is like normal, but checks comments: `15; // fifteen` is not equal to `15;`

I'm not extremely happy with these names... Ideas?

matozoid:
I've made a mental mistake here. By having the generator work on a model created by introspecting all nodes, I've made a compilable project a requirement for running the generator. But one of the main advantages of the generator is being able to fix compile problems caused by modifying the code of a node.  So I have to cut this into two steps:

1. introspect the nodes into a static metamodel.
2. use the metamodel to generate stuff.

1 can be run when the code is still compiling. 2 depends on 1 only, and does not need to compile the nodes anymore.

1 will always lag behind a bit since it is the model of the last compilable source code. There are two solutions I can see:

a. modify the static metamodel by hand
b. use JavaParser or JavaSymbolSolver to "introspect" nodes without the need to compile, instead of using reflection.

Maybe both...

ftomassetti:
In general you want to have a model and use it to generate a portion of Java classes, while defining manually other methods, correct?

For example, I would like to generate all the setters but maybe some convenience constructors would be written manually.

How would you merge the manual changes with the changes obtained from regenerating the code from the model?

So, on GitHub  patch modify a piece of code and move a generated method around or change a generated method. What happens when regenerating the code from the model next time?

Maybe we could have a model (initially obtained from the code and then modified manually). From such model we could generate code and compare the generated methods with the existing methods in the repository and then override single methods (while preserving their position in the file) or report the ones which are different. Some could be different because they have been manually changed for some reason (maybe the JavaDoc was changed, maybe the indentation was changed, maybe the logic was changed)

matozoid:
What I was planning to do (and the current code doesn't do that yet) is to annotate all generated methods. We would run through all the nodes, look at all the methods, and all of them that are annotated with @Generated(more, specific, info) would be regenerated.

Alternatively we could recognize methods by name as you seem to imply.

The visitors are currently modified in place: read them with JavaParser, change stuff, and write them back out in the same location.

I like the idea of reporting which have been changed manually - but git can also show me the diff :)

matozoid:
@ftomassetti - could you take a look at https://github.com/matozoid/javaparser/blob/8c42b2c4013fa95ebe3852ccecfc937580e023ff/javaparser-metamodel/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java ?

It's a mostly generated file that should be the entry point to the meta model. It contains instances of https://github.com/matozoid/javaparser/blob/8c42b2c4013fa95ebe3852ccecfc937580e023ff/javaparser-metamodel/src/main/java/com/github/javaparser/metamodel/ClassMetaModel.java for every node (and nodelist) and those should contain instance of https://github.com/matozoid/javaparser/blob/8c42b2c4013fa95ebe3852ccecfc937580e023ff/javaparser-metamodel/src/main/java/com/github/javaparser/metamodel/FieldMetaModel.java for their fields.

These classes are currently written for my use case, but I imagine we could use them for the formatter as well. My question to you: how would you want to use this model to find the information you want?

ftomassetti:
I would tend to organize the things like in metamodels that I have worked with, for example the Eclipse Modeling Framework. I am not saying that is the way to go or we should do this, I am just stating my opinion based on things that are familiar to me, so that they make me feel comfortable because I do not have to think too much to understand different ways of doing things :D 

In my view the metamodel should be as simple and as abstract as possible and it should contain only the information strictly needed to define what information is contained by each element.

I would rename `ClassMetaModel` into `NodeType`: the fact that we generate a class from it is secondary, in principle we are describing the types of nodes. What can be calculated should be calculated imho, for example `qualifiedClassName` can be calculated from `className` and `packageName`. 

Very, very minor: ideally I would also remove `packageName` and create a `NodeTypeGroup`. Each `NodeType` would have a reference to a `NodeTypeGroup`. Then incidentally we would map each `NodeTypeGroup` to a package.

I would rename `FieldMetaModel` in `Property`. All properties have a name and all properties can have a multiplicity ONE or MANY. They could contain values of two types:

* primitive values (typically booleans or strings, maybe also Modifier): I would call them Attributes
* other nodes. They would contain the children of the node. I would call them References.

so for example a ClassOrInterface could have:

* Attribute(name="interface", type=Boolean, multiplicity=ONE)
* Attribute(name="modifiers", type=Modifier, multiplicity=MANY)
* Reference(name="simpleName", type=theSimpleNameNodeType, multiplicity=ONE)
* Reference(name="members", type=theMemberNodeType, multiplicity=MANY)

We could also add another flag for properties: `derived`. Derived properties cannot be set but only calculated. This is for example the case of the "maximum common type" of a FieldDeclaration. That type is calculated looking at the types of the VariableDeclaration inside the FieldDeclaration.

Does it make any sense?

Again, let me stress that I am not in any way saying that your approach is wrong, I just report what I would tend to do.

And sorry for being terribly late.

ftomassetti:
Also, a property should be optional/required and all our nodes will probably have the property "range" (maybe we could treat it differently from other properties)

matozoid:
- Indeed I would like to say Node instead of Class - but NodeList is in there. I'm considering booting it out of the model so we can talk about nodes. What do you think?
- I appended "MetaModel" everywhere in an attempt to avoid confusion.
- "Node" and "Property" are fine names.
- The model could be simplified a lot - I was following our "just decorate everything with helpers" philosophy.
- as much as possible of the calculating was moved to the generator. I'm indeed not sure that is such a great thing. It would certainly make the generator simpler not to do so.
- how does NodeTypeGroup work? Is it like a qualified name where they point to their next component? What is the advantage?

This is very interesting feedback and I'll be rebuilding stuff.

ftomassetti:
* In my opinion NodeList should not appear in the metamodel: we will generate it later, for references with multiplicity MANY. I understand that in practice you want to change same names to avoid confusion and adding a suffix makes sense
* in the EMF or similar worlds what I would do if I was afraid to much logic would appear in the generator is having intermediate transformation: I start with very abstract nodes, I do some transformation and I get a more explicit bunch of nodes automatically, I give those nodes to the generator. In this case I am not sure this is necessary. Maybe just writing a few helper classes for calculating things like the return type of methods (like `Optional<NodeList<ParameterDecl>>` depending on the type of the reference, its multiplicity and it being optional or not). You could even just put those methods in the Reference class (`calculateValueType`)
* the NodeTypeGroup could be organized in hierarchies if you want or just contain a qualified name. The advantage would be close to zero: is just that you would have a reference to the whole group of nodes

matozoid:
- OK. Makes sense, with the exception that we also use sets so "multiplicity" is too abstract.
- How would you go about transforming these nodes? Is that like defining your own VisitorGeneratorNodeModel and VisitorgeneratorPropertyModel and building a transformer?
- I see. I don't think our package structure is very useful except for locating the classes - I do want to add some meta-information like "isExpression" "isComment" "isStatement" which should help with what one would want to do with the packages, I think.

So... I'll reduce the amount of fields to the absolute minimum, add the functionality we lose as methods, use different terminology, ditch NodeListMetaModel, and improve the link from a property to another NodeType. Next to that, I think it would be good to move it to a separate module to avoid confusing people.

matozoid:
Okay, done :) The generator is ugly, but the result is pretty nice if I say so myself.

I wanted to avoid some casting for references and attributes, so instead I gave each Property(MetaModel) a field "type" that is directly the class of the property (minus wrapping generics) and an optional nodeReference to the metamodel of the property type in case it is a node. So "prp.nodeReference.isPresent" means "is property prp a node?"

matozoid:
See #718

matozoid:
@DeepSnowNeeL - you were into creating visitors a while ago, would you like to review this too? There are also some code generation additions (especially SourceRoot) that you may have ideas for.

ftomassetti:
Travis is sad :(

We have ~50 commits and ~130 files changed. How do we tackle this? Is there anything we should extract into a smaller PR? Where should we start looking?

matozoid:
Travis is happy :)

Here goes:

# new modules
- javaparser-generator-utils contains some utilities that I reused while writing all the generation code.
- javaparser-metamodel-generator has a class with a main method that will introspect the AST nodes in core to generate:
- javaparser-metamodel: a huge bunch of mostly generated code that describes the AST nodes.

- javaparser-visitor-generator: code that generates two visitors using the metamodel, and a go at a generic "visitor generator" for users. You can find a new, generated HashCodeVisitor in core. This could be a separate PR, although that would remove the actual use case of the metamodel from this PR.

A discussion on whether we need all of these modules is fine.

# classes of interest
- SourceRoot is a fun class that takes away the hassle of locating a java file, updating it and writing it out again
- MetaModelGenerator is the ugly class that generates the metamodel
- JavaParserMetaModel is the huge class that is the "face' of the metamodel.
- BaseNodeMetaModel and PropertyMetaModel define the content of the metamodel
- HashCodeVisitorGenerator is the most complex usage of the metamodel for now.

The generators can be run by simply running the main class. They will figure out where to read and write their artifacts.

ftomassetti:
I like having multiple modules, however we should invest in documentation to not scare new contributors. We would need to explain the role of each modules and what is generated code and should not be changed (maybe add a comment like "Hey, this is generated!" in all generated files, if it is not already present)

matozoid:
It's interesting - there is an @Generated annotation Java. I looked into it, and you are supposed to pass it the fqn of the generator. That makes it so ugly :( :(

matozoid:
So... no complaints?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -218,7 +219,7 @@ public final Node setComment(final Comment comment) {
     }
 
     public boolean hasJavaDocComment() {
-        return hasComment() && getComment() instanceof JavadocComment;
+        return hasComment() && getComment().get() instanceof JavadocComment;
```
arturbosch: This may lead to a NoSuchElementException. I will take a look at the changes a bit later and mainly focus on code quality as I don't have an usecase for the metamodel yet :)

ftomassetti:
~~It is a bit too big for me to review it: if we wait for me to take a look at it this could mean waiting for weeks.~~

Yes, it looks good to me. I forgot I already look into it! Time to rest a little bit :D 

I have faith in you :D

matozoid:
Please don't! :D 

Maybe I should put all this stuff in a separate project as "experimental?"

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -167,6 +168,23 @@ public ClassOrInterfaceDeclaration setTypeParameters(final NodeList<TypeParamete
         return this;
     }
 
+    /**
+     * Try to find a {@link ConstructorDeclaration} with no parameters by its name
+     *
+     * @return the methods found (multiple in case of polymorphism)
+     */
+    public Optional<ConstructorDeclaration> getDefaultConstructor() {
+        for (BodyDeclaration<?> bd : getMembers()) {
```
arturbosch: Can be replaced in a more functional way:

```
return getMembers().stream()
                .filter(bd -> bd instanceof ConstructorDeclaration)
                .map(bd -> (ConstructorDeclaration) bd)
                .filter(cd -> cd.getParameters().isEmpty())
                .findFirst();
```

javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
```diff
@@ -363,8 +363,8 @@ public String getDeclarationAsString(boolean includingModifiers, boolean includi
 
     @Override
     public JavadocComment getJavadocComment() {
-        if (getComment() instanceof JavadocComment) {
-            return (JavadocComment) getComment();
+        if (getComment().isPresent() && getComment().get() instanceof JavadocComment) {
```
arturbosch:
```
 if (getComment().isPresent() && getComment().get() instanceof JavadocComment) {
           return (JavadocComment) getComment().get();
 }
```
We are doing a lot of same stuff here in each node. Let's introduce a utility method?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitorAdapter.java
```diff
@@ -1657,11 +1657,11 @@ public R visit(NodeList n, A arg) {
     }
 
     private void visitComment(Node n, A arg) {
-        if (n.getComment() != null) {
-            Comment result = (Comment) n.getComment().accept(this, arg);
+        n.getComment().ifPresent(c -> {
+            Comment result = (Comment) c.accept(this, arg);
```
arturbosch: Can be replaced with:

```
n.getComment()
  .map(comment -> (Comment) comment.accept(thus, arg))
  .ifPresent(result -> n.setComment(result))

```

javaparser-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
```diff
@@ -0,0 +1,269 @@
...
+        SourceRoot sourceRoot = new SourceRoot(root);
+
+        CompilationUnit javaParserMetaModel = sourceRoot.parse(METAMODEL_PACKAGE, "JavaParserMetaModel.java", javaParser).get();
+
+        generateClassMetaModels(javaParserMetaModel, sourceRoot);
+
+        sourceRoot.saveAll();
+    }
+
+    private void generateClassMetaModels(CompilationUnit javaParserMetaModelCu, SourceRoot sourceRoot) throws NoSuchMethodException {
```
arturbosch: This methods seems quite heavy 30+ LOC. Maybe you are doing two different things and it is splitable ? :)

javaparser-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
```diff
@@ -0,0 +1,269 @@
...
+    }
+
+    private boolean isNode(Class<?> c) {
+        return Node.class.isAssignableFrom(c);
+    }
+
+    private void generateFieldMetaModels(Class<?> c, ClassOrInterfaceDeclaration classMetaModelClass, String classMetaModelFieldName, NodeList<Statement> initializeFieldMetaModelsStatements) throws NoSuchMethodException {
+        List<Field> fields = new ArrayList<>(Arrays.asList(c.getDeclaredFields()));
+        fields.sort(Comparator.comparing(Field::getName));
+        for (Field field : fields) {
```
arturbosch: same here 30 LOC maybe you would like fields.forEach(field -> newMethod(field)) but I'm not quite sure how expensive it would be to extract parts of this methods...

javaparser-metamodel/src/test/java/com/github/javaparser/metamodel/PropertyMetaModelTest.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Optional;
+
+import static org.junit.Assert.assertEquals;
+
+public class PropertyMetaModelTest {
+    @Test
+    public void whenPropertyIsVerySimpleThenTypeInfoIsCorrect() {
+        PropertyMetaModel bert = new PropertyMetaModel(null, "bert", int.class, Optional.empty(), false, false, false, false);
+        assertEquals("int", bert.getTypeName());
+        assertEquals("int", bert.getTypeNameGenericsed());
```
arturbosch: assertEquals("int", bert.getTypeNameGenericsed()); what does sed stand for? GenericSeed or GenericSet?

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/HashCodeVisitorGenerator.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Generates JavaParser's HashCodeVisitor.
+ */
+public class HashCodeVisitorGenerator extends VisitorGenerator {
+    public HashCodeVisitorGenerator(JavaParser javaParser, SourceRoot sourceRoot, JavaParserMetaModel javaParserMetaModel) {
+        super(javaParser, sourceRoot, "com.github.javaparser.ast.visitor", "HashCodeVisitor", "Integer", "Void", true, javaParserMetaModel);
+    }
+
+    @Override
+    protected void generateVisitMethodBody(BaseNodeMetaModel node, MethodDeclaration visitMethod, List<PropertyMetaModel> propertyMetaModels, CompilationUnit compilationUnit) {
```
arturbosch, line 26: This long method would also benefit from extracting some functionality. Each if-else seems to be a candidate for extracting to a new method with speaking name :)

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/VisitorGenerator.java
```diff
@@ -0,0 +1,85 @@
...
+        this.sourceRoot = sourceRoot;
+        this.pkg = pkg;
+        this.visitorClassName = visitorClassName;
+        this.returnType = returnType;
+        this.argumentType = argumentType;
+        this.createMissingVisitMethods = createMissingVisitMethods;
+        this.javaParserMetaModel = javaParserMetaModel;
+    }
+
+    public void generate() throws IOException {
```
arturbosch, line 43: Long method and NestedBlockDepth is four :)

arturbosch (commented):
Overall quite impressiv PR and a lot of work. Good job! :) I mainly focused on code quality. I hope some comments help!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: hasComment() actually does a check to see if the comment exists, so no worries here :)

matozoid: @arturbosch if you have experience with metamodels in general, you could take a look at what I wrote above to see if it makes sense.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
matozoid: Nice!

javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
matozoid: I've just managed to merge all of that code into a single default method. I'm stunned that some part of Java makes things this pretty :-O

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/GenericVisitorAdapter.java
matozoid: This even: :)
```java
        n.getComment()
                .map(comment -> (Comment) comment.accept(this, arg))
                .ifPresent(n::setComment);
```

javaparser-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
matozoid: For both of these: I'm quite new to writing generator code with JP, and right now I'm convinced there is no way to make it really pretty. Maybe @DeepSnowNeeL could add some thoughts? He seems busy though...

javaparser-metamodel/src/test/java/com/github/javaparser/metamodel/PropertyMetaModelTest.java
matozoid: I just realised the right term is "generified" :)

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/VisitorGenerator.java
matozoid: Are you just quoting Sonar now? :)

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/HashCodeVisitorGenerator.java
matozoid: Disagree, since I don't think exploding a simple algorithm into pieces makes it clearer. It currently shows the decision tree and the different patterns quite clearly, I think.

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/VisitorGenerator.java
arturbosch: No way, I don't like sonar that much anyway ... :), my focus in my masterthesis (master overall) is code quality and bad smells ;)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
arturbosch: I can give it a try :)

javaparser-visitor-generator/src/main/java/com/github/javaparser/generator/visitor/VisitorGenerator.java
matozoid: This one was refactorable, but I'm not sure it got prettier in the progress...

matozoid: You ARE Sonar :-O

matozoid:
Fixes #719 too

matozoid:
Fixes #720 too

matozoid:
That was quite enough time for feedback :)

matozoid:
Closes #718

Issue #644, opened 2016-12-16
Add visitor property to PrettyPrinter


xiewenya:
How about make visitor configurable in class PrettyPrinter so that we can easily customize the printer. Or maybe add a abstract printer class

matozoid:
Hmmm, I'm in an architectural discussion with myself about this now :)

Is your request...
1. I would like to use print my AST in a slightly different way by adapting the current PPVisitor?
2. I would like to print the AST in a drastically different way by writing a different visitor?
3. I would like to change the behaviour of toString() ?

Because in my mind, the PrettyPrinter is a package we offer that will print your AST, and its behaviour is changed by changing its configuration, not by passing a completely different visitor.

But, since we are going to work on generating more visitors, we also have to solve the problem of "which visitor do we use for toString, hashcode, etc" - so we're probably going to make that configurable as well.

xiewenya:
nice to learn how you think :P
now I understand why the visitor is final in the first place.

I would like to print the AST into h5 format, that's why a configurable visitor would make it much easier.

matozoid:
Okay, my advice would be to subclass a ...VisitorAdapter and write your own thing, then use the visitor pattern: call accept on a node with that visitor.  That's what the PrettyPrinter does too.

But you do have a point that if someone wants to modify the PP visitor even a little bit, the PrettyPrinter class becomes useless. Okay, we'll pick it up :)

PR #659, opened 2016-12-19
#644 add visitor property to both PP and PPC

xiewenya:


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinter.java
```diff
@@ -28,18 +28,24 @@
  */
 public class PrettyPrinter {
     private final PrettyPrinterConfiguration configuration;
+    private PrettyPrintVisitor visitor;
```
matozoid: If the visitor is in the configuration, then we don't really need to add it here

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinter.java
```diff
@@ -28,18 +28,24 @@
...
         this(new PrettyPrinterConfiguration());
     }
 
     public PrettyPrinter(PrettyPrinterConfiguration configuration) {
         this.configuration = configuration;
+        this.visitor = configuration.getVisitor();
     }
 
     public String print(Node node) {
         final PrettyPrintVisitor visitor = new PrettyPrintVisitor(configuration);
```
matozoid: We can move this initialization of visitor the the field in PPConfiguration. Then we just call get on the configuration here.

matozoid: Oh I see, the visitor needs the configuration :(

matozoid: Damn this mutability!

matozoid: So I guess this can be solved by setting the configuration on the visitor when it is initialized or set in the configuration. So that will need a setter on the visitor.

xiewenya:
Sorry, I accidentally upload an unfinished version earlier.
The class of the visitor is added as a property of PPC, and the PP creates new visitor instance using reflection every time print() is called. 
It seems a little overkilled, any idea to simplify it?

matozoid:
Yes, just put a setter for the configuration on the visitor.

No need for workarounds with reflection since this is our code :)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -24,6 +24,7 @@
 public class PrettyPrinterConfiguration {
     private boolean printComments = true;
     private String indent = "    ";
+    private Class<? extends PrettyPrintVisitor> visitorClass;
```
matozoid: set this to new PrettyPrinterVisitor() to get a default

matozoid: No wait, you made it a class. Don't make it a class. We don't need a new instance for every print.

xiewenya: the previous version also create new instance for each call, did I miss something?

```
 public String print(Node node) {
        final PrettyPrintVisitor visitor = new PrettyPrintVisitor(configuration);
        node.accept(visitor, null);
        return visitor.getSource();
    }
```

And if the visitor is shared among different print call, test cases fails.
For example, if I add a visitor to PP class, and pass it to node.accept()
```
public class PrettyPrinter {
    private final PrettyPrinterConfiguration configuration;
    private PrettyPrintVisitor visitor;

    public PrettyPrinter() {
        this(new PrettyPrinterConfiguration());
    }

    public PrettyPrinter(PrettyPrinterConfiguration configuration) {
        this.configuration = configuration;
        visitor = new PrettyPrintVisitor(configuration);
    }

    public String print(Node node) {
        node.accept(visitor, null);
        return visitor.getSource();
    }
}
```

For test case `retainAll()`, it reports

    java.lang.AssertionError: 
    Expected :['int a;' REMOVAL in list at 0, 'int c;' REMOVAL in list at 1, 'int e;' REMOVAL in list at 2]
    Actual   :['intint aint;' REMOVAL in list at 0, 'intint aint;intint cint;' REMOVAL in list at 1, 'intint aint;intint cint;intint eint;' REMOVAL in list at 2]

That‘s why I used reflection. So I'm a little confused here, maybe we can discuss it on gitter?

matozoid: Hmmm, I see. Sure, let's go to gitter.

matozoid: After some thinking: maybe an overridable factory method on the configuration would be best, like "createVisitor(config)"

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -51,8 +51,7 @@
  */
 public class PrettyPrintVisitor implements VoidVisitor<Void> {
     private final PrettyPrinterConfiguration configuration;
```
matozoid: I guess you can make this protected too

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -42,4 +45,20 @@ public PrettyPrinterConfiguration setPrintComments(boolean printComments) {
...
     }
+
+    public PrettyPrinterConfiguration setVisitorClass(Class<? extends PrettyPrintVisitor> visitorClass) {
+        this.visitorClass = visitorClass;
+        return this;
+    }
+
+    public PrettyPrintVisitor createVisitor() {
+        try {
+            Constructor<? extends PrettyPrintVisitor> constructor =
```
matozoid: Please no reflection like this - there is nothing here that requires us to subvert type safety. Why not configure things with a Function<PPC, PPV> ?

xiewenya: Did you mean using generics as follows or moving createVisitor to an extra factory class? Could you be more specific?
```
public class PrettyPrinterConfiguration<T extends PrettyPrintVisitor> {
    private Class<T> visitorClass = (Class<T>) PrettyPrintVisitor.class;

    public T createVisitor() {
        try {
            Constructor<T> constructor =
                    visitorClass.getConstructor(PrettyPrinterConfiguration.class);
            return constructor.newInstance(this);
        } catch (Exception e) {
            assert false;
            return null;
        }
    }
}
```

matozoid:
@xiewenya - are you happy with this?

xiewenya:
made SourcePrinter Class public and protected in PPV so that the sourcePrinter can also be extended. For example I need to add \<span\> to line start or use  \<br\> for line separator

matozoid:
@xiewenya - good work, is it still a draft or should we merge it?

xiewenya:
@matozoid sure, it's enough for this issue

matozoid:
Thanks you very much for your submission :)

xiewenya:
Thanks for your help too!

Issue #347, opened 2016-07-09
Write some JavaDoc for each Node type


matozoid:
Just a simple one-liner that says what syntax maps to this node type and a link to the [JLS](https://docs.oracle.com/javase/specs/jls/se8/html/index.html)

gs0510:
Hi! I'd be able to do this. Can you please assign me to this issue. Thanks!

matozoid:
@gs0510 that would be great! You've got an invite.

gs0510:
@matozoid  I can see the javadoc for Node types [here](http://static.javadoc.io/com.github.javaparser/javaparser-core/3.0.0-alpha.2/com/github/javaparser/ast/nodeTypes/package-summary.html). How do I get edit access to this doc. Also, I see a class JavaDocComment and NodeWithJavaDoc, am I supposed to use this class for setting the comment for the JavaDoc? Thanks!

matozoid:
@gs0510 - the task is to take all sub-classes of "Node" and document them. This is done by writing javadoc, which is Java's way of documenting source code. So we want to describe what syntax each node represents and where that is described in the [Java Language Specification](https://docs.oracle.com/javase/specs/jls/se8/html/index.html). This requires detailed knowledge of the Java programming language. Since "how do I write JavaDoc" is basic knowledge, I have to ask if you are aware of the size of this task?

matozoid:
Hmmm, you seem to have some compiler writing experience. Have a go at it then:
https://docs.oracle.com/javase/specs/jls/se8/html/index.html
http://www.oracle.com/technetwork/articles/java/index-137868.html

matozoid:
@gs0510 - how is it going?

matozoid:
Assuming @gs0510 is not doing the job.

PR #665, opened 2016-12-19
Javadoc for all nodes

matozoid:
Since we're all opening half-finished PR's, I'm going to open this one too.

--> issue #347


matozoid:
@SmiddyPence @ftomassetti - what do you think of the level of documentation?
I'm trying to...
- document in practical terms what a specific node represents
- give some examples and point out what parts correspond to the node and its fields
- document on class level only to prevent endless repetition of nearly identical text
- not link to the JLS, because keeping that up to date is horror. It doesn't add that much for the casual user anyway

Issue #670, opened 2016-12-21
Reintroduce ImportDeclaration level functionality


matozoid:
Namely, a factory method for creating imports, and having ImportDeclaration visitable.

Issue #660, opened 2016-12-19
ImportDeclaration to have getImportAsString method to get full import


nishant-labs:
`ImportDeclaration` class to have a `getImportAsString` method which will return fully qualified name of import irrespective of implementation of `ImportDeclaration`.

`public abstract String getImportAsString(){ }`

   E.g. 
    For code `import java.util.List;`
    output: `java.util.List`
    For code `import static java.lang.System.out;`
    output: `java.lang.System.out`
    For code `import java.util.*;`
    output: `java.util.*`
    For code `import static java.lang.System.*;`
    output: `java.lang.System.*`

matozoid:
Okay, one question: why would you leave "static" out, and ".*" in?

nishant-labs:
usually static is not considered as the part of import, its an access modifier, Additionally it does not provide much information about the class(except static members can be used without qualification).

The idea behind adding `".*"` was to make `getImportAsString` method more meaningful. Once `".*"` is removed from import declaration, it is no longer a valid import(in terms of class/package information)

ftomassetti:
This method could be useful, we could add it but we have just to be sure the name is very clear.
What about `getImportedElement`?

matozoid:
@ftomassetti what would you mean with "element" in this case? And certainly it would be `getImportedElementAsString` then? :)

edit: my point is: this part of the declaration doesn't really have a name. It's a type or a package or type and a method.

matozoid:
@nishantsinha-5 - maybe I'm nitpicking now, but * can officialy not be part of a qualified name since it is made up of identifiers, and * is not an identifier. (See https://docs.oracle.com/javase/specs/jls/se8/html/jls-6.html )

Alright, maybe the most important question is: what do you want to use this for? Do you simply want to print it out? Because if you want to interpret it, using the info in the subclasses is much easier.

nishant-labs:
@ftomassetti is there any specific reason/logic behind using `Element`

@matozoid 
You are right,  * is not a part of qualified name, what are you suggesting !!
I am not sure but then we can think of below implementation.
for code `java.util.*`
Shall we use use `java.util` as the value of `getImportAsString` and use `abstract boolean isAsterisk();` method to get asterisk info.

> ~~"Alright, maybe the most important question is: what do you want to use this for? Do you simply want to print it out? Because if you want to interpret it, using the info in the subclasses is much easier.

Let me give information about what I am doing.

[...]

PR #673, opened 2016-12-23
Revert back to a single node per import.

matozoid:
Deals with #660 and #670


matozoid:
@xiewenya and @nishantsinha-5 - please check

xiewenya:
That's much better than adpoting one of the faulty cases. The user can now decide how they will use the name field.

PR #679, opened 2016-12-30
Fix Issue 677: range of AnnotationMemberDeclaration

### LINKED ISSUE #677: The range of the AnnotationMemberDeclaration is wrong
It should be corrected like this:
```
AnnotationMemberDeclaration AnnotationTypeMemberDeclaration(ModifierHolder modifier):
{
	Type type;
	SimpleName name;
	Expression defaultVal = null;
}
{
  type = Type() name = SimpleName() "(" ")" [ defaultVal = DefaultValue() ] ";"

  {
	Position begin = modifier.begin.orIfInvalid(type.getBegin().get());
    return new AnnotationMemberDeclaration(range(begin, tokenEnd()), modifier.modifiers, modifier.annotations, type, name, defaultVal);
  }
}
```

ftomassetti:
Fix #677 and a test


matozoid:
+1

PR #689, opened 2017-01-02
Assert not nullness of comments

### LINKED ISSUE #680: Comment and JavadocComment accept a null content
I think they should complain if built with a null string

matozoid:
Fixes #680

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
@@ -27,6 +27,8 @@ import com.github.javaparser.ast.observer.ObservableProperty;
 
 import java.util.Optional;
 
+import static com.github.javaparser.utils.Utils.assertNotNull;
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
```diff
@@ -61,7 +63,7 @@ public final String getContent() {
      */
     public Comment setContent(String content) {
         notifyPropertyChange(ObservableProperty.CONTENT, this.content, content);
-        this.content = content;
+        this.content = assertNotNull(content);
```
ftomassetti, line 66: maybe this assert not null should be used before notifying the property change, otherwise we risk notifying about a change that will not happen

matozoid: I agree, but that makes the location of just about every assertNotNull wrong :-O New issue?

PR #690, opened 2017-01-02
issue682: override Comment.remove

### LINKED ISSUE #682: Node.remove seems not to work with comments
Probably because the comment could be a child of the parent of the commented node.

Note that it does not explode, it just returns false because it cannot remove the comment. So it is not 100% a bug, more a missing feature.

ftomassetti:
Fix #682


matozoid:
+1

ftomassetti:
Taking advantage of the +1 and merging

PR #693, opened 2017-01-02
issue686: add AnnotationMemberDeclaration.removeDefaultValue

### LINKED ISSUE #686: Consider Adding AnnotationMemberDeclaration.removeDefaultValue
Similar to #681

ftomassetti:
fix #686

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
@@ -134,6 +134,14 @@ public final class AnnotationMemberDeclaration extends BodyDeclaration<Annotatio
         return type;
     }
 
+    public Expression removeDefaultValue() {
+        Expression res = defaultValue;
+        if (res != null) {
+            setDefaultValue(null);
+        }
+        return res;
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


ftomassetti:
rebased

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
```diff
@@ -142,6 +150,9 @@ public Type getType() {
      */
     public AnnotationMemberDeclaration setDefaultValue(Expression defaultValue) {
         notifyPropertyChange(ObservableProperty.DEFAULT_VALUE, this.defaultValue, defaultValue);
+        if (this.defaultValue != null) {
+            this.defaultValue.setParentNode(null);
+        }
```
matozoid, line 155: I guess this is the code that we have to put in all setters?

matozoid:
+1

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationMemberDeclaration.java
ftomassetti: yes, but maybe we can write an helper method in Node

Issue #695, opened 2017-01-07
getScope of MethodCallExpr should return Optional<Expression>


arturbosch:
Like FieldAccessExpr.getScope() returns an Optional, I think MethodCallExpr should also return one.

Is there a specific reason for this difference? 

If not I can provide a PullRequest. Already tested in the source code and only four Visitors needed to be adjusted.

matozoid:
The reason is that we forgot. Please open a PR, tomorrow we'll release version 3.0.0 so this is just on time :)

PR #697, opened 2017-01-07
Change return type of MethodCallExpr.getScope() to Optional<Expression>

arturbosch:
Scope can be null - #695


matozoid:
Thank you very much :)

Issue #325, opened 2016-06-29
Can it parse JavaDoc? 


ipolevoy:
I wonder if it is possible to get a representation of JavaDoc from classes and methods, i.e.:

``` java

/**
* Foo does foo. 
*
*  <p> hello world</p>
*
*/
public  class Foo extends Bar{

   /**
   *  This method blah, blah
   *  @param foo - this is a foo
   *  @return initialized Bar
   */
   public Bar getBar(Foo foo) { ... }
}

```

Specifically, is there a way to parse the Java sources and get JavaDoc in object form, including method names, parameters and return values. 

The goal of this is to write a generator of documentation for web services.

ftomassetti:
Currently JavaParser:
- recognize comments
- associate comments to elements
- distinguish between normal comments and JavaDoc comments

However it does not parse the content of JavaDoc comments. You could use JavaParser to get the AST + the JavaDoc comments associated to the different elements and then parse yourself the content.
If you would also send a PR with that contribution it would be absolutely great.

rpau:
Hi Igor,

This is a feature I implemented in a very old fork of javaparser (https://github.com/rpau/javalang) for the walkmod project. It is a project with LGPL license, and probably your transformation can be shared as a walkmod plugin. However, our version you do not know the comments (excluding Javadoc) associated to an specific element. 

If you are interested in collaborate with Javaparser to include this particular feature, we can talk about :)

ipolevoy:
thanks, guys. I will explore.

ipolevoy:
Just to give you more context. I was thinking of adding web services support for self-documenting, where JavaDoc can be dynamically converted to a simple webapp for developers to see service end points documentation, and potentially directly invoke methods. I'm a main contributor to http://javalite.io. If you have better ideas for implementation than parsing JavaDoc, that would also be great. 
Example  of an API app: https://github.com/javalite/activeweb-rest

ftomassetti:
I think this is a very interesting topic and I would be very interested in hearing more about this project.

For self-documentation I guess there are different strategies possible, but the simplest one is extracting JavaDoc. It should be reasonably easy to do.

ftomassetti:
And of course if @rpau can contribute her code adapting the license it would be great!

ipolevoy:
@ftomassetti do you think you can provide a Hello World to pull the JavaDoc? This would save me some time:)

ftomassetti:
That would not save time to me :)
[...]

PR #699, opened 2017-01-08
Issue433: parsing Javadoc comments

### LINKED ISSUE #433: Parsing JavaDoc
If possible to improve the parsing of a javaDoc comment Block, by adding javaDoc annotation like @return, @param ... .
I think is the last missing things to javaparser.
In the same idea, add some methode to the Generic comment like  isTodoComment() and method like that.
I will try to create a little POC if you think this feature is interesting

ftomassetti:
We parse Javadoc comments to produce a `JavadocDocument`.

A `JavadocDocument` is composed by a `JavadocDescription` and a list of `JavadocBlockTags` (e.g. `@param myParam amazing param`).

A `JavadocDescription` is a list of text blocks and `JavadocInlineTags` (e.g., `{@see foo}`)

We have a list of well-known block and line tags and we try to recognize, however the user can also use arbitrary tags (type `UNKNOWN`).

Fix #433


ftomassetti:
Mmm + 3.9% to coverage?

matozoid:
Nice. I have to take some time to look at this carefully though.

ftomassetti:
Yes, it would benefit from a good revision and probably also some additional tests

javaparser-core/src/main/java/com/github/javaparser/javadoc/JavadocDocument.java
```diff
@@ -0,0 +1,96 @@
...
+    public JavadocDocument addBlockTag(String tagName) {
+        return addBlockTag(tagName, "");
+    }
+
+    /**
+     * Return the text content of the document. It does not containing trailing spaces and asteriks
+     * at the start of the line.
+     */
+    public String toText() {
+        StringBuffer sb = new StringBuffer();
```
arturbosch: A StringBuffer is synchronized, was this by a mistake? (I just scanned the modified files...)

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocDescription.java
```diff
@@ -0,0 +1,107 @@
...
+    public JavadocDescription() {
+        elements = new LinkedList<>();
+    }
+
+    public void addElement(JavadocDescriptionElement element) {
+        this.elements.add(element);
+    }
+
+    public String toText() {
+        StringBuffer sb = new StringBuffer();
```
arturbosch: Replace with StringBuilder here too

javaparser-core/src/main/java/com/github/javaparser/javadoc/JavadocDocument.java
ftomassetti: Right, will fix. Thanks!

matozoid:
Alright, mostly hygiene feedback:
- please give every class a Javadoc that gives the user an idea what its use is. If it represents some kind of text (like JavadocBlockTag) show a little example. Just enough to trigger "oh, that's what you mean!"
- the StringBuffer remark from @arturbosch 
- I think we can create a new Utils method that converts CONSTANT_TEXT to constantText for use in our enums
- `JavadocDocument` is a silly name :) What about simply `Javadoc`? It seems to be a central class for everything Javadoc.
- how about NodeWithJavadoc - maybe juggle the methods there to have `getJavadocRaw()` `getJavadocParsed()` `setJavadoc(String)` and `setJavadoc(JavadocDocument)` ? Not so sure about the naming...

matozoid:
Another one:
* To make JavadocParser "equal" to JavaParser, it could reside in the same package, and offer a static `parse` method. And/or JavaParser can have a `parseJavadoc(String)` method.

ftomassetti:
Implement `setJavadoc(JavadocDocument)` would not be trivial because to convert JavadocDocument to a string we would need to know the proper indentation, so it would leave this out of this PR

ftomassetti:
Thank you to all of you for the comments. I tried to address them and rebase the PR on master. When updating `NodeWithJavadoc` I also tried to normalize `Javadoc` to be written always as `Javadoc` and not `JavaDoc` some times

matozoid:
`setJavadoc(String indentation, Javadoc)` - only half serious this ;)

I am slightly bothered by not having a way to put a Javadoc back in the AST though...

ftomassetti:
I will add `setJavadoc(String indentation, Javadoc)`. Then as a followup we can write a `setJavadoc(Javadoc)` which gets the indentation from the context of the node.
Would that work for you guys?

matozoid:
+1 from me!

ftomassetti:
Good, then I will rebase it on master and wait for the tests to pass before merging

arturbosch:
Looks awesome! +1

ftomassetti:
thanks!

matozoid:
See #325

PR #702, opened 2017-01-09
Let FieldAccessExpr implement NodeWithSimpleName

arturbosch:
I think like MethodCallExpr a FieldAccessExpr should also implement NodeWithSimpleName as this can be handy when trying to resolve method or field chains (eg. obj.a.b.method().stuff()).

I deprecated the old get-/setField methods with a reference to the new methods. All references to getField inside javaparser were replaced by getName.

I will deliver some tests tomorrow for this changes.


matozoid:
Makes sense. Thanks!

Issue #696, opened 2017-01-07
Add NodeWithOptionalScope


matozoid:
There are about four node types that would benefit from this.

Issue #700, opened 2017-01-08
Consider adding more remove methods


ftomassetti:
* `ArrayCreationlevel.removeDimension`
* `CompilationUnit.removePackageDeclaration`

matozoid:
How about we add remove methods for everything that's optional?

ftomassetti:
Yes, also because when there are overloaded setters passing null to remove the child is cumbersome (I have to cast the null)

PR #703, opened 2017-01-10
Introduce NodeWithOptionalScope interface and migrate apt expressions

arturbosch:
Proposal for #696 

Do you have any ideas for default methods for the new interface?

I found out that MethodReferenceExpr had also a setScope method which did not return an Optional, so this PR will break the api. What do you think?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
@@ -23,6 +23,7 @@ package com.github.javaparser.ast.expr;
 
 import com.github.javaparser.Range;
 import com.github.javaparser.ast.NodeList;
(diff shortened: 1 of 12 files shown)


matozoid:
Thanks!

MethodReference always has a scope, so it shouldn't implement this interface. We could have NodeWithScope, but that would be for one node only, so don't bother.

No ideas for default methods here. You?

arturbosch:
Ok, no interface for MethodReference!

Maybe something like removeScope (will set scope to null)?
I don't really know. My use cases are to ask if the scope is a NameExpr or another method/field chain etc but they would be too specific. I use javaparser more to traverse the tree and not (yet) manipulate elements.

matozoid:
Oh yes, remove would be fine, we would have less to do for #700 then :)

Well, "remove" and MethodReference, and that's it then :)

matozoid:
Thanks @arturbosch !

PR #708, opened 2017-01-12
Check that users don't put empty strings in places where they make no sense

### LINKED ISSUE #502: Check that users don't put empty strings in places where they make no sense.
Like, a VariableDeclaratorId should never be "". Any empty string that causes nonsense output should be caught.

arturbosch:
Proposal to fix #502 

Changes are mostly in SimpleName as this now handles all identifiers :)

In StringLiteralExpr.setValue() was a nullCheck. I'm not sure if non-empty check would be right here as some JavadocExtractorTest-Cases fail then?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
@@ -141,6 +141,7 @@ public final class EnumDeclaration extends TypeDeclaration<EnumDeclaration> impl
     }
 
     public EnumConstantDeclaration addEnumConstant(String name) {
+        assertNonEmpty(name);
         EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(name);
         getEntries().add(enumConstant);
         enumConstant.setParentNode(this);
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
@@ -36,6 +36,7 @@ import java.util.EnumSet;
 import java.util.LinkedList;
 import java.util.List;
 
+import static com.github.javaparser.utils.Utils.assertNonEmpty;
 import static com.github.javaparser.utils.Utils.assertNotNull;
 
 /**
@@ -109,12 +110,12 @@ public final class EnumDeclaration extends TypeDeclaration<EnumDeclaration> impl
         return getEntries().get(i);
     }
 
-    public EnumDeclaration setEntry(int i, EnumConstantDeclaration element){
+    public EnumDeclaration setEntry(int i, EnumConstantDeclaration element) {
         getEntries().set(i, element);
         return this;
     }
 
-    public EnumDeclaration addEntry(EnumConstantDeclaration element){
+    public EnumDeclaration addEntry(EnumConstantDeclaration element) {
         getEntries().add(element);
         return this;
     }
@@ -140,7 +141,7 @@ public final class EnumDeclaration extends TypeDeclaration<EnumDeclaration> impl
     }
 
     public EnumConstantDeclaration addEnumConstant(String name) {
-        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(assertNotNull(name));
+        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(assertNonEmpty(name));
         getEntries().add(enumConstant);
         enumConstant.setParentNode(this);
         return enumConstant;
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
@@ -141,7 +141,7 @@ public final class EnumDeclaration extends TypeDeclaration<EnumDeclaration> impl
     }
 
     public EnumConstantDeclaration addEnumConstant(String name) {
-        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(assertNonEmpty(name));
+        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(name);
         getEntries().add(enumConstant);
         enumConstant.setParentNode(this);
         return enumConstant;
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/SimpleName.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/SimpleName.java
@@ -28,7 +28,7 @@ import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
 
-import static com.github.javaparser.utils.Utils.assertNotNull;
+import static com.github.javaparser.utils.Utils.assertNonEmpty;
 
 /**
  * A name that consists of a single identifier.
@@ -49,6 +49,7 @@ public class SimpleName extends Node implements NodeWithIdentifier<SimpleName> {
 
     public SimpleName(Range range, final String identifier) {
         super(range);
+        assertNonEmpty(identifier);
         setIdentifier(identifier);
     }
 
@@ -70,7 +71,7 @@ public class SimpleName extends Node implements NodeWithIdentifier<SimpleName> {
     @Override
     public SimpleName setIdentifier(final String identifier) {
         notifyPropertyChange(ObservableProperty.IDENTIFIER, this.identifier, identifier);
-        this.identifier = assertNotNull(identifier);
+        this.identifier = assertNonEmpty(identifier);
         return this;
     }
 }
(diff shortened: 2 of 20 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithAnnotations.java
```diff
@@ -66,7 +64,7 @@ default N addAnnotation(AnnotationExpr element) {
      */
     default NormalAnnotationExpr addAnnotation(String name) {
         NormalAnnotationExpr normalAnnotationExpr = new NormalAnnotationExpr(
-                parse(name), new NodeList<>());
+                Name.parse(name), new NodeList<>());
```
matozoid, line 67: Huh? :D

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithName.java
```diff
@@ -40,7 +38,7 @@
 
     @SuppressWarnings("unchecked")
     default N setName(String name) {
-        return setName(parse(name));
+        return setName(Name.parse(name));
```
matozoid, line 44: Do you hate static imports? :)

matozoid:
Uh oh - please leave a comment on the issue that you want to work on it - I would have had the time to mention that I wasn't sure if this had to be implemented like this, or that we could put this in the validation framework instead.

The difference with nulls is that using a null value where it is not supposed to be used is a type error. It should have been a compiler error in Java. But an empty string? Not sure.

matozoid:
A `StringLiteralExpr` is a what you get from `String x = "hello world!";`. That is absolutely allowed to be empty.

javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
```diff
@@ -140,7 +141,7 @@ public EnumDeclaration setImplementedTypes(NodeList<ClassOrInterfaceType> implem
     }
 
     public EnumConstantDeclaration addEnumConstant(String name) {
-        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(assertNotNull(name));
+        EnumConstantDeclaration enumConstant = new EnumConstantDeclaration(name);
```
matozoid, line 145: What's happening here?

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
```diff
@@ -50,6 +50,13 @@
         return o;
     }
 
+    public static String assertNonEmpty(String id) {
+        if (id == null || id.isEmpty()) {
```
matozoid: A generic util method shouldn't be written for id's.

matozoid:
Can you refer to the issue itself? This is all stuff that is mentioned in https://github.com/javaparser/javaparser/blob/master/CONTRIBUTING.md

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithName.java
arturbosch: Well, I found both ways to use the parse method in source code, so I decided to also use my preferred way where you immediately see where the method comes from :D

javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
arturbosch: As most expressions, statements and declarations have a SimpleName and all string aka (name, field, identifier) are passed to a new SimpleName instance many checks were not necessary now.

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
arturbosch: You mean the parameter name? Well i couldn't come up with a proper name ... or you mean something else? I could rename it to string or something

arturbosch:
Your right, I did not pay enough attention to the contribution guide. I'm sorry, will get better at it!

I think everywhere where we use a simple name it is right to assume it should be not empty as a class, method, variable, annotation etc in java should also have a name and led to compile errors. 

With StringLiteralExpr, yeah right, also though about the same after I made the PR...

I will add headers to new files I created and link to the issue in the description.

javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
matozoid: Hmmm, see it as a method precondition: parameter name should not be null. It is definitely not required, but it makes the stack trace clearer: it points to this location instead of the constructor of SimpleName. The user will then have to figure out why that constructor is called at all, and deduce from that nullpointer that they shouldn't have passed in a null String.

I say leave it in, the more asserts the merrier :)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/SimpleName.java
```diff
@@ -69,8 +69,9 @@ public final String getIdentifier() {
 
     @Override
     public SimpleName setIdentifier(final String identifier) {
+        assertNonEmpty(identifier);
         notifyPropertyChange(ObservableProperty.IDENTIFIER, this.identifier, identifier);
-        this.identifier = assertNotNull(identifier);
+        this.identifier = identifier;
```
matozoid, line 74: <3 more asserts <3

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithName.java
matozoid: Cool then :)

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
matozoid: That and the error message - dunno, I guess "String s" and "a string was unexpectedly empty" or "This string may not be empty."

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/NameTest.java
```diff
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
- * Copyright (C) 2011, 2013-2016 The JavaParser Team.
+ * Copyright (C) 2011, 2013-2017 The JavaParser Team.
```
matozoid, line 3: Nice find! We'll have to automate updating the headers when we have the generator stuff.

matozoid:
By now I agree that this is a good thing to assert. We'll see what the effect is, I have no idea if users are temporarily putting empty values in names - I hope not because their code will suddenly start failing :)

Thanks for doing the work, by the way. Once we can get past the last one or two discussion points, this will go into 3.0.1 or 3.1.

javaparser-core/src/main/java/com/github/javaparser/ast/body/EnumDeclaration.java
arturbosch: Yeah I struggled with that decision. Double check or confuse the user... As the JIT will ignore the second check I think your right. I will fix some more places.

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/NameTest.java
arturbosch: That's a task I would like to participate in. I need that tool too :). I think also about a tutorial for this case, so we can 'kill two birds with one stone' as we are saying in germany :P

arturbosch:
Btw that you mentioned 3.0.1, how many more features do we need for that release :P ? (Just asking as I work with the 3.0.1-SNAPSHOT for now (I need NodeWithOptionalScope && NodeWithSimpleName for FieldAccess), but there is no hurry ^^)

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
arturbosch: Ok will change!

matozoid:
Well, releasing is so easy that I got into the habit of doing it weekly :) Here is proof: https://mvnrepository.com/artifact/com.github.javaparser/javaparser-core

arturbosch:
Added more assertions when a string should be used as an identifier

matozoid:
Looks cool! +1

Is this ready to merge?

arturbosch:
I would say yes! Stupid null check tests just for the sake of it, are mostly a waste of time :)

arturbosch:
Uff, this is an example that I'm pretty new to github. The last commit should be a new pull request.
I need to create a new branch to make let github know it should be a new PR? (Edit: jep, new branch... sorry for this one!)

Well ... I have the use case to analyze missing @param tags in javadoc. I realized that the Javadoc class does not provide getters for the tags and description.

ftomassetti:
Yes, one branch = one PR
We have been all learning :)
If you have trouble with git feel free to jump on gitter and ask: we could may not super reactive but we would try to help

matozoid:
@ftomassetti I assume those getters are fine. Merging!

matozoid:
@arturbosch - I've just asked sonatype to publish 3.0.1. That may take somewhere between zero seconds and three days.

PR #709, opened 2017-01-14
Make print & println methods public

ghost:
The printer field of PrettyPrintVisitor is protected，some times we want to extends that class and print some texts manually before invoke super.accept(n, arg).


matozoid:
Okay, I guess we need to see the pretty printer as something that users will regularly want to extend. Thanks!

PR #717, opened 2017-01-20
Parsing Javadoc should not throw UnsupportedOperationException and StringIndexOutOfBoundException

arturbosch:
Changes:
- JavadocInlineTag should not throw UnsupportedOperationException 
- empty lines between block tags should not be treated as block tags and throw StringIndexOutOfBoundException
-  braces of inline tags now belong to JavadocSnippet with same behaviour as JavadochBlockTags 
-  fixed some style violations reported by intellij.


javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
```diff
@@ -87,7 +87,7 @@ public JavadocInlineTag(Type type, String content) {
 
     @Override
     public String toText() {
-        throw new UnsupportedOperationException();
+        return "{@" + type.keyword + this.content;
```
ftomassetti: should we have a closing brace here?

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
```diff
@@ -97,8 +97,7 @@ public boolean equals(Object o) {
 
         JavadocInlineTag that = (JavadocInlineTag) o;
 
-        if (type != that.type) return false;
-        return content.equals(that.content);
+        return type == that.type && content.equals(that.content);
```
ftomassetti: If we allow for the Unknown type we should probably store the actual name of the tag used and use it also in the equals method and the toString method

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
arturbosch: Yeah it would be nice. The content has the closing brace. I will check if I can extract it.

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
arturbosch: The actual name is stored in the keyword attribute of the enum. I guess its alright for the equals and hashcode method. In toString we can change that. Also in JavadocBlockTag toString?

arturbosch:
I will also rename this PR as I found a StringIndexOutOfBoundException when trying to parse empty block tags. Will commit later...

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
ftomassetti: Yes, probably I should have done that. I think it is wrong the closing brace is part of the content

arturbosch: Mh, your second sentence sounds like the brace should stay with the content ?

ftomassetti: No, sorry, I meant that I am afraid I could not do the parsing correctly and maybe the brace end up inside the content, while it should not.

matozoid:
Nice, looks like we have something for a 3.0.2 release this weekend :)

arturbosch:
I've updated the description to note the three changes in parsing javadoc.
Hope that's enough for 3.0.2 ;)

matozoid:
@ftomassetti is this +1 for you?

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
```diff
@@ -87,7 +87,7 @@ public JavadocInlineTag(Type type, String content) {
 
     @Override
     public String toText() {
-        throw new UnsupportedOperationException();
+        return "@" + type.keyword + this.content;
```
ftomassetti: I think we should save the name of the tag and use it instead of keyword. Consider what happens if I invoke `fromText("{@foo something}")`

```
public static JavadocDescriptionElement fromText(String text) {
        if (!text.startsWith("{@")) {
            throw new IllegalArgumentException(String.format("Expected to start with '{@'. Text '%s'", text));
        }
        if (!text.endsWith("}")) {
            throw new IllegalArgumentException(String.format("Expected to end with '}'. Text '%s'", text));
        }
        text = text.substring(2, text.length() - 1);
        String tagName = nextWord(text);
        Type type = Type.fromName(tagName);
        String content = text.substring(tagName.length());
        return new JavadocInlineTag(type, content);
    }
```

The tagName will be `foo` and the type will be `UNKNOWN`. We should save the tagName and store it to use it in toText, otherwise:

`JavadocInlineTag.fromText("{@foo something}").toText()` would produce `{@unknown something}` instead of `{@foo something}`.

It would be very nice if you could add a test for this case.

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
ftomassetti: Yes, good point, in `JavadocBlockTag` we should use the tagName in the equals method, while we already use it in the `toText` method

ftomassetti:
I am sorry for being unresponsive: I added two more comments.
You are fixing the mess I did, so I should be more grateful and reply faster :D

arturbosch:
Hey thx for the feedback! I will have the time tomorrow to hopefully finish this one :)

arturbosch:
I had to do `assertEquals(tag, "@foo something");` in the test case without the braces because of the JavadocDescription.parseText() functionality. It seems kinda wrong, but changing the description parseText method seems not the scope of this PR and well it works.

ftomassetti:
I would say we should merge this one, after we have rebased it on top of master (feel free to ask for help if rebasing is new/scaring)

arturbosch:
Rebased, now lets wait for CI :)

matozoid:
I see +1 from everybody?

ftomassetti:
Yes, go for it and thank you guys for your patience on this one!

matozoid:
Thank you both for making something good :)

PR #734, opened 2017-01-28
Issue 666 generate nodelists

matozoid:
@ftomassetti - I would like you to check:
- if you could use the metamodel directly, so we can remove getNodeLists from the AST
- if you are taking typeArguments into account correctly. I had to add a null check in `registerForSubtree`
- the changes in nodelists - there seem to be a lot more now.

I'd like to have this in the next release because it renames one of the new modules. Releasing now would push one single version of that module, never to be updated again.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
@@ -18,7 +18,6 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  */
-
 package com.github.javaparser.ast;
 
 import com.github.javaparser.JavaParser;
@@ -34,13 +33,11 @@ import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
 import com.github.javaparser.utils.ClassUtils;
-
 import java.util.Arrays;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
-
 import static com.github.javaparser.utils.Utils.assertNotNull;
 
 /**
@@ -78,8 +75,7 @@ public final class CompilationUnit extends Node {
         this(null, packageDeclaration, imports, types);
     }
 
-    public CompilationUnit(Range range, PackageDeclaration packageDeclaration, NodeList<ImportDeclaration> imports,
-                           NodeList<TypeDeclaration<?>> types) {
+    public CompilationUnit(Range range, PackageDeclaration packageDeclaration, NodeList<ImportDeclaration> imports, NodeList<TypeDeclaration<?>> types) {
         super(range);
         setPackageDeclaration(packageDeclaration);
         setImports(imports);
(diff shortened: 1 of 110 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -46,51 +44,9 @@
...
- * Any change in the tree is sent as an event to any observers watching.
- * <h2>Visitors</h2>
- * <p>The most comfortable way of working with an abstract syntax tree is using visitors.
- * You can use one of the visitors in the visitor package, or extend one of them.
- * A visitor can be "run" by calling accept on a node:
- * <pre>node.accept(visitor, argument);</pre>
- * where argument is an object of your choice (often simply null.)
- *
- * @author Julio Vilmar Gesser
- */
```
matozoid, line 91: Hmm, interesting side-effect of our commentsinserter here

matozoid:
Oh, @ftomassetti , also if we should change the return type to `List<Optional<NodeList<?>>>` so that we can handle typeArguments without introducing nulls there.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -332,10 +324,7 @@ public ClassOrInterfaceDeclaration addInterface(String name) {
      * @return the newly created class
      */
     public ClassOrInterfaceDeclaration addInterface(String name, Modifier... modifiers) {
-        ClassOrInterfaceDeclaration classOrInterfaceDeclaration = new ClassOrInterfaceDeclaration(
-                Arrays.stream(modifiers)
-                        .collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))),
-                true, name);
+        ClassOrInterfaceDeclaration classOrInterfaceDeclaration = new ClassOrInterfaceDeclaration(Arrays.stream(modifiers).collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))), true, name);
```
ftomassetti, line 327: too long? Is this line generated?

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -305,10 +300,7 @@ public ClassOrInterfaceDeclaration addClass(String name) {
      * @return the newly created class
      */
     public ClassOrInterfaceDeclaration addClass(String name, Modifier... modifiers) {
-        ClassOrInterfaceDeclaration classOrInterfaceDeclaration = new ClassOrInterfaceDeclaration(
-                Arrays.stream(modifiers)
-                        .collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))),
-                false, name);
+        ClassOrInterfaceDeclaration classOrInterfaceDeclaration = new ClassOrInterfaceDeclaration(Arrays.stream(modifiers).collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))), false, name);
```
ftomassetti, line 303: too long? Is this line generated?

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -397,10 +384,7 @@ public AnnotationDeclaration addAnnotationDeclaration(String name, Modifier... m
      * @param className the class name (case-sensitive)
      */
     public Optional<ClassOrInterfaceDeclaration> getClassByName(String className) {
-        return getTypes().stream().filter(type -> type.getNameAsString().equals(className)
-                && type instanceof ClassOrInterfaceDeclaration && !((ClassOrInterfaceDeclaration) type).isInterface())
-                .findFirst()
-                .map(t -> (ClassOrInterfaceDeclaration) t);
+        return getTypes().stream().filter( type -> type.getNameAsString().equals(className) && type instanceof ClassOrInterfaceDeclaration && !((ClassOrInterfaceDeclaration) type).isInterface()).findFirst().map( t -> (ClassOrInterfaceDeclaration) t);
```
ftomassetti, line 387: too long? Is this line generated?

matozoid:
@ftomassetti about generated: yes, all nodes had their formatting ruined because they get PrettyPrinted now.

Issue #538, opened 2016-11-07
Feature to print out AST tree structure


matozoid:
(taken from #537)

`Node.toStrucString(): String`
to print out the tree structure of the AST's subtree rooted at a given node
- this method is useful to learn the tree structure of a node to see what it has and determine how to best process it

DeepSnowNeeL:
Can be done with a visitor, but this would mean another visitor to maintain,

here is an example :

Usage : 

``` java
parse.accept(new TreeStructVisitor() {

            @Override
            public void out(Node n, int indentLevel) {
                System.out.println(StringUtils.repeat('\t', indentLevel) + n.getClass().getSimpleName());
            }
        }, 0);
```

Visitor :

``` java
public abstract class TreeStructVisitor implements VoidVisitor<Integer> {

        public abstract void out(Node n, int indentLevel);

        @Override
        public void visit(final AnnotationDeclaration n, final Integer arg) {
            out(n, arg);
            visitComment(n.getComment(), arg + 1);
            visitAnnotations(n, arg + 1);
            n.getNameExpr().accept(this, arg + 1);
            if (n.getMembers() != null) {
                for (final BodyDeclaration<?> member : n.getMembers()) {
                    member.accept(this, arg + 1);
                }
            }
        }

        @Override
        public void visit(final AnnotationMemberDeclaration n, final Integer arg) {
            out(n, arg);
            visitComment(n.getComment(), arg + 1);
            visitAnnotations(n, arg + 1);
            n.getType().accept(this, arg + 1);
            n.getDefaultValue().ifPresent(d -> d.accept(this, arg + 1));
        }

        @Override
        public void visit(final ArrayAccessExpr n, final Integer arg) {
            out(n, arg);
            visitComment(n.getComment(), arg + 1);
            n.getName().accept(this, arg + 1);
            n.getIndex().accept(this, arg + 1);
        }

        @Override
        public void visit(final ArrayCreationExpr n, final Integer arg) {
            out(n, arg);
            visitComment(n.getComment(), arg + 1);
            n.getElementType().accept(this, arg + 1);
            for (ArrayCreationLevel level : n.getLevels()) {
                level.accept(this, arg + 1);
            }
            n.getInitializer().ifPresent(i -> i.accept(this, arg + 1));
        }

        @Override
        public void visit(final ArrayInitializerExpr n, final Integer arg) {
            out(n, arg);
            visitComment(n.getComment(), arg + 1);
            if (n.getValues() != null) {
[...]

PR #735, opened 2017-01-28
Add a tree "visualizer"

matozoid:
A visitor that prints a simple hierarchical tree of the AST.

About: #538


sourabhsparkala:
Hello,

Is this feature implemented. What I need is a way to print the Compilation Unit in the Tree Structure format, or an XML format.

Regards
Sourabh

Issue #738, opened 2017-01-30
MethodDeclaration and ConstructorDeclaration share many fields -> new SuperType?


arturbosch:
This two behave quite similar in java and also share a lot of fields in javaparser. I often want treat both the same in my smell detection code but then need to work with a BodyDeclaration and make casts. NodeWithDeclaration etc do not help that much here. That's quite a pain.

How about we introduce a super type for both of these. This is what both share:
```java
    private EnumSet<Modifier> modifiers;

    private NodeList<TypeParameter> typeParameters;

    private SimpleName name;

    private NodeList<Parameter> parameters;

    private NodeList<ReferenceType> thrownExceptions;

    private BlockStmt body;
```

Spoon (other java AST tool) has a CtExecutable node which do exactly the same as I propose. 
What do you think? I can also work on this issue..

matozoid:
+1. Let's make sure we have a good name for it though, what do you suggest?

arturbosch:
How about CallableDeclaration?

matozoid:
Okay, if that goes in before the weekend I can still regenerate the metamodel for it.

PR #755, opened 2017-02-04
Create a super type for Constructor- and MethodDeclaration - #738

arturbosch:
I hope its not to late @matozoid 

One thing that I don't like in this implementation is, that i've created setXXInternal(..) methods for all setter which return the declaration itself eg.:
```
    @Override
    public ConstructorDeclaration setModifiers(final EnumSet<Modifier> modifiers) {
        setModifiersInternal(modifiers);
        return this;
    }

    protected void setModifiersInternal(final EnumSet<Modifier> modifiers) {
        assertNotNull(modifiers);
        notifyPropertyChange(ObservableProperty.MODIFIERS, this.modifiers, modifiers);
        this.modifiers = modifiers;
    }
``` 
Another solution would be to write
```
    @Override
    public T setModifiers(final EnumSet<Modifier> modifiers) {
        ...
        return (T) this;
    }
```
with unchecked casts ... someone likes unchecked casts? What if one a user would extend CallableDeclaration and do not a subtype as T? I would like to hear your opinion?

Referenced issues #738


matozoid:
We've been using two approaches until now: do the unchecked cast, or, well, take a look at `setAnnotations` in `Type` and `UnionType`

I prefer moving to the last one, because it would help us get rid of the `<?>` on some nodes (like TypeDeclaration)

matozoid:
`setNameInternal`: try `setName`, and calling it with `super.setName` - that's one less naming convention to worry about.

arturbosch:
That would work if i remove all the interfaces from CallableDeclaration and put them on the subtypes. That's also in Type and UnionType. Do you mean this way?

matozoid:
Yes. I can see that it affects quite a lot of setters, but getting rid of useless `<?>` is a good thing. I'll open an issue to do this for other nodes too.

matozoid:
Alternatively, implement `CallableDeclaration` as an interface like the NodeWith... interfaces.

arturbosch:
I moved the interfaces to the subtypes and renamed the internal methods. I think having the attributes in one place is nicer than just an interface.

matozoid:
+1 from me :)

matozoid:
Sorry for breaking your branch :-|

arturbosch:
I hope I fixed it though the GitHub GUI. First time! :D. Are we waiting for more feedback ?

matozoid:
If there is no feedback for a few days, then that's all the feedback we're going to get :)

Issue #598, opened 2016-11-24
Generate remove methods for nodes


matozoid:
Instead of having a fancy introspection remove() method, generate a remove method per node type by using the metamodel that is being considered here: #353

ftomassetti:
This would be a good idea!
Should we assign it to milestone 4? Create a label for metamodel stuff?

matozoid:
My planning goes no further than 3.0 ;) We've got so many ideas now that we could plan 3.1, 3.2, 3.3 instead of a big jump to 4.0. But planning what is going to be in 3.1 is impossible because us volunteers do whatever we like anyway ;)

A label? I'm using one for Java 9 so I know when we can claim Java 9 support. How would it help with metamodeling?

matozoid:
@ftomassetti @DeepSnowNeeL can you write a sample remove() method for one type of node, non-trivial? So that it is clear what to generate?

ftomassetti:
The way remove works is basically call some setter in the parent with a null value. Note that the method to call depends on the parent: the same Node could be potentially used in different roles. This is what would make difficult this method I think

matozoid:
Would it help to call `parent.remove(this)` so that the logic can be moved to the node that knows what children it has?

ftomassetti:
Yes, using `parent.remove(this)` would help

PR #765, opened 2017-02-07
Add a replace method to nodelist & generate remove methods

matozoid:
I missed this a few times: a method to replace a node in a nodelist with another node. Note that it works on `.equals()`, not on `==`


matozoid:
Accidentally pushed #598 into this

matozoid:
@DeepSnowNeeL look! The Node.remove() method has been replaced by a code generator that is shorter than the remove method :)

matozoid:
#700 in here now too. Note that the return values of specific returnXxx() methods were inconsistent: I found void and "the thing we just removed." To be consistent I return "this" now, but "the thing we just removed" makes sense too. Please convince me if you think that's better :)

(Looks like @ftomassetti made that remove method.)

matozoid:
All the metamodel is getting merged quickly because nobody understand what I'm doing anyway :D

Issue #727, opened 2017-01-28
Generate ModifierVisitor


matozoid:
This one is tricky because it has some additional logic that says "if visiting this property returns null, then the current node becomes damaged, so better return null myself too." I think that can be generated by saying "if a property is required, but visiting it returns null, then I should return null too"

PR #766, opened 2017-02-07
Issue 727 generate modifiervisitor

matozoid:
See #727


matozoid:
@arturbosch - I had to struggle with #755 because it didn't follow the usual patterns after all. That's nice, but a lot of code is generated and that doesn't play well with different approaches to the same thing. Anyway, I've learned: let everyone run the metamodel generator and the core generator before opening a PR :)

Anyway, I had to take body out of `CallableDeclaration` because a `BlockStmt` is not the same as an `Optional<BlockStmt>`

PR #769, opened 2017-02-09
Fix getJavadocComment to not  throw ClassCastException 

arturbosch:
if a comment but not a javadoc comment is present. Also removing unnecessary optional wrapping.


matozoid:
+1, and thanks for the cleanup.

Issue #762, opened 2017-02-06
Reintegrate javaparser-generator-utils with javaparser-core


matozoid:
These utilities are generally useful for anybody using JavaParser.

Issue #761, opened 2017-02-06
Integrate javaparser-metamodel in javaparser-core


matozoid:
This would give us an opportunity to use the metamodel for...
- pseudo-reflection by adding getters to the nodes that return information from the metamodel.
- observers: these use an enum (`ObservableProperty`) for properties now, but could use PropertyMetaModel fields from `JavaParserMetaModel` instead

PR #772, opened 2017-02-10
Issue 762 761 reintegrate modules

matozoid:
#762 #761 

Also adds getMetaModel to the nodes to get access to it directly.

Also refactors the tree structure for StringLiteralExpr to avoid having a node that is not abstract, but is extended.


matozoid:
The build has become a rather hilarious mess now.
Step 1: compile core.
Step 2: generate the metamodel INTO CORE.
Step 3: run the core generators USING THE OLD METAMODEL BECAUSE THE NEW ONE IS NOT COMPILED YET

It's probably best to run the steps manually if you change the AST.

PR #775, opened 2017-02-15
Issue705: building a Concrete Syntax Model

ftomassetti:
Let's try to converge now :)

This patch introduces a Concrete Syntax Model which could be used for pretty printing and for lexical preserving printing.

I have not replaced the PrettyPrintVisitor because the CSM does not support yet that configuration and for compatibility reason we could have the two mechanisms, deprecating the PrettyPrinter and PrettyPrintVisitor until the next major version.

To support the CSM a few changes have been necessary, most importantly we needed to introduce derive properties.

Currently we have two large source files pretty printed with the CSM and with the PrettyPrintVisitor and we get exactly the same output. We could introduce more tests to battle test the CSM.


matozoid:
It's HUGE

Let's merge this thing - it doesn't break the existing API does it?

ftomassetti:
Yes, it is huge, because to get the CSM to work I needed to do some changes, in the meantime other changes were happening and I had to adapt. I have built over a couple of months but if I memory assist me it should not have breaking change aside two methods of WildcardType renamed (getExtendedTypes and getSuperTypes into getExtendedType and getSuperType). The reason I renamed it is to distinguish them from the getExtendedTypes and getSuperTypes used in other nodes. They are different because in the case of the WildcardType they return a single node, while in the other cases they return a nodelist.

If you want I can start separating the changes to the generators (to support derived properties). Now that it is rebased after the metamodel generation I think I can keep it up to date for a while if we want to dissect it and merge a few pieces separately.

matozoid:
Can you re-add those renamed methods as @Deprecated and make them simply call them under the new name?

ftomassetti:
Just thought that it would be great to have a tool to tell you if a patch breaks the public API...

ftomassetti:
Methods from WildcardType readded, marking them as deprecated

matozoid:
Would you like to open an issue to create that tool? Sounds like a "fun side project" :)

ftomassetti:
That idea is now #776 :D

matozoid:
I've had a look at the changes as far as they concern the generators. In short: I'm fine with this going to master.

Longer: I think some code, especially `ObservableProperty`, can benefit from the fact that the metamodel is available *at runtime*. A lot of stuff seems to be generated that is already available. `getNodeLists()` is another instance of that.

I think it's cool that you worked with the generators, and I'm curious if you liked or hated the experience.

ftomassetti:
Yes, I think ObservableProperty was created when the metamodel was not there and now it is outdated because we have a full blown metamodel, which is very good in my opinion.

About working with generators I have liked. I have found them different from approaches I am used to, but I think they are practical and the things work. Maybe for version 4 we can sit down and see what we can refactor but for now it is good.

I think the PR is reasonably safe because we are not removing the PrettyPrintVisitor (but we should probably mark it as deprecated). 

If you are ok with this I am happy to have this merged, but if you feel uncomfortable let me know and I will see if I can extract some pieces.

Either way I am looking forward to go back to the lexical preservation stuff and get it done.

* Comments
* Observers
* Metamodel
* Lexical preservation

With all this stuff we should have solid basis to do great work. It has been a long road since you put the code on GitHub :D

matozoid:
The metamodel stuff is only in the betas - so if you think the generators should be different then now is the time :)

I'm very comfortable with not having two developers develop huge chunks of code without merging, please do.

ftomassetti:
Ok, then let's merge it (after solving new conflict) and if something will need to be cleaned after we will do that

PR #779, opened 2017-02-19
Old release notes

matozoid:
I found some historical release notes - let me know if it's fun to add these :)


ftomassetti:
It is nice, it gives a sense of the history of the project

PR #781, opened 2017-02-21
Issue 763 sourceroot improvements

matozoid:
Instead of storing CompilationUnits and accumulating the Problems in a list, this stores the ParseResults, which is a much more solid approach.

This is new enough to not require deprecation, right? :)


ftomassetti:
+1

PR #654, opened 2016-12-18
issue124: initial implementation of Lexical Preservation

ftomassetti:
Initial work on Lexical Preservation.

I started cleaning up my experiment and move to using tokens. My experiment had more code and several tests, I have removed them for now to start from something cleaner.

The next steps will be to write some tests and build from there.

In this way we can discuss and see how work proceed.

PR #783, opened 2017-02-21
Fix range calculated for catch clauses

ftomassetti:
Resolve some range issues as reported by @cal101 

There are tests verifying this but they will be merged as part of #654


matozoid:
+1 undoubtedly correct.

PR #785, opened 2017-02-21
The NodeList now can notify about a replacement inside a list

ftomassetti:
Added a specific notification when an element is replaced in a list (before it was a combination of removal + addition)


javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
```diff
@@ -526,10 +526,10 @@ private static void initializePropertyMetaModels() {
         unaryExprMetaModel.getDeclaredPropertyMetaModels().add(unaryExprMetaModel.expressionPropertyMetaModel);
         unaryExprMetaModel.operatorPropertyMetaModel = new PropertyMetaModel(unaryExprMetaModel, "operator", com.github.javaparser.ast.expr.UnaryExpr.Operator.class, Optional.empty(), false, false, false, false, false);
         unaryExprMetaModel.getDeclaredPropertyMetaModels().add(unaryExprMetaModel.operatorPropertyMetaModel);
-        unaryExprMetaModel.prefixPropertyMetaModel = new PropertyMetaModel(unaryExprMetaModel, "prefix", boolean.class, Optional.empty(), false, true, false, false, false);
-        unaryExprMetaModel.getDerivedPropertyMetaModels().add(unaryExprMetaModel.prefixPropertyMetaModel);
         unaryExprMetaModel.postfixPropertyMetaModel = new PropertyMetaModel(unaryExprMetaModel, "postfix", boolean.class, Optional.empty(), false, true, false, false, false);
         unaryExprMetaModel.getDerivedPropertyMetaModels().add(unaryExprMetaModel.postfixPropertyMetaModel);
+        unaryExprMetaModel.prefixPropertyMetaModel = new PropertyMetaModel(unaryExprMetaModel, "prefix", boolean.class, Optional.empty(), false, true, false, false, false);
```
matozoid, line 531: Maybe I should do some sorting so that these properties don't keep moving around...

matozoid:
+1

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
ftomassetti: This is now #789

PR #790, opened 2017-02-21
Workaround for error in composition of JavaToken for < and <<

ftomassetti:
Tests will be committed as part of #654 because they reuse classes introduced there.

This error was found because of bugs reported by @cal101

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
@@ -37,7 +37,18 @@ public class JavaToken {
     }
 
     public JavaToken(Token token) {
-        this(Range.range(token.beginLine, token.beginColumn, token.endLine, token.endColumn), token.kind, token.image);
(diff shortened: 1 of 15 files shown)


matozoid:
Oh ugh, is this for the horrible unreadable exception code in the grammar?

ftomassetti:
Yes, I think is for that case. My understanding is that when recognizing type parameters if we see a token like ">>>" or ">>" we pretend is a ">" and rewind the character stream. However the value passed to JavaToken is incorrect. So we check the token kind and adapt the text accordingly.

ftomassetti:
rebased to avoid conflicts

matozoid:
Sorry for more comments, but can you put a bit of explanation in the code? I just know someone will come across it in five years and that person will be utterly confused.

ftomassetti:
No problem, I was very surprised when I saw this!

ftomassetti:
Comment added, rebased

matozoid:
That's a lovely comment!

PR #791, opened 2017-02-22
Add new GenericListVisitorAdapter

Gottox:
This Visitor generates a flattened List<R> of all results.

I found this helpful as I need to map the whole tree recursively to another format. The GenericVisitorAdapter stopped at the first visit, that returned a nonNull.


matozoid:
Very nice! +1

ftomassetti:
The build seem fine on Travis but somehow it was not updated here, I think we can go ahead and merge it. Kudos to @Gottox for having added tests

matozoid:
And kudos for using the VisitorGenerator, good to know it is usable :)

PR #792, opened 2017-02-22
Refine CSM

ftomassetti:
Refinements to CSM either suggested by @cal101 or emerged while working on lexical preservation.

Also:
* introduced check to verify all node classes are covered by the CSM (idea from @matozoid)
* reorganize ObservableProperty to make easier to get values using reflection


matozoid:
I see you have a not-yet-conscious need for PropertyMetaModel to be able to get and set the property on an actual node!

ftomassetti:
Well, probably at some point I should drop the ObservableProperty and just use PropertyMetaModel. I would prefer not doing that right now because I would like to merge the lexical preservation, but if you think we should bite the nail and do it I will go for it.

matozoid:
If you're certain that this will get cleaned up afterwards then +1 to this PR.

ftomassetti:
Added #794 to deal with ObservableProperty, rebasing and merging as soon as tests pass

PR #793, opened 2017-02-22
Fix modifyList(Optional) infinite recursion

icmdaf:
`modifyList(Optional)` calls `modifyList(Optional)` recursively, instead of `modifyList(NodeList)` as it should.


matozoid:
Wow! How did that ever work? Well, I guess it didn't :)

PR #796, opened 2017-02-23
Add maven-source-plugin

icmdaf:
Useful for third party users, enables viewing generated source when debugging, regardless of build system etc.

Will cause a `...-sources.jar` file to be generated next to the binary jar.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -166,6 +166,18 @@
                     <target>${java.version}</target>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
+                        <goals>
+                            <goal>jar</goal>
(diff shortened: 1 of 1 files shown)


matozoid:
Thanks! However, if I look at http://central.maven.org/maven2/com/github/javaparser/javaparser-core/3.1.0/ we already upload the sources. What problem are you experiencing?

icmdaf:
When building locally using `mvn package` these source jars weren't being generated; in my case I wanted to build a version that includes my fix from PR #793 without having to wait for a release — Is there another `mvn` command I could use instead to build the source jars as they appear in the repository?

matozoid:
I have no idea - we can merge this in if you want, I assume it won't hurt the release process.

icmdaf:
It would make my life easier because I wouldn't have to keep it in my branch and merge it myself 😃 
But if nobody else needs it (and I don't know how it affects the source jar for the release)…

matozoid:
Okay, if the next release goes well it stays in :)

Issue #554, opened 2016-11-10
Support module-info.java files.


matozoid:
These files contain a new mini-language.

gunnarmorling:
Has work on this began already?

matozoid:
We're waiting for the specification to stabilize - has that happened?

gunnarmorling:
It's not finalized yet (in Early Draft state currently).

But I think module-info.java syntax is fairly stable by now. Smaller
adjustments may be done down the road, but I wouldn't expect any
fundamental changes at this point.

matozoid:
Let's see... This seems to contain the specs: http://openjdk.java.net/projects/jigsaw/spec/

So alright, work can be started here. It would be very nice to have Java 9 support finished before it's out.

gunnarmorling:
Cool. The grammar for module descriptors is here:
http://cr.openjdk.java.net/~mr/jigsaw/spec/lang-vm.html#jigsaw-1.1

matozoid:
Here's a little braindump:
- module info is a little language that has overlap with Java, but adds a lot of keywords. Therefore we cannot make it part of the main grammar - we don't want normal Java files to fail because of reserved words that are intended for module info. So we'll need a second grammar.
- the second grammar will reuse a lot of stuff from the main grammar (imports, type names, annotations...) so we'll have to find a reuse mechanism. JavaCC does not seem to have one.
- for the main JavaParser interface we'll define an additional `parseModuleInfo` method next to the usual `parse` method, with the same amount of variants, and an instance parser method variant. So no funky "what are we parsing here?" logic.
- The grammar for module info starts with `CompilationUnit`, so we can simply add a new `ModuleInfo` field to `CompilationUnit`. I'm a little sad that two different parse rules will be able to produce a `CompilationUnit` now, this might cause confusion.
- `SourceRoot` has the option to parse a directory full of files. We'll add an overridable heuristic to it that checks the file name by default. `module-info.java` will trigger the module info parser. It should be overridable since the JLS is not exact: "A compilation unit (JLS 7.3) may contain a module declaration, in which case the filename of the compilation unit is *typically* module-info.java."

gunnarmorling:
I don't know JavaCC, but instead of having a separate grammar, couldn't you have two alternative definitions for `CompilationUnit`: `TypeDeclaration` | `ModuleDeclaration`? Like it's done for `TypeDeclaration` itself which can be a class/interface | enum | annotation type.

[...]

Issue #594, opened 2016-11-23
Create validation framework


matozoid:
The idea:
```java
Validator v = new Java14Validator();
List<Problem> problems = v.validate(compilationUnit);
```
Problems could be "annotations are not supported in Java 1.4" or "try has no catch or finally" and so on.

It would be nice of a validator was simply a collection of validationRules that are applied to the supplied compilationUnit. A validationRule would probably contain some kind of visitor.

Why? There can be many problems with an AST, and currently there is no consistent way to report those. With this framework, every non-parse error (there is no AST to validate until we're done parsing) can be detected on demand by the user.

This splits JavaParser errors in these categories:
- environmental errors (file not found...)
- parse errors
- semantic errors (found by validators)
- user errors (triggered by asserts: using empty names, using null where not allowed, ...)

ftomassetti:
Should this be a separate project under the JavaParser umbrella?
This could benefit from typechecks I think

matozoid:
What do you mean with type checks? In what way?

Separate project: I feel that reporting problems for a certain Java version is core business. But only the validators for Java versions.

ftomassetti:
Type checks: checks for example if you try to assign an int to a String field

matozoid:
Oh alright, that goes into another module or project.

matozoid:
@ftomassetti - how would you put type checking in here? That seems JSS territory?

ftomassetti:
Yes, it should be based on JSS. We could implement another Validator to.

Some things we could verify:
* verify assignments involve compatible types
* all references are resolved
* a class do not extend itself
* in catch clauses Exceptions are used and not other types

So we are getting closer and closer to get our own compiler...

matozoid:
Okay, shall we open an issue on JSS for this?

PR #800, opened 2017-02-24
Support for Java 9 project Jigsaw module-info.java files

matozoid:
See #554.

This is a first draft. It seems to parse and print files quite well.

It has one big disadvantage: there is now a second grammar, and it contains a copy of the first. The differences are small, so that should be fixable with some trickery.

CSM or symbol solver support is not there yet.

I think the most interesting thing to look at is the unit test: https://github.com/javaparser/javaparser/pull/800/files#diff-32ad75ad8b271ad317da30653e5b9c83


matozoid:
Well, that's embarrassing.

matozoid:
Build fails because the CSM wants me to implement module info. Can @ftomassetti help a bit?

ftomassetti:
Sure: to implement it or make CSM not complain?
In the latter case you may want to go to ConcreteSyntaxModel.java and changes these lines:

```
        List<String> unsupportedNodeClassNames = JavaParserMetaModel.getNodeMetaModels().stream()
                .filter(c -> !c.isAbstract() && !Comment.class.isAssignableFrom(c.getType()) && !concreteSyntaxModelByClass.containsKey(c.getType()))
                .map(nm -> nm.getType().getSimpleName())
```

Maybe you want to create a common ancestor for the ModuleInfo AST classes and put them in a set and make CSM skipping those

matozoid:
No no, to implement them. I can probably do it myself if I get the time, but I'm lazy. The PPV is already implemented.

ftomassetti:
You got your first PR :D 

https://github.com/matozoid/javaparser/pull/1

I just sketch something immediately because in the week-end I will unable to do much coding :(

ftomassetti:
Do you think we will have many conflicts with #654 ?

matozoid:
@ftomassetti - about conflicts: there is one extra parameter to the CU constructor, a bunch of extra nodes, and I've changed the names of the classes that JavaCC generates. So that may cause a lot of conflicts, but they should be simple to fix (mostly name changes.)

matozoid:
@ftomassetti - I've added a test to `ModuleDeclarationTest`, but it complains that `ModuleProvidesStmt` doesn't have a `hasName` method. Name is required, so I don't know why I would ask that. No time to figure it out right now.

ftomassetti:
@matozoid perfect. I do not plan to touch #654 anymore. If there are improvements to do (aside those that will emerge during code review) I will do as separate PRs

matozoid:
This works, but it isn't pretty. The problem is with the 10 additional keywords. They have to be specified in the list of keyword tokens. When you do so, all the following tokens will get their ID increased by 10 (because we just inserted 10 new ones.) Now we have a problem with the Concrete Syntax Model, which uses the token ID's from the normal grammar, and starts outputting the wrong tokens for the module info grammar.

The quick fix is to define 10 nonsense keywords in the normal grammar. They are named "qqqqqqqqqqqqqqqqqqqqqqq...."

I'll try to merge it back into one grammar, and redefine IDENTIFIER to match those ten keywords too... If that doesn't work either I'm out of options.

matozoid:
@ftomassetti what the heck does `java.lang.UnsupportedOperationException: kept token(() vs ChildTextElement{foo}` mean? I'm getting that from one of the unit tests.

ftomassetti:
It means that according to the difference it has calculated it is expecting to find a certain token and to keep it. It finds instead a token of a different type so it does not know what to do. Note that it is fine to find extra whitespace tokens or comments and they can be skip.

matozoid:
@ftomassetti it seems to be an EOL issue again, since travis doesn't complain and it build on my linux machine.

ftomassetti:
You mean the failures in Appveyor? Also, isn't surprising that coverage is down by almost 12%?

matozoid:
Coverage is explained by having a second copy of the grammar, which causes a second set of generated code.

matozoid:
This was rather surprising: removed the second grammar, and made all the module-info keywords match as identifiers too.

```java
String Identifier():
{
    String ret;
}
{
  (<MODULE> | <REQUIRES> | <TO> | <WITH> | <OPEN> | <OPENS> | <USES> | <EXPORTS> | <PROVIDES> | <TRANSITIVE> |
   <IDENTIFIER>) { ret = token.image; }
  { return ret; }
}
```

So this seems to work correctly!

ftomassetti:
Great!

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
```diff
@@ -37,7 +37,8 @@
     VOLATILE,
     SYNCHRONIZED,
     NATIVE,
-    STRICTFP;
+    STRICTFP,
+    TRANSITIVE;
```
gunnarmorling, line 41: I don't think it should be treated as a regular modifier.

javaparser-core/src/main/javacc/java_1_8.jj
```diff
@@ -25,11 +25,11 @@ options {
   JAVA_UNICODE_ESCAPE=true;
   COMMON_TOKEN_ACTION=true;
   JDK_VERSION = "1.8";
```
gunnarmorling, line 27: How comes that's art of java_1_8.jj? I'd have assumed you'd create a new file java_9.jj instead.

gunnarmorling (commented):
Hey @matozoid looks good to me after a quick look. Put two comments inline, mainly I was wondering why there is not a new version of the grammar file for Java 9.

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
matozoid: I put it there because JP matches all modifiers in every location where modifiers are allowed. The idea is to check if the correct modifiers are used by validating the AST, which is not implemented yet: #594 . Or did you mean something else?

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: Well, looking at the changes in Java 9 it seems that it is a superset of 8, so we can keep a single grammar (much less maintenance work!) - but it should be renamed yes :)

matozoid:
@gunnarmorling thanks for the feedback - I was hoping you could check this branch out and try it on what you are thinking of using it for (if anything.)

javaparser-core/src/main/javacc/java_1_8.jj
matozoid: I'll not rename it right away because these comments might disappear...

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
gunnarmorling: Ah, I see.

I expected this to resemble `javax.lang.model.element.Modifier` which doesn't contain `TRANSITIVE`. I.e. the "module-only" modifiers are kept separately there. There are separate enums for the different module instruction modifiers, e.g. [here](http://download.java.net/java/jdk9/docs/api/java/lang/module/ModuleDescriptor.Requires.Modifier.html) for `Requires`.

gunnarmorling:
I checked it out and parsed one test descriptor, all looking good.

I noticed though that `JavaParser#parse(InputStream)` doesn't like module descriptors as input, I had to use `parseModuleInfo()`. Is that expected?

javaparser-core/src/main/javacc/java_1_8.jj
gunnarmorling: Sounds good.

matozoid:
@gunnarmorling currently it is, since it is a separate production, but I'll look into that tomorrow. Thanks for testing!

gunnarmorling:
I see. Thanks a lot for working on this, @matozoid!

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
matozoid: Hmmm, that page says "mandated" is a modifier. I didn't see that in the JLS...

javaparser-core/src/main/javacc/java.jj
```diff
@@ -1318,15 +1329,30 @@ CompilationUnit CompilationUnit():
...
+        modifier = Modifiers()
+        (
+            tn = ClassOrInterfaceDeclaration(modifier) { types = add(types, tn); }
+        |
+            tn = EnumDeclaration(modifier) { types = add(types, tn); }
+        |
+            tn = AnnotationTypeDeclaration(modifier) { types = add(types, tn); }
+        |
+            module = ModuleDeclaration(modifier)
+        |
```
matozoid, line 1350: This is what I did to make CompilationUnit() match all possible content. So the grammar thinks it's okay to mix classes, multiple module declarations, enums and annotation declarations in the same file. Yet again, a validator could check if a module declaration is mixed with type declarations and flag a warning or error.

matozoid:
Okay, the main JavaParser.parse(...) methods should now parse module declarations, and the grammar file no longer contains the java version, which wasn't helping anyone anyway.

matozoid:
I think this will do as a nice preview. If anything changes in the specs we can change it later.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
```diff
@@ -25,23 +25,21 @@ public static void main(String[] args) throws Exception {
...
+        new EqualsVisitorGenerator(sourceRoot).generate();
+        new VoidVisitorAdapterGenerator(sourceRoot).generate();
+        new VoidVisitorGenerator(sourceRoot).generate();
+        new GenericVisitorGenerator(sourceRoot).generate();
+        new HashCodeVisitorGenerator(sourceRoot).generate();
+        new CloneVisitorGenerator(sourceRoot).generate();
+        new TreeStructureVisitorGenerator(sourceRoot).generate();
+        new ModifierVisitorGenerator(sourceRoot).generate();
+
+        new GetNodeListsGenerator(sourceRoot).generate();
```
ftomassetti, line 39: Why they are divided in two blocks? Is this worth a comment to explain what the two groups of generators represent?

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
```diff
@@ -16,8 +16,8 @@
  * Generates JavaParser's CloneVisitor.
  */
 public class CloneVisitorGenerator extends VisitorGenerator {
-    public CloneVisitorGenerator(JavaParser javaParser, SourceRoot sourceRoot) {
-        super(javaParser, sourceRoot, "com.github.javaparser.ast.visitor", "CloneVisitor", "void", "A", true);
+    public CloneVisitorGenerator(SourceRoot sourceRoot) {
+        super(sourceRoot, "com.github.javaparser.ast.visitor", "CloneVisitor", "Visitable", "Object", true);
```
ftomassetti, line 20: Here I miss Kotlin named parameters :)

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -293,10 +293,7 @@ public static Statement parseStatement(final String statement) {
 
     private static <T extends Node> T simplifiedParse(ParseStart<T> context, Provider provider) {
         ParseResult<T> result = new JavaParser(new ParserConfiguration()).parse(context, provider);
-        if (result.isSuccessful()) {
-            return result.getResult().get();
-        }
-        throw new ParseProblemException(result.getProblems());
+        return result.getResult().orElseThrow(() -> new ParseProblemException(result.getProblems()));
```
ftomassetti, line 296: nice

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithAnnotations.java
```diff
@@ -60,27 +60,53 @@ default N addAnnotation(AnnotationExpr element) {
...
-     * Annotates this and automatically add the import
+     * Annotates this node and automatically add the import
      *
      * @param clazz the class of the annotation
-     * @return the {@link NormalAnnotationExpr} added
+     * @return this
      */
-    default NormalAnnotationExpr addAnnotation(Class<? extends Annotation> clazz) {
+    default N addAnnotation(Class<? extends Annotation> clazz) {
         ((Node) this).tryAddImportToParentCompilationUnit(clazz);
```
ftomassetti, line 94: Is this wise? Should the addImport be optional? What if the annotation is in the same package?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitor.java
```diff
@@ -1097,5 +1092,79 @@ public Visitable visit(final LineComment n, final A arg) {
...
+        n.setWithTypes(withTypes);
+        n.setComment(comment);
+        return n;
+    }
+
+    @Override()
+    public Visitable visit(ModuleUsesStmt n, A arg) {
+        Type type = (Type) n.getType().accept(this, arg);
+        Comment comment = n.getComment().map( s -> (Comment) s.accept(this, arg)).orElse(null);
+        if (type == null)
```
ftomassetti, line 1150: nitpick: even in generated code I would really love to see the then statement between braces

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -64,17 +65,17 @@ private static CsmElement memberAnnotations() {
     }
 
     private static CsmElement annotations() {
-        return list(ObservableProperty.ANNOTATIONS, none(), none(), newline());
+        return list(ObservableProperty.ANNOTATIONS, space(), none(), newline());
```
ftomassetti, line 68: I am surprised this change does not cause tests to fail

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -64,17 +65,17 @@ private static CsmElement memberAnnotations() {
...
     }
 
     private static CsmElement typeParameters() {
-        return list(ObservableProperty.TYPE_PARAMETERS, CsmElement.sequence(CsmElement.comma(), CsmElement.space()), CsmElement.token(ASTParserConstants.LT),
-                CsmElement.sequence(CsmElement.token(ASTParserConstants.GT), CsmElement.space()));
+        return list(ObservableProperty.TYPE_PARAMETERS, CsmElement.sequence(CsmElement.comma(), CsmElement.space()), CsmElement.token(GeneratedJavaParserConstants.LT),
```
ftomassetti, line 72: Maybe we could use static imports for the token constants

javaparser-core/src/main/javacc/java.jj
```diff
@@ -2101,7 +2112,19 @@ SimpleName SimpleName():
...
+}
+
+
+String Identifier():
+{
+    String ret;
+}
+{
+    // Make sure the module info keywords don't interfere with normal Java parsing by matching them as normal identifiers.
+  (<MODULE> | <REQUIRES> | <TO> | <WITH> | <OPEN> | <OPENS> | <USES> | <EXPORTS> | <PROVIDES> | <TRANSITIVE> |
```
ftomassetti, line 2126: in this case what would be the token type? If the topken type is not IDENTIFIER we could run into issues with the CSM

ftomassetti (commented):
Looks good, my only real concern is about the token type being IDENTIFIER when parsing identifiers that are keywords in the context of module declarations

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
matozoid: True true, but if you really like those then the latest intellij can help you too :)

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithAnnotations.java
matozoid: This functionality wasn't changed (what I did was correct the use of the "add" and "addAndGet" pattern) but it is a good idea: #803

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitor.java
matozoid: Go submit a PR for your nitpicking :-P

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
matozoid: I was surprised it was working at all :D

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
matozoid: I'm a big fan of static imports, and I thought you were not using them here because there are constants coming in from several sources. PR please :)

javaparser-core/src/main/javacc/java.jj
matozoid: That's what I was worried about too, but the tests kept running. Would you like to suggest a test for this?

Anyway, the fix would be to override the token type to IDENTIFIER right here, I think.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitor.java
matozoid: Actually, I have a better answer. The code in the generator is a little more readable because it doesn't contain the braces :)

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
gunnarmorling: Not sure whether you clicked that option or not when creating this PR, but since recently one can have other contributors push to the same PR. I.e. @ftomassetti could add this kind of changes right to this PR (it's a checkbox when submitting a PR). Found it quite useful myself.

matozoid: I missed it completely, but that's good to know!

javaparser-core/src/main/javacc/java.jj
matozoid: @ftomassetti - do you have an idea how to test this? It is the last thing before I pull this.

ftomassetti: I would:
* parse something that has a name
* change it to a name corresponding to a module keyword
* print it using lexical preservation

Do the viceversa (starting from something having a name like "requires").

These tests would be very similar to the ones in the lexicalpreservation package

matozoid: @ftomassetti I added the following tests and they work - am I doing this right?

```java
    @Test
    public void printAModuleInfoSpecificKeywordUsedAsIdentifier1() {
        considerCode("class module { }");

        cu.getClassByName("module").get().setName("xyz");

        assertEquals("class xyz { }", lpp.print(cu));
    }

    @Test
    public void printAModuleInfoSpecificKeywordUsedAsIdentifier2() {
        considerCode("class xyz { }");

        cu.getClassByName("xyz").get().setName("module");

        assertEquals("class module { }", lpp.print(cu));
    }
```

ftomassetti:
+1 from me: the tests in LexicalPreservingPrinterTest.java are exactly what I meant

matozoid:
Thank you!

gunnarmorling:
Awesome, thanks a lot for making it happen, @matozoid and @ftomassetti! Any idea when we can get it in a release?

matozoid:
Yes, next Sunday. This project releases every Sunday unless something gets really in the way.

matozoid:
I'm not sure what you want to do it with it, but remember that this is only partial Java 9 support :) Here's the rest: https://github.com/javaparser/javaparser/labels/Java%209%20support

Issue #753, opened 2017-02-02
Add other output formats for TreeStructureVisitor


matozoid:
Try...
- [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) 
- XML 
- HTML
- yaml 
- JSON

matozoid:
```xml
<root type='MethodCallExpr'>
    <arguments type='EnclosedExpr'>
        <inner type='EnclosedExpr'>
            <inner type='BinaryExpr' operator='PLUS'>
                <left type='IntegerLiteralExpr' value='1'/>
                <right type='IntegerLiteralExpr' value='1'/>
            </inner>
        </inner>
    </arguments>
    <name type='SimpleName' identifier='println'/>
    <scope type='FieldAccessExpr'>
        <name type='SimpleName' identifier='out'/>
        <scope type='NameExpr'>
            <name type='SimpleName' identifier='System'/>
        </scope>
    </scope>
</root>
```

matozoid:
```
root (BinaryExpr)
	operator: PLUS
	left (IntegerLiteralExpr)
		value: 1
	right (IntegerLiteralExpr)
		value: 1
```

ryan-beckett:
@matozoid: Is there any rhyme or reason to the XML format you've created or just an example. Before I work on this I'd like to agree upon a syntax. I'll post a sample format that covers most of the node types that we can review before I implement it.

ryan-beckett:
@matozoid: Also, would it suffice to add these formatters as XMLStructureVisitor, DOTStructureVisitor, etc?

matozoid:
Okay, the full story: I was working on this before, trying to get a basic framework together that would make it easy to build these things. I generated `TreeStructureVisitor` for that. By now I think I should have used the `TreeVisitor` or a custom simple recursive one. I think the user wants to specify how to output a single node, so it should have a method that receives that node, and you can do anything there. Maybe also a context or a stack for each level. The whole idea (read #754 too) is that people can quickly get an overview of what the node structure looks like. Since it doesn't have the purpose of getting processed, the format can be arbitrary.

Anyway, the current `TreeStructureVisitor` should be trashed.

My plan was to create a new `TreeStructureVisitor` and give it a callback that will create the output. Pass different callbacks, get a different output format. It currently uses extension.

[...]

PR #811, opened 2017-03-02
[DRAFT] issue753: New formatting framework

ryan-beckett:
A few thoughts:

1. `TreeStructureVisitor` still uses the visitor pattern, but the API changed slightly. A user has to invoke `TreeStructureVisitor.getResult()` after calling `Visitable.accept(Visitor)` to retrieve the formatted string instead of the output being dumped to the console. This is also nice since we don't know what the user wants to do with the output (i.e. send it to a browser or file).

2. Since JSON requires the keys of object members to be unique, I had to append indices to each alike key. Shouldn't be a big issue for anyone looking to further process the JSON. E.g. 
`
"ClassOrInterfaceDeclaration" : {
    "MethodDeclaration1": "",
    "MethodDeclaration2": "",
    "MethodDeclaration3": ""
}
`

3. I've included default, XML, and JSON implementations, and the generated output for the latter two checkout in online syntax checkers. YAML, HTML, and DOT are coming up soon. Check out the tests to get a feel of everything.

4. I felt like `com.github.javaparser.utils.format` was an appropriate place to stash the framework.


ftomassetti:
I do not think that appending indexes is a good idea. If we have a name clash we could perhaps using an array:

```
{"methodDeclaration": [ method1, method2, method3 ] }
```

matozoid:
Great! Thanks for putting it up for feedback.

Here it is :)
- about 1 and 2: why I wanted to move away from this visitor was because it took away control over how I wanted to print node lists - so I was considering the output with indices and everything and thought: nah, I need (for XML) an open and close element, and @ftomassetti's brace syntax for JSON, etc. So I would expect `FormatInstructions` to have `enterNodeList` and `exitNodeList` too.

Do you have sample output?

- package: you couldn't know this, but `utils` is the place where everything goes that we want to reuse but don't have a good location for. In an ideal world it wouldn't exist :) How about a subpackage of `printer` ?

- if you keep using this visitor: have you considered switching to `GenericVisitor`? That has a return value, so you could avoid the `getResult()` step. Just an idea, because `PrettyPrintVisitor` does exactly what you do.

Do go on! 👍

ryan-beckett:
@matozoid : Thanks for the feedback. I'll see what I can do. Here's the [JSON](https://gist.github.com/ryan-beckett/62d14f8af5f913fd609e42914ca5fed8)  and [XML](https://gist.github.com/ryan-beckett/b85423496e67be81cc6cc26120171d09) output from the tests.

@ftomassetti: Can further clarify what you mean by using arrays? Can you checkout the output linked above just to make sure we're on the same page, because I'm not sure what you're suggesting. I don't like using indices either, but I can't think of a better way.
`

ftomassetti:
@ryan-beckett when outputting JSON you can use JSON arrays to group together several elements. For example:
`{ "foo" : [1, 2, 3] }`

In this case the property `foo` has as value an array containing 1, 2, 3. I am suggesting you use arrays to group different elements and associate them to a single property. In this way you would not need to use names like `MethodDeclaration1`, `MethodDeclaration2`, `MethodDeclaration3` etc. You would just use `MethodDeclaration` and associate to that name an array instead of a single value.
Does this help?

ryan-beckett:
@ftomassetti: I understand now. Thanks.

@matozoid: Yeah, I'll have to implement some real tests. I don't think a `GenericVisitor` will work because I need to be able to call `Instructions.postProcess()` before the final output is given to the caller, and only once. For example, post-processing is necessary with JSON because the final attribute of any object can't be proceeded by a comma, and there's no way to circumvent this with current visitor API. My solution was to use regex to remove those commas, and it works, so I'm ok with that. I'm sure other formats will require some cleaning up at the end as well. I even considered a `preProcess()` method, but it's not needed for now.

I'll see what I can cook up tomorrow. I work night shift by the way, hence my late night/early morning responses.

If you don't mind me asking, @ftomassetti and @matozoid, are you two professional developers? Or were you at some point?

matozoid:
Using regexes to fix a weakness in your approach? Come on, I can't accept that as the official API in a serious library.

I'm a Java consultant for 16 years now, doing this in my spare time because this is more fun :)

matozoid:
@ryan-beckett - okay, I tried putting the stuff in my mind into code, the result is here: https://github.com/matozoid/javaparser/tree/issue_754_more_structure_output/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/treestructure

It wasn't very easy, so excuse me for putting the easy tag on this, @ryan-beckett .

What's unfinished in my stuff:
- the two pieces of code look very much alike, it probably has a nice abstract base class hiding in there somewhere
- tests have the wrong name
- XML stuff doesn't use the Java XML API - looks like a good fit for SAX.
- there's no pretty printing, but maybe we don't care (every IDE has a "format source" option)
- ... maybe more?

matozoid:
#753 <-- mentioning the issue links it.

ryan-beckett:
@matozoid: Consultant, huh? Cool. Thanks for sharing. Your code is a lot more extensive than what I have, and thus my need to use regex to remove commas. I was trying to stay aligned with using the current `TreeStructureVisitor` class and use ideas from your **serious** library. I have no problem re-designing it. The interface I defined was a little limiting for a user looking to implement a custom format, yes, but it was simplistic and enough to get the job done. I tried implementing JSON as any other user would given that API, and I was able to do it successfully.... round about, but successfully. 

However, I can appreciate your desire for better code, so I'll try something more robust. In fact, I just saw that project is being used in commercial products! Thanks for your feedback.

**Edit:**

On second thought, your design is just out flat better. I'll do it that way.

matozoid:
Cool! Thanks for staying constructive :)

ryan-beckett:
@matozoid: I notice you're using `getValue(Node)` in your example XML dump, but this method doesn't exist.

`  for (PropertyMetaModel attributeMetaModel : attributes) {
            builder.append(attribute(attributeMetaModel.getName(), attributeMetaModel.getValue(node).toString()));
        }`

Shall I just create a utility method to get an attribute value given the attribute model and node?

Also, what is multiplicity?

`List<PropertyMetaModel> subNodes = allPropertyMetaModels.stream().filter(PropertyMetaModel::isNode).filter(p -> p.getMultiplicity() == ONE).collect(toList());`

This too doesn't exist in the code.

Lastly, `PropertyMetaModel.isAttribute` doesn't exist. Would `isSingular` suffice?

`        List<PropertyMetaModel> attributes = allPropertyMetaModels.stream().filter(PropertyMetaModel::isAttribute).filter(p -> p.getMultiplicity() == ONE).collect(toList());`

I really don't see the need to create an abstract base class for different formats to extend. What functionality would really be useful to put in it? The formats are very different and really comes down to writing a custom `output(Node)` method. I just don't see the need in complicating things.

matozoid:
Hi, that stuff only exists on my branch. I'll see if I can turn it into a PR quickly.

ryan-beckett:
Ah. If you don't PR it by tomorrow, I'll incorporate your branch into mine and try implementing the YAML and DOT formats. Yours work already, so no need to them over. Thanks.

matozoid:
Okay, multiplicity ONE is identical to `isSingular()` and multiplicity MANY is identical to `!isSingular()` so I won't be adding that.

matozoid:
Here's the PR: https://github.com/javaparser/javaparser/pull/869

Better just copy the files from my branch if you like them so you don't import the messy stuff too.

ryan-beckett:
@matozoid: Sounds good. Thanks.

matozoid:
Base class: go ahead and do it your way, if the users are happy then we will have saved some effort :)

"Multiplicity" simply means if a field contains one thing, or multiple. So anything that is a NodeList or an EnumSet has multiplicity "many."

ryan-beckett:
@matozoid: I understand now. Thanks. I should be good from here.

matozoid:
Okay, PR is merged, have fun :)

PR #825, opened 2017-03-05
issue822: make addModifier notify about the change

### LINKED ISSUE #822: Add modifier does not notify about the change
In NodeWithModifiers:

```
    default N addModifier(Modifier... modifiers) {
        getModifiers().addAll(Arrays.stream(modifiers)
                .collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))));
        return (N) this;
    }
```

This does not cause a notification. We could refactor it to invoke setModifiers, which does. The problem is we cannot call notifyChange directly from here I think.

ftomassetti:
fix #822

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithModifiers.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithModifiers.java
@@ -45,8 +45,10 @@ public interface NodeWithModifiers<N extends Node> {
 
     @SuppressWarnings("unchecked")
     default N addModifier(Modifier... modifiers) {
-        getModifiers().addAll(Arrays.stream(modifiers)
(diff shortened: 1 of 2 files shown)


matozoid:
Is this because EnumSet is being a pain? Do you think we should change it to a NodeList in the future?

ftomassetti:
Well, the issue is just that we have this method in NodeWithModifiers wich should call notifyPropertyChange. The problem is that you cannot do that from NodeWithModifiers because it is not extending Node (even if the name let us understand a NodeWithModifiers is a Node). So we have to translate the addModifier in a setModifiers, because that method is implemented in the concrete class and it can invoke the Node.notifyPropertyChange method.
The problem here it is not EnumSet, is the lack of multiple inheritance that force NodeWithModifiers to be an interface and not a class extending Node.

ftomassetti:
(rebased)

matozoid:
Well, +1 then.

Issue #826, opened 2017-03-05
Fix exception handling in 3.1.1


matozoid:
Just noticed that I introduced a bug: the static parse method will not throw an exception if there is a result, but also problems.

PR #828, opened 2017-03-05
Issue 594 validation framework

matozoid:
See #594. This will...
- enable support for different Java versions
- enable validating an AST without having to go through parsing (validating a hand made AST for example)
- make the grammar cleaner

Also includes a fix for #826


ftomassetti:
I find it a great idea! +1

matozoid:
Efficient day today!

PR #829, opened 2017-03-05
Intersection type

ftomassetti:
Following an issue reported by @cal101 more issues were found (https://github.com/ftomassetti/javaparser/issues/8)

* there is an issue in how IntersectionType used in casts is parsed
* there is an issue in how IntersectionType is printed
* there is an issue in how the lambda parameters without type are printed

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -803,14 +803,12 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
         printJavaComment(n.getComment(), arg);
         printAnnotations(n.getAnnotations(), false, arg);
         printModifiers(n.getModifiers());
-        if (n.getType() != null) {
-            n.getType().accept(this, arg);
-        }
+        n.getType().accept(this, arg);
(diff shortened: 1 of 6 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -810,7 +810,9 @@ public void visit(final Parameter n, final Void arg) {
             printAnnotations(n.getVarArgsAnnotations(), false, arg);
             printer.print("...");
         }
-        printer.print(" ");
+        if (n.getType() != null && !(n.getType() instanceof UnknownType)) {
```
matozoid: Null checks should not be necessary in JavaParser.

matozoid:
Good find, I had no idea!

+1 once Travis runs

ftomassetti:
null checks removed. All merit of @cal101 !

ftomassetti:
Rebased & tests passing. Merging.

matozoid:
Hey, now that you're working on this part - can you make the term "multicast" disappear?

matozoid:
Too late :)

ftomassetti:
eh eh, my sixth sense made me hurry up to avoid the work :)

PR #839, opened 2017-03-06
Check root argument of SourceRoot must be a directory

arturbosch:
Or else `tryToParse` will fail on strage paths like `/home/artur/HelloWorld.java/HelloWorld.java`.

Also cleaned up `save` method.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
@@ -7,8 +7,6 @@ import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.printer.PrettyPrinter;
 
 import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.UnsupportedEncodingException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -128,14 +126,9 @@ public void saveAll(Path root) throws IOException {
...
-        path.getParent().toFile().mkdirs();
-
+        Files.createDirectories(path.getParent());
         final String code = new PrettyPrinter().print(cu);
-        try (PrintWriter out = new PrintWriter(path.toFile(), UTF8.toString())) {
-            out.println(code);
-        } catch (UnsupportedEncodingException e) {
-            throw new RuntimeException(e);
-        }
+        Files.write(path, code.getBytes());
```
matozoid: Oops, pass a character set to getBytes() please.

matozoid:
Cool, there is so much stuff they sneaked into Java 8 that I still don't know...

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
arturbosch: Argh, I thought utf-8 is always default ... seems not :(

arturbosch:
There is too much stuff to know everything :). Btw its java 7 stuff :D

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -50,6 +46,9 @@
     private JavaParser javaParser = new JavaParser();
 
     public SourceRoot(Path root) {
+        if (Files.isRegularFile(root)) {
```
jeandersonbc: another straightforward way of saying this would be `if (!Files.isDirectory(root))` but I guess this is just a minor comment. Overall, looks good to me.

matozoid: Are you sure that's the same thing?

arturbosch: I think your right! What I wanted to say is: it must be a directory, so writing `!directory` is more cleaner to read, thx

jeandersonbc: @matozoid I'm not 100% sure if one is the complement of the other (ie, `!Files.isDirectory` == `Files.isRegularFile`) but this is apparently what @arturbosch meant initially. I assumed this by the exception message.

jeandersonbc:
Feel free to pull the trigger if it's okay to you @matozoid :-) This one looks fine for me.

matozoid:
Thanks @arturbosch !

PR #846, opened 2017-03-07
Relativize the path of current visited file in SourceRoot

arturbosch:
Fixing that all filenames got resolved to `root/filename.java` and throwing exceptions on parsing.
Aligning with the behaviour of `parse with callback`.


matozoid:
Thanks! Can you throw in a simple unit test? There is a `SourceRootTest` already.

arturbosch:
Done :)

matozoid:
Thanks!

PR #861, opened 2017-03-13
Issue 109 left hand expressions

matozoid:
See #109 

This was so much easier than when I tried fixing it in the grammar, thank you new validators and OpenJDK test set.

### DIFF
--- a/javaparser-testing/src/test/java/com/github/javaparser/ast/validator/BaseJavaValidatorTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/ast/validator/BaseJavaValidatorTest.java
@@ -12,6 +12,7 @@ import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
 import static com.github.javaparser.ParseStart.EXPRESSION;
 import static com.github.javaparser.ParseStart.STATEMENT;
 import static com.github.javaparser.Providers.provider;
+import static com.github.javaparser.utils.TestUtils.assertNoProblems;
 import static com.github.javaparser.utils.TestUtils.assertProblems;
 
 public class BaseJavaValidatorTest {
@@ -72,6 +73,6 @@ public class BaseJavaValidatorTest {
     @Test
     public void leftHandAssignmentCanBeInBraces() {
         ParseResult<Expression> result = new JavaParser().parse(EXPRESSION, provider("(i) += (i) += 1"));
-        assertProblems(result);
+        assertNoProblems(result);
     }
 }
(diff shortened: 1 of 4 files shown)


javaparser-testing/src/test/resources/com/github/javaparser/bulk_test_results/openjdk_test_results.txt
```diff
@@ -473,6 +473,10 @@ Lexical error at line 8, column 9.  Encountered: "#" (35), after : ""
 openjdk/langtools/test/tools/javac/rawDiags/Error.java
 (line 9,col 17) Parse error. Found <EOF>, expected "("
 
+openjdk/langtools/test/tools/javac/StoreClass.java
+(line 35,col 9) Illegal left hand side of an assignment.
```
ftomassetti, line 477: Not sure if this file should be marked for removal

javaparser-testing/src/test/java/com/github/javaparser/ast/validator/BaseJavaValidatorTest.java
```diff
@@ -54,4 +57,21 @@ public void localInterface() {
...
+    @Test
+    public void leftHandAssignmentCannotBeEmptyBraces() {
+        ParseResult<Expression> result = new JavaParser().parse(EXPRESSION, provider("()=3"));
+        assertProblems(result, "(line 1,col 1) Illegal left hand side of an assignment.");
+    }
+
+    @Test
+    public void leftHandAssignmentCanBeInBraces() {
+        ParseResult<Expression> result = new JavaParser().parse(EXPRESSION, provider("(i) += (i) += 1"));
+        assertProblems(result);
```
ftomassetti: maybe "assertNoProblems" would be more intuitive in this case?

ftomassetti:
Looks good!

javaparser-testing/src/test/resources/com/github/javaparser/bulk_test_results/openjdk_test_results.txt
matozoid: Removal? Huh?

javaparser-testing/src/test/java/com/github/javaparser/ast/validator/BaseJavaValidatorTest.java
matozoid: I can add that.

javaparser-testing/src/test/resources/com/github/javaparser/bulk_test_results/openjdk_test_results.txt
ftomassetti: I mean: do we need to commit a file calculating the results of parsing the JDK?

matozoid: Why I'm doing that is so that I can see how a change in the grammar affects the parse results. If it wasn't checked in, I couldn't use the diff tools to see what the effect of a grammar or validation change is. It's invaluable to me :-)

ftomassetti: Right, make sense then

matozoid:
Travis seems kind of broken lately.

PR #863, opened 2017-03-15
Make helper methods to access literal values

### LINKED ISSUE #663: Make helper methods to access literal values
All literal integers, doubles, etc, are stored as strings. Add methods to get and set these as their native type.

arturbosch:
Fixes #663


matozoid:
Check feedback in the issue :)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
```diff
@@ -76,7 +76,17 @@ public boolean remove(Node node) {
     }
 
     public int asInt() {
-        return Integer.parseInt(value);
+        String result = value.replaceAll("_", "");
+        if (result.startsWith("0x")) {
+            return Integer.parseUnsignedInt(result.substring(2), 16);
```
matozoid: Check out https://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html#decode(java.lang.String)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
```diff
@@ -77,13 +77,13 @@ public boolean remove(Node node) {
 
     public int asInt() {
         String result = value.replaceAll("_", "");
-        if (result.startsWith("0x")) {
+        if (result.startsWith("0x") || result.startsWith("0X")) {
```
matozoid: Aw, just uppercase the thing on the first line :)

matozoid:
Can you add a test for this format?
`float x=0x1.fffffffffffffp10f;`

https://blogs.oracle.com/darcy/entry/hexadecimal_floating_point_literals

javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
arturbosch: Do you mean uppercase 'value' or try to uppercase the second char if the size is > 1 and its not a number :) ? Your sure if thats more readable than wo startsWith checks?

arturbosch:
argh more of this cryptic syntax xD, will see later if that can be parsed

javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
```diff
@@ -78,6 +78,10 @@ public boolean remove(Node node) {
         return super.remove(node);
     }
 
+    public String asString() {
+        return value;
+    }
```
matozoid: That's not right, is it? This should have the same (un)escaping as `asChar()`.

Interesting question: so what should `StringLiteralExpr(String)` do?

arturbosch: Oh year forgot that, that would also be my question, does StringLiteral need such methods (well like you said yes). If `asString` should always escape, `StringLiteralExpr(String)` shouldn't. Maybe we add documentation to `getValue` that it always returns unescaped strings?

matozoid: Should it? What I see is that what takes two constructors for `CharLiteralExpr` should take two constructors for `StringLiteralExpr` too, but we have a typing problem now:
`CharLiteral(String) --> StringLiteral(String) // literal source code`
`CharLiteral(char) --> StringLiteral(String???) // translates special characters into source code escapes`

I don't have any good solutions though. Add an `isLiteralSourceCode` parameter? Maybe deprecate all the string constructors and use escaping as the default?

matozoid: Adding some documentation to all this stuff is a very good idea though :-)

arturbosch: Are there any cases where you want the unescaped value? If there is no demand for special treatment of escaped and unescaped value, maybe we should escape the 'value' in the constructor? and return in the asXXX methods always the escaped one?

matozoid: If we do that, `StringLiteralExpr(String)` would do something different than `StringLiteralExpr.setValue(String)` which is unexpected. We can't get `setValue` to escape strings since all setters simply set the field without extra's, don't want to make an exception there.

Okay, after a lot of thinking: let's do your thing.
- make the constructor escape only the values that will create an illegal string constant (so \n and \r as `StringLiteralExpr.escape` does? Try reusing that and marking it deprecated)
- but also add a `setEscapedValue` or `setValueSafe` or whatever you think is clearest that does what the constructor does.
- `StringLiteralExpr(Range, String)` should stay as it is now!

arturbosch: ok done, remarks :D ? Sorry for merging and not rebasing >.<

matozoid:
+1 I'll leave you in peace now ;)

Issue #873, opened 2017-03-19
Add basic source levels for major Java versions


matozoid:
They can have validations for a few of the major things. This will force us to think about how we implement reuse in the validators.

Issue #787, opened 2017-02-21
Allow more modifiers for methods in interfaces


matozoid:
Check the JLS

matozoid:
http://aruld.info/private-interface-methods-in-java-9/
http://openjdk.java.net/jeps/213
https://bugs.openjdk.java.net/browse/JDK-8071453

matozoid:
Looks like the only things added are...
```java
interface X {
   private xxx() {...}
   private static yyy() {...}
}
```

PR #874, opened 2017-03-20
Issue 873 all source levels & remaining Java 9 work

matozoid:
See #873. Validations for older versions are pretty much guesswork. Version 8 and up should be correct, meaning: they should not report problems where there are none.

To be clear: this will allow the user to set the expected Java version, and it will report problems found in the AST when taking that version into account. So some version will complain that they do not support annotations, or that the `assert` keyword does not exist, that kind of thing.


matozoid:
Also implements #787 and #872 , thereby completing Java 9 support.

PR #879, opened 2017-03-29
Issue 878

### LINKED ISSUE #878: Let setters do nothing when set value == current value
Greetings. 

I am using an extended ModifierVisitor to change (add/change and remove code) an AST. I also wanted to use the AstObserver (with propagation) to monitor the changes I make to the AST.

My approach is to only override the ModifierVisitor methods that I'm interested in (eg. BlockStmt). However, the ModifierVisitor methods I don't override reconstruct each Node to essentially create an identical Node. As an example here is a code snippet:

public Visitable visit(final BinaryExpr n, final A arg) {
        Expression left = (Expression) n.getLeft().accept(this, arg); 
        Expression right = (Expression) n.getRight().accept(this, arg); 
        Comment comment = n.getComment().map(s -> (Comment) s.accept(this, arg)).orElse(null);
        if (left == null)
            return right;
        if (right == null)
            return left;
        n.setLeft(left);
        n.setRight(right);
        n.setComment(comment);
        return n;
    }

The problem is this reconstruction triggers change events to the AstObserver even though no changes have actually occurred.

I'm just wondering if perhaps I'm misusing either ModifierVisitor/AstObserver or if anyone has recommendations on a better approach.

Thanks!

ftomassetti:
Ignore setters setting the same value already present (fix #878 )

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/node/PropertyGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/node/PropertyGenerator.java
@@ -60,6 +60,7 @@ public class PropertyGenerator extends NodeGenerator {
 
         final BlockStmt body = setter.getBody().get();
         body.getStatements().clear();
+
         if (property.isRequired()) {
             Class<?> type = property.getType();
             if (property.isNonEmpty() && property.isSingular()) {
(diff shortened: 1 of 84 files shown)


matozoid:
Nice! +1

ftomassetti:
Thanks, merging

Issue #882, opened 2017-04-01
CallableDeclaration: Add method to retrieve signature


fruffy:
In Java a signature is defined as 
> the method name and the number and type of its parameters. Return types and thrown exceptions are not considered to be a part of the method signature.  

However, this feature is not present in CallableDeclaration.
getDeclarationAsString() always includes the return type.
I ran into trouble here when comparing the MethodDeclaration of JavaParser vs the MethodDeclaration of Java-Symbol-Solver, which both have a different naming behaviour.

So I added a method to all CallableDeclaration types which is very similar to getDeclarationAsString()
```
    @Override
    public String getSignature() {
        StringBuilder sb = new StringBuilder();
        sb.append(getName());
        sb.append("(");
        boolean firstParam = true;
        for (Parameter param : getParameters()) {
            if (firstParam) {
                firstParam = false;
            } else {
                sb.append(", ");
            }
            sb.append(param.getType().toString(prettyPrinterNoCommentsConfiguration));
            if (param.isVarArgs()) {
                sb.append("...");
            }
        }
        sb.append(")");
        return sb.toString();
}
```
Is it sensible to add this method or should getDeclarationAsString() just be amended with a fourth parameter?

matozoid:
Thanks for the feedback!

Lots of remarks here:
- `getDeclarationAsString()` is something I've been tolerating in the codebase for a while, but I'm not happy with it
- It would be nice to have `getSignature()` return a new `Signature` type. Functionally meaningful information shouldn't be passed around in a string.
- instead of `toString`, use `asString`. `toString` does pretty printing, `asString` tries to give the most useful representation of the node.
- printing a list like that can be done easily with `SeparatedItemStringBuilder`
- what is the actual problem we're solving between JP and JSS? I haven't used JSS lot - do you have an idea @ftomassetti ?

matozoid:
@FRuffy - could you take a look at the PR and check if it is correct and sufficient for your needs?

fruffy:
Yes, it works great. thanks a lot! There is one thing, I had to add a temporary toString() method to make it comparable with the javasymbolsolver `getSignature()` call. I am going to make a suggestion to implement the `Signature` type in javasymbolsolver too.
```
@Override
public String toString() {
	StringBuilder parameterBuilder = new StringBuilder();
	parameterBuilder.append(name + "(");
    for (int i = 0; i < parameterTypes.size(); i++) {
        if (i != 0) {
        	parameterBuilder.append(", ");
[...]

PR #890, opened 2017-04-13
Add a getter for method or constructor signatures.

matozoid:
See #882


matozoid:
If more is needed, create issues.

Issue #842, opened 2017-03-07
Save filepath/filename in CompilationUnit


shibumi:
It would be nice if JavaParser.parse(Filename) could return the filepath/filename in the returned Object.

matozoid:
On one hand: this would mean refactoring `SourceRoot`, on the other hand: this would enable more validations (class matches filename, package declaration wrong, etc.)

shibumi:
At the moment we have an "information leak". When i for example iterate over a big java project and produce CompilationUnits for every file and sort these units into a bigger modell, I would like to have the filepath/filename for every CompilationUnit. I am sure that also other projects would profit from this changes.

matozoid:
Yes, it's probably a good thing and this will be built. In the end it is a design decision, but since our approach is "the Abstract Syntax Christmas Tree" full of decorations, this can go in too.

shibumi:
@matozoid cool awesome!

PR #893, opened 2017-04-14
Store path of source file on CompilationUnit

matozoid:
See #842 

It's a bit of a simplistic implementation. I could have stored the location of files loaded using a classloader for example, but I'm assuming that users who want that information have a very specific use case that they can build themselves.

`SourceRoot` has minor breakage: instead of a map of `path`->`ParseResult`, you now get a list of `ParseResult`. The `CompilationUnit` in the result has the path already.

This makes me wonder if the path shouldn't be stored in the `ParseResult`. Does anybody have a suggestion?


DeepSnowNeeL:
I think compilation unit is better to store the path, so it opens up things
like compilationUnit.save() , which would be nonsense to do on a "parse
result"

matozoid:
Oh right, I could put save on there as well.

ftomassetti:
What is the status of this one? Do you need a review? Are you still working on this one?

matozoid:
Yes, I'm cleaning up smaller stuff for now, but I'll get back to this one soon.

One thing I was pondering is whether you should be able to ask the compilation unit for the source directory, and I think that's a good idea:

`file location = source directory + package + file name`

I had some doubt about if it is smart to use the `PackageDeclaration` to figure out the package, and I think it is. It should be correct almost always, and when it isn't a big exception is fitting.

This would make interoperability with `SourceRoot` a lot easier too.

PR #898, opened 2017-04-16
Use annotations for metamodel generation

matozoid:
Inspired by @ftomassetti's annotation usage.


javaparser-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/InitializePropertyMetaModelsStatementsGenerator.java
```diff
@@ -71,21 +72,11 @@ public void generateDerivedProperty(Class<?> nodeClass, Method method, ClassOrIn
...
-                (c == ArrayCreationExpr.class && name.equals("levels")) ||
-                (c == FieldDeclaration.class && name.equals("variables")) ||
-                (c == IntersectionType.class && name.equals("elements")) ||
-                (c == UnionType.class && name.equals("elements")) ||
-                (c == VariableDeclarationExpr.class && name.equals("variables"));
+        return field.isAnnotationPresent(NonEmptyProperty.class);
     }
 
     private boolean isNonEmpty(Method method) {
-        return true;
```
matozoid, line 88: This is actually false, "non empty" means "this list should have at least one element" or "this string should have at least one character." That was badly documented so I fixed that.

Issue #713, opened 2017-01-19
Fix AppVeyor trouble


matozoid:
It hasn't been working for ages. What I remember from a discussion with @DeepSnowNeeL was that all the test files that are loaded use unix EOL's, and JavaParser uses the platform specific EOL. So the tests should make sure that, I guess, the files are loaded and then converted to platform specific.

Disadvantage: any tests that test what the EOL character should be will never fail anymore.

PR #901, opened 2017-04-17
Support parsing generic type argument lists containing primitives

batterseapower:
Obviously these argument lists don't really have a meaning in current Java:

- They may be supported in the future
- Third party tools that extend Java in some way may be able to give this construct meaning so being able to parse it is useful for them
- JavaParser can already *write* source files that contain primitive types in type argument lists, it just can't *parse* the files that it writes

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java5Validator.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java5Validator.java
@@ -3,6 +3,7 @@ package com.github.javaparser.ast.validator;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.nodeTypes.NodeWithTypeArguments;
+import com.github.javaparser.ast.type.PrimitiveType;
 import com.github.javaparser.ast.type.Type;
 
 import java.util.Optional;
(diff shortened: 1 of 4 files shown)


batterseapower:
Not really clear to me what the AppVeyor problem is. The tests run fine for me on OS X. For at least one of the AppVeyor issues (package-info) the problem seems to be that the BDD framework's "expected" output is wrong (doesn't match what I see in pretty_printing_scenarios.story).

matozoid:
Thanks! This is cool :-) Never mind appveyor, that has been broken for a long time now (and I can build on windows without a problem.) See #713.

The one argument that already makes this a good PR is that it moves checks from the parser to the validator, making JavaParser more robust. +1 for me, will leave it open in case someone has something more to say about it.

matozoid:
Oh, one thing: `SingleNodeTypeValidator` seems to be a better fit for the validator.

batterseapower:
I'm using TreeVisitorValidator because the genericsWithoutDiamondOperator validator uses it. I had a play with changing both to SingleNodeTypeValidator, but it's not obvious how to make it work because NodeWithTypeArguments does not extend Node.

matozoid:
Ah right, sorry, it will indeed not work with those interfaces. All set then.

matozoid:
Thanks!

batterseapower:
Thanks for the merge!

matozoid:
Release will probably be next Sunday :)

Issue #902, opened 2017-04-22
Generate "range" constructors


matozoid:
In preparation for letting nodes point to tokens.

Issue #770, opened 2017-02-10
Add annotation to methods and classes that were generated


matozoid:
... to indicate that they were.

Java has a `@Generated` annotation which requires the generator class as a parameter. This is messy, but I think it's the best option: no additional dependencies, no custom annotation, and it is a sensible thing to mention what generated the code.

```java
@Generated("com.github.javaparser.generator.core.node.GetNodeListsGenerator")
@Override
public List<NodeList<?>> getNodeLists() {
    return Arrays.asList(getMembers(), getAnnotations());
}
```

It might be possible to automate this in the `VisitorGenerator` class, not sure about the `NodeGenerator`

arturbosch:
+1, they would also indicate not to try to refactor generated code

PR #904, opened 2017-04-23
Issue 902 generate range constructors

matozoid:
See #902 and #770

This got a little out of hand because I found all kinds of blemishes and also improved the generators a bit.

Some grammar improvements sneaked in too.


ftomassetti:
Nice idea to use the `@Generated` annotation.
Yes, it went a bit out of hand :D

matozoid:
Sorry! It's the best way to get a PR accepted though: change each and every file in the project ;)

ftomassetti:
Eh eh, that is true, and I kind of remember that someone did much worse a few months ago... so I am definitely not complaining here :D

matozoid:
And there it goes!

PR #908, opened 2017-04-26
Fixes #907 allowing javadoc block parsing over multiple lines.

### LINKED ISSUE #907: Javadoc block tag parsing issues
Hi guys,

I'm having issues with parsing of Javadoc block tags. Given the following javadoc:

    /*
    * Add a field to this and automatically add the import of the type if needed
    *
    * @param typeClass the type of the field
    * continued in a second line
    * @param name the name of the field
    */

I would expect two block tags for the two params, however with version 3.2.3-SNAPSHOT I get three block tags, where the "continued in a second line" is parsed as `JavadocBlockTag{type=UNKNOWN, content='JavadocDescription{elements=[JavadocSnippet{text='in a second line'}]}', name=Optional.empty}`.

I have extended the unit tests to showcase the problem: https://github.com/sischnei/javaparser/commit/6c450c880c6dca8f24c43c91ab680fc08a04ce25

I also attempted to solve the issue and contribute a fix, but I'm worried that the problem is a little bit more complicated. I was thinking about adding all lines that do not confirm with `JavadocParser::isABlockLine` to the last block, however the `JavadocParser::isABlockLine` only checks if the line starts with an '@' symbol. So with my solution, a second line starting with eg. a "@see" would not be correctly associated with the last "@param" block. 

I hope you understand what I mean. Basically I'd suggest extending also `JavadocParser::isABlockLine` to only consider real blocks (like "@param", "@return",...). What do you think? Should I go on and create a pull request?

sischnei:
Allows javadoc blocks to extend over multiple lines and parse them correctly as one block.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
@@ -26,10 +26,10 @@ import com.github.javaparser.javadoc.Javadoc;
 import com.github.javaparser.javadoc.JavadocBlockTag;
 import com.github.javaparser.javadoc.description.JavadocDescription;
 import com.github.javaparser.utils.Utils;
-
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import static com.github.javaparser.utils.Utils.nextWord;
@@ -38,6 +38,9 @@ import static com.github.javaparser.utils.Utils.nextWord;
  * The class responsible for parsing the content of JavadocComments and produce JavadocDocuments.
  */
 class JavadocParser {
+
+    private static String BLOCK_TAG_PREFIX = "@";
+    private static Pattern BLOCK_PATTERN = Pattern.compile("^" + BLOCK_TAG_PREFIX, Pattern.MULTILINE);
 
     public static Javadoc parse(JavadocComment comment) {
         return parse(comment.getContent());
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
```diff
@@ -57,9 +60,15 @@ public static Javadoc parse(String commentContent) {
             blockLines = Collections.emptyList();
         } else {
             descriptionText = trimRight(String.join("\n", cleanLines.subList(0, indexOfFirstBlockTag)));
-            blockLines = cleanLines.subList(indexOfFirstBlockTag, cleanLines.size()).stream()
-                    .filter(Utils.STRING_NOT_EMPTY)
-                    .collect(Collectors.toList());
+            String tagBlock = cleanLines.subList(indexOfFirstBlockTag, cleanLines.size())
```
ftomassetti: I am a bit confused by the fact you first take everything and put it into a string and later you seem to split it again and remove empty lines. Could you add some comments to explain what is going on.

I mean, I should have written way more comments when I wrote the original version of this cose, but let's not focus on that...

javaparser-testing/src/test/java/com/github/javaparser/JavadocParserTest.java
```diff
@@ -111,6 +111,26 @@ public void parseParamBlockTags() {
     }
 
     @Test
+    public void parseMultilineParamBlockTags() {
+        String text = "\n" +
```
ftomassetti, line 115: Nice test

javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
sischnei: I do the combining and splitting because the combined block makes it easier to split according to the "javadoc rules". When working with each line on its own, I think I would have needed a weird looping, handling the backtracking etc. I hope my comments make it clearer!

matozoid:
Okay, it's release time and I think this one still has a discussion open, so it's not going in.

sischnei:
What can I improve to have this integrated?

matozoid:
Agree or disagree on the comments in #907 :-)

I'm already releasing, but we have weekly releases so no big deal.

matozoid:
Hmmm, looks like there is no real discussion there now that I read carefully.

matozoid:
+1 :-)

PR #910, opened 2017-04-26
Make sure files are always generated with a unix end of line character.

matozoid:
Adds an option to the pretty printer to set the end of line character, and makes `SourceRoot` always use `\n`. This way files should always be generated with the same EOL character.

The other PR will change this code around again.


matozoid:
@DeepSnowNeeL - if you merge this in, code generation should no longer cause a huge amount of changes on your work.

DeepSnowNeeL:
I'll try it tonight :)

DeepSnowNeeL:
Looks like it's ok, the files generated doesn't show up in my git unstaged changes and they does have linux ending now

DeepSnowNeeL (approved):

Issue #913, opened 2017-05-04
FieldAccessExpr.scope should not be optional


matozoid:
It currently is. I don't think accessing a field of nothing is a useful expression.

PR #915, opened 2017-05-07
Make field access scope not optional

matozoid:
See #913

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
@@ -27,11 +27,13 @@ import com.github.javaparser.ast.expr.Expression;
 import java.util.Optional;
 
 /**
- * Represents a node which can have a scope expression eg. method calls (object.method()).
+ * Represents a node which has an optional scope expression eg. method calls (object.method()).
  */
 public interface NodeWithOptionalScope<N extends Node> {
 
     Optional<Expression> getScope();
 
     N setScope(Expression scope);
+    
+    N removeScope();
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
@@ -27,22 +27,11 @@ import com.github.javaparser.ast.expr.Expression;
 import java.util.Optional;
 
 /**
- * Represents a node which can have a scope expression eg. method calls or
- * field accesses (object.method(), object.field).
+ * Represents a node which can have a scope expression eg. method calls (object.method()).
  */
 public interface NodeWithOptionalScope<N extends Node> {
 
     Optional<Expression> getScope();
 
     N setScope(Expression scope);
-
-    /**
-     * Removes the scope by calling {@link #setScope} with null.
-     *
-     * @return this node
-     */
-    default N removeScope() {
-        return setScope(null);
-    }
-
 }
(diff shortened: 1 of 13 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
```diff
@@ -27,22 +27,11 @@
 import java.util.Optional;
 
 /**
- * Represents a node which can have a scope expression eg. method calls or
- * field accesses (object.method(), object.field).
+ * Represents a node which can have a scope expression eg. method calls (object.method()).
```
ftomassetti: Maybe we can stress the "can" and say explicitly that this is interface is for the nodes which may or may not have a scope. It should not be used for the nodes that must have a scope

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
```diff
@@ -27,22 +27,11 @@
...
  */
 public interface NodeWithOptionalScope<N extends Node> {
 
     Optional<Expression> getScope();
 
     N setScope(Expression scope);
-
```
ftomassetti, line 38: Why do we want to remove it?

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
```diff
@@ -501,7 +501,7 @@ private static void initializePropertyMetaModels() {
         enclosedExprMetaModel.getDeclaredPropertyMetaModels().add(enclosedExprMetaModel.innerPropertyMetaModel);
         fieldAccessExprMetaModel.namePropertyMetaModel = new PropertyMetaModel(fieldAccessExprMetaModel, "name", com.github.javaparser.ast.expr.SimpleName.class, Optional.of(simpleNameMetaModel), false, false, false, false, false);
         fieldAccessExprMetaModel.getDeclaredPropertyMetaModels().add(fieldAccessExprMetaModel.namePropertyMetaModel);
-        fieldAccessExprMetaModel.scopePropertyMetaModel = new PropertyMetaModel(fieldAccessExprMetaModel, "scope", com.github.javaparser.ast.expr.Expression.class, Optional.of(expressionMetaModel), true, false, false, false, false);
+        fieldAccessExprMetaModel.scopePropertyMetaModel = new PropertyMetaModel(fieldAccessExprMetaModel, "scope", com.github.javaparser.ast.expr.Expression.class, Optional.of(expressionMetaModel), false, false, false, false, false);
```
ftomassetti, line 504: Looking at this list of boolean flags I wonder if we should create enums for all of these options, so that the code becomes much more readable

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -334,7 +334,8 @@ private static CsmElement typeArguments() {
 
         concreteSyntaxModelByClass.put(FieldAccessExpr.class, CsmElement.sequence(
                 CsmElement.comment(),
-                CsmElement.conditional(SCOPE, IS_PRESENT, CsmElement.sequence(CsmElement.child(SCOPE), CsmElement.token(GeneratedJavaParserConstants.DOT))),
+                CsmElement.child(SCOPE),
```
ftomassetti, line 337: It seems correct to me

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -577,8 +577,7 @@ public void visit(final EnclosedExpr n, final Void arg) {
     @Override
     public void visit(final FieldAccessExpr n, final Void arg) {
         printJavaComment(n.getComment(), arg);
-        if (n.getScope().isPresent())
-            n.getScope().get().accept(this, arg);
+            n.getScope().accept(this, arg);
```
ftomassetti: Is the indentation off?

ftomassetti (approved):
Looks good to me

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
matozoid: My thinking was that they get generated anyway, but that's bad thinking now you mention it.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithOptionalScope.java
matozoid: Rephrasing that.

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
matozoid: Hmmm, I didn't generate this for readability, but you do have a point. I did spend too much time rummaging through these because they are so meaningless.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: Whoops, this visitor is of course not generated :(

Issue #922, opened 2017-05-12
All MethodCallExpr should take either SimpleName or a String


ftomassetti:
So that they are consistent

matozoid:
The constructors you mean? Yes, they are a bit of a mess.

ftomassetti:
Yes, I meant the constructors. I am the worst issue opener on earth :D

PR #925, opened 2017-05-12
Add missing constructors for MethodCallExpr

matozoid:
See #922

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
@@ -65,6 +65,14 @@ public final class MethodCallExpr extends Expression implements NodeWithTypeArgu
         this(null, scope, new NodeList<>(), new SimpleName(name), new NodeList<>());
     }
 
+    public MethodCallExpr(final Expression scope, final SimpleName name) {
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
```diff
@@ -65,6 +65,14 @@ public MethodCallExpr(final Expression scope, final String name) {
         this(null, scope, new NodeList<>(), new SimpleName(name), new NodeList<>());
     }
 
+    public MethodCallExpr(final Expression scope, final SimpleName name) {
```
ftomassetti, line 68: Looks good. Eventually we could remove some of these constructors, but for now adding these two seems the way to go

ftomassetti:
+1

javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
matozoid: Well, I don't think any of these are doing any harm?

matozoid: Ever since I put every name and identifier into a node, I've been trying to give `String` shortcuts to reduce the pain.

PR #929, opened 2017-05-14
Handling more cases of indent/unindent reconciliation

ftomassetti:
While working on a small example I found an issue with indentation and it snowballed in a larger set of changes. A bunch of new tests were added (we have now 200 tests just for printing: i.e., pretty printing + lexical preservation).

There are a lot of heuristics and tentatives of doing the most sensible thing in each situation. I will try to add more comments, simplify and polish things following the advices I will receive for this PR.

I wish things could be simpler, but I could not find a way to make the implementation easier and still have all the tests pass.


matozoid:
Well, I'm glad you're picking this up. I'm not familiar with the code yet.

ftomassetti:
It should be now ready for review. Even if I am aware this stuff is difficult to review. Anyway I would base the solution to the other lexical preservation bugs upon this branch in case it cannot be merged soonish.

matozoid:
Can I merge this then?

ftomassetti:
This was included in #932

Issue #926, opened 2017-05-12
LexicalPreservingPrinter BlockStmt addStatement is not added on a new line


adhesivee:
I currently have JavaParser working with `LexicalPreservingPrinter`. The problem that I have is that when I add a statement in `MethodDeclaration` it is added to end of the line, instead of a new line. I have method:
```
public void someMethod() {
   String test = "";
}
```

When I want to add a statement it will result in:
```
public void someMethod() {
   String test = "";String test2 = "";
}
```

My snippet to add the test2.
```
MethodDeclaration method = /* get method declaration */;
BlockStmt blockStmt = method.getBody()
                    .get();
Statement statement = JavaParser.parseStatement("String test2 = \"\"");
blockStmt.addStatement(blockStmt);
```

ftomassetti:
Working on solving this one on top of PR #929 
The bad thing is that I have a stack overflow because we are building the NodeText for a bunch of nodes at the same time and there are issues related to indentation, that is the root of all evil...

ftomassetti:
Resolved that particular issue, I am getting closer to a solution but there are still issues related to guess what... indentation :D

Currently I am getting:

```
    public void someMethod() {
        String test = "";
        String test2 = "";
        }
```

which is correct beside the last line that should have an indentation of 4 instead of 8.

ftomassetti:
This should now be solved in the next release.

PR #930, opened 2017-05-15
Lexical Preservation, issue 926

ftomassetti:
This PR is built on top of #929 

It solves a few issues:
* stack overflow in findIndentation
* special handling for the type in VariableDeclaration/FieldDeclaration
* refinements to indentation reconciliation

A new test specific for the case reported in #926 is added.

One step (one test?) at the time we will get it right...


matozoid:
Ummm, you know best! :-)

ftomassetti:
Well, I can continue working on other Lexical Preservation fixes on top of these two PRs (when time allows). Then when you are ready to do a review we evaluate the situation and perhaps we merged the changes that are ready. In the meantime if other contributors want to take a look and comment I would include their feedback.
Does it sound good?

matozoid:
My attitude is: you are making lex pres better, so it's +1 by default. I'm planning to do clean up the API for 4.0, so before that it's all yours :)

matozoid:
Merge?

ftomassetti:
This was included in #932

PR #932, opened 2017-05-19
Fix issue 866: LexicalPreservingPrinter: Unsupported operation - remove

### LINKED ISSUE #866: LexicalPreservingPrinter: Unsupported operation - remove()
Not sure about this one, if it even is supposed to work like that, but I just put together some quick test of a specific use case I would like to use the LexicalPreservingPrinter for.

Basically "moving" some Annotations around by first removing them and adding them again.
The test I put together (copy of LexicalPreservingPrinterTest.handleOverrideAnnotation) fails with...
```
java.lang.UnsupportedOperationException: removed child(MarkerAnnotationExpr) vs TokenTextElement(73) {protected}

	at com.github.javaparser.printer.lexicalpreservation.Difference.apply(Difference.java:520)
	at com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter$1.concreteListChange(LexicalPreservingPrinter.java:164)
....
```

The test case
```
    @Test
    public void moveOverrideAnnotations() {
        String code = "public class TestPage extends Page {" + EOL +
                EOL +
                "   protected void test() {}" + EOL +
                EOL +
                "   protected @Override void initializePage() {}" + EOL +
                "}";

        Pair<ParseResult<CompilationUnit>, LexicalPreservingPrinter> result = LexicalPreservingPrinter
                .setup(ParseStart.COMPILATION_UNIT, Providers.provider(code));

        CompilationUnit cu = result.a.getResult().get();

        cu.getTypes().stream()
                .forEach(type -> {
                    type.getMembers().stream()
                            .forEach(member -> {
                                if (member instanceof MethodDeclaration) {
                                    MethodDeclaration methodDeclaration = (MethodDeclaration) member;
                                    if (methodDeclaration.getAnnotationByName("Override").isPresent()) {
                                        methodDeclaration.getAnnotations().stream()
                                                .forEach(anno -> anno.remove());

                                        methodDeclaration.addMarkerAnnotation("Override");
                                    }
                                }
                            });
                });
        assertEquals("public class TestPage extends Page {" + EOL +
                EOL +
                "   @Override" + EOL +
                "   protected void test() {}" + EOL +
                EOL +
                "   @Override" + EOL +
                "   protected void initializePage() {}" + EOL +
                "}", result.b.print(cu));
    }
```

Kind regards

ftomassetti:
This fix #866. 
The PR is built on top of #929 and #930 

We introduce a new CsmElement named CsmMix to represent elements that can have any order. They require a specific handling in Difference.apply. We basically considere the CsmMix before the change, after the change and the order of elements in the text trying to figure out the most natural way to change the text.

Additionally we need also to adapt space around in certain cases.

These three PRs combines should fix a few things in the lexical preservation. It is probably not 100% perfect but the number of tests is growing (203) and it should eventually stabilize. There are hard things related to indentation and space reconciliation.

In the long term I hope we root out all bugs, we grow our test base and then in 4.0 we can improve the whole design.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
@@ -60,6 +60,9 @@ public class ConcreteSyntaxModel {
         return list(ObservableProperty.MODIFIERS, space(), none(), space());
     }
 
+    /**
+     * Build a mix collecting all the elements specified.
+     */
     private static CsmElement mix(CsmElement... elements) {
         return new CsmMix(Arrays.asList(elements));
     }
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
@@ -34,13 +34,11 @@ import com.github.javaparser.ast.type.*;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
 import com.github.javaparser.printer.concretesyntaxmodel.CsmConditional;
 import com.github.javaparser.printer.concretesyntaxmodel.CsmElement;
+import com.github.javaparser.printer.concretesyntaxmodel.CsmMix;
 import com.github.javaparser.printer.concretesyntaxmodel.CsmToken;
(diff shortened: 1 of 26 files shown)


matozoid:
Mergey merge merge?

ftomassetti:
Rebased after merging #933

Issue #812, opened 2017-03-02
Store begin and end token in every node


matozoid:
This would simplify the lexical preservation code, and is also more useful to users than the long list of tokens we currently supply, and it can be used to examine or display the input when a parse error was recovered. This is intimately tied to begin and end position, so we have to see what a nice API would be.

This change would affect every node constructor (time to build a generator?) and just about every production in the grammar.

matozoid:
This is now top priority:
- it would (hopefully) simplify lexical preservation code
- it would help @jeandersonbc put something sensible inside bad nodes
- it would make a big but necessary change in the grammar, making it more stable for the future (for the people working on forks)

matozoid:
I'm thinking to put the information in a `Range`-like object with the same constraints: either the whole object isn't there, or it is there and all its fields are there.
```
node.range.begin
node.range.end
node.tokens.begin
node.tokens.end
```

We would start getting some inconsistencies:
- when someone makes a new nodes and set the position, there may not be a token.
- when someone parses some code and changes the position, the token will still contain the old position, and the other way around.

PR #933, opened 2017-05-19
Issue 812 begin end token

matozoid:
Finally finished this: in addition to `range` we now have `tokenRange`. `range` keeps location information by using line/column position. `tokenRange` keeps location information by pointing to the begin/end `JavaToken` (which has line/column information inside.)

We could now deprecate `range`.

There are some API breakages (like `Problem` indicating the location as a `TokenRange` instead of a `Range`) and I may spend some more time there.

@ftomassetti this should be a lot of help with the lexical preservation code.

See #812.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
@@ -56,7 +56,7 @@ public abstract class NodeGenerator extends Generator {
                     throw new AssertionError(f("Wanted to regenerate a method with signature %s in %s, but it wasn't there.", callable.getSignature(), containingClassOrInterface.getNameAsString()));
                 });
     }
-
+    
     private void addMethod(
             ClassOrInterfaceDeclaration containingClassOrInterface,
             CallableDeclaration<?> callable,
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -47,7 +47,6 @@ import java.util.TreeSet;
 import static com.github.javaparser.ParseStart.*;
 import static com.github.javaparser.Problem.PROBLEM_BY_BEGIN_POSITION;
 import static com.github.javaparser.Providers.*;
-import static com.github.javaparser.Range.range;
 import static com.github.javaparser.utils.Utils.assertNotNull;
 
 /**
@@ -112,8 +111,7 @@ public final class JavaParser {
      * The start indicates what can be found in the source code (compilation unit, block, import...)
      *
      * @param start refer to the constants in ParseStart to see what can be parsed.
-     * @param provider refer to Providers to see how you can read source.
-     * The provider will be closed after parsing.
+     * @param provider refer to Providers to see how you can read source. The provider will be closed after parsing.
      * @param <N> the subclass of Node that is the result of parsing in the start.
      * @return the parse result, a collection of encountered problems, and some extra data.
      */
@@ -134,9 +132,14 @@ public final class JavaParser {
             return new ParseResult<>(resultNode, parser.problems, parser.getTokens(),
                     parser.getCommentsCollection());
         } catch (ParseException p) {
-            final Token token = p.currentToken;
-            final Range range = range(token.beginLine, token.beginColumn, token.endLine, token.endColumn);
-            parser.problems.add(new Problem(makeMessageForParseException(p), range, p));
+            TokenRange tokenRange = null;
+            if (p.currentToken != null) {
+                if (p.currentToken instanceof GeneratedJavaParser.CustomToken) {
+                    final JavaToken token = ((GeneratedJavaParser.CustomToken) p.currentToken).javaToken;
+                    tokenRange = new TokenRange(token, token);
+                }
+            }
+            parser.problems.add(new Problem(makeMessageForParseException(p), tokenRange, p));
             return new ParseResult<>(null, parser.problems, parser.getTokens(), parser.getCommentsCollection());
         } catch (Exception e) {
             final String message = e.getMessage() == null ? "Unknown error" : e.getMessage();
@@ -213,8 +216,7 @@ public final class JavaParser {
      * Parses the Java code contained in the {@link InputStream} and returns a
      * {@link CompilationUnit} that represents it.
      *
-     * @param in {@link InputStream} containing Java source code.
-     * It will be closed after parsing.
+     * @param in {@link InputStream} containing Java source code. It will be closed after parsing.
      * @param encoding encoding of the source code
      * @return CompilationUnit representing the Java source code
      * @throws ParseProblemException if the source code has parser errors
@@ -228,8 +230,7 @@ public final class JavaParser {
      * {@link CompilationUnit} that represents it.<br>
      * Note: Uses UTF-8 encoding
      *
-     * @param in {@link InputStream} containing Java source code.
-     * It will be closed after parsing.
+     * @param in {@link InputStream} containing Java source code. It will be closed after parsing.
      * @return CompilationUnit representing the Java source code
      * @throws ParseProblemException if the source code has parser errors
      */
@@ -241,8 +242,7 @@ public final class JavaParser {
      * Parses the Java code contained in a {@link File} and returns a
      * {@link CompilationUnit} that represents it.
      *
-     * @param file {@link File} containing Java source code.
-     * It will be closed after parsing.
+     * @param file {@link File} containing Java source code. It will be closed after parsing.
      * @param encoding encoding of the source code
      * @return CompilationUnit representing the Java source code
      * @throws ParseProblemException if the source code has parser errors
@@ -257,8 +257,7 @@ public final class JavaParser {
      * {@link CompilationUnit} that represents it.<br>
      * Note: Uses UTF-8 encoding
      *
-     * @param file {@link File} containing Java source code.
-     * It will be closed after parsing.
+     * @param file {@link File} containing Java source code. It will be closed after parsing.
      * @return CompilationUnit representing the Java source code
      * @throws ParseProblemException if the source code has parser errors
      * @throws FileNotFoundException the file was not found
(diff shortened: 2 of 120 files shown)


matozoid:
Merge?

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
```diff
@@ -56,7 +56,7 @@ protected void replaceWhenSameSignature(ClassOrInterfaceDeclaration containingCl
                     throw new AssertionError(f("Wanted to regenerate a method with signature %s in %s, but it wasn't there.", callable.getSignature(), containingClassOrInterface.getNameAsString()));
                 });
     }
-
+
```
ftomassetti, line 59: Wearing my extra-douche-bag hat I would say not spaces on a blank line. Maybe at some point we could have some automated process remove these things. For now I would not bother changing it.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -112,8 +111,7 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
      * The start indicates what can be found in the source code (compilation unit, block, import...)
      *
      * @param start refer to the constants in ParseStart to see what can be parsed.
-     * @param provider refer to Providers to see how you can read source.
-     * The provider will be closed after parsing.
+     * @param provider refer to Providers to see how you can read source. The provider will be closed after parsing.
```
ftomassetti, line 114: Does closing apply to all providers?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -24,17 +24,29 @@
...
 public class JavaToken {
+    public static final JavaToken INVALID = new JavaToken();
+
     private final Range range;
     private final int kind;
     private final String text;
     private final Optional<JavaToken> previousToken;
     private Optional<JavaToken> nextToken = Optional.empty();
 
+    private JavaToken() {
```
ftomassetti, line 43: good that this is private if it builds an invalid JavaToken

javaparser-core/src/main/java/com/github/javaparser/Problem.java
```diff
@@ -69,21 +69,21 @@ public String getMessage() {
      * @return the message plus location information.
      */
     public String getVerboseMessage() {
-        return getLocation().map(l -> l.begin + " " + message).orElse(message);
+        return getLocation().map(l -> l.getBegin().getRange().begin + " " + message).orElse(message);
```
ftomassetti, line 72: it seems more complicate than before to access the begin value

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
```diff
@@ -0,0 +1,38 @@
+package com.github.javaparser;
+
+import static com.github.javaparser.utils.Utils.assertNotNull;
+
+/**
+ * The range of tokens covered by this node.
```
ftomassetti, line 6: Could range let users think it is just about knowing the position of the tokens and not the tokens itself?

ftomassetti: Could it makes sense to call it TokenSequence to distinguish it more from Range?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -437,7 +459,7 @@ public boolean remove() {
      * it will try to remove its parent instead,
      * until it finds a node that can be removed,
      * or no parent can be found.
-     * 
+     * <p>
```
ftomassetti, line 462: This is professional usage of JavaDoc :D

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithTokenRange.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.github.javaparser.Position;
+import com.github.javaparser.Range;
+import com.github.javaparser.TokenRange;
+import com.github.javaparser.ast.Node;
+
+import java.util.Optional;
+
+/**
+ * A node that has a Range, which is every Node.
```
ftomassetti, line 11: A node that has a _TokenRange_

javaparser-core/src/main/java/com/github/javaparser/ast/type/TypeParameter.java
```diff
@@ -72,8 +73,13 @@ public TypeParameter(final String name, final NodeList<ClassOrInterfaceType> typ
         this(null, new SimpleName(name), typeBound, new NodeList<>());
     }
 
+    /**
+     * @deprecated range shouldn't be in utility constructors.
```
ftomassetti, line 77: Why not?

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
```diff
@@ -47,8 +47,7 @@
     public static <E> boolean isNullOrEmpty(Collection<E> collection) {
         return collection == null || collection.isEmpty();
     }
-
-
+
```
ftomassetti, line 50: Nice

javaparser-core/src/main/javacc/java.jj
```diff
@@ -171,6 +164,7 @@ TOKEN_MGR_DECLS :
 {
     private List<JavaToken> tokens = new ArrayList<JavaToken>();
     private CommentsCollection commentsCollection = new CommentsCollection();
+    private JavaToken homeToken;
```
ftomassetti, line 167: What is the home token?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
```diff
@@ -88,23 +86,31 @@ static void addMultipleModifier(GeneratedJavaParser generatedJavaParser, EnumSet
             modifiers.add(m);
     }
 
+    static TokenRange range(JavaToken begin, JavaToken end) {
```
ftomassetti, line 89: maybe `tokenRange`, to make clear it does not return a `Range`?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
```diff
@@ -88,23 +86,31 @@ static void addMultipleModifier(GeneratedJavaParser generatedJavaParser, EnumSet
...
+        return new TokenRange(begin, end);
+    }
+
+    static TokenRange range(Node begin, Node end) {
```
ftomassetti, line 93: same as above

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public JavaToken getBegin() {
+        return begin;
+    }
+
+    public JavaToken getEnd() {
+        return end;
+    }
+
+    public Range getRange() {
```
ftomassetti, line 27: what about `toRange`? It is more of a conversion than extracting a value. Just a thought, not feeling strongly about this at all

ftomassetti (approved):
It looks good.
All of my comments are really minor. Considering that and the fact that this is a large patch I would go on and merge it. If later we decide to act on some comments we can do follow up PRs. The only thing is that we decide to go for any renaming we should do this before releasing the next version.

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
matozoid: I do feel strongly about naming, thanks :)

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
matozoid: That feels like it has its own set of tokens, where it really is just two pointers indicating a range in the full list of tokens.

Actually, I would say that adding Range was a mistake and it should have been this implementation from the start - so this should become the new Range :)

javaparser-core/src/main/java/com/github/javaparser/Problem.java
matozoid: Indeed, we now have...

tokenRange.begin.range.begin
tokenRange.begin.range.end
tokenRange.end.range.begin
tokenRange.end.range.end

For now the normal Range is still there.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
matozoid: Yes actually, there is always a stream inside somehow :) I get your point though.

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
matozoid: You know what? I'll leave it in beta for a release or two so we can discuss naming.

javaparser-core/src/main/java/com/github/javaparser/ast/type/TypeParameter.java
matozoid: This is a utility constructor for the grammar, I made it pick the usual longer constructor instead.

Why no range in the normal constructors? Because it seems to be a marginal use case to me, shuffling nodes around. Am I wrong?

javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java
matozoid: :-D

javaparser-core/src/main/javacc/java.jj
matozoid: The very first token in the file. Should I document it?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
matozoid: That would make the grammar more verbose - this method is called in every `new` :-O I think I should be able to clean up some stuff here, but that's for another day.

Issue #948, opened 2017-06-03
I guess LamdaExpr's Range is wrong


Yangseon-Ryu:
I am using javaparser 3.1.3. and I make source code line counting using Node class's Range.

just run javaparser to parsing javaparser 3.1.3's code, 

then in com.github.javaparser.CommentsInserter's Line 102~107

```java
            commentsInsideChild.addAll(
                    commentsToAttribute.stream()
                            .filter(c -> c.getRange().isPresent())
                            .filter(c -> PositionUtils.nodeContains(child, c,
                                    configuration.isDoNotConsiderAnnotationsAsNodeStartForCodeAttribution()))
                            		.collect(Collectors.toList()));
```
- first filter argument "c -> c.getRange().isPresent()" com.github.javaparser.ast.expr.LambdaExpr's  getRange - Optional[(line 104,col 37)-(line 104,col 37)]
- second filter argument "c -> PositionUtils.nodeContains(child, c,                              configuration.isDoNotConsiderAnnotationsAsNodeStartForCodeAttribution())" com.github.javaparser.ast.expr.LambdaExpr's getRange - Optional[(line 105,col 37)-(line 105,col 37)]

Is it right?

matozoid:
Indeed!

```java
import com.github.javaparser.TokenRange;
import com.github.javaparser.ast.expr.Expression;
import static com.github.javaparser.JavaParser.parseExpression;

public class Issue948 {
    public static void main(String[] args) {
        Expression expression = parseExpression("x -> y");
        System.out.println(expression.getRange().get());
        TokenRange tokenRange = expression.getTokenRange().get();
        System.out.println(tokenRange.getBegin() +"-"+ tokenRange.getEnd());
    }
}
```

This shows that the range is the range of the variable name, not the whole expression. Will look into this.

matozoid:
Okay, fixed. It will be in tomorrow's release.

PR #949, opened 2017-06-03
Fix bug in lambda expression range (also remove some unused code)

matozoid:
See #948


matozoid:
@ftomassetti - this may be of interest to you, since you're a huge user of range info.

Issue #589, opened 2016-11-23
Make Compilation Unit parse errors recoverable


matozoid:
Make a recovery strategy that skips to the end of the file, storing the whole file contents in a BrokenCompilationUnit or so. The result of this would be that parsing will always have a result.

Exceptions: when parsing a file and the file is not found, there is nothing to parse.

tamtam22:
When analyzing some java files (e.g. from an APK) which are decompiled wrongly, it throws ParseProblemException, which is expected as it didn't fit any known tokens/syntax. e.g.

~~~~
if (this.down.getDownloadUrl().contains("url")) {
    /* anonymous class already generated */.start();
} else {
    /* anonymous class already generated */.start();
}
~~~~

To overcome this, I put a catch on ParseProblemException in try-catch block as I still want to parse my other java files. However, I believe this is definitely not the fix. Even though the problematic java files are caught in ParseProblemException and the catch block does nothing, at least it doesn't stop the program from terminating nor stop reading other java files. 

However, it later encountered StackOverflowError 

~~~~
Exception in thread "main" java.lang.StackOverflowError
	at com.github.javaparser.printer.PrettyPrintVisitor.visit(PrettyPrintVisitor.java:52)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.accept(ClassOrInterfaceType.java:89)
	at com.github.javaparser.printer.PrettyPrintVisitor.visit(PrettyPrintVisitor.java:275)
	at com.github.javaparser.printer.PrettyPrintVisitor.visit(PrettyPrintVisitor.java:52)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.accept(ClassOrInterfaceType.java:89)
	at com.github.javaparser.printer.PrettyPrintVisitor.visit(PrettyPrintVisitor.java:275)
	at com.github.javaparser.printer.PrettyPrintVisitor.visit(PrettyPrintVisitor.java:52)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.accept(ClassOrInterfaceType.java:89)
.....
~~~~

So i think it would be great if parsing for this file can be skipped correctly and sacrifice parsing this file rather than terminating half-way through

matozoid:
This is noted, and recovery is going up in importance as you are the second user to explicitly request it.

@jeandersonbc - will you find time for this soon? Otherwise I'll start on it.

For the stack overflow: can you send the source file that causes this? It should not happen (well, only in extreme cases.)

[...]

PR #952, opened 2017-06-03
Recover from errors at CU & Statement level

matozoid:
See #589 and #581.

Here's the implementation of the lamest version of recovering from parse errors: handle the exception at the very top - the compilation unit. It recovers from any error by moving forward to the end of the file. The CompilationUnit it produces has defaults for all fields, and a new `isBad` flag is set.

For discussion:
- currently I'm just setting a flag, but I actually *know* which error happened to get this flag set. I think I should store the error (the `Problem`) right here in the tree. That would be a beautiful thing, but would also cause another `List` to appear in every `Node`. It would be beautiful because you can directly ask a node for its problems, instead of the situation we have now: a list of problems with no clear link to the AST. (Another solution would be to link every `Problem` to an `Optional<Node>`... Hmmm, that seems nice too...)
- currently the `isBad` flag is only on `CompilationUnit`. When I keep going this way, the flag will be on `Statement` and uh, one other node type. I think it would be a good idea to simply move it to `Node`, since all nodes can theoretically be recovered.

New thoughts:
- I'll put the flag at `Node` level so you can traverse the AST and look for the bad nodes.
- I've added #955 to make the parse problems point to the nodes that can't parse.
- Recoverting `Statement` is different from recovering `CompilationUnit` since `Statement` is not a leaf node - it is abstract. And I don't know which child of `Statement` to instantiate because there was a parse error. So there I am making a new node type: `UnparsableStatement` which always has the unparsable flag set.

More thoughts:
- recovering a statement by skipping to ";" is easy. Recovering a statement by skipping to "}" is harder, since this token does not belong to the statement. This PR only does the easy thing.


javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -90,23 +90,27 @@
     @InternalProperty
     private Storage storage;
 
+    @InternalProperty
+    private boolean isBad;
+
     public CompilationUnit() {
-        this(null, null, new NodeList<>(), new NodeList<>(), null);
+        this(null, false, null, new NodeList<>(), new NodeList<>(), null);
```
ftomassetti: for the boolean parameter could be useful to use enums, so that it is more clear what it means

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -547,6 +551,11 @@ public CompilationUnit setStorage(Path path) {
         return this;
     }
 
+    @Override
+    public boolean isBad() {
```
ftomassetti: what about "containsParsingErrors" or something more explicit?

pom.xml
```diff
@@ -3,8 +3,8 @@
     <modules>
         <module>javaparser-core</module>
         <module>javaparser-testing</module>
-        <module>javaparser-core-generators</module>
-        <module>javaparser-metamodel-generator</module>
+        <!--<module>javaparser-core-generators</module>-->
```
ftomassetti: should we just remove them?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
```diff
@@ -148,4 +144,63 @@ static JavaToken nodeListBegin(NodeList<?> l) {
...
+     * This is the code from ParseException.initialise, modified to be more horizontal.
+     */
+    static String makeMessageForParseException(ParseException exception) {
+        final StringBuilder sb = new StringBuilder("Parse error. Found ");
+        final StringBuilder expected = new StringBuilder();
+
+        int maxExpectedTokenSequenceLength = 0;
+        TreeSet<String> sortedOptions = new TreeSet<>();
+        for (int i = 0; i < exception.expectedTokenSequences.length; i++) {
+            if (maxExpectedTokenSequenceLength < exception.expectedTokenSequences[i].length) {
```
ftomassetti, line 140: maybe you could use `Math.min` here

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
```diff
@@ -148,4 +144,63 @@ static JavaToken nodeListBegin(NodeList<?> l) {
...
+                maxExpectedTokenSequenceLength = exception.expectedTokenSequences[i].length;
+            }
+            for (int j = 0; j < exception.expectedTokenSequences[i].length; j++) {
```
ftomassetti, line 143: Could you use a for each here?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
```diff
@@ -148,4 +144,63 @@ static JavaToken nodeListBegin(NodeList<?> l) {
...
+            }
+        }
+
+        for (String option : sortedOptions) {
+            expected.append(" ").append(option);
+        }
+
+        sb.append("");
+
+        Token token = exception.currentToken.next;
```
ftomassetti, line 154: Maybe you could add some comment here because it is starting to be a long method

ftomassetti (commented):
It seems in order to me, I just added a few comments. However I confess I am out of my depth here.

Maybe we could add a few more tests to better illustrate the feature

pom.xml
matozoid: I'll consider it. I think it's doing more harm than good to have them here.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
matozoid: Sure, could you take a look at the points in the PR description?

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
matozoid: I could, but should I? This is copied from the JavaCC source. Maybe it should be an issue on their project.

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
matozoid: "I could, but should I? This is copied from the JavaCC source. Maybe it should be an issue on their project."

javaparser-core/src/main/javacc_support/com/github/javaparser/GeneratedJavaParserSupport.java
matozoid: "I could, but should I? This is copied from the JavaCC source. Maybe it should be an issue on their project."

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
matozoid: Okay, let's see if we go with a boolean or a list first.

matozoid: Enum is being added.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
matozoid: It's `getParsed` -> PARSED/UNPARSABLE now.

matozoid:
@martinlippert - if you're interested in influencing the API, here's your chance :)

martinlippert:
@matozoid thanks for the update, much appreciated. Great to see the progress in this area, really great. Unfortunately I am heads down in other work for our upcoming release at the moment, so don't have a chance to take a look right away. But will take a closer look again once I am back to work on this area.

revusky:
Hi Danny, 

I ran across this 3-year-old conversation and thought to say some things... first of all, I don't think you can get very far on this problem with legacy JavaCC. For one thing, Legacy JavaCC has no concept of rewinding back to a previous point in the parse, say, so your range of strategies for dealing with error recovery are pretty darned limited. (Slim and none, as they say...) 

 I am trying to resuscitate that old JavaCC dinosaur as an actively developed project, [JavaCC 21](https://github.com/javacc21/javacc21 )and have been working on providing some machinery to deal with these sorts of problems. Here is something I wrote about this topic, just a couple of weeks ago: https://javacc.com/2020/04/23/fault-tolerant-parsing-progress/
And then there is a new feature that I added as a result of my fault-tolerant/error recovery sort of work:
https://javacc.com/2020/05/03/new-experimental-feature-attempt-recover/

I would be quite happy if you dropped me a note at revusky@javaccNOSPAM.com since I am mostly an email guy.

MysterAitch:
Hi @revusky , I've opened #2659 as discussion here in a closed PR is likely to drop off the radar.

It sounds like there's likely to be some benefits to switching, but it's not something I fully understand at the moment.

Issue #959, opened 2017-06-14
Version constant


ftomassetti:
Many users have issues with JSS because they use the wrong version of JP. Maybe we could have a check if we had a constant in JP telling us which version of JP we are using. On the JSS side we could use a static initializer to throw an exception if needed.

What do you guys think?

matozoid:
Do you know of any maven tools to do this easily?

matozoid:
Personally I think it's a beginner's mistake that shouldn't be fixed with more complexity on our side - I have no idea why suddenly this minor change triggered three issues though :-/

ftomassetti:
Beginner mistakes have a cost on our time :)
It is also something surprising for more people that they do not need to use JavaParser when using JavaSymbolSolver.
Maybe it could be solved at the maven/gradle level checking if a dependency appear twice, it could be worth looking into that.

PR #965, opened 2017-06-17
Add build information so java-symbol-solver can detect a JP version mismatch

matozoid:
See #959 

It generates this file:

```java

/**
 * Some information that was available when this library was built by Maven.
 */
public class JavaParserBuild {
    public static final String PROJECT_VERSION = "3.2.8-SNAPSHOT";
    public static final String PROJECT_NAME = "javaparser-core";
    public static final String PROJECT_BUILD_FINAL_NAME = "javaparser-core-3.2.8-SNAPSHOT";
    public static final String MAVEN_VERSION = "3.3.9";
    public static final String MAVEN_BUILD_VERSION = "Apache Maven 3.3.9 (bb52d8502b132ec0a5a3f4c09453c07478323dc5; 2015-11-10T17:41:47+01:00)";
    public static final String MAVEN_BUILD_TIMESTAMP = "2017-06-17T14:38:52Z";
    public static final String JAVA_VENDOR ="Oracle Corporation";
    public static final String JAVA_VENDOR_URL ="http://java.oracle.com/";
    public static final String JAVA_VERSION ="1.8.0_131";
    public static final String OS_ARCH ="amd64";
    public static final String OS_NAME ="Linux";
    public static final String OS_VERSION ="4.4.0-72-generic";
}
```


matozoid:
So now you all know what OS I'm running :)

DeepSnowNeeL:
Why not using the meta-inf/manifest.mf with implementation version,
buildtime and all that stuff ? It might be more "standard" ? :)

matozoid:
@DeepSnowNeeL - yeah, I guess @ftomassetti could read `/META-INF/maven/com.github.javaparser/javaparser-core/pom.properties`. That would require no change in JP, but everyone who wants to do this would have to write the same bit of code. Let's see what he thinks.

ftomassetti:
I think I would love to have a way to understand if the version of JP being used is the one I expect in JSS. I would think to do it using a static initializer like:

```
static {
    if (JavaParser.getVersion().equals("3.5.7")) {
        throw new RuntimeException("Hey, check your classpath! You seem to have the wrong JP version");
    }
}
```

Is there a way I could be able to easily read the value from manifest.mf at runtime?

matozoid:
Well, the value is in that properties file I mention, so it would be "open that properties file from that exact path and read a specific key". Just take a look in the JP jar.

matozoid:
Meh, I'll merge this. It's not like it's a major change or anything.

PR #974, opened 2017-06-23
Added documentation to the CompilationUnit.Storage#save and changed l…

Bohdankm22:
…amda expression

Lambda expression has to work the same way because in the save(Function<CompilationUnit, String> makeOutput) method the same argument is passed to the function makeOutput.


Bohdankm22:
Not sure why one of CI builds was failed. There are no code changes (only refactored lambda expression). The pull request is about the adding java doc on one of methods.

matozoid:
Thanks, good job!

Appveyor has been broken for a long time - something with end-of-line characters breaking tests on Windows. It's not caused by this PR.

Issue #931, opened 2017-05-18
Add Node.replace method


ftomassetti:
Followint a question I have received I realized that it would be nice to have a way to replace a given element (for example a statement or an expression).

This would work similarly to remove because it is about setting the node in the parent.

matozoid:
I guess this is rather easy to do, since you just have to copy a generator and adjust it a bit

PR #975, opened 2017-06-25
Add replace method generator and its results

matozoid:
See #931


matozoid:
Bit late for a release now. Maybe tomorrow.

Issue #979, opened 2017-06-26
FieldDeclaration with multiple VariableDeclarators: Type of Declarator may point to wrong parent.


Gottox:
Testcode:
```java
        CompilationUnit compilationUnit = parse("" +
                "class A {\n" +
                "    int a, b;\n" +
                "}");

        BodyDeclaration<?> declaration = compilationUnit.getType(0).getMembers().get(0);
        FieldDeclaration fieldDeclaration = (FieldDeclaration) declaration;
        VariableDeclarator var1 = fieldDeclaration.getVariables().get(0);
        VariableDeclarator var2 = fieldDeclaration.getVariables().get(1);
        assertEquals(var1, var1.getType().getParentNode());
        assertEquals(var2, var2.getType().getParentNode());
```
AssertionError on the first `assertEquals`:
```
java.lang.AssertionError: 
Expected :a
Actual   :b
```

matozoid:
Ah yes, the type node is reused for all variables, so first it is the type of `a` so it has parent `a`, then `b` is initialized and its type is set to exact the same type, but we now set its parent to `b` and, being the same object, parent `a` is overwritten... Oops!

PR #981, opened 2017-06-27
Make sure types are not reused

matozoid:
Also fix WildcardType losing main annotations when cloning.

See #979


ftomassetti:
This one seems to be out of date and the build fails

ftomassetti:
Aside from that it looks good

PR #982, opened 2017-06-28
Make scripts to run the generator phases instead of running them automatically

matozoid:
By popular request :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,8 +3,6 @@
     <modules>
         <module>javaparser-core</module>
         <module>javaparser-testing</module>
-        <module>javaparser-core-generators</module>
-        <module>javaparser-metamodel-generator</module>
     </modules>
 
     <groupId>com.github.javaparser</groupId>
--- a/run_core_generators.sh
+++ b/run_core_generators.sh
@@ -3,13 +3,18 @@
 # Runs all the code generators.
 # If the node structure was changed, run_metamodel_generator.sh first!
 
+# Remember current directory
 pushd javaparser-core-generators
 
+# Generate code
 mvn clean package
 
+# Go back to previous directory
 popd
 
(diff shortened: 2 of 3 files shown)


run_core_generators.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
```
ftomassetti, line 1: What about windows?

run_core_generators.sh
```diff
@@ -0,0 +1,15 @@
...
+
+# Runs all the code generators.
+# If the node structure was changed, run_metamodel_generator.sh first!
+
+pushd javaparser-core-generators
```
ftomassetti: Maybe pushd and popd are not so commonly used and they could deserve a comment

ftomassetti (commented):
Also, should we add some comment in the documentation?

pom.xml
```diff
@@ -3,8 +3,6 @@
     <modules>
         <module>javaparser-core</module>
         <module>javaparser-testing</module>
-        <module>javaparser-core-generators</module>
```
ftomassetti, line 6: Ok, we do not want to distribute the generators but will IDE consider this modules if we remove them from the POM?

run_core_generators.sh
matozoid: It's best if a Windows user contributes those, I think.

pom.xml
matozoid: Well, they are not really modules anymore. They are now stand-alone projects that use the `javaparser-parent` module as their parent. So I guess in intellij for example, you would have to treat them as separate projects.

run_core_generators.sh
matozoid: I documented the whole thing now.

Issue #992, opened 2017-07-08
The parent of the javaparser-core-generators is not updated


ftomassetti:
It is 3.2.9-SNAPSHOT. Idealy we should change it automatically

matozoid:
Hmmm, what I can do is make them part of the build again, but take out the part where they are run during the build. Then we do the running from the scripts, or with a special profile or so.

matozoid:
Let's not tag stuff "easy" that we don't want new people to do :-D

ftomassetti:
Right, I will use the tag easy more carefully

PR #993, opened 2017-07-09
Issue 992 module redo

matozoid:
See #992 

All modules are part of the build again.

The code generation steps are now only enabled when their maven profiles are enabled.

Result: parent and modules are seen as one and stay consistent (no version number deviations, no forgetting to build loose modules) and no confusion when code is being generated in odd places while building.


ftomassetti:
We should probably document this somewhere, also for my old-man very limited memory :)

matozoid:
This is documented on https://matozoid.github.io/2017/04/27/code-generation-steps.html and in the two scripts. I can document more, but please point out where :-)

ftomassetti:
I would just add a pointer to that resource in the README

Issue #988, opened 2017-07-03
Deal with the updated parsing method for "require transitive"


matozoid:
"A further ten character sequences are restricted keywords: open, module, requires, transitive, exports, opens, to, uses, provides, and with. These character sequences are tokenized as keywords solely where they appear as terminals in the ModuleDeclaration and ModuleDirective productions (§7.7). They are tokenized as identifiers everywhere else, for compatibility with programs written prior to Java SE 9. There is one exception: immediately to the right of the character sequence requires in the ModuleDirective production, the character sequence transitive is tokenized as a keyword unless it is followed by a separator, in which case it is tokenized as an identifier."

PR #994, opened 2017-07-09
Add workaround for JPMS edge case

matozoid:
"requires transitive;" now requires a module called transitive. Which will never exist. But it needs exceptions all over the parser anyway. Thanks JPMS.

See #988

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
@@ -35,7 +35,7 @@ public class JavaToken {
     public static final JavaToken INVALID = new JavaToken();
 
     private final Range range;
-    private final int kind;
+    private int kind;
     private final String text;
     private final Optional<JavaToken> previousToken;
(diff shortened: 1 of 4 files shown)


matozoid:
Seriously, why create such an archaic grammar for something as straightforward as this?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -110,6 +110,10 @@ public int getKind() {
         return kind;
     }
 
+    public void setKind(int kind) {
```
ftomassetti: Is it necessary to make the Token mutable? Could we instead instantiate a new token when we need to change the kind?

ftomassetti (approved):

matozoid:
@ftomassetti I tried hard to make `kind` final again, but it turns out that the `JavaToken` is already stored in two places when I change the kind. One was fixable, the other was scary, so in the end I decided to only reduce acces to `package` for the setter.

ftomassetti:
Makes sense

PR #998, opened 2017-07-12
Code cleanups in CSM

matozoid:
I ran all of intellij's inspections on core, and it pointed out some CSM stuff first. @ftomassetti please carefully check if I didn't destroy anything!


ftomassetti:
It seems ok, I am thinking about cases in which we would need to old value of nodelists or properties. I think we could have multiple changes that are in progress and not yet saved in the nodes and that was the original thiking. However if they are unused and all tests pass...

matozoid:
@ftomassetti I will absolutely understand if you say "I'm not sure this is a good idea" :-)

Issue #1001, opened 2017-07-12
Inspect Intellij inspections on CSM & lex pres


matozoid:
There are quite a few "condition is always false" "wrong type in collection" and other warnings on this code. Needs a careful inspection!

PR #1006, opened 2017-07-18
Issue 1001: Fixing warning in the printer package

ftomassetti:


matozoid:
See #1001

matozoid:
Hmmm, what I do is select "analyze/inspect code" in Intellij, and run it on the printer package in this branch. This gives me 20 "possible bugs" of which two are "condition ... is always true", one is "equals() between inconvertible objects"  and one "collection may not contain element of type bla"

ftomassetti:
Ok, now they are 4 (in the whole printer package), all of which are not bugs but things I have written in a certain way for clarity

screen shot 2017-07-19 at 17 35 11

ftomassetti:
There are 14 of such warnings in other packages but I did not consider them

matozoid:
That's cool, the plan is not to remove all warnings, only to check what they are about.

Issue #845, opened 2017-03-07
LexicalPreservingPrinter with ModifierVisitor throws IllegalArgumentException


ThLeu:
Hey thanks a lot for getting the lexical preservation going, the work you guys put into the JavaParser is greatly appreciated!

I am currently trying to play around with it and it seems to me that a lot of changes to the code are not yet supported (with Visitors but also with CompilationUnit directly).
Naturally I tried some changes with the ModifierVisitor but that seems to not work at all.
Here is the simplest example I came up with which does not work for me.

```
try (BufferedReader reader = Files.newBufferedReader(someFilePath)) {
   Pair<ParseResult<CompilationUnit>, LexicalPreservingPrinter> result = LexicalPreservingPrinter
      .setup(ParseStart.COMPILATION_UNIT, Providers.provider(reader));
   cu = result.a.getResult().get();
   cu.accept(new ModifierVisitor(), null);
} catch (IOException e) {
   e.printStackTrace();
}
```

I always end up getting the IllegalArgumentException, which I attached.
[stacktrace.txt](https://github.com/javaparser/javaparser/files/825738/stacktrace.txt)

Also, can you say in what state the Lexical Preservation currently is? Are there still a lot of open TODOs you know about or should it mainly be bugfixing? It would be nice to know what to expect and not to expect to work.

Kind regards

matozoid:
I can answer about the state. I try to mention it in the release notes: https://github.com/javaparser/javaparser/blob/master/changelog.md

ThLeu:
Ah yes thanks, I didnt even realise that it was mentioned as in beta but makes sense of course.
Still, from the unit tests I can see that some changes are supported already (via CompilationUnit directly) and I am not sure if I should report unsupported nodes which throw an exception or if it might be too early.

ftomassetti:
hi @ThLeu LexicalPreservation should reasonably work, but as with all new features problems are to be expected: we figured out that the JavaParser community finds always applications that we did not think of.

For examples while using LexicalPreservation I wrote a large number of tests but it never came to me to use the ModifierVisitor. The thing is also that LexicalPreservation use a lot of metaprogramming, so it tends also to be sensible to little inconsistencies. In this case it seems to fail on the method `com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter.findNodeListName` which use reflection and expect methods to have a certain name.

[...]

PR #1007, opened 2017-07-18
Fix 845 and a bunch of issues with ranges

ftomassetti:


matozoid:
See #845 - thanks, looks like it was a horrible debugging session :-)

ftomassetti:
It was not as terrible as it seems, I think I started weeks ago going through the wrong path but today I saw it was about ranges and it was reasonably quick to fix

Issue #1002, opened 2017-07-17
Categorize tokens


matozoid:
We're suggesting more and more that people take a look at the tokens when they want to work with comments. It would be nice if we could ask a token if it is a comment (meaning: it is javadoc, a line comment, or a block comment.) We could do that with a `isComment` method, but the thing we're doing seems to be categorization of tokens. The JLS has some categories already: https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.5

If we do this, we also make syntax highlighting a lot easier to do.

Create a method `JavaToken.getCategory()` that returns an enum of WHITESPACE, COMMENT, IDENTIFIER, KEYWORD, LITERAL, SEPARATOR, OPERATOR (maybe more?) using the `kind` field and consider if subcategorizations are useful.

PR #1008, opened 2017-07-18
Categorize tokens

matozoid:
See #1002


javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -139,10 +139,6 @@ public String toString() {
         return text;
     }
 
-    /**
-     * Check if the position is usable. Does not know what it is pointing at, so it can't check if the position is after
-     * the end of the source.
-     */
     public boolean valid() {
```
ftomassetti: Why removing the comment? Is the position that is not valid or the token itself?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -159,5 +155,19 @@ public JavaToken orIfInvalid(JavaToken anotherToken) {
...
+    public enum Category {
+        WHITESPACE, COMMENT, IDENTIFIER, KEYWORD, LITERAL, SEPARATOR, OPERATOR
+    }
+
+    public JavaToken.Category getCategory() {
+        return TokenTypes.getCategory(kind);
+    }
 
-}
+    public boolean isEndOfLineToken() {
```
ftomassetti: Should end of line be a category? A subcategory?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -159,5 +155,19 @@ public JavaToken orIfInvalid(JavaToken anotherToken) {
...
+        return TokenTypes.isEndOfLineToken(kind);
+    }
+
+    public boolean isSpaceOrTab() {
+        return TokenTypes.isSpaceOrTab(kind);
+    }
+}
```
ftomassetti: newline :)

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -7,12 +7,20 @@
  * Complements GeneratedJavaParserConstants
  */
 public class TokenTypes {
-    public static boolean isWhitespace(int tokenType) {
-        return tokenType >= EOF && tokenType <= ZERO_WIDTH_NO_BREAK_SPACE;
+    public static boolean isWhitespace(int kind) {
```
ftomassetti, line 10: nice

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -22,12 +30,18 @@ public static boolean isEndOfLineCharacter(int tokenType) {
         }
     }
 
-    public static boolean isWhitespaceOrComment(int tokenType) {
-        return isWhitespace(tokenType) || isComment(tokenType);
+    public static boolean isWhitespaceOrComment(int kind) {
+        switch (getCategory(kind)) {
```
ftomassetti: nice. Maybe we can have a `isWhitespaceOrComment` method on the category and change this to:

`return getCategory(kind).isWhitespaceOrComment()`

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -62,13 +76,16 @@ public static boolean isSpaceOrTab(int tokenType) {
         }
     }
 
-    public static boolean isComment(int tokenType) {
-        return tokenType == SINGLE_LINE_COMMENT
-                || tokenType == MULTI_LINE_COMMENT
-                || tokenType == JAVA_DOC_COMMENT;
+    public static boolean isComment(int kind) {
+        return getCategory(kind)== JavaToken.Category.COMMENT;
```
ftomassetti: for the other categories you are using just the name (COMMENT instead of JavaToken.Category.COMMENT), also a space before `==`

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -62,13 +76,16 @@ public static boolean isSpaceOrTab(int tokenType) {
...
     }
 
+    /** @deprecated use eolTokenKind */
     public static int eolToken() {
```
ftomassetti: Deprecated annotation

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -81,7 +98,193 @@ public static int eolToken() {
         throw new AssertionError("Unknown EOL character sequence");
     }
 
-    public static int spaceToken() {
+    public static int spaceTokenKind() {
         return SPACE;
     }
+
+    /** @deprecated use spaceTokenKind */
+    public static int spaceToken() {
```
ftomassetti: Deprecated annotation

ftomassetti (approved):
Looks good, just some nitpick comments

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: Oh, this was a copy/paste error from `Range`, I think. I really should write the correct documentation.

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: According to the JLS it's part of whitespace (although it's not even a token according to the JLS) - I was thinking about subcategories but there is not much to subcategorize. Maybe have a `JlsCategory` and a `FineGrainedCategory` ?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: This is valid Java, we can parser it :-)

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
matozoid: Oh yeah, nice!

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
matozoid: The people who decided that that annotation cannot have a reason field should be spanked.

matozoid:
@ftomassetti thanks to your comments the category enum is now more detailed, and is has getters for the more general categories. For completeness I implemented getters for every possible category.

ftomassetti:
Nice, thank you for considering them

Issue #1013, opened 2017-07-21
Add to ObjectCreationExpr Javadoc what "scope" is for


matozoid:
```java
class B { class C { public void a() { new B().new C(); } } }
```

`new B()` is the scope

PR #1015, opened 2017-07-21
Fix bad Javadoc

matozoid:
See #1013

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/ObjectCreationExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/ObjectCreationExpr.java
@@ -44,9 +44,12 @@ import javax.annotation.Generated;
 import com.github.javaparser.TokenRange;
 
 /**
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/ObjectCreationExpr.java
```diff
@@ -44,9 +44,12 @@
 import com.github.javaparser.TokenRange;
 
 /**
- * A constructor call. <br/>In <code>new HashMap.Entry<String, Long>(15) {public String getKey() {return null;}};</code>
- * HashMap is the scope, Entry is the type, String and Long are type arguments, 15 is an argument, and everything in { }
+ * A constructor call.
+ * <br/>In <code>new HashMap.Entry&lt;String, Long>(15) {public String getKey() {return null;}};</code>
```
ftomassetti, line 48: should > be &gt; ? (after the word `Long` and before `(15)` )

matozoid: Can be, but HTML is fine with this. Keeps it a little more readable for people reading it from their IDE.

Issue #1024, opened 2017-07-26
Grammar for float literals


ftomassetti:
Why is this expressed in this way?
```
  < #DECIMAL_FLOATING_POINT_LITERAL:
        (<DECIMAL_LITERAL>)+ "." (<DECIMAL_LITERAL>)* (<DECIMAL_EXPONENT>)? (["f","F","d","D"])?
      | "." (<DECIMAL_LITERAL>)+ (<DECIMAL_EXPONENT>)? (["f","F","d","D"])?
      | (<DECIMAL_LITERAL>)+ <DECIMAL_EXPONENT> (["f","F","d","D"])?
      | (<DECIMAL_LITERAL>)+ (<DECIMAL_EXPONENT>)? ["f","F","d","D"]
  >
```

I do not understand the "+" and "*" followin the DECIMAL_LITERAL, because DECIMAL_LITERAL should represent already a sequence of digits, not just one digit.

matozoid:
Stop it with all the grammar bugs already :-D

ftomassetti:
As soon as you stop with the comments bugs :D
Anyway I am not sure it is really a bug because it works anyway, but I thought it was the case to look into this and clarify

PR #1025, opened 2017-07-26
Simplify the float/double grammar

matozoid:
See #1024 

Simply did what @ftomassetti indicated, which seems logical, and no tests fail.


ftomassetti (approved):
Looks fine to me and more importantly to the tests

PR #1026, opened 2017-07-26
Improve Sourceprinter performance and allow to ignore javadoc

arturbosch:


javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrintVisitorTest.java
```diff
@@ -130,4 +130,13 @@ public void printLambdaIntersectionTypeReturn() {
...
     }
+
+    @Test
+    public void printClassWithoutJavaDocButWithComment() {
+        String code = "/** javadoc */ public class A { \n// stuff\n}";
+        CompilationUnit cu = JavaParser.parse(code);
+        PrettyPrinterConfiguration ignoreJavaDoc = new PrettyPrinterConfiguration().setPrintJavaDoc(false);
+        String content = cu.toString(ignoreJavaDoc);
+        assertEquals("public class A {\n    // stuff\n}\n", content);
+    }
```
matozoid: Use the Utils.EOL constant instead of \n so the test doesn't fail on Windows :-(

PR #1028, opened 2017-08-01
Extended LexicalPreservingPrinter to deal with final and abstract modifiers on a class.

glimmerveen:
When using javaparser's LexicalPreservingPrinter I noticed that adding an annotation to a class that has either a final or abstract modifier would result in an UnsupportedOperationException.

I have extended the toToken method to handle these modifiers as well and added two basic tests to verify if it works. All other tests work as well, but I don't have the overview to see possible negative consequences of this patch.


ftomassetti:
It looks good to me, thank you @glimmerveen for this PR!
I will go on and merge it since @matozoid is in vacation

PR #1033, opened 2017-08-10
[DRAFT] Issue 1005 cursor api

matozoid:
This thing is slowly taking shape. I'd like to know if this API looks okay.

One thing that is currently badly designed is dealing with the start and end of the token list: especially the find functions simply fall off the begin or end of the list, rendering the cursor useless. I'm wondering if this is okay (just make a clone of the cursor before searching if you care about the current location) or if I should use some kind of sentinel token there, or maybe have a "reset" method or whatever.

Oh, what's the point of this thing:
- enabling reformatting of the source code behind the AST
- doing more challenging work with comments, inserting them at funny places etc.

It will probably stay in beta for a while after it gets into a release, it's just too hard to think up a whole new API correctly at the first attempt.

### DIFF
--- a/javaparser-core/src/main/java/Issue1029.java
+++ b/javaparser-core/src/main/java/Issue1029.java
@@ -1,0 +1,38 @@
+import com.github.javaparser.GeneratedJavaParserConstants;
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.JavaToken;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.body.BodyDeclaration;
+import com.github.javaparser.ast.visitor.TreeVisitor;
+
+public class Issue1029 {
+    public static void main(String[] args) {
+        BodyDeclaration<?> mainMethod = JavaParser.parseBodyDeclaration("int main(){\n" +
+                "          return 0;\n" +
+                "     }");
+
+        new TreeVisitor() {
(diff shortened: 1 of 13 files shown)


matozoid:
Also, @ftomassetti , is directly manipulating the token list going to impact lexical preservation? What should we tell users?
- when you want to modify the AST without changing formatting, use lex pres
- when you want to modify the source code but do not change the AST, modify the token list
- when you want both, tough luck
?

ftomassetti:
Well, lexical preservation in practice is based on having a `NodeText` for each node. `NodeText` is a list of `TextElement`. Each `TextElement` represents either a token or the position of a child node. When we modify the Node the changes are reflected in the `NodeText` using observers. So if you manipulate tokens directly the lexical preservation would not be aware of it. Maybe we need just one way of storing tokens, like using NodeText or the list of tokens in all situations.

matozoid:
Hmmmm, do you think I should build this cursor API on top of the lexical preservatin code then? Or maybe build a second API? :)

ftomassetti:
I think we should have only one data structure that keeps the updated relation between a node and its tokens. We could use the NodeText we have in LPP or something else, but I think we should unify that.

matozoid:
It would make sense to use the LPP structure. I'll drop this token list cursor idea.

PR #1034, opened 2017-08-12
Use tokenrange to construct lex pres data

matozoid:


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -65,6 +67,7 @@
 
     /**
      * Parse the code and setup the LexicalPreservingPrinter.
+     * @deprecated just use the other constructor.
```
matozoid, line 70: @ftomassetti - I can't see a good reason to have this constructor, is it okay to deprecate?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -201,7 +205,7 @@ public void process(Node node) {
 
         // We go over tokens and find to which nodes belong. Note that we start from the most specific nodes
         // and we move up to more general nodes
-        for (JavaToken token : documentTokens) {
+        for (JavaToken token : root.getTokenRange().get()) {
```
matozoid, line 208: @ftomassetti Well, having token ranges didn't help much to reduce the code here :-)

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/TokenTextElement.java
```diff
@@ -21,57 +21,41 @@
...
 
 class TokenTextElement extends TextElement {
-
-    private final int tokenKind;
-    private final String text;
-
-    public static TokenTextElement newLine() {
-        return new TokenTextElement(TokenTypes.eolTokenKind(), EOL);
-    }
+    private final JavaToken token;
```
matozoid, line 29: I'll directly wrap a JavaToken here, so the lex pres code gets a little more connected to the rest.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/TokenTextElement.java
```diff
@@ -81,27 +65,25 @@ public boolean equals(Object o) {
...
-        return result;
+        return token.hashCode();
     }
 
     @Override
     public String toString() {
-        return "TokenTextElement(" + tokenKind +
-                ") {" + text + '}';
+        return "TokenTextElement(" + token.getKind() +
+                ") {" + getText() + '}';
```
matozoid, line 81: @ftomassetti I could use JavaToken's toString here - are you very attached to this representation?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
ftomassetti: Well, until now it was the easiest way to setup LPP. The user would call this method and get the AST with LPP enabled. How the user should get the same result? Should he parse the code and then call `new LexicalPreservingPrinter(parseResult)`?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
ftomassetti: Well, now we do not need a ParseResult anymore, just the AST is enough to enable LPP, right?

javaparser-core/src/main/java/com/github/javaparser/TokenRange.java
```diff
@@ -1,13 +1,14 @@
...
+import java.util.Iterator;
 import java.util.Optional;
 
 import static com.github.javaparser.utils.Utils.assertNotNull;
 
 /**
  * The range of tokens covered by this node.
  */
-public class TokenRange {
+public class TokenRange implements Iterable<JavaToken> {
```
ftomassetti, line 11: Nice

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
ftomassetti: Oh, ok, with the new changes the Node is enough to start LPP, right

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/TokenTextElement.java
ftomassetti: ok

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/TokenTextElement.java
ftomassetti: I am ok with changing it, if it shows the token kind

ftomassetti (approved):

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
matozoid: Uhuh, we might even refactor this code into the node constructors ("I know my start token, my end token, and the start and end token of my children, so I can deduce which tokens are mine")

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
matozoid: Yes, that is an advantage.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/TokenTextElement.java
matozoid: "Ok, that's great!" or "ok, if you insist..." ?

ftomassetti: The first one: it is great!

PR #1036, opened 2017-08-12
Make a token list transformable into a TokenRange

matozoid:


matozoid:
Not much going on here

PR #1058, opened 2017-08-15
Make final usage consistent on nodes

matozoid:
Those asking for the reduction of finals are not going to like this :-) Anyway, we were not being consistent here, so now we are.


ftomassetti:
+1

PR #1059, opened 2017-08-16
Talk explicitly about top level types.

matozoid:


jimshowalter:
That fixes the confusion.

Issue #1061, opened 2017-08-17
Make equals/hashcode configurable


lodart:
Hi ! When parsing this portion of code : 
```java
public class Test {
    String a = "google";

    public void test() throws MalformedURLException {
        new URL(a);
        String a = "yahoo";
        new URL(a);
    }
}
```
I realised both `NameExpr` `a` share the same HashCode, so I can't use them as a key in a HashMap. `.equals` returns also true. Is this normal behaviour ?

Thank you :)

matozoid:
I wanted to say "yes, that is expected behaviour, but you can change it by changing the visitors used by `Node`" - but it looks like that wasn't implemented yet. Putting it on the backlog for next release!

lodart:
Thank you :)

matozoid:
Oh, I've been thinking this over for a bit: when I do what I originally planned, things will break horribly. A lot of code depends on the default behaviour of equals and hashcode in `Node`. So I'm going to see if I can make a `VisitorMap` :-)

matozoid:
@lodart would it be possible for you to take the source code from this repo: https://github.com/matozoid/javaparser.git

branch `visitor_map`

and see if this works for you?

lodart:
Unfortunately, I don't think it has worked :/ [Here's what I got](https://lut.im/vSlEhtcpgw/gCdfvZLDoz71sqph). I must have done something wrong, though. I struggled adding the library to IntelliJ (I'm used to just copying Maven dependencies :) ).
So I cloned the repo (the branch `visitor_map`), I did `mvn clean install`, then I imported the module and added the module dependency. Then, [I removed some stuff that didn't compile](https://lut.im/IbPBA8iWw3/Hcx9zZjmzVYh9tFd) and here I am ...

lodart:
Oh I didn't see #1062 ! It works nicely :) Thank you so much !

matozoid:
Do use Maven or Gradle or so. Here's an example setup: https://github.com/matozoid/java-logic-positivizer

PR #1062, opened 2017-08-17
Visitor map

matozoid:
Allows overriding hashcode and equals for nodes used as keys in a `Map` by wrapping the map in a special map that wraps all keys in a wrapper node that catches `hashcode()` and `equals()` and sends to the requested visitor.

It sounds a lot more complicated than it is.

### DIFF
--- a/javaparser-testing/src/test/java/com/github/javaparser/utils/VisitorMapTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/utils/VisitorMapTest.java
@@ -1,0 +1,36 @@
+package com.github.javaparser.utils;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.visitor.ObjectIdentityEqualsVisitor;
+import com.github.javaparser.ast.visitor.ObjectIdentityHashCodeVisitor;
+import org.junit.Test;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class VisitorMapTest {
+    @Test
+    public void normalEqualsDoesDeepCompare() {
+        CompilationUnit x1 = JavaParser.parse("class X{}");
+        CompilationUnit x2 = JavaParser.parse("class X{}");
+
+        Map<CompilationUnit, Integer> normalMap = new HashMap<>();
+        normalMap.put(x1, 1);
+        normalMap.put(x2, 2);
+        assertEquals(1, normalMap.size());
+    }
+
+    @Test
+    public void objectIdentityEqualsDoesShallowCompare() {
+        CompilationUnit x1 = JavaParser.parse("class X{}");
+        CompilationUnit x2 = JavaParser.parse("class X{}");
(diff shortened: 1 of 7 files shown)


javaparser-testing/src/test/java/com/github/javaparser/utils/VisitorMapTest.java
```diff
@@ -0,0 +1,36 @@
...
+public class VisitorMapTest {
+    @Test
+    public void normalEqualsDoesDeepCompare() {
+        CompilationUnit x1 = JavaParser.parse("class X{}");
+        CompilationUnit x2 = JavaParser.parse("class X{}");
+
+        Map<CompilationUnit, Integer> normalMap = new HashMap<>();
+        normalMap.put(x1, 1);
+        normalMap.put(x2, 2);
+        assertEquals(1, normalMap.size());
```
matozoid, line 23: @lodart - when I use a normal map, a node with the same structure is `equals` and gets the same `hashcode` so only one can be used as a map key.

javaparser-testing/src/test/java/com/github/javaparser/utils/VisitorMapTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Test
+    public void objectIdentityEqualsDoesShallowCompare() {
+        CompilationUnit x1 = JavaParser.parse("class X{}");
+        CompilationUnit x2 = JavaParser.parse("class X{}");
+
+        Map<CompilationUnit, Integer> normalMap = new VisitorMap<>(new HashMap<>(), new ObjectIdentityHashCodeVisitor(), new ObjectIdentityEqualsVisitor());
+        normalMap.put(x1, 1);
+        normalMap.put(x2, 2);
+        assertEquals(2, normalMap.size());
```
matozoid, line 34: @lodart when I wrap the map in the `VisitorMap` and give it visitors that calculate everything on object identity (`==`) then `x1` and `x2` are seen as different, and we get two entries in the map.

matozoid:
See #1061

lodart:
I might not have tested it enough, because I think I found a bug. 
Before, different objects were considered the same key, but thanks to your  VisitorMap, @matozoid, this is no longer the case. The problem is that now, even the same objects produces several keys ! 
If I take your Test example : 
```java
CompilationUnit x1 = JavaParser.parse("class X{}");
Map<CompilationUnit, Integer> normalMap = new VisitorMap<>(new HashMap<>(), new ObjectIdentityHashCodeVisitor(), new ObjectIdentityEqualsVisitor());
normalMap.put(x1, 1);
normalMap.put(x1, 2);
System.out.println(x1 == x1);
System.out.println(normalMap.size());
System.out.println(normalMap.get(x1));
```
it returns : 
```
true
2
null
```
Did I miss something ? Do _I_ have to implement an `ObjectIdentityHashCodeVisitor` and a `ObjectIdentityEqualsVisitor` ?

matozoid:
Hmm, that does sound like a bug yes - can you move this to a separate issue?

lodart:
Sure !

Issue #1065, opened 2017-08-23
VisitorMap duplicates keys for same object. 


lodart:
Before, different objects were considered the same key, but thanks to your  VisitorMap, @matozoid, this is no longer the case. The problem is that now, even the same objects produces several keys ! 
If I take your Test example : 
```java
CompilationUnit x1 = JavaParser.parse("class X{}");
Map<CompilationUnit, Integer> normalMap = new VisitorMap<>(new HashMap<>(), new ObjectIdentityHashCodeVisitor(), new ObjectIdentityEqualsVisitor());
normalMap.put(x1, 1);
normalMap.put(x1, 2);
System.out.println(x1 == x1);
System.out.println(normalMap.size());
System.out.println(normalMap.get(x1));
```
it returns : 
```
true
2
null
```

lodart:
In both `ObjectIdentityHashCodeVisitor` and `ObjectIdentityEqualsVisitor`, there doesn't seem to be a `visit` method for `Expression`s

matozoid:
Okay, problem found. It will be in next weekend's release, or you can copy `VisitorMap.java` to your own project to get it now :-)

lodart:
Thank's, but I think there is still a problem :/ `System.out.println(normalMap.get(x1));` still returns me `null` ...
I think you forgot  to change 
```java
public V get(Object key) { return (V) innerMap.get(key) }
```
to
```java
public V get(Object key){return (V) innerMap.get(new EqualsHashcodeOverridingFacade((N) key));}
```
`containsKey(Object)` is also still bugged.

matozoid:
I'm sorry for putting this thing together so quickly, this is a little embarrassing. Would you be interested in building a PR for this so it is correct right away? :-)

lodart:
I'm on it ! This will be my first PR ever, so be indulgent ;)

matozoid:
You can now put "contributed to an open source project" on your resume ;-)

matozoid:
So do you think we're done with this issue?

PR #1067, opened 2017-08-24
Fix containsKey, get, putAll and Equals of VisitorMap

lodart:
See #1065
There is still `putAll` to fix, but I don't know how to do that.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/VisitorMap.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/VisitorMap.java
@@ -99,7 +99,7 @@ public class VisitorMap<N extends Node, V> implements Map<N, V> {
 
     @Override
     public void putAll(Map<? extends N, ? extends V> m) {
-        innerMap.putAll(m);
+        m.forEach((key, value) -> this.put(key,value));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


matozoid:
So you fixed putall too? Well done :-) I might complain about a lack of unit tests, but it's not like I wrote a lot myself...

lodart:
Here you go ! Tell me if I do anything wrong, I'm not used to work like that :)

matozoid:
You're doing perfectly well, actually! Would you like to do some refactorings that I thought of?
- we don't need to pass `innerMap` in the constructor. It's no use outside the `VisitorMap`. Remove that parameter and just `new` a `HashMap` in the field declaration.
- that `innerMap` has been declared as a raw type. Maybe we can declare it `Map<EqualsHashcodeOverridingFacade, V>` or `Map<EqualsHashcodeOverridingFacade, ?>` to get rid of some casts?
- really every method that accesses `innerMap` passing a key should be fixed, so also `remove`.

lodart:
Unfortunately, I can't make `remove` to work correctly, and I don't know why :/ Neither did I succeed removing the casts... I failed you master (branch, lol) !

matozoid:
Well, that's for another day then :-) Thank you for your contribution!

PR #1071, opened 2017-08-26
Fix remove and clean code

matozoid:
Hey @lodart - here's some stuff that Intellij suggested to fix, and I fixed remove. I think this is complete now :-)


javaparser-testing/src/test/java/com/github/javaparser/utils/VisitorMapTest.java
```diff
@@ -18,52 +19,62 @@ public void normalEqualsDoesDeepCompare() {
...
+        CompilationUnit x1 = JavaParser.parse("class X{}");
+        VisitorMap<CompilationUnit, Integer> map = new VisitorMap<>(new ObjectIdentityHashCodeVisitor(), new ObjectIdentityEqualsVisitor());
+        map.put(x1, 1);
+        assertTrue(map.containsKey(x1));
+        
+        map.remove(x1);
+        
+        assertFalse(map.containsKey(x1));
+    }
 }
```
ftomassetti: newline :)

matozoid: I left it there for you to complain about :-)

matozoid:
@ftomassetti - do you think this is a good approach? I feel mixed about solving this by making special versions of collections. Can't think of another way though.

ftomassetti:
I also have troubles thinking of an alternative, so I would say this seems a reasonable solution and the only I can think about

lodart:
Thank's guys ! :)

Issue #1046, opened 2017-08-14
Use threads to visit directories and parse files in parallel in SourceRoot


jimshowalter:
We're getting:

[info] Parsed 26810 files
[info] ELAPSED: 70 (383 files/second)

We wrote a regex-based tool that isn't as powerful as a Java compiler front-end, but still does a lot of work, and with that we're getting:

26707 all Java files
32581 total files
[info] ELAPSED: 19 (1714 files/second)

It's likely that much of the difference in speed is due to the 1000 threads and 500 open files allowed in the faster one.

ftomassetti:
Well, a regex-based tool does probably a fraction of the work of JavaParser so I would say it is a pretty good comparison for JavaParser

matozoid:
Good suggestion though - this would be a nice one for a user contribution since the issue is constrained to one file.

jimshowalter:
I hope to have a user contribution for you soon.

Regarding "a pretty good comparison for JavaParser", don't sell yourselves short. The parser is fine. But you're running it synchronously. On an 8-way box, expect to see a corresponding 8x improvement in speed.

Our Jenkins server runs on a 48-core box.

Threads are good.

jimshowalter:
It's ironic that you labelled this easy, because I just blew six hours on it. Everything I tried just made it slower. It threaded, and it wrote out the number of parses done in background vs. synchronous (it throttled), but the thread performance was abysmal compared to no threading. I used the same pattern we've used numerous times in other tools at my company.

matozoid:
:-D yes, it's one of the worst labels we have. It should be "a good first contribution" or so. We had more easy tasks explode like this

ryan-beckett:
@jimshowalter: What do you, "synchronously"? Single-threaded?

__Edit:__ Nevermind, I understand.

jimshowalter:
The idea is to visit directories on separate threads and parse files on separate threads, up to a tunable limit. I wrote a tool that does that for a different application. The trick is to avoid needing to synchronize, and the way to do that is to make each visited directory a separate object that contains a set of files parse in that directory, and a set of subdirectories. At the end, you visit the tree and collect all parse files. This took our parsing from minutes to seconds.

Unfortunately, when I tried applying the same approach using javaparser, it made things worse. It's possible javaparser consumes so much memory that trying to background it is doomed.

matozoid:
That reminded me to build this thing: https://github.com/javaparser/javaparser/pull/1069 :-)

[...]

PR #1073, opened 2017-08-28
Use threads to visit directories and parse files in parallel

ryan-beckett:
SourceRoot #1046

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -95,6 +95,15 @@ public final class JavaParser {
         JavaParser.staticConfiguration = staticConfiguration;
     }
 
+    /**
+     * Get the non-static configuration for this parser.
+     *
+     * @return The non-static configuration for this parser.
+     */
+    public ParserConfiguration getParserConfiguration() {
+        return this.configuration;
+    }
+
     private GeneratedJavaParser getParserForProvider(Provider provider) {
         if (astParser == null) {
             astParser = new GeneratedJavaParser(provider);
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
@@ -1,10 +1,11 @@
 package com.github.javaparser.utils;
 
-import com.github.javaparser.JavaParser;
-import com.github.javaparser.ParseProblemException;
-import com.github.javaparser.ParseResult;
-import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.printer.PrettyPrinter;
+import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
+import static com.github.javaparser.Providers.provider;
+import static com.github.javaparser.utils.CodeGenerationUtils.fileInPackageRelativePath;
+import static com.github.javaparser.utils.CodeGenerationUtils.packageAbsolutePath;
+import static com.github.javaparser.utils.SourceRoot.Callback.Result.SAVE;
+import static com.github.javaparser.utils.Utils.assertNotNull;
 
 import java.io.IOException;
 import java.nio.file.FileVisitResult;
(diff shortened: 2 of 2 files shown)


matozoid:
Cool stuff! Thanks!

Now I'll get to the nitpicking :-)

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -68,16 +73,54 @@ public SourceRoot(Path root) {
...
+                try {
+                    Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
+                        @Override
+                        public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
+                                throws IOException {
+                            if (!dir.equals(AsynchronousParse.this.path)) {
+                                AsynchronousParse w = new AsynchronousParse(dir);
+                                w.fork();
+                                walks.add(w);
+                                return FileVisitResult.SKIP_SUBTREE;
```
matozoid: This piece of code is duplicated below, any chance of cleaning that up? (only if it improves things of course)

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -45,7 +49,7 @@
     private final Path root;
     private final Map<Path, ParseResult<CompilationUnit>> cache = new HashMap<>();
     private JavaParser javaParser = new JavaParser();
-    private Function<CompilationUnit, String> printer = new PrettyPrinter()::print; 
+    private Function<CompilationUnit, String> printer = new PrettyPrinter()::print;
```
matozoid, line 52: Easy on the reformat button please!

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -92,30 +135,73 @@ public SourceRoot parse(String startPackage, JavaParser javaParser, Callback cal
...
+                try {
+                    Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
+
+                        @Override
+                        public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs)
+                                throws IOException {
+                            if (!dir.equals(AsynchronousParse.this.path)) {
+                                AsynchronousParse w = new AsynchronousParse(dir);
+                                w.fork();
+                                walks.add(w);
```
matozoid: I'm actually completely unfamiliar with this API, so I can't really judge if this is well written. Works well for the code generation parts of JP though. @ftomassetti ? @jimshowalter?

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -95,6 +95,15 @@ public static void setStaticConfiguration(ParserConfiguration staticConfiguratio
         JavaParser.staticConfiguration = staticConfiguration;
     }
 
+    /**
+     * Get the non-static configuration for this parser.
+     *
+     * @return The non-static configuration for this parser.
+     */
+    public ParserConfiguration getParserConfiguration() {
+        return this.configuration;
```
matozoid, line 104: Ah well, why not?

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -92,30 +135,73 @@ public SourceRoot parse(String startPackage, JavaParser javaParser, Callback cal
...
+                        public FileVisitResult visitFile(Path absolutePath, BasicFileAttributes attrs)
+                                throws IOException {
+                            if (!attrs.isDirectory() && absolutePath.toString().endsWith(".java")) {
+                                Path localPath = root.relativize(absolutePath);
+                                Log.trace("Parsing %s", localPath);
+                                final ParseResult<CompilationUnit> result = new JavaParser(
+                                        SourceRoot.this.javaParser.getParserConfiguration()).parse(COMPILATION_UNIT,
+                                                provider(absolutePath));
+                                result.getResult().ifPresent(cu -> cu.setStorage(absolutePath));
+                                if (callback.process(localPath, absolutePath, result) == SAVE) {
```
matozoid: Does this change mean that the code in the callbacks must be threadsafe? That's going to mess with our users - most are junior level. There should either be a big flashing warning, or we could make a switch for parallelizing that is off by default, or we can have two methods - `parse` and `parseParallelized`, or maybe something else...?

PR #1077, opened 2017-08-29
Use threads to visit directories and parse files in parallel in SourceRoot #1046

ryan-beckett:
@matozoid:

1. I abstracted away some of the redundant code I had.

2. In this PR, other than switching the cache to a ConcurrentHashMap and the addition of my methods, the original SourceRoot code is functionally the same. I refactored a bit, but the copying of the parser no longer exists in **tryToParse()** and **parse()**.

3. In the javadocs for **parseParallelized(String, JavaParser, Callback)**, I've noted that the callback must be threadsafe. I've also noted in the javadocs for both parallelized methods that a new instance of the parser that's instantiated with the internal parser's configuration is used to parse each file.

4. Although variable at times, I'm seeing nearly a 50% speed up on my machine (and it's an ancient CPU) for ~20k files, with a marginal memory increase for both parse and cached parse.

5. Before taking steps to ensure thread safety, I was getting parse exceptions. Since implementing these steps, I've seen no such exceptions and all tests are passing.

I can't think of any other issues.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -95,6 +95,15 @@ public final class JavaParser {
         JavaParser.staticConfiguration = staticConfiguration;
     }
 
+    /**
+     * Get the non-static configuration for this parser.
+     *
+     * @return The non-static configuration for this parser.
+     */
+    public ParserConfiguration getParserConfiguration() {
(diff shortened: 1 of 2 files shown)


ftomassetti:
Why was this one closed? Too much stuff going on, I am a bit lost :D

matozoid:
@ryan-beckett is a bit trigger happy with the pr close button 😁

PR #1078, opened 2017-08-29
PrettyPrint option to wrap-and-column-align parameters of method calls

tarilabs:
This option wrap-and-column align parameters passed inside method calls.

Example.
source
```
myMethod(1, 2, 3, 5, Object.class);
```
pretty-printed with this option
```
myMethod(1,
         2,
         3,
         5,
         Object.class);
```

It use the configured indentation, followed by space-padding in order to align to the required column. In other words, leverage configured indentation as much as possible, but in order to align to column in the final parts space is used. The test included highlight this behaviour.

I have also "in the works" an additional configuration option to wrap-align chained method calls like fluent DSL or Java 8 stream API call like. Kindly let me know if sensible to include in this PR or better to manage as a separate PR later.

Thanks!

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser.printer;
 
+import com.github.javaparser.Position;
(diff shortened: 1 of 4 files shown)


matozoid:
Thanks! Looks fine to me :-)

In general small PR's are preferred, it makes reviewing a lot easier. I'll pull this one so it's easier for you to make a follow-up PR.

PR #1080, opened 2017-08-30
Update stubparser from upstream javaparser

wmdietl:


wmdietl:
Sorry, pull request in wrong direction...

matozoid:
Well, it's interesting what you're doing alright :-) Let me add some comments...

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/SimpleVoidVisitor.java
```diff
@@ -0,0 +1,337 @@
...
+import com.github.javaparser.ast.body.*;
+import com.github.javaparser.ast.comments.BlockComment;
+import com.github.javaparser.ast.comments.JavadocComment;
+import com.github.javaparser.ast.comments.LineComment;
+import com.github.javaparser.ast.expr.*;
+import com.github.javaparser.ast.modules.*;
+import com.github.javaparser.ast.stmt.*;
+import com.github.javaparser.ast.type.*;
+
+public class SimpleVoidVisitor<A> implements VoidVisitor<A> {
```
matozoid, line 13: Is this visitor interesting for the core library? How would you describe what to use it for?

javaparser-core/src/main/java/com/github/javaparser/ast/body/CallableDeclaration.java
```diff
@@ -140,6 +148,15 @@ public T setParameters(final NodeList<Parameter> parameters) {
         return (T) this;
     }
 
+    public CallableDeclaration<T> setReceiverAnnotations(NodeList<AnnotationExpr> receiverAnnotations) {
+        this.receiverAnnotations = receiverAnnotations;
+        return this;
+    }
+
+    public NodeList<AnnotationExpr> getReceiverAnnotations() {
+        return receiverAnnotations;
```
matozoid, line 157: Is this something we should add to the core library?

javaparser-core/src/main/java/com/github/javaparser/ast/type/ArrayType.java
```diff
@@ -118,6 +120,10 @@ public static Type wrapInArrayTypes(Type type, List<ArrayBracketPair>... arrayBr
                 }
             }
         }
+        // The above code concatenates the arrayBracketPairLists, because they represent different
+        // parts of the array rather than two sources of information about the same parts of the
+        // array.
+        type.setArrayBracketPairs(arrayAnnotations);
```
matozoid, line 126: The array bracket pairs were removed from the API because they were error prone and probably-usually not the best way to deal with arrays. That's why we introduced the wrapping `ArrayType`. That one has annotations for every bracket level. What is the reason you're adding the bracket pairs back to the API?

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
```diff
@@ -147,6 +147,13 @@ public Boolean visit(final CompilationUnit n, final Visitable arg) {
         return true;
     }
 
+    // TODO Implement EqualsVisitor for StubUnit
```
matozoid, line 150: That should get auto-generated when you add it to the list of node types in the metamodel generator.

matozoid:
Additionally: is there anything you've noticed about the library that you would have done differently?

wmdietl:
Thanks for the comments, @matozoid !
@Bohdankm22 will prepare a proper pull request against javaparser with the generally-useful changes. He'll comment on the other parts as well. He's travelling this week.

javaparser-core/src/main/java/com/github/javaparser/ast/body/CallableDeclaration.java
Bohdankm22: I think it would be good to have the receiver annotations list field. Because in case when receiver annotations passed to the method (for example) the javaparser keeps them at a regular parameter with name "this". The signature of the method is different when we pass the receiver annotations, but I think it should be the same.

matozoid: That's an interesting one. Made a suggestion in #1129

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/SimpleVoidVisitor.java
Bohdankm22: Yes, the Checker Framework uses it once.
It is extended to another visitor class that overrides only needed methods for those case.
You could check the code here: https://github.com/typetools/checker-framework/blob/master/framework/src/org/checkerframework/framework/stub/StubUtil.java#L211-L333

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
Bohdankm22: Thank you for pointing to that!

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/SimpleVoidVisitor.java
matozoid: Made an issue for it here: https://github.com/javaparser/javaparser/issues/1140

Bohdankm22: @matozoid thank you!

javaparser-core/src/main/java/com/github/javaparser/ast/body/CallableDeclaration.java
matozoid: Made a PR here: https://github.com/javaparser/javaparser/pull/1194

Issue #918, opened 2017-05-07
Override NodeList.toString


ftomassetti:
So that we can print something nicer

matozoid:
Any suggestions on format?

matozoid:
I guess simply toString the items in the list and append them.

ftomassetti:
that would do

PR #1081, opened 2017-08-30
Give NodeList a toString

matozoid:
See #918

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
@@ -32,6 +32,7 @@ import com.github.javaparser.metamodel.InternalProperty;
 import java.util.*;
 import java.util.function.*;
 import java.util.stream.Collector;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -472,4 +473,9 @@ private void setAsParentNodeOf(Node childNode) {
             childNode.setParentNode(getParentNodeForChildren());
         }
     }
+
+    @Override
+    public String toString() {
+        return innerList.stream().map(Node::toString).collect(Collectors.joining());
```
ftomassetti: should this be `joining(", ")` ?

ftomassetti: I would also consider adding some delimiters, useful in particular for empty lists. For example: "[" and "]"

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeListTest.java
```diff
@@ -219,4 +219,11 @@ public void replace() {
         assertEquals("z", list.get(1).asString());
         assertEquals("c", list.get(2).asString());
     }
+
+    @Test
+    public void toStringTest() {
+        final NodeList<Name> list = nodeList(new Name("abc"), new Name("bcd"), new Name("cde"));
+
+        assertEquals("abcbcdcde", list.toString());
```
ftomassetti: This could be instead be "[abc, bcd, cde]" with the changes I suggested

matozoid: Well, I did exactly what you agreed to in the issue, but I'll go fix it again then ;-)

PR #1082, opened 2017-08-30
Upgrade JavaCC

matozoid:
An update was overdue. Also: someone took over development of the JavaCC maven plugin, so I switched to that.

See #899


ftomassetti:
This makes sense to me but this require rebase/manual merging

Issue #546, opened 2016-11-07
PrettyPrinter: specify the indentation style for switch/case


matozoid:
```java
switch {
    case:
```
or
```java
switch {
case:
```

Issue #541, opened 2016-11-07
PrettyPrinter: support different indent characters


matozoid:
I'm going through the options in https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties to make a minimal set of features that a pretty printer needs.

I actually think this one is already implemented. We can specify the indentation string, which can be a tab or a certain amount of spaces. Maybe specifying the tab size in spaces is necessary for other formatting options, like "how many tabs do I use to indent the second parameter of a method if it gets its own line?"

matozoid:
Yeah, this is fine as it is.

PR #1085, opened 2017-08-31
PrettyPrint option to wrap-and-column-align chained method calls

tarilabs:
This option wrap-and-column align chained method calls.

This can be helpful in case of using "fluent DSL" or API similar to Java 8's Stream API.

Example.
source
```
IntStream.range(0, 10).filter(x -> x % 2 == 0).map(x -> x * 10).forEach(System.out::println);
```
pretty-printed with this option
```
IntStream.range(0, 10)
         .filter(x -> x % 2 == 0)
         .map(x -> x * IntStream.of(1, 3, 5, 1)
                                .sum())
         .forEach(System.out::println);
```
It use the configured indentation, followed by space-padding in order to align to the required column. In other words, leverage configured indentation as much as possible, but in order to align to column in the final parts space is used. The test included highlight this behaviour.

This PR follows-up on #1078 .


javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -97,21 +99,50 @@ public void printUseTestVisitor(){
     
     @Test
     public void prettyColumnAlignParameters() {
+        PrettyPrinterConfiguration config = new PrettyPrinterConfiguration();
+        config.setIndent("\t");
+        config.setColumnAlignParameters(true);
```
ftomassetti: should we have two versions of these test? With and without the column align parameters flag set

matozoid:
Any chance we can get you to look at issues #541 - #546 ? :-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -30,6 +30,7 @@
     private boolean printComments = true;
     private boolean printJavaDoc = true;
     private boolean columnAlignParameters = false;
+    private boolean columnAlignFirstMethodChain = false;
```
matozoid: I'm fine with all these options enabled by default - for configuration options we say "you get everything, if you don't want it turn it off." It will surprise some users, but I will put a clear comment in the release notes.

tarilabs:
Hi, I have received notification by email from GH about:

> I'm fine with all these options enabled by default [...]

but said comment is not available here, hence I assume it has been deleted because it is actually preferred these options to not be enabled by default.

Personally, I agree with having them disabled by default. This is because I find these options helpful for my personal case where I need some specific code-style applied (hence I'm proposing to be merged in the upstream JavaParser, so others can benefit _if_ they also need them) but in general and for ensuring easier testing, I believe these should be disabled by default.

Besides of the above, @matozoid @ftomassetti kindly let me know if there's anything else missing in this PR so to facilitate it could be merged, thanks again!

matozoid:
@tarilabs yes, I deleted it because I changed my mind. I guess I'm the future we can look at it and reduce their impact (like only wrapping when lines are long or there are many method calls)

javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
matozoid: @tarilabs could you address this question? It's policy not to merge PR's with open questions.

tarilabs: @ftomassetti trying to address this question but a bit undecided: the full test suite (I mean all the other tests but this method) are actually testing "without" this flag. Proof is, if I force this flag as the default, a few tests break in string comparison. Therefore, the full test suite is acting as "flag=false", while this test is ensuring the pretty printer is acting as required when the "flag=true".

Anyway, if we prefer to have a specific test for the highlighted case in both variant, kindly let me know if you prefer for it to be:

- testing the other variant within the existing test method
- as a separate test method

(I would refrain from using JUnit "parametrized" because it would require the combinatorial of the several options..)

I have no problem adding additional test and assertion, just kindly let me know which strategy is preferred ;-) thanks!

ftomassetti: As a separate test method would be perfect. Thank you for your work on this one!

ftomassetti:
+1 from me, thanks

matozoid:
Thanks! Any more coming?

tarilabs:
@matozoid not for the moment; for the next features I'm working on and that I believe could benefit the upstream javaparser as well, I will need a bit more time to make up some experiment and then propose it back to you. I'll be in touch on gitter I think, once I have the basics working. Thanks to all of you!

Issue #1079, opened 2017-08-30
Parse java files from a zip or jar


RedBlackTreei:
for example I have a jar called `commons-collections4-4.1-sources.jar`, How can I get the classes and methods, I have read the wiki, but I  didn't find anything about this.

In addition, does javaparser support fuzzy matching like eclipse or IDEA(eg. *Facade)

matozoid:
- this parser parses source files only. Maybe you want https://github.com/javaparser/javasymbolsolver ?
- that's functionality for you to build, it's not the job of a parser, or maybe there is something in (see previous question)

ryan-beckett:
@Matozoid: Does adding JAR support into the source root sound like something worthwhile?

matozoid:
@ryan-beckett well, what would it do? JP's task is to parse java files, and `SourceRoot` is built for one thing: look at source files under a root directory.

ryan-beckett:
It would extract the files then parse through them, but then does **saveAll()** write source back to its file then zip the files? It's stuff like that's kind of hassling when thinking about the API in my opinion. Whatever, I just want your thoughts. There are other things I can work on.

matozoid:
Oh right, I completely missed that we're talking about a jar of source files! Sorry!

Here's what I wrote to do the test of the full openjdk source code (which disappeared, so it's not working right now) - https://github.com/javaparser/javaparser/blob/master/javaparser-testing/src/test/java/com/github/javaparser/manual/BulkParseTest.java

A bit disappointing maybe that I just unzip the whole thing, then go through the extracted files with `SourceRoot`. Hmmm, I guess you could mess with that unzip method to make it pass every inputstream directly to a `JavaParser` instance so we don't need a temporary file on disk... That does sound rather cool, doesn't it?

If `SourceRoot` is a root directory for source files, would `SourceJar` or `SourceZip` or `SourceArchive` be the same thing for an archive with source code? Where we only support reading with a callback? Sounds like a fun project.

ryan-beckett:
> If SourceRoot is a root directory for source files, would SourceJar or SourceZip or SourceArchive be the same thing for an archive with source code? Where we only support reading with a callback? Sounds like a fun project.

@matozoid: Do you want to create a **Source** interface/base class? Or, does adding a new **SourceZip** class suffice for now?  Also, should this class support everything the **SourceRoot** currently does, or does supporting a non-cached parse w/ callback suffice for the moment?

[...]

PR #1088, opened 2017-08-31
Parse java files from a zip or jar

ryan-beckett:
See #1079 

@matozoid: Let me know if you want API changes and/or additions. I edited your **BulkParseTest** to accommodate **SourceZip** and used the JavaParser ZIP (via GitHub download) to test against. Everything works as expected.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2016 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ * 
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,26 @@
...
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.utils;
+
+public class SourceZip {
+
```
matozoid, line 25: Something went wrong with pushing here? :-)

ryan-beckett: Yeah, let me see.

Issue #1074, opened 2017-08-28
Pass ParserConfiguration to SourceRoot instead of a JavaParser


matozoid:
There's nothing interesting a user can do to a `JavaParser` instance. All configuration is in `ParserConfiguration`.

Issue #941, opened 2017-05-30
Figure out where the openJDK zip downloads went


matozoid:
and use that to fix the BulkParseTest

matozoid:
Maybe use the Ubuntu package? https://packages.ubuntu.com/xenial/java/openjdk-9-source

PR #1089, opened 2017-08-31
Parse java files from a zip or jar

ryan-beckett:
See #1079 

@matozoid: Let me know if you want API changes and/or additions. I edited your BulkParseTest to accommodate SourceZip and used the JavaParser ZIP (via GitHub download) to test against. Everything works as expected.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
@@ -1,0 +1,142 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2016 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.utils;
+
+import static com.github.javaparser.utils.Utils.assertNotNull;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseResult;
+import com.github.javaparser.ast.CompilationUnit;
+import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
+import static com.github.javaparser.Providers.provider;
+
+/**
+ * A collection of Java source files and its sub-directories located in a ZIP or JAR file on the file system.
+ * Files can be parsed with a callback.
+ */
+public class SourceZip {
+
+    private final Path zipPath;
+    private JavaParser javaParser;
+
+    /**
+     * Create a new ZIP parser. A default instance of {@link JavaParser} will be used to parse the ZIP.
+     *
+     * @param zipPath The absolute path of ZIP file to parse.
+     */
+    public SourceZip(Path zipPath) {
+        assertNotNull(zipPath);
+        if (Files.isDirectory(zipPath)
+                || !(zipPath.toString().endsWith(".zip") || zipPath.toString().endsWith(".jar"))) {
+            throw new IllegalArgumentException("SourceZip only parses ZIP and JAR files.");
+        }
+        this.zipPath = zipPath.normalize();
+        this.javaParser = new JavaParser();
+        Log.info("New source zip at \"%s\"", this.zipPath);
+    }
+
+    /**
+     * Create a new ZIP parser.
+     *
+     * @param zipPath The absolute path of ZIP file to parse.
+     * @param javaParser The parser used to parse the ZIP files.
+     */
+    public SourceZip(Path zipPath, JavaParser javaParser) {
+        this(zipPath);
+        assertNotNull(javaParser);
+        this.javaParser = javaParser;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Create a new ZIP parser. A default instance of {@link JavaParser} will be used to parse the ZIP.
+     *
+     * @param zipPath The absolute path of ZIP file to parse.
+     */
+    public SourceZip(Path zipPath) {
+        assertNotNull(zipPath);
+        if (Files.isDirectory(zipPath)
+                || !(zipPath.toString().endsWith(".zip") || zipPath.toString().endsWith(".jar"))) {
+            throw new IllegalArgumentException("SourceZip only parses ZIP and JAR files.");
```
matozoid, line 58: Hm, not sure, do we want people to be able to use funny extensions on zip files, or should we protect them from making a mistake like this?

ryan-beckett: I say we don't protect against it.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    /**
+     * Create a new ZIP parser.
+     *
+     * @param zipPath The absolute path of ZIP file to parse.
+     * @param javaParser The parser used to parse the ZIP files.
+     */
+    public SourceZip(Path zipPath, JavaParser javaParser) {
+        this(zipPath);
```
matozoid, line 72: As a general rule, make constructors with less parameters call constructors with more parameters, so that you get one simple, general purpose constructor and a bunch of "utility" constructors that default some behaviour.

javaparser-testing/src/test/java/com/github/javaparser/utils/SourceZipTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.Problem;
+import com.github.javaparser.ast.validator.Java9Validator;
+
+public class SourceZipTest {
+
+    private final static String PROJECT_NAME = "JavaParser";
+    private final static String PROJECT_TEST_RESULT_FILE = PROJECT_NAME + "_Test_Results.txt";
+    private final static String ZIP_URL = "https://github.com/javaparser/javaparser/archive/master.zip";
```
matozoid, line 50: Let's put a tiny zip with some sources in test/resources so we don't need an internet connection while building.

ryan-beckett: ok.

javaparser-testing/src/test/java/com/github/javaparser/utils/SourceZipTest.java
```diff
@@ -0,0 +1,102 @@
...
+        Path zipPath = workdir.resolve(ZIP_NAME);
+        if (Files.notExists(zipPath)) {
+            Log.info("Downloading " + PROJECT_NAME);
+            download(new URL(ZIP_URL), zipPath);
+        }
+        bulkTest(new SourceZip(zipPath), PROJECT_TEST_RESULT_FILE);
+    }
+
+    private void bulkTest(SourceZip sourceZip, String testResultsFileName) throws IOException {
+        sourceZip.setJavaParser(new JavaParser(new ParserConfiguration().setValidator(new Java9Validator())));
```
matozoid, line 68: What you could do instead is change the existing BulkParseTest like this and call that the test. You will have to figure out #941 then though :-(

matozoid: I mean - this class is way too huge to be called a unit test, it writes a file with parse problems so that the parser can be verified.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,142 @@
...
+        Log.info("Parsing zip at \"%s\"", zipPath);
+        ZipFile zipFile = new ZipFile(zipPath.toFile());
+        for (ZipEntry entry : Collections.list(zipFile.entries())) {
+            if (!entry.isDirectory() && entry.getName().endsWith(".java")) {
+                final ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT,
+                        provider(zipFile.getInputStream(entry)));
+                callback.process(Paths.get(entry.getName()), result);
+            }
+        }
+        zipFile.close();
```
matozoid, line 95: Can we use a try-with-resources here?

matozoid:
I love how short the code is!

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,142 @@
...
+        Log.info("New source zip at \"%s\"", this.zipPath);
+    }
+
+    /**
+     * Create a new ZIP parser.
+     *
+     * @param zipPath The absolute path of ZIP file to parse.
+     * @param javaParser The parser used to parse the ZIP files.
+     */
+    public SourceZip(Path zipPath, JavaParser javaParser) {
```
matozoid, line 71: If you don't mind - I was planning to deprecate passing a `JavaParser` to `SourceRoot` (#1074) so maybe we can prevent that for this class and pass `ParserConfiguration` instead.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Tries to parse all '.java' files in the ZIP located at this <i>SourceZip</i>'s path and passes them one by one
+     * to the provided callback.
+     *
+     * @param callback A callback method for the post-processing of a parsed ZIP file.
+     * @return This {@link SourceZip} object.
+     */
+    public SourceZip parse(Callback callback) throws IOException {
+        assertNotNull(callback);
+        Log.info("Parsing zip at \"%s\"", zipPath);
```
matozoid, line 86: What if it's a jar? ;-)

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
ryan-beckett: I ended up removing that constructor anyways.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceZip.java
ryan-beckett: Technically, a JAR is an extension of a ZIP, so I feel like that's OK, but we can also name the class SourceArchive. It's up to you. I'm fine with SourceZip though.

PR #1115, opened 2017-09-04
Issue 941 fix bulkparsetest

matozoid:
Actually, the issue still isn't fixed, but there is another target for bulk testing now.

@ryan-beckett `SourceZip` died horribly with out of memory while parsing the JDK's `src.zip`. I completely forgot that using a callback makes sure that only one AST is in memory at a time, so I added that functionality back :-)


javaparser-testing/src/test/java/com/github/javaparser/manual/BulkParseTest.java
```diff
@@ -69,8 +73,29 @@ public void bulkTest(SourceRoot sourceRoot, String testResultsFileName) throws I
             }
             return DONT_SAVE;
         });
+        writeResults(results, testResultsFileName);
+    }
+
+    public void bulkTest(SourceZip sourceRoot, String testResultsFileName) throws IOException {
+        sourceRoot.setJavaParser(new JavaParser(new ParserConfiguration().setValidator(new Java9Validator())));
```
matozoid: @ryan-beckett I kind of regret not having that base class now - I had to write the same thing twice for directories and zip files...

matozoid:
@ryan-beckett it works wonderfully though!

matozoid:
Now found a location for the old test suite, which wasn't the JDK source but the langtools repository source. Fixes #941

Issue #921, opened 2017-05-12
Use enums or helper methods to create static and asterisk imports


ftomassetti:
This method call is not obvious:

```
compilationUnit.addImport("java.util", false, true);
```

Alternatives:
```
compilationUnit.addImport("java.util", NO_STATIC, ASTERISK);
compilationUnit.addAsteriskImport("java.util");
```

ftomassetti:
Similarly for `ClassOrInterfaceDeclaration`, the flag telling us if it is a class or an interface

matozoid:
It may be flagged easy, but be sure to *deprecate* the methods with the booleans, not *remove*.

matozoid:
F*** this "asterisk" nonsense, the name is "on demand" and now you're getting it :-)

matozoid:
I'll skip the interface flag since that's not easy to do without breaking stuff, and we have #967 on the backlog.

matozoid:
I currently have
```java
    public enum ScopeSize {
        /** import of a single type */
        SINGLE,
        /** import of a package (gets a * at the end) */
        ON_DEMAND
    }

    public enum ScopeStaticness {
        /** imports a type */
        TYPE,
        /** imports static methods (adds the "static" keyword) */
        STATIC
    }
```
... but I don't like the names "ScopeStaticness" and "ScopeSize". Any ideas @ftomassetti or others?

ftomassetti:
ImportLevel?
ImportGranularity?
ImportScope?

matozoid:
Okay, I guess `ImportDeclaration.Scope` isn't bad. How about `ImportDeclaration.Staticness`? :-/

ftomassetti:
that also sounds reasonable

PR #1118, opened 2017-09-05
[DRAFT] Introduce enums for ImportDeclaration

matozoid:
Draft to show @ftomassetti what he's asking for :-)

See #921 

Edit: it's probably way more than you asked for...


matozoid:
I think I'm taking this too far, the code is not getting any clearer this way, just more verbose.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -311,18 +310,24 @@ else if (clazz.isAnonymousClass() || clazz.isLocalClass())
         return addImport(clazz.getName());
     }
 
+    public CompilationUnit addImport(String name, boolean isStatic, boolean isAsterisk) {
```
ftomassetti, line 313: I would deprecate this method in favor of the next one

javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
```diff
@@ -20,53 +20,72 @@
...
     private boolean isAsterisk;
 
-    private ImportDeclaration() {
-        this(null, new Name(), false, false);
+    public ImportDeclaration() {
+        this(null, new Name(), TYPE, false);
     }
 
-    @AllFieldsConstructor
     public ImportDeclaration(Name name, boolean isStatic, boolean isAsterisk) {
```
ftomassetti, line 72: I would deprecate it

javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
```diff
@@ -97,9 +116,12 @@ public boolean isAsterisk() {
         return isAsterisk;
     }
 
-    @Generated("com.github.javaparser.generator.core.node.PropertyGenerator")
+    /**
+     * Use setStaticness instead
```
ftomassetti, line 120: The comment should be changed

javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java
```diff
@@ -97,9 +116,12 @@ public boolean isAsterisk() {
...
+     */
+    @Deprecated
```
ftomassetti, line 122: I would not deprecate it, maybe I would add isType or isInstance instead

javaparser-core/src/main/java/com/github/javaparser/ast/observer/ObservableProperty.java
```diff
@@ -35,7 +35,7 @@
...
...LECTOR(Type.SINGLE_REFERENCE), STATEMENT(Type.SINGLE_REFERENCE), STATEMENTS(Type.MULTIPLE_REFERENCE), STATIC(Type.SINGLE_ATTRIBUTE), STATICNESS(Type.SINGLE_ATTRIBUTE), SUPER_TYPE(Type.SINGLE_REFERENCE), TARGET(Type.SINGLE_REFERENCE), THEN_EXPR(Type.SINGLE_REFERENCE), THEN_STMT(Type.SINGLE_REFERENCE), THIS(Type.SINGLE_ATTRIBUTE), THROWN_EXCEPTIONS(Type.MULTIPLE_REFERENCE), TRY_BLOCK(Type.SINGLE_REFERENCE), TYPE(Type.SINGLE_REFERENCE), TYPES(Type.MULTIPLE_REFERENCE), TYPE_ARGUMENTS(Type.MULTIPLE_REFERENCE), TYPE_BOUND(Type.MULTIPLE_REFERENCE), TYPE_PARAMETERS(Type.MULTIPLE_REFERENCE), UPDATE(Type.MULTIPLE_REFERENCE), VALUE(Type.SINGLE_REFERENCE), VALUES(Type.MULTIPLE_REFERENCE), VARIABLE(Type.SINGLE_REFERENCE), VARIABLES(Type.MULTIPLE_REFERENCE), VAR_ARGS(Type.SINGLE_ATTRIBUTE), VAR_ARGS_ANNOTATIONS(Type.MULTIPLE_REFERENCE), WITH_TYPES(Type.MULTIPLE_REFERENCE), ELSE_BLOCK(Type.SINGLE_ATTRIBUTE, true), EXPRESSION_BODY(Type.SINGLE_REFERENCE, true), MAXIMUM_COMMON_TYPE(Type.SINGLE_REFERENCE, true), POSTFIX(Type.SINGLE_ATTRIBUTE, true), PREFIX(Type.SINGLE_ATTRIBUTE, true), THEN_BLOCK(Type.SINGLE_ATTRIBUTE, true), USING_DIAMOND_OPERATOR(Type.SINGLE_ATTRIBUTE, true), RANGE, COMMENTED_NODE;
```
ftomassetti, line 38: Maybe we should add newlines to make easier spotting differences

javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java1_0Validator.java
```diff
@@ -76,7 +81,7 @@
             (n, reporter) -> reporter.report(n, "For-each loops are not supported.")
     );
     protected final Validator noStaticImports = new SimpleValidator<>(ImportDeclaration.class,
-            ImportDeclaration::isStatic,
+            n -> n.getStaticness() == STATIC,
```
ftomassetti, line 84: I would leave it as it was, because I would not deprecate isStatic

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1384,7 +1384,7 @@ public void visit(NodeList n, Void arg) {
     public void visit(final ImportDeclaration n, final Void arg) {
         printJavaComment(n.getComment(), arg);
         printer.print("import ");
-        if (n.isStatic()) {
+        if (n.getStaticness() == ImportDeclaration.Staticness.STATIC) {
```
ftomassetti, line 1387: I would leave the call as it was

run_core_generators.sh
```diff
@@ -7,7 +7,7 @@
 pushd javaparser-core-generators
 
 # Generate code
-mvn clean package -P run-core-generators
+mvn clean package -P run-core-generators -DskipTests
```
ftomassetti, line 10: right!

ftomassetti (approved):

javaparser-core/src/main/java/com/github/javaparser/ast/observer/ObservableProperty.java
matozoid: This is what our pretty printer does :-/

matozoid:
@ftomassetti so do you think we should change the field in ImportDeclaration from boolean to enum, or just keep the enums as utils? I personally think the enums don't add much - except maybe in the addImport method (but to add two enums just to make one method more readable? Hm...)

ftomassetti:
Well I think that readability is important and I know of companies having as a rule to create enums to avoiding booleans as method arguments, especially when you have more than one of them. I think the main value is having enums as arguments, while the fact we use them or not in fields is still useful but less important imho

matozoid:
@ftomassetti I'm not very happy with this:
- we're forced to invent terms that don't exist
- the enums are clumsy to use. Intellij nowadays even tells you what parameter name a boolean has :-)
- too big a change for the 3 line.

ftomassetti:
Then we should probably drop it

matozoid:
I'll keep the issues open for another try another day ;-)

Issue #1122, opened 2017-09-07
try with resources also accepts a variable since Java 9


matozoid:
```java
        // resource as variable
        DataInputStream is = new DataInputStream(new FileInputStream("x"));
        try (is) {
            while (true) {
                is.getChar();  // method not found
            }
        } catch (EOFException e) {
        }
```

PR #1125, opened 2017-09-07
Try with variable resource

matozoid:
See #1122


matozoid:
@ftomassetti this will impact JSS! A node was added and the `try` resources can now be more types of expressions. Check http://cr.openjdk.java.net/~iris/se/9/java-se-9-pr-spec-01/java-se-9-jls-pr-diffs.pdf

matozoid:
@ftomassetti I can avoid adding the new node if I make the type of the resources in the try `Node` instead of `Expression`...

javaparser-core/src/main/java/com/github/javaparser/ast/expr/QualifiedNameExpr.java
```diff
@@ -0,0 +1,127 @@
...
+import com.github.javaparser.metamodel.NameExprMetaModel;
+import javax.annotation.Generated;
+import static com.github.javaparser.utils.Utils.assertNotNull;
+import com.github.javaparser.metamodel.QualifiedNameExprMetaModel;
+
+/**
+ * Whenever a Name is used in an expression, it is wrapped in QualifiedNameExpr.
+ * <br/>In <code>try(a.b.c){}</code> a.b.c is a Name inside a QualifiedNameExpr.
+ */
+public final class QualifiedNameExpr extends Expression implements NodeWithName<QualifiedNameExpr> {
```
ftomassetti: This seems a big change!

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
```diff
@@ -57,7 +58,7 @@
  */
 public final class TryStmt extends Statement {
 
-    private NodeList<VariableDeclarationExpr> resources;
+    private NodeList<Expression> resources;
```
ftomassetti, line 99: Is this change specific for Java 9?
Looking here: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html
It says: "The try-with-resources statement is a try statement that declares one or more resources".
Should we need to guide users to use only VariableDeclarationExpr unless they are using Java 9?

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
```diff
@@ -57,7 +58,7 @@
  */
 public final class TryStmt extends Statement {
```
ftomassetti, line 97: Should this be named TryWithResourcesStatement ? Or are we using one statement for the TryStatement and the TryWithResources statement?

ftomassetti (approved):
I think it is interesting case in which we need to balance support for Java 9 and "preference for Java 8", in the sense that by default I think we should help users doing the right thing in Java 8 and make possible to do the right thing in Java 9.

Perhaps we can have different setters for the expression of the TryStmt, one with a default name, taking a VariableDeclaration and one with a clear name like "setExpressionJava9" or something that makes obvious this would not work in Java 8. Maybe we could even add annotation to methods that exist only to support Java 9. Would that make sense?

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
matozoid: There's only one indeed.

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
matozoid: Something like adding a validation that adds a Problem when parsing with Java 8 and before and we find something that is not a variable decl? Sure.

matozoid:
@ftomassetti the most important question is: shall I lower the resources even further, all the way to type `Node` and get rid of the new `QualifiedNameExpr`, or keep it this way?

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
ftomassetti: I would think we should have two separate statements but I realize this would be a breaking change

ftomassetti:
I would not use Node. Stupid question: how QualifiedNameExpr is different from FieldAccessExpr? Is it like a FieldAccessExpr where the parent can only be a FieldAccessExpr or a NameExpr?

matozoid:
The JLS (take a look) explicitly states that IDENTIFIER . IDENTIFIER . IDENTIFIER etc. should be allowed. I have no idea why. I'm not sure why matching `Expression` doesn't match this sequence. I'll go back and see.

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
matozoid: Hm, this is arbitrary: the try body, the final block and the catch statements are all optional, but we don't make separate nodes for those cases?

ftomassetti:
Perhaps we can use an expression and then check:
* if the language level is set to 8 then we will look for a VariableDeclarationExpr and produce an error
* if the language level is set to 9 then we will look for a NameExpr or a FieldAccessExpr having the form IDENTIFIER (. IDENTIFIER) *.

This would feel like something more structural than the problems we trigger during validation, and we know the language level only during validation, correct?

matozoid:
It sounds like there is some confusion about validations: the parser and the validations generate problems, and if there is any problem, parsing will fail. Doing those validations you suggest is indeed what we should do, and when done in a validator they will be as strong as a parse error - but better: we will have a complete AST, and we can run them again whenever we like. My intention is to weaken the grammar and move checks towards the validations so that we will get more AST to work with.

Also: validations are the only way to do different Java versions without making (and maintaining, ugh!) multiple grammars.

I'm going to ponder the extra methods for Java 9 - interesting idea!

matozoid:
Okay, one issue solved: `try(a.b.c){}` does indeed resolve to a `FieldAccessExpr`. I must have run the tests before regenerating the grammar or so, so that I didn't notice. Well, good :-)

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/TryStmt.java
matozoid: Added.

javaparser-testing/src/test/java/com/github/javaparser/ast/validator/Java7ValidatorTest.java
```diff
@@ -39,6 +39,18 @@ public void tryWithoutAnything() {
...
     @Test
+    public void tryWithResourceVariableDeclaration() {
+        ParseResult<Statement> result = javaParser.parse(STATEMENT, provider("try(Reader r = new Reader()){}"));
+        assertNoProblems(result);
+    }
+
+    @Test
+    public void tryWithResourceReference() {
+        ParseResult<Statement> result = javaParser.parse(STATEMENT, provider("try(a.b.c){}"));
+        assertProblems(result, "(line 1,col 1) Try with resources only supports variable declarations.");
```
matozoid, line 50: @ftomassetti check this out - much better than "PARSE ERROR AT 1,1" right?

matozoid:
@ftomassetti I just found a problem with the java9-method idea: setters, getters, adders, etc. are generated, so they will always match the type of the list: Expression. All I can think of now is documenting this very well.

What was your idea with annotations?

matozoid:
@ftomassetti I rewrote the Javadoc for `TryStmt`.

javaparser-testing/src/test/java/com/github/javaparser/ast/validator/Java7ValidatorTest.java
ftomassetti: Nice!

ftomassetti:
The idea with annotations was to mark with things like `@Java9` methods intended to be used only when building Java 9 code

matozoid:
This has become a lot better thanks to your feedback, @ftomassetti :-)

ftomassetti:
Thank you! I am just sorry for the delayed answers

matozoid:
I don't know how to improve this further without breaking things, so in it goes.

Issue #1056, opened 2017-08-15
Can you use <?> to get rid of the raw type warning for this signature?


jimshowalter:
VoidVisitorAdapter.public void visit(NodeList n, final A arg) {

matozoid:
I tried a lot of things there, but didn't succeed. If there is a way, I don't know it.

jimshowalter:
You can't add <?> in the base and propagate that down the inheritance?

matozoid:
I can't remember what exactly the problems were, but I know I tried everything I could. So unless someone contributes a fix it will stay this way.

matozoid:
Did you experiment with this? Do you hate Java yet? ;-)

jimshowalter:
I haven't tried it yet.

matozoid:
It would be nice if there was some kind of fix, because I wasn't too happy with this solution :-)

ryan-beckett:
@matozoid: Were these testing problems? Changing the current code to:

`VoidVisitor.void visit(NodeList<?> n, A arg);`
`VoidVisitorAdapter.public void visit(NodeList<?> n, A arg)`

causes no compilation issues.

PR #1126, opened 2017-09-10
[DRAFT] Attempt to fix raw NodeList type in visitors

matozoid:
@ryan-beckett @jimshowalter here's my attempt to get give NodeList a type parameter in the visitors. Currently fails in the `ModifierVisitor`. Any ideas?

See #1056


ftomassetti:
What is the state of this one?

matozoid:
No response from @ryan-beckett @jimshowalter - I'll delete it when nothing happened next Sunday, ok?

ryan-beckett:
I'll take a look at it.

ryan-beckett:
@ryan-beckett, @jimshowalter: The only way I can see to get rid of that warning is to change `NodeList.set(int, N)` to `NodeList.set(int, Node)`, which breaks type-safety. I can't think of an easy fix.

matozoid:
@jimshowalter does not seem active anymore, nobody knows a way to fix, I'll close the original issue too. Thanks for looking, @ryan-beckett

Issue #1128, opened 2017-09-11
Missing method addConstructor on EnumDeclaration


DeepSnowNeeL:
Hey guys, i'm updating from `3.0.0-RC.4` to `3.3.5`

And I used to do :
```
EnumDeclaration enumType....;
ConstructorDeclaration ctor = enumType.addConstructor(Modifier.PRIVATE);
```
but it looks like addConstructor for EnumDeclaration has been deleted ?

That would be great to restore it :)

matozoid:
Hmmm, odd. Let's investigate!

DeepSnowNeeL:
I just created https://github.com/javaparser/javaparser/pull/1130 to fix it :)

PR #1130, opened 2017-09-11
Missing method addConstructor on EnumDeclaration #1128

DeepSnowNeeL:
Here is the fix :)


matozoid:
Any idea why they disappeared in the first place? Maybe more stuff is broken.

DeepSnowNeeL:
I can't find when we deleted it from EnumDeclaration so I suppose it was in TypeDeclaration before and since it has been moved to ClassOrInterfaceDeclaration we didn't copy it in EnumDeclaration at that time.

matozoid:
I don't have access to an IDE right now - do we cover all TypeDeclaration subclasses with this PR?

Never mind, have access now

matozoid:
Hmmm, it seems I moved them from `NodeWithMembers` to `ClassOrInterfaceDeclaration` around 3.2.2. I think we should stuff these in a `NodeWithConstructors` interface. Do you want to do that, or shall I pull and do the work afterwards?

matozoid:
I'll give it a go then :-)

PR #1134, opened 2017-09-11
Remove get nodelists

matozoid:
This was bothering me for a long time :-)


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -556,24 +556,20 @@ public void register(AstObserver observer, ObserverRegistrationMode mode) {
...
-            if (nl != null)
-                nl.register(observer);
-        });
+        for (PropertyMetaModel property : getMetaModel().getAllPropertyMetaModels()) {
+            if (property.isNodeList()) {
+                NodeList<?> nodeList = (NodeList<?>) property.getValue(this);
+                if (nodeList != null)
+                    nodeList.register(observer);
+            }
+        }
```
matozoid, line 565: @ftomassetti here's the magic!

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
```diff
@@ -56,7 +56,7 @@ protected void replaceWhenSameSignature(ClassOrInterfaceDeclaration containingCl
                     throw new AssertionError(f("Wanted to regenerate a method with signature %s in %s, but it wasn't there.", callable.getSignature(), containingClassOrInterface.getNameAsString()));
                 });
     }
-    
+
```
ftomassetti, line 59: Good eye

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: Magic indeed!

javaparser-core/src/main/java/com/github/javaparser/ast/ArrayCreationLevel.java
```diff
@@ -128,12 +128,6 @@ public ArrayCreationLevel setAnnotations(final NodeList<AnnotationExpr> annotati
         return this;
     }
 
-    @Override
-    @Generated("com.github.javaparser.generator.core.node.GetNodeListsGenerator")
-    public List<NodeList<?>> getNodeLists() {
```
ftomassetti, line 133: Are we removing this public method? Could it be useful to someone? Maybe we could implement it once in Node using the metamodel

matozoid: The reason I'm removing it is because I can't imagine it being useful to anyone - do you have use cases? If so, it would be nice to at least rename it, I remember people being confused over what it does.

ftomassetti: Well I guess we can always remove it and put it back if people complain. Anyway for "smart" usages the result can be obtained from the metamodel I guess

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
matozoid: Thank you!

Issue #1072, opened 2017-08-28
Create VisitorList and VisitorSet like VisitorMap


matozoid:
... for completeness.

matozoid:
Mistake in the issue name, I meant `VisitorList` not `VisitorMap`

PR #1145, opened 2017-09-18
[DRAFT] Generate equals & hashcode visitors that ignore comments #729

ryan-beckett:
**Thoughts**

1. In `HashCodeVisitor`, generated methods for `JavadocComment`, `LineComment`, etc. return 0 for the hash code, as well as any other fields whose sole field is a comment. E.g., `NullLiteralExpr`, `UnparsableStmt`, etc.
2. All existing tests pass, and I've added tests to show equals and hashcode visitors ignore comments. Additionally, after writing the new tests, I used the current implementations of `EqualsVisitor` and `HashCodeVisitor`, and the new tests failed as expected, except for the following test:

```
    @Test
    public void testEqualsWithDifferentComments() {
        CompilationUnit p1 = JavaParser.parse("/* a */ class X { /** b */} //c");
        CompilationUnit p2 = JavaParser.parse("/* b */ class X { }  //c");
        assertEquals(p1.hashCode(), p2.hashCode());
        assertEquals(p1.getComments().size(), 3);
        assertEquals(p2.getComments().size(), 2);
    }
```
The current implementation of `HashCodeVisitor` effectively ignores comments in this example, even though it processes them. The two parsed sources are varied enough that they should produce different hashes, right? Also, the last two assertions guarantee the comments are attached to the nodes. Anyways, it has no bearing on my work, but I had expected for this test to fail. I don't really think this is a big deal, but I figured I'd point it out.


javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/HashCodeVisitorGenerator.java
```diff
@@ -22,19 +22,29 @@ public HashCodeVisitorGenerator(SourceRoot sourceRoot) {
...
         final BlockStmt body = visitMethod.getBody().get();
         body.getStatements().clear();
 
         final SeparatedItemStringBuilder builder = new SeparatedItemStringBuilder("return ", "* 31 +", ";");
-        final List<PropertyMetaModel> propertyMetaModels= node.getAllPropertyMetaModels();
-        if (propertyMetaModels.isEmpty()) {
+        final List<PropertyMetaModel> propertyMetaModels = node.getAllPropertyMetaModels();
+        String ndName = node.getTypeName();
+        if (ndName.equals("JavadocComment") || ndName.equals("LineComment") || ndName.equals("BlockComment")
+                || propertyMetaModels.isEmpty()) {
```
matozoid, line 36: This can be done in a nicer way: there is exactly one meta model instance for every node and property, so you can directly compare `node` to one of the fields in `JavaParserMetaModel`. (I'm writing this without checking, so if it turns out it makes no sense I'll see where I went wrong ;-) )

matozoid:
@ryan-beckett oops - miscommunication: these have to be new visitors. The existing equals and hashcode are fine. (Except for that test you wrote, please open an issue for that)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/HashCodeVisitorGenerator.java
```diff
@@ -22,19 +22,29 @@ public HashCodeVisitorGenerator(SourceRoot sourceRoot) {
...
             builder.append("0");
         } else {
             for (PropertyMetaModel field : propertyMetaModels) {
                 final String getter = field.getGetterMethodName() + "()";
+                if (getter.equals("getComment()")) {
```
matozoid, line 41: Same thing here: you should be able to find the `PropertyMetaModel` for `comment` in `Node`.

javaparser-testing/src/test/java/com/github/javaparser/ast/visitor/EqualsVisitorTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+
+public class EqualsVisitorTest {
```
matozoid, line 32: Yeah, I guess this could use a few tests.

ryan-beckett:
@matozoid: Will do.

matozoid:
@ryan-beckett as for a use case for this: there was a discussion about how to implement equals and hashcode for nodes, and we couldn't find a good answer because everyone has different requirements (do you want to check if it is exactly the same piece of code, or if the structure is the same, or if the structure and the naming is the same, or if all the nodes are the same but they can come from different files?) So the current ones were chosen as a good default, with the idea that you could override them. Now, that led to this: https://github.com/javaparser/javaparser/issues/1061

So: these visitors are supposed to be used with `VisitorMap` and friends (#1072) for when people want to compare AST's in a different way.

ryan-beckett:
@matozoid: Yeah, I read the discussion. Would `NoCommentEqualsVisitor` suffice as a name?

matozoid:
@ryan-beckett yes, that naming would match `ObjectIdentityEqualsVisitor`.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/HashCodeVisitorGenerator.java
ryan-beckett: @matozoid: I was hoping you brought this up. If hypothetically, the class name of one those nodes changed, this code would break. Thanks!

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/HashCodeVisitorGenerator.java
ryan-beckett: @matozoid: I haven't tested it, but something like this:

```
if (field.equals(JavaParserMetaModel.nodeMetaModel.commentPropertyMetaModel)) {
}
```

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/HashCodeVisitorGenerator.java
matozoid: No problem. If you ever think "hey, I should use reflection on the nodes to do x" - don't do it, use the meta model :-)

PR #1150, opened 2017-09-19
Make sure tokenranges are not created when not storing tokens.

### LINKED ISSUE #1149: Let storeTokens also influence tokenrange creation.

matozoid:
Fixes #1149 . @arturbosch can you take a look?

### DIFF
--- a/javaparser-testing/src/test/java/com/github/javaparser/ParserConfigurationTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/ParserConfigurationTest.java
@@ -1,0 +1,17 @@
+package com.github.javaparser;
+
+import com.github.javaparser.ast.CompilationUnit;
+import org.junit.Test;
+
+import static com.github.javaparser.Providers.provider;
+import static org.junit.Assert.assertFalse;
+
+public class ParserConfigurationTest {
+    @Test
+    public void storeNoTokens() {
+        ParseResult<CompilationUnit> result = new JavaParser(new ParserConfiguration().setStoreTokens(false)).parse(ParseStart.COMPILATION_UNIT, provider("class X{}"));
+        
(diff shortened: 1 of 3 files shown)


javaparser-testing/src/test/java/com/github/javaparser/ParserConfigurationTest.java
```diff
@@ -0,0 +1,17 @@
...
+import static com.github.javaparser.Providers.provider;
+import static org.junit.Assert.assertFalse;
+
+public class ParserConfigurationTest {
+    @Test
+    public void storeNoTokens() {
+        ParseResult<CompilationUnit> result = new JavaParser(new ParserConfiguration().setStoreTokens(false)).parse(ParseStart.COMPILATION_UNIT, provider("class X{}"));
+        
+        assertFalse(result.getTokens().isPresent());
+        assertFalse(result.getResult().get().getTokenRange().isPresent());
```
arturbosch: Maybe add something like `result.getResult.get().getNodesByType(Node.class).stream().allMatch(node -> !node.getTokenRange().isPresent())` (with the premise of extending the test code :))?

arturbosch (commented):
The changes look okay to me.

javaparser-testing/src/test/java/com/github/javaparser/ParserConfigurationTest.java
matozoid: Done!

matozoid:
This one is still not working well...

arturbosch:
What do you mean? Are there still JavaToken objects?

matozoid:
No, the grammar assumes there are always token ranges and runs into endless nullpointers when parsing more complex classes. Working on it, but it will not only drop `TokenRange` but also `Range`.

PR #1152, opened 2017-09-20
Cache the newline split pattern in SourcePrinter for performance reasons

arturbosch:
More information on java not caching patterns: https://stackoverflow.com/questions/13420321/does-pattern-compile-cache

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
@@ -24,6 +24,7 @@ package com.github.javaparser.printer;
 import java.text.Normalizer;
 import java.util.Deque;
 import java.util.LinkedList;
+import java.util.regex.Pattern;
 
 import com.github.javaparser.Position;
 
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
```diff
@@ -78,14 +79,19 @@ public SourcePrinter println() {
...
-    
+
     private StringBuilder bufAppend(final String arg) {
         updateCursor(arg);
         return buf.append(arg);
     }
 
+    private static final Pattern SPLIT_NEWLINE_PATTERN;
+    static {
+        SPLIT_NEWLINE_PATTERN = Pattern.compile("\r\n|\r|\n");
```
matozoid: Egh, end of line characters defined somewhere in the code. Will have to look into that later.

matozoid: Won't a field initializer do instead of a static block?

arturbosch: A static block does only initialize the field if this classes is loaded. If the user does not use it, it won't be loaded at all.

matozoid: How does the performance gain compare to the loss of readability?

arturbosch: Your right, the static block does do nothing in this case as this is not a static class and the pattern is always used.

arturbosch: I haven't found such string somewhere in the code base. Tell me if you find something :)

matozoid:
Hmmm, if I look at the code it seems like there is no point in all this end of line matching. All the sourceprinter ever receives is commands to print a piece of text without end of lines, and only the println method ever insert an EOL. @tarilabs - can you explain the reasoning behind this?

tarilabs:
The change suggested by @arturbosch is correct; and sorry I missed that in the first place :(

@matozoid IIRC there is no check that the passed argument does not contain a newline. Additionally, please consider for instance when multi-line comments is supplied to the printer, it would "miss" some new lines if I remove this check (unless I overlooked something).

matozoid:
@tarilabs ahhh, you're right, comments can interfere, and they can have different kinds of EOL characters. Good thing I'm refactoring all that comment stuff :-) Thanks for the answer!

PR #1153, opened 2017-09-20
Disconnect LexicalPreservingPrinter and preparation

matozoid:
Hey @ftomassetti - if I store the NodeText directly on the nodes instead of keeping a map inside the printer, I can disconnect the setup from the printing. Somehow all methods turned static now :-)


ftomassetti:
Look awesome: I looked in all the changes and everything seems great. I wonder if it would also reduce memory & elaboration time. Good call!

ftomassetti (approved):

matozoid:
@ftomassetti I think it did improve performance because of a silly thing: `Node` was used as the key of the map, so it got a lot of calls to hashcode and equals, and those visit the node and all of its children :-O - but that could also have been fixed by using the `VisitorMap`.

Issue #1154, opened 2017-09-20
Import methods from Java Symbol Solver's "Type"


matozoid:
These are possible:

```java
    boolean isArray();
    int arrayLevel();
    boolean isPrimitive();
    boolean isNull();
    boolean isReference(); // This is a confusing one...
    boolean isConstraint();
    boolean isReferenceType();
    boolean isVoid();
    boolean isTypeVariable();
    boolean isWildcard();
    ArrayType asArrayType();
    ReferenceType asReferenceType();
    TypeParameterDeclaration asTypeParameter();
    PrimitiveType asPrimitive();
    Wildcard asWildcard();
```
... these I don't know. @ftomassetti ?
```java
    String describe();
    Type replaceTypeVariables(TypeParameterDeclaration tp, Type replaced, Map<TypeParameterDeclaration, Type> inferredTypes);
    Type replaceTypeVariables(TypeParameterDeclaration tp, Type replaced);
    boolean mention(List<TypeParameterDeclaration> typeParameters);
    boolean isAssignableBy(Type other);
    TypeVariable asTypeVariable();
    LambdaConstraintType asConstraintType();
```

Also the question is how much this helps to integrate JSS. Not much I think?

ftomassetti:
Agreed that `isReference` and `isReferenceType` are confusing :)

Maybe we can start looking for a common interface defining a Type and see how to move from there.

The problem is that "resolved types" will have some extra methods (like, get all ancestors). I do not know if the implementation in JP should throw exception for those, or maybe look into the data to see if they find a delegate (from JSS). Something like:

```
List<Type> getAllAncestors() {
   if (data.get(TypeSolver).isPresent()) {
    return data.get(TypeSolver).getAllAncestors();
   }
  throw new IllegalStateException("Symbol resolution not setup");
}
```

Or maybe we could have a Type interface and a SolvedType interface, which would extend Type.

ElvisMachuca:
I have an issue with statement "if (data.get(TypeSolver).isPresent()){ ...}"

Can anyone say if TypeSolver has had an update? --Not Referring to the validity of "TypeNotFound" but rather, the input that is missing....is it harmonic or gaussian, then a single parser for this case, has to make the distintion of its solver.

Rewritten:

List<Type> getAllAncestors() {
   InputTypeStream ts = data.resolve().getInputTypeStream();
   if (data.get(ts.getChannel()).isPresent()) {
    return data.get(ts.getChannel()).getAllAncestors();
[...]

PR #1155, opened 2017-09-20
Generate typecasting code

matozoid:
This generates `isSomeType()` `asSomeType()` and `ifSomeType(Consumer)` methods for most of the type hierarchies of `Node`. `Node` itself is excluded since it would become enormously bloated with methods.

See #1154

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
@@ -6,6 +6,7 @@ import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.metamodel.BaseNodeMetaModel;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
 import com.github.javaparser.utils.Log;
+import com.github.javaparser.utils.Pair;
 import com.github.javaparser.utils.SourceRoot;
 
 import java.io.IOException;
@@ -25,18 +26,23 @@ public abstract class NodeGenerator extends Generator {
     public final void generate() throws Exception {
         Log.info("Running %s", getClass().getSimpleName());
         for (BaseNodeMetaModel nodeMetaModel : JavaParserMetaModel.getNodeMetaModels()) {
-            CompilationUnit nodeCu = sourceRoot.parse(nodeMetaModel.getPackageName(), nodeMetaModel.getTypeName() + ".java");
-            ClassOrInterfaceDeclaration nodeCoid = nodeCu.getClassByName(nodeMetaModel.getTypeName()).orElseThrow(() -> new IOException("Can't find class"));
-            generateNode(nodeMetaModel, nodeCu, nodeCoid);
+            Pair<CompilationUnit, ClassOrInterfaceDeclaration> result = parseNode(nodeMetaModel);
+            generateNode(nodeMetaModel, result.a, result.b);
         }
         after();
     }
 
+    protected Pair<CompilationUnit, ClassOrInterfaceDeclaration> parseNode(BaseNodeMetaModel nodeMetaModel) throws IOException {
+        CompilationUnit nodeCu = sourceRoot.parse(nodeMetaModel.getPackageName(), nodeMetaModel.getTypeName() + ".java");
+        ClassOrInterfaceDeclaration nodeCoid = nodeCu.getClassByName(nodeMetaModel.getTypeName()).orElseThrow(() -> new IOException("Can't find class"));
+        return new Pair<>(nodeCu, nodeCoid);
+    }
+    
     protected void after() throws Exception {
 
     }
 
-    protected abstract void generateNode(BaseNodeMetaModel nodeMetaModel, CompilationUnit nodeCu, ClassOrInterfaceDeclaration nodeCoid);
+    protected abstract void generateNode(BaseNodeMetaModel nodeMetaModel, CompilationUnit nodeCu, ClassOrInterfaceDeclaration nodeCoid) throws Exception;
 
     /**
      * Utility method that looks for a method or constructor with an identical signature as "callable" and replaces it
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
@@ -239,78 +239,45 @@ public abstract class BodyDeclaration<T extends BodyDeclaration<?>> extends Node
 
     @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
     public void ifAnnotationDeclaration(Consumer<AnnotationDeclaration> action) {
-        if (isAnnotationDeclaration()) {
-            action.accept(asAnnotationDeclaration());
-        }
     }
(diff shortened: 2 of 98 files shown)


matozoid:
Maybe `Node` should have these methods, but only for direct subclasses to keep things under control.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
```diff
@@ -25,18 +26,23 @@ protected NodeGenerator(SourceRoot sourceRoot) {
...
+        ClassOrInterfaceDeclaration nodeCoid = nodeCu.getClassByName(nodeMetaModel.getTypeName()).orElseThrow(() -> new IOException("Can't find class"));
+        return new Pair<>(nodeCu, nodeCoid);
+    }
+    
     protected void after() throws Exception {
 
     }
 
-    protected abstract void generateNode(BaseNodeMetaModel nodeMetaModel, CompilationUnit nodeCu, ClassOrInterfaceDeclaration nodeCoid);
+    protected abstract void generateNode(BaseNodeMetaModel nodeMetaModel, CompilationUnit nodeCu, ClassOrInterfaceDeclaration nodeCoid) throws Exception;
```
ftomassetti, line 45: I am not a fan of checked exceptions and they seem to being less used.

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
```diff
@@ -124,4 +125,191 @@ public boolean replace(Node node, Node replacementNode) {
...
     }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public boolean isAnnotationDeclaration() {
+        return false;
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public AnnotationDeclaration asAnnotationDeclaration() {
+        return (AnnotationDeclaration) this;
```
ftomassetti: should this throw an exception?

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
```diff
@@ -124,4 +125,191 @@ public boolean replace(Node node, Node replacementNode) {
...
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public boolean isAnnotationMemberDeclaration() {
+        return false;
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public AnnotationMemberDeclaration asAnnotationMemberDeclaration() {
+        return (AnnotationMemberDeclaration) this;
```
ftomassetti: should this throw an exception? [the same for all the other similar cases]

javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
```diff
@@ -168,4 +167,45 @@ public boolean replace(Node node, Node replacementNode) {
...
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public JavadocComment asJavadocComment() {
+        return (JavadocComment) this;
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
+    public void ifBlockComment(Consumer<BlockComment> action) {
+        if (isBlockComment()) {
```
ftomassetti: this one could be empty I guess

ftomassetti (commented):
I would expect these methods to be declared at an high level (on Node, Statement or Expression). To be implemented with the default behavior (return false, throw exception, do nothing) and then overriden in specific cases

ftomassetti:
I think these can be useful and in JSS we adopt this style, I have just a few comments on the implementation and I like the use of consumers. I should do the same in JSS

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
matozoid: It will throw a `ClassCastException` when `this` is not an `AnnotationDeclaration`. Is that okay? I saw you were explicitly using a different exception, but I thought `ClassCastException` was perfectly applicable.

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
matozoid: Same comment then :-)

javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
matozoid: Oh right, that's a nice approach :-)

matozoid:
@ftomassetti I also like the consumers, but I was also wondering if this is the best functional approach. Just thought that `asXyz()` could be implemented with an `Optional` : `Optional<IfStmt> asIfStmt()` so you could do `abc.asIfStmt.map(i -> .....)`. It would also stop throwing exceptions... What do you think?

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
ftomassetti: I think ClassCastException would be acceptable, however I think the code could be slightly surprising: a causal reader could wonder why we try to cast if we are sure the cast is going to fail. We could make it explicit that this method will always fail

ftomassetti:
I like the current proposal, your alternative with Optional is also valid but I find the current one more obvious

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
matozoid: It's an easy way to say "whatever, all these exceptions stop the generator" so it practically turns checked exceptions into runtime ones ;-)

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
matozoid: Would you accept...
- base method only throws an `IllegalStateException` (because we *do* support this operation, but not in this specific state)
- specific node returns `this`.

So more or less what you have but another exception?

matozoid: Latest commit does that ^^^^

javaparser-core/src/main/java/com/github/javaparser/ast/comments/Comment.java
matozoid: Also implemented

javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
ftomassetti: Yes, that would be great

ftomassetti:
Looks good to me

matozoid:
Well, I doubt anybody else is going to add a comment, so here it goes.

PR #1167, opened 2017-09-26
Widen the statement you can parse with parseStatement

### LINKED ISSUE #1166: Invoking super constructor
Invoking the `super` constructor is not accepted, like so: 
```
constructorDeclaration.createBody()
                .addStatement("super();")
```
It will result in a `ParseProblemException`, because it expects that the `(` would be a `.` or `::`.

For now I can use a workaround, using `new ExplicitConstructorInvocationStmt()` and attach to the block statement

matozoid:
Fixes #1166


matozoid:
Not much going on here, just expanded the manual statement parsing rule to also include explicit constructor invocations :-)

PR #1169, opened 2017-09-27
Issue 1154 - Bridging JP & JSS

ftomassetti:
This thing snowballed... basically I brought declarations & types interfaces from JSS.

We have the concept of "ResolvedDeclarations" and "ResolvedTypes". ResolvedDeclarations can be obtained by JP declarations or by class files or through reflection.

How do we get a ResolvedDeclaration from a JP Declaration? Calling the resolve method. This resolve method will look into the data map of the node for a SymbolResolver which will do the mapping. 

Similarly we have ResolvedTypes, which have more features. They can also be obtained from JP type in the same way (calling the resolve method).

It is larger than expected but it compiles at least and it would probably be more obvious to use. Also, it would remove some names confusion (basically JSS stuff is prefixed with "Resolved").

What do you guys think?


javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
```diff
@@ -55,7 +58,8 @@
  *
  * @author Julio Vilmar Gesser
  */
-public final class ClassOrInterfaceType extends ReferenceType implements NodeWithSimpleName<ClassOrInterfaceType>, NodeWithAnnotations<ClassOrInterfaceType>, NodeWithTypeArguments<ClassOrInterfaceType> {
+public final class ClassOrInterfaceType extends ReferenceType<ResolvedReferenceType> implements NodeWithSimpleName<ClassOrInterfaceType>,
+        NodeWithAnnotations<ClassOrInterfaceType>, NodeWithTypeArguments<ClassOrInterfaceType> {
```
matozoid: Looks like the diff will be smaller if you run the code generator script.

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
```diff
@@ -55,7 +58,8 @@
  *
  * @author Julio Vilmar Gesser
  */
-public final class ClassOrInterfaceType extends ReferenceType implements NodeWithSimpleName<ClassOrInterfaceType>, NodeWithAnnotations<ClassOrInterfaceType>, NodeWithTypeArguments<ClassOrInterfaceType> {
+public final class ClassOrInterfaceType extends ReferenceType<ResolvedReferenceType> implements NodeWithSimpleName<ClassOrInterfaceType>,
```
matozoid: We're trying to get rid of the `<?>` stuff on nodes, I think this will reintroduce one :-(

javaparser-core/src/main/java/com/github/javaparser/ast/type/Type.java
```diff
@@ -288,4 +292,19 @@ public void ifVoidType(Consumer<VoidType> action) {
...
+        if (symbolResolver == null) {
+            throw new IllegalStateException("Symbol resolution not configured");
+        }
+        return symbolResolver;
+    }
+
+    private static final DataKey<SymbolResolver> SYMBOL_RESOLVER_KEY = new DataKey<SymbolResolver>() {
+        @Override
+        public int hashCode() {
+            return super.hashCode();
```
matozoid: Huh? :-D

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/AccessLevel.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.javaparser.resolution.declarations;
+
+/**
+ * @author Federico Tomassetti
+ */
+public enum AccessLevel {
```
matozoid: Didn't we have this one already?

ftomassetti: Probably, as a first step I just imported stuff from JSS, renamed things and made it compile, now it is time for polishing and removing duplication

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedUnionType.java
```diff
@@ -0,0 +1,48 @@
...
+ * A union type is defined in java as list of types separates by pipes.
+ *
+ * @author Federico Tomassetti
+ */
+public class ResolvedUnionType implements ResolvedType {
+    private List<ResolvedType> elements;
+
+    public ResolvedUnionType(Collection<ResolvedType> elements) {
+        if (elements.size() < 2) {
+            throw new IllegalArgumentException("An union type should have at least two elements. This has " + elements.size());
```
matozoid, line 40: Oh, I need that one too: https://github.com/javaparser/javaparser/issues/1170

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/AccessLevel.java
ftomassetti: Just removed and renamed the related interface `HasAccessLevel` to `HasAccessSpecifier`

javaparser-core/src/main/java/com/github/javaparser/ast/type/Type.java
ftomassetti: Removed :D

javaparser-testing/src/test/java/com/github/javaparser/ast/type/TypeTest.java
```diff
@@ -53,7 +53,7 @@ public void arrayType() {
         assertTrue(type.isArrayType());
         ArrayType arrayType = type.asArrayType();
         final ArrayType[] s = new ArrayType[1];
-        type.ifArrayType(t -> s[0] = t);
+        type.ifArrayType(t -> s[0] = (ArrayType)t);
```
matozoid: Huh? The point is that a cast isn't necessary

matozoid:
I think it's great :-)

How about storing the resolved type in the data field too? Or do we run into the known memory overflows then?

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
matozoid: Oh, if you need help with this, I have some strategies.

javaparser-testing/src/test/java/com/github/javaparser/ast/type/TypeTest.java
ftomassetti: Yes, but I could not convince the compiler that was the case... it kept telling me it was an Object, so I just made it happy to move forward this thing and get initial feedback, just in case we figured out immediately I was going in the wrong direction

matozoid: Hmmmm, that needs an issue, although I don't know how to reproduce it.

javaparser-core/src/main/java/com/github/javaparser/ast/type/ClassOrInterfaceType.java
ftomassetti: I am all ears :D

javaparser-testing/src/test/java/com/github/javaparser/ast/type/TypeTest.java
ftomassetti: If you remove it and it does compile it for you then it has probably something to do with my checkout (even if I have no idea what it could be)

ftomassetti:
We have all sort of caching and at the same time I would love to support the possibility of performing changes in the code and have the symbols to be re-resolved automatically. Probably we need to pick one of these two directions... or find some consistent approach anyway

javaparser-testing/src/test/java/com/github/javaparser/ast/type/TypeTest.java
matozoid: Yeah, I don't check in broken code ;-) Also, Travis is completely fine with it, and since it is a test I would prefer if you don't change it :-)

ftomassetti: I removed the hack and it is failing also on the server. Really strange...

ftomassetti: This works:
```
arrayType.ifArrayType(t -> s[0] = t);
```

but this does not:
```
type.ifArrayType(t -> s[0] = t);
```

However both versions of ifArrayType have the same signature...

matozoid:
About (ArrayType) - I have exactly the same Java version as Travis. I'm baffled.

javaparser-testing/src/test/java/com/github/javaparser/ast/type/TypeTest.java
ftomassetti: This also works:
```
        Consumer<ArrayType> c = t -> s[0] = t;
        type.ifArrayType(c);
```

ftomassetti: Oh my...

This works:
```
    @Test
    public void arrayType() {
        Type<?> type = parseType("int[]");
        assertTrue(type.isArrayType());
        ArrayType arrayType = type.asArrayType();
        final ArrayType[] s = new ArrayType[1];
        type.ifArrayType(t -> s[0] = t);
        assertNotNull(s[0]);
    }
```

This does not:
```
    @Test
    public void arrayType() {
        Type type = parseType("int[]");
        assertTrue(type.isArrayType());
        ArrayType arrayType = type.asArrayType();
        final ArrayType[] s = new ArrayType[1];
        type.ifArrayType(t -> s[0] = t);
        assertNotNull(s[0]);
    }
```

`Type` != `Type<?>`...

ftomassetti:
Fixing the ArrayType issue by making Type NOT parametric: two birds with one stone

matozoid:
@ftomassetti I noticed `UnknownType` is not resolvable - but that's the type of a lambda parameter! That should get resolved with type inference, shouldn't it?

matozoid:
Am I understanding right that the solver gets linked to each and every node? Or only types? It's not really something that belongs to a specific node, so I'm unsure this is the right approach.

matozoid:
I was going to solve the type parameter problem for `Type`, but I see you already got it :-)

ftomassetti:
About unknown type, you are right

ftomassetti:
About the solver, every node on which we can invoke "solve" should be able to get the solver in some way. Maybe we can just store it in the root node or somewhere else where every node that needs it can find it

matozoid:
Uhuh, I've been trying to solve the problem where a node needs some kind of global setting before. Never found a good solution. Storing it in the `CompilationUnit` may be our best bet.

ftomassetti:
In the meantime I am also experimenting on the JSS side, by removing the classes ported to JP in this patch and adapt as needed. It is taking some effort

matozoid:
Here, I got you a `findCompilationUnit` method to get to the place to store stuff quickly :-)

How are you planning to set it initially though?

ftomassetti:
Ok, all the necessary conversions have been done in JSS (branch jpsmoothing) so that it now all tests are passing now.

On the JP side I think we need to add those solved methods to extract the information from the SymbolSolver. However JP will not know anything about how the SymbolSolver is built and injected. This implementation will stay in JSS. Maybe in JSS we could expose a facade that call JavaParser and then inject also the solver before returning the AST.

ftomassetti:
I think this will need a few iterations to complete the API. We should probably revise and merge this, implement necessary support in JSS (what is in branch jpsmoothing + code to inject the SymbolSolver), and then comes back here and complete the API, for example to calculate the type of expressions.

ftomassetti:
Examples on tests in JSS using the new API:
```java
    @Test
    public void resolveMethodDeclaration() throws ParseException {
        TypeSolver typeSolver = new ReflectionTypeSolver();

        CompilationUnit cu = parseSample("SymbolResolverExample");
        new JavaSymbolSolver(typeSolver).inject(cu);

        MethodDeclaration methodDeclaration = cu.getClassByName("A").get().getMethods().get(0);
        ResolvedMethodDeclaration resolvedMethodDeclaration = methodDeclaration.resolve();
        assertEquals("foo", resolvedMethodDeclaration.getName());
        assertEquals("A[]", resolvedMethodDeclaration.getReturnType().describe());
        assertEquals("java.lang.String[]", resolvedMethodDeclaration.getParam(0).getType().describe());
        assertEquals("int[]", resolvedMethodDeclaration.getParam(1).getType().describe());
    }

    @Test
    public void resolveArrayType() throws ParseException {
        TypeSolver typeSolver = new ReflectionTypeSolver();

        CompilationUnit cu = parseSample("SymbolResolverExample");
        new JavaSymbolSolver(typeSolver).inject(cu);

        MethodDeclaration methodDeclaration = cu.getClassByName("A").get().getMethods().get(0);
        ResolvedType resolvedType = methodDeclaration.getType().resolve();
        assertEquals("A[]", resolvedType.describe());
    }
```

matozoid:
Looks cool :-) I'll pull it and go through it later this week.

ftomassetti:
Great! I will then merge & tests things on the JSS side and then come back with more methods to add to the API, all backed by the `getSymbolSolver()` method

matozoid:
I think this one could use a link to something that describes how to set up stuff: `throw new IllegalStateException("Symbol resolution not configured");`

ftomassetti:
Yes, most definitely

matozoid:
What do you think about adding an option to `ParseConfiguration` so that `JavaParser` will add a specific `TypeSolver` to everything it parses?

ftomassetti:
Yes, it makes a lot of sense

PR #1174, opened 2017-09-27
Work with jdk9

### LINKED ISSUE #1172: Ensure compile with JDK 9
We should check if Travis supports it already

matozoid:
Fixes #1172 

It actually didn't compile with JDK 9 because of an outdated plugin...

Took the opportunity to update a few other plugins as well, including the JUnit plugin!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: java
 install: mvn install -DskipTests=true
 script: mvn test
 jdk:
+  - oraclejdk8
   - oraclejdk9
 notifications:
   webhooks:
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: java
 install: mvn install -DskipTests=true
 script: mvn test
 jdk:
-  - oraclejdk8
+  - oraclejdk9
 notifications:
   webhooks:
     urls:
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -79,7 +79,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-enforcer-plugin</artifactId>
-                <version>1.4.1</version>
+                <version>3.0.0-M1</version>
                 <executions>
                     <execution>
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -3,7 +3,7 @@ language: java
 install: mvn install -DskipTests=true
 script: mvn test
 jdk:
-  - oraclejdk8
+  - oraclejdk9
```
ftomassetti: I would not replace it but add it: Travis should be able to run the tests using both the JDKs so we verify it stays compatible with JDK8 and JDK9 at the same time

javaparser-core/pom.xml
```diff
@@ -79,7 +79,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-enforcer-plugin</artifactId>
-                <version>1.4.1</version>
+                <version>3.0.0-M1</version>
```
ftomassetti, line 82: This M1 should scare us?

.travis.yml
matozoid: How would I do that?

javaparser-core/pom.xml
matozoid: Not as much as 1.4.2-jenkins1 ;-) After some hunting for release notes I found this: http://maven.40175.n5.nabble.com/ANN-Apache-Maven-Enforcer-Plugin-3-0-0-M1-Released-td5911744.html

.travis.yml
ftomassetti: I think you can just add the second line without removing the first one

matozoid: It worked :-O

javaparser-core/pom.xml
matozoid: Anyway, the alternatives are worse in my opinion: either take out the JDK 8 API enforcement for core, or don't compile with JDK9 at all.

matozoid:
Assuming okay, @ftomassetti please comment if you disagree so I can do more.

ftomassetti:
It seems perfect to me

PR #1178, opened 2017-10-07
Injection symbol resolver

ftomassetti:
Let's make possible to configure the parser to inject a SymbolResolver in all parsed CompilationUnits

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -45,6 +45,7 @@ import com.github.javaparser.resolution.SymbolResolver;
 import java.io.*;
 import java.nio.charset.Charset;
 import java.nio.file.Path;
+import java.util.Optional;
 
 import static com.github.javaparser.ParseStart.*;
 import static com.github.javaparser.Problem.PROBLEM_BY_BEGIN_POSITION;
@@ -340,7 +341,6 @@ public final class JavaParser {
     private static <T extends Node> T simplifiedParse(ParseStart<T> context, Provider provider) {
         ParseResult<T> result = new JavaParser(staticConfiguration).parse(context, provider);
         if (result.isSuccessful()) {
-            considerInjectingSymbolResolver(result, staticConfiguration);
             return result.getResult().get();
         }
         throw new ParseProblemException(result.getProblems());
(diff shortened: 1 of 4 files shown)


javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -336,6 +340,7 @@ public static Statement parseStatement(final String statement) {
     private static <T extends Node> T simplifiedParse(ParseStart<T> context, Provider provider) {
         ParseResult<T> result = new JavaParser(staticConfiguration).parse(context, provider);
         if (result.isSuccessful()) {
+            considerInjectingSymbolResolver(result, staticConfiguration);
```
matozoid: This is already done in the other method.

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -124,4 +125,18 @@ public ParserConfiguration setLexicalPreservationEnabled(boolean lexicalPreserva
     public boolean isLexicalPreservationEnabled() {
         return lexicalPreservationEnabled;
     }
+
+    /**
+     * Retrieve the SymbolResolver to be used while parsing, if any.
+     */
+    public SymbolResolver getSymbolResolver() {
+        return symbolResolver;
```
matozoid: No nulls please, `Optional.ofNullable` to the rescue :-)

matozoid: Background: not a single method in JavaParser returns null anymore since 3.0.0. Mixing `Optional` and null will break the rule, and people would again have to null-check everything.

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -124,4 +125,18 @@ public ParserConfiguration setLexicalPreservationEnabled(boolean lexicalPreserva
...
+    }
+
+    /**
+     * Set the SymbolResolver to be injected while parsing.
+     */
+    public void setSymbolResolver(SymbolResolver symbolResolver) {
```
matozoid: return this; for chaining.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -665,4 +665,5 @@ public boolean replace(Node node, Node replacementNode) {
         }
         return super.replace(node, replacementNode);
     }
+
```
matozoid, line 668: :-P

matozoid (changes requested):
This review feature sucks!

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
ftomassetti: Right, my bad

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
ftomassetti: Ok

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
ftomassetti: Right, doing

matozoid (approved):

ftomassetti:
Given it is approved I will merge it.
Thanks!

matozoid:
Yep, you know when the next release is :-)

PR #1181, opened 2017-10-08
Small ParseResult improvements

matozoid:
A nice improvement found while writing a new article.


javaparser-core/src/main/java/com/github/javaparser/ParseResult.java
```diff
@@ -76,6 +77,15 @@ public boolean isSuccessful() {
     }
 
     /**
+     * Calls the consumer with the result if parsing was succesful.
```
ftomassetti, line 80: Nice

PR #1192, opened 2017-10-16
[DRAFT] Convert Javadoc to Markdown

matozoid:
I'm interested in feedback on a little dilemma: on one hand, it's easier to write Javadoc this way. On the other, it's really easy to completely destroy the formatting by pressing "format file" in an IDE. Hm.


ftomassetti:
Personally I love Markdown, however not many developers are used to use Markdown in Javadoc comments so this is an issue for a project that is "popular" and getting many contributions from occasional contributors. It is also true that few of those contributors write much Javadoc anyway or write Javadoc that requires special formatting so maybe this is not an issue...

matozoid:
@ftomassetti the nice thing about this plugin is that you can mix with standard Javadoc, so that's not too big an issue.

matozoid:
Too much risk of accidental breakage due to source code formatting

PR #1193, opened 2017-10-17
Use symbol resolver

ftomassetti:
Using the SymbolResolver in more cases:
* to find declarations
* to get resolved types from types
* to calculate the types of expressions

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
@@ -22,7 +22,15 @@
 package com.github.javaparser.resolution;
 
 import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.resolution.types.ResolvedType;
 
 public interface SymbolResolver {
+    /**
+     * Perform the resolution. For a reference it would find the corresponding
+     * declaration while for types it would find the corresponding resolved types.
+     */
     <T> T resolve(Node node, Class<T> resultClass);
+
+    ResolvedType calculateType(Expression expression);
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
@@ -23,14 +23,20 @@ package com.github.javaparser.resolution;
 
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.resolution.types.ResolvedType;
 
 public interface SymbolResolver {
     /**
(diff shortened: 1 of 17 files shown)


javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
```diff
@@ -22,7 +22,21 @@
...
+    /**
+     * For a reference it would find the corresponding
+     * declaration.
+     */
+    <T> T resolveDeclaration(Node node, Class<T> resultClass);
+
+    /**
+     * For types it would find the corresponding resolved types.
+     */
+    <T> T toResolvedType(Type javaparserType, Class<T> resultClass);
```
matozoid, line 39: I feel that `resolveType` means the same thing as `toResolvedType` and is more in line with `resolveDeclaration`.

javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
```diff
@@ -22,7 +22,21 @@
...
+
+    ResolvedType calculateType(Expression expression);
```
matozoid, line 41: ... or maybe I just don't understand this very well ;-)

javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
ftomassetti: I am not very good with naming, especially in English. The difference I see is that I can have a reference (es. "foo") and from that I want to find the correspinding declaration (es. "private int foo;"). This is what `resolveDeclaration` does. `toResolvedType` instead takes a JP type and give me a JSS type. Sure, to build the JSS type it does all the necessary resolving. However to me it seems more like a conversion. Making we can calling `correspondingResolvedType` or something like that.

javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
ftomassetti: ...or maybe I am just terrible at giving names :D

javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
matozoid: I'm mostly thinking about "discoverability" - if I type a variable and a dot in my IDE and scroll through the autosuggested methods, the methods from this interface would not show up next to each other.

But that's a small thing. Keep it as it is then :-)

matozoid (approved):

javaparser-core/src/main/java/com/github/javaparser/resolution/SymbolResolver.java
ftomassetti: I could make the two methods start with get (getDeclaration, getResolvedType) or with to (toDeclaration, toResolvedType)

ftomassetti:
Well, given it is approved let's merge it

PR #1194, opened 2017-10-17
Redo receiver parameters

### LINKED ISSUE #1129: Improve receiver parameter support
Here's an idea to improve our current hacky receiver parameter support, inspired by @Bohdankm22:
- receiver parameters currently influence the method signature. They should be filtered out when creating a `CallableDeclaration.Signature`
- we could add a helper method or two that parses the receiver parameter (can we have more than one?) correctly (it is a qualified name, not a simple name like the others) and returns it.
- taking the receiver parameter out of the parameter list would mean losing the position in that list. This may lead to problems for lexical preservation. @ftomassetti ?

matozoid:
Fixes #1129 

Receiver parameter support was a bit of a hack, and thanks to new insights from @Bohdankm22 I've managed to make a good implementation. It breaks the API, but I don't think anybody is using this anyway :-D

The receiver parameter is always the first parameter of a method. It is always called "`this`" and may be qualified. It can have annotations. JavaParser now sees it as a separate node, separate from `Parameter`, and stores it in an optional field in `MethodDeclaration`. This is much better than the previous approach where it was stored as a parameter, forcing a potentially qualified name into a parameter's simple name.

Arbitrary decisions: the type of the receiver parameter can have array brackets, but it cannot have ellipsis (`...`) or modifiers.


javaparser-core/src/main/java/com/github/javaparser/ast/ArrayCreationLevel.java
```diff
@@ -76,11 +76,13 @@ public ArrayCreationLevel(TokenRange tokenRange, Expression dimension, NodeList<
     }
 
     @Override
+    @Generated("com.github.javaparser.generator.core.node.AcceptGenerator")
```
ftomassetti, line 81: I would love a tool to identify these changes and hide them :D

javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
```diff
@@ -67,45 +67,50 @@
...
+        this(null, modifiers, new NodeList<>(), new NodeList<>(), type, new SimpleName(name), parameters, new NodeList<>(), new BlockStmt(), null);
     }
 
     @AllFieldsConstructor
-    public MethodDeclaration(final EnumSet<Modifier> modifiers, final NodeList<AnnotationExpr> annotations, final NodeList<TypeParameter> typeParameters, final Type type, final SimpleName name, final NodeList<Parameter> parameters, final NodeList<ReferenceType> thrownExceptions, final BlockStmt body) {
-        this(null, modifiers, annotations, typeParameters, type, name, parameters, thrownExceptions, body);
+    public MethodDeclaration(final EnumSet<Modifier> modifiers, final NodeList<AnnotationExpr> annotations, final NodeList<TypeParameter> typeParameters, final Type type, final SimpleName name, final NodeList<Parameter> parameters, final NodeList<ReferenceType> thrownExceptions, final BlockStmt body, ReceiverParameter receiverParameter) {
+        this(null, modifiers, annotations, typeParameters, type, name, parameters, thrownExceptions, body, receiverParameter);
```
ftomassetti: I wonder if it would make sense to exclude receiverParameter from the all fields constructor because it is so rarely used

javaparser-core/src/main/java/com/github/javaparser/ast/body/ReceiverParameter.java
```diff
@@ -0,0 +1,221 @@
...
+    public ReceiverParameter(Type type, String name) {
+        this(null, new NodeList<>(), type, new Name(name));
+    }
+
+    @AllFieldsConstructor
+    public ReceiverParameter(NodeList<AnnotationExpr> annotations, Type type, Name name) {
+        this(null, annotations, type, name);
+    }
+
+    /**This constructor is used by the parser and is considered private.*/
```
ftomassetti: Should we change the style of the comment?

ftomassetti (commented):
I am too tired to do a proper review but I what I saw seems good. I just added a few comments

javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
matozoid: That constructor is used to generate te all field + range constructor. It needs to have all fields.

But I did unneccesarily break backwards compatibility here, so I should add a constructor with the previous signature.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ReceiverParameter.java
matozoid: In what way? More end-of-lines you mean?

ftomassetti: Yes

javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
ftomassetti: Makes sense

javaparser-core/src/main/java/com/github/javaparser/ast/ArrayCreationLevel.java
matozoid: There must be about 400 of them? :-D

javaparser-core/src/main/java/com/github/javaparser/ast/body/ReceiverParameter.java
matozoid: Tell me how it works :-)

This is the current code:
`.setJavadocComment("This constructor is used by the parser and is considered private.");`

matozoid: Made a hack:

`.setJavadocComment("\n     * This constructor is used by the parser and is considered private.\n     ");`

PR #1195, opened 2017-10-17
Add functional visitor methods

### LINKED ISSUE #804: Add a lambda visitor syntax
Add a method with approximate syntax `apply(Class<T extends Node>, Function<T, ?> f)` to `Node` which visits the whole tree and calls the function for all nodes matching the class.

This could be implemented with a more generic  `apply(Function<T, ?> f)`.

Maybe find the correct name for this method, and if there is an existing approach in the functional programming world.

matozoid:
Fixes #804 

I tried to define a small set of generally useful method for traversing the AST with lambda syntax.

Thoughts:
- there is no traversing to the parent
- there is no way to stop traversing
- there are two `find`s that return all matches but no `findOne` that returns only the first match
- the existing `getChildNodesByType` is close to this functionality, so I deprecated it and added `find(Class)` - which is a specific but common case of `find(Class, Predicate)`.
- either `walk(Consumer)` or `walk(Class, Consumer)` is redundant. The first is the general case, the second the more specific, but probably more common case. I would say we keep both.
- the argument that normal visitors have (`arg`) is not really needed here, since it's common to use variables defined outside the visitor function (as seen in the unit tests.)
- it would be nice to have some kind of `map` function. I guess that would be a kind of `ModifierVisitor`


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +678,50 @@ protected SymbolResolver getSymbolResolver() {
...
+    /**
+     * Walks the AST, calling the consumer for every node.
+     */
+    public void walk(Consumer<Node> consumer) {
+        new TreeVisitor() {
+            @Override
+            public void process(Node node) {
+                consumer.accept(node);
+            }
+        }.visitPreOrder(this);
```
ftomassetti: should be the order (preOrder/postOrder) a flag?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +678,50 @@ protected SymbolResolver getSymbolResolver() {
...
+     */
+    public <T extends Node> void walk(Class<T> nodeType, Consumer<T> consumer) {
+        new TreeVisitor() {
+            @Override
+            public void process(Node node) {
+                if (nodeType.isInstance(node)) {
+                    consumer.accept(nodeType.cast(node));
+                }
+            }
+        }.visitPreOrder(this);
```
ftomassetti: should be the order (preOrder/postOrder) a flag?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +678,50 @@ protected SymbolResolver getSymbolResolver() {
...
+    }
+
+    /**
+     * Walks the AST, returning the all nodes of type "nodeType".
+     */
+    public <T extends Node> List<T> find(Class<T> nodeType) {
```
ftomassetti: Maybe findAll would make it clear that the result is a list

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeTest.java
```diff
@@ -327,17 +329,47 @@ public void removingTheSecondOfAListOfIdenticalStatementsDoesNotMessUpTheParents
...
-    
+
     @Test
     public void cantFindCompilationUnit() {
         VariableDeclarator x = new VariableDeclarator();
         assertFalse(x.findCompilationUnit().isPresent());
     }
+
+    @Test
+    public void walk1() {
```
ftomassetti: Maybe simpleWalk or genericWalk

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeTest.java
```diff
@@ -327,17 +329,47 @@ public void removingTheSecondOfAListOfIdenticalStatementsDoesNotMessUpTheParents
...
+        Expression e = parseExpression("1+1");
+        StringBuilder b = new StringBuilder();
+        e.walk(n -> b.append(n.toString()));
+        assertEquals("1 + 111", b.toString());
+    }
+
+    @Test
+    public void walk2() {
```
ftomassetti: maybe classSpecificWalk

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeTest.java
```diff
@@ -327,17 +329,47 @@ public void removingTheSecondOfAListOfIdenticalStatementsDoesNotMessUpTheParents
...
+        Expression e = parseExpression("1+1");
+        StringBuilder b = new StringBuilder();
+        e.walk(IntegerLiteralExpr.class, n -> b.append(n.toString()));
+        assertEquals("11", b.toString());
+    }
+
+    @Test
+    public void find1() {
```
ftomassetti: maybe conditionalFind

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeTest.java
```diff
@@ -327,17 +329,47 @@ public void removingTheSecondOfAListOfIdenticalStatementsDoesNotMessUpTheParents
...
+        Expression e = parseExpression("1+2+3");
+        List<IntegerLiteralExpr> ints = e.find(IntegerLiteralExpr.class, n -> n.asInt() > 1);
+        assertEquals("[2, 3]", ints.toString());
+    }
+
+    @Test
+    public void find2() {
```
ftomassetti: maybe unconditionalFind

ftomassetti (approved):
good stuff, now we just need to adjust the book :D

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: Maybe, I didn't do that right now because it would make code less concise, assuming that pre-order is the most common case. You can always use the `TreeVisitor` for others. But I could make another overloaded method :-)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: Since I'm thinking of adding find(One), you are right.

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeTest.java
matozoid: Sorry, lazy! :-D

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/TreeVisitor.java
```diff
@@ -23,57 +23,145 @@
...
-     *
      * @see <a href="https://en.wikipedia.org/wiki/Pre-order">Pre-order traversal</a>
      */
     public void visitPreOrder(Node node) {
-        process(node);
-        new ArrayList<>(node.getChildNodes()).forEach(this::visitPreOrder);
+        new PreOrderIterator(node).forEachRemaining(this::process);
+    }
+
+    public static class BreadthFirstIterator implements Iterator<Node> {
```
matozoid: @ftomassetti I turned the tree walking algorithms into iterators so you don't have to visit every node when you only wanted a few, like for a `findFirst`.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +683,91 @@ protected SymbolResolver getSymbolResolver() {
     // We need to expose it because we will need to use it to inject the SymbolSolver
     public static final DataKey<SymbolResolver> SYMBOL_RESOLVER_KEY = new DataKey<SymbolResolver>() {
     };
+
+
+    public enum TreeTraversal {
+        PREORDER, BREADTHFIRST, POSTORDER
+    }
+
+    public Iterator<Node> treeIterator(TreeTraversal traversal) {
```
matozoid: @ftomassetti a method to pick an iterator based on traversal algorithm

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +683,91 @@ protected SymbolResolver getSymbolResolver() {
...
+                return new TreeVisitor.PostOrderIterator(this);
+            case PREORDER:
+                return new TreeVisitor.PreOrderIterator(this);
+            default:
+                throw new IllegalArgumentException("Unknown traversal choice.");
+        }
+    }
+    
+    public Iterable<Node> treeIterable(TreeTraversal traversal) {
+        return () -> treeIterator(traversal);
```
matozoid: @ftomassetti an iterable is usable in a for-each loop, so I added this. Funky syntax!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +683,91 @@ protected SymbolResolver getSymbolResolver() {
...
+    }
+    
+    public Stream<Node> treeStream(TreeTraversal traversal) {
+        return StreamSupport.stream(Spliterators.spliteratorUnknownSize(treeIterator(traversal), NONNULL | DISTINCT), false);
```
matozoid: An iterator can be turned into a stream, although it won't be parrallellizable.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +683,91 @@ protected SymbolResolver getSymbolResolver() {
...
+    }
+    
+    /**
+     * Walks the AST, calling the consumer for every node.
+     */
+    public void walk(Consumer<Node> consumer) {
+        treeIterable(TreeTraversal.PREORDER).forEach(consumer);
```
matozoid: @ftomassetti nice and short.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: nice!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: Quite classy code!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: Perhaps you could move the logic into the TreeTraversal enum. Like:

```
enum TreeTraversal {
 BREADTHFIRST {
   Iterator<Node> iterator(Node node) { return TreeVisitor.BreadthFirstIterator(node); }
}
;
abstract Iterator<Node> iterator(Node node);
}
```

```
public Iterator<Node> treeIterator(TreeTraversal traversal) {
return traversal.iterator(this);
}
```

This would avoid the switch but maybe it is just overeegineering

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: Wow!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
ftomassetti: Nice

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -677,4 +683,91 @@ protected SymbolResolver getSymbolResolver() {
...
+    public <T extends Node> List<T> findAll(Class<T> nodeType) {
+        final List<T> found = new ArrayList<>();
+        walk(nodeType, found::add);
+        return found;
+    }
+
+    /**
+     * Walks the AST, returning all nodes of type "nodeType" that match the predicate.
+     */
+    public <T extends Node> List<T> findAll(Class<T> nodeType, Predicate<T> predicate) {
```
ftomassetti: Now we should rebrand JavaParser as the _functional_ parser for Java. It will bring us success and glory!

ftomassetti (approved):
Wow, the quality of this codebase is skyrocketing!
It looks very different from the project you found and put on GitHub :D

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: I Googled it :-D

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: I doubt we'll get away with having everything mutable ;-)

matozoid:
@ftomassetti I agree it's cool, but I think we have too many options now :-O What do you think would be the most useful set of functionality? TreeVisitor, Iterators, Iterables, walk, find, Stream...

matozoid:
I've added iterators for  "iterate from inside a hierarchy upwards to the root node" and "iterate over direct children" so most traversal logic is now captured in iterators :-O

matozoid:
Okay, I'm going to...
- keep the Iterator classes public and put them in `Node`
- hide the factories for iterators and iterables so they become private helper methods
- keep `stream` public since it's fancy
- create an API of `walk` (looks at every node) `findFirst` (stops on predicate match) `findAll` (collects all that match predicate) with overloads that either
  - select the traversal algorithm
  - select the type of node to visit
  - have no parameters
- `findParent` which specifically finds a parent of a certain type.

IrfanKhanKakar:
Hi, 
my eclipse give me an Error in (   import com.github.javaparser ) in import or  i write (  	CompilationUnit compilationUnit = JavaParser.parse(sources)   ) in my java code,

there is any  Jar file ,  or libraries ?????

ftomassetti:
@IrfanKhanKakar could you please stop adding comments inside totally unrelated issues? This is third one you add and it is getting annoying

Issue #1198, opened 2017-10-18
Prettier imports


leus:
Can we add an option to order imports during pretty print?

Also, specifying order (like IntelliJ and Eclipse) would be a plus.

ftomassetti:
Could changing the order of imports affect the program? Like if you have asterisks imports and specific imports and one override the other.

Like:
import c.d.*;
import a.b.MyClass;

If we have both a.b.MyClass and c.d.MyClass does the order of the imports matter?

leus:
It may. But this looks like a good candidate for a `PrettyPrinterConfiguration` option.

PR #1199, opened 2017-10-19
Allow to order imports in pretty print

leus:
Order imports by putting static imports first, then ordering by name. Use `PrettyPrinterConfiguration().setOrderImports(true)` to enable.

See #1198


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1380,8 +1380,23 @@ public void visit(TypeExpr n, Void arg) {
...
+                int sort = Integer.compare(left.isStatic() ? 0 : 1, right.isStatic() ? 0 : 1);
+                if (sort == 0) {
+                    sort = left.getNameAsString().compareTo(right.getNameAsString());
+                }
+                return sort;
+            });
+            for (Object node : modifiableList) {
+                ((Node) node).accept(this, arg);
+            }
+        } else {
```
matozoid, line 1409: The import stuff could be extracted into a separate method so that the common case doesn't get hidden away at the end.

javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrintVisitorTest.java
```diff
@@ -139,4 +139,31 @@ public void printClassWithoutJavaDocButWithComment() {
...
         assertEquals(String.format("public class A {%s    // stuff%s}%s", EOL, EOL, EOL), content);
     }
+
+    @Test
+    public void printImportsDefaultOrder() {
+        String code = "import x.y.z;import a.b.c;import static b.c.d;class c {}";
+        CompilationUnit cu = JavaParser.parse(code);
+        String content = cu.toString();
+        assertEquals(String.format("import x.y.z;%1$s" +
+                "import a.b.c;%1$s" +
```
matozoid, line 155: I saw this and was thinking: maybe we can create an assertEquals method that replaces all EOL characters with the EOL constant before calling JUnit's assertEquals? Then we can simply pass "`class X{\n}`".

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
leus: I'm still struggling with this, since we don't want to mutate the original list. Perhaps something like
``` java
public void visit(NodeList n, Void arg) {
    NodeList<ImportDeclaration> sortedList = processOrderOfImports(n);
    for (Object node : sortedList) {
        ((Node) node).accept(this, arg);
    }
}
```
...and move the sort logic to `processOrderOfImports()`? The problem is that this method is used everywhere. Is it possible to differentiate a list of imports before getting here?

javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrintVisitorTest.java
leus: We already have a method (`com.github.javaparser.utils.Utils.normalizeEolInTextBlock()`) so it shouldn't be difficult. I can add a new pull request.

matozoid: Cool, or put it in this one.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: Hmmm, I see. Leave it like this then :-)

javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrintVisitorTest.java
matozoid: A new PR it is, because I'm merging this one :-)

matozoid:
Thank you very much!

Issue #1047, opened 2017-08-14
Test failures in javaparser


jimshowalter:
Tried building javaparser and got a test failure:
Failed tests: removeAllOnRequiredProperty(com.github.javaparser.ast.NodeTest): expected:<class X {[(..)
testSimplePropertyWithGenericReplace(com.github.javaparser.ast.ReplaceNodeTest): expected:<package z;[(..)
testListProperty(com.github.javaparser.ast.ReplaceNodeTest): expected:<package x;[(..)

I'm on Windows 10. Is it possible the tests assume a Linux-style OS and file system?

matozoid:
They shouldn't, but they do. This has been one the backlog for quite a while because none of our contributors seem to be working on Windows. See also #713

jimshowalter:
I can try to fix them.

matozoid:
Do try to fix them the right way, like described in #713 :-)

Issue #1200, opened 2017-10-19
EOL inconsistences when parsing from file


leus:
When reading a file with EOL different than the current system, block comments are preserved as-is and this may lead to generated files with different EOL.

This causes the test "Scenario: we can parse a package-info file" fail under Windows.

leus:
By the way, this fixes AppVeyor :) #713

PR #1201, opened 2017-10-19
Normalize EOL inside comments

leus:
Normalize EOL characters inside comment text.

Also, there are some test fixes for EOL differences under Windows.

See #1200

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
@@ -27,7 +27,7 @@ import static com.github.javaparser.utils.Utils.EOL;
 import static com.github.javaparser.utils.Utils.assertNotNull;
 
 public class PrettyPrinterConfiguration {
-    private boolean normalizeEolInComments = false;
+    private boolean normalizeEolInComment = false;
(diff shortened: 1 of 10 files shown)


leus:
Fixes #713

leus:
Fixes #1047

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -44,6 +45,15 @@ public PrettyPrinterConfiguration setIndent(String indent) {
         return this;
     }
 
+    public boolean isNormalizeEolInComments() {
```
ftomassetti: maybe in this case the name should be `normalizeEolInComment` or `isEolInCommentsNormalizationEnabled`, not sure

ftomassetti (approved):
Good stuff!

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
leus: I think the first one is more consistent

matozoid:
Appveyor is working again! @leus, you are my hero! It was broken for so long...

PR #1206, opened 2017-10-21
Fix eol fixing method

matozoid:
@leus - that EOL code needed fixing :-)

Can you think of a better name for "assertEqualsNoEol"?


leus:
That was embarrasing :-)

My take is `assertEqualsNormalizingNewlines` (because EOL are compared, just normalized before comparison.)

matozoid:
Well, that's a mouthful! If you're still working on it, go ahead and rename the method to that.

PR #1208, opened 2017-10-21
[DRAFT] Move as much as possible to JUnit 5

matozoid:
We had the JUnit 5 dependency already, but I suddenly noticed that the JBehave stuff was still pulling in JUnit 4. Tried to upgrade as much as possible, but we're stuck with JUnit 4 compatibility mode until JBehave upgrades.

--> https://github.com/valfirst/jbehave-junit-runner/issues/106

Most noticable things: Before -> BeforeEach, After -> AfterEach, @Test(expected =...) -> assertThrows(), @RunWith -> some kind of extension stuff, assertTrue/False -> assertEquals(true/false, .


matozoid:
A big decrease in coverage. Have to find out if some tests stopped working...

matozoid:
Let's try this again in a year or two, when JUnit 5 is well supported :-/

PR #1215, opened 2017-10-25
Refining integration of resolution methods

ftomassetti:
As I write tests on the JSS side I check if something is still missing


matozoid:
If there is nothing major in here, feel free to pull at any time. Except Sunday evening ;-)

matozoid:
I've set the coveralls limit to -0.1% so it won't report the build as broken for decreases this small!

ftomassetti:
Added a few missing bits

PR #1223, opened 2017-11-03
Use the table drive code generator

### LINKED ISSUE #1221: Try out javacc 7 table driven code generation
See the javacc wiki for the new secret option

matozoid:
Well, this seems to work just fine. Should be faster.

Fixes #1221


matozoid:
Doesn't actually work.

PR #1226, opened 2017-11-04
make also VariableDeclarator resolvable

ftomassetti:


matozoid:
+1

matozoid:
How about `ReceiverParameter`?

matozoid:
@ftomassetti - does this go into today's release?

ftomassetti:
I would not consider the ReceiverParameter because any reference is to "this" in any case. Do you agree with this?

matozoid:
@ftomassetti - maybe it would be good to be able to solve this:
```java
class A {
    class B {
        B(A A.this) {
            // 8-O
        }
    }
}
```

ftomassetti:
But this would be solved in the same way if there was or there was not a receiver parameter declaration, correct?

Inside the constructor `this` would have the same meaning. The receiver parameter is not introducing any new value or any alias or an existing value, if I understand correctly.

matozoid:
Here's a little test program:
```java

public class A {
    class B {
        B(A A.this) {
            A a = A.this;
            B b = this;
            System.out.println(a);
            // com.github.javaparser.A@677327b6
            System.out.println(b);
            // com.github.javaparser.A$B@14ae5a5
        }
    }

        A a = new A();
        B b = a.new B();
    }
}
```

So if I'm in an inner class, the receiver parameter always refers to the outer class, but `this` points to the inner class?! `B(B this){}` is illegal for some reason.

I'm not telling you this is something that has to be resolved, I haven't worked with JSS enough to judge that, but I think it's an obscure and tricky feature, not a simple one. (I may have suggested it was simple. I was wrong :-D )

Here's another question: receiver parameters exist for annotating `this`. Should these annotations show up in JSS?

ftomassetti:
These annotations should show up in JSS, somehow. Not sure how :)
I should understand receiver parameter better myself but it seems to me that if you remove them the logic of the methods does not change, this is why we could ignore them in the resolution phase. We just need to provide a way to get their annotations

matozoid:
So, do you want to pull this thing or not? ;-)

ftomassetti:
Yes and I will pull it :D

PR #1240, opened 2017-11-11
Stop preprocessing unicode escapes

### LINKED ISSUE #209: JavaParser and Unicode Escaping
Hi

is it intended that parsing and serializing turns

```
public class Sample {

    public static final String SAMPLE="\u00DCbung";

}
```

into

```
public class Sample {

    public static final String SAMPLE = "Übung";
}
```

### LINKED ISSUE #600: JavaParser interprets Unicode escape sequences in comments
In JavaParser 3.0.0 alpha 10 Unicode escape sequences are interpreted. This results in the following behavior:

```java
JavaParser.parse("class Foo { /* \\u002e */ }").toString()
```
expected result:
```java
class Foo {
    /* \u002e */
}
```

actual result:
```java
class Foo {
    /* . */
}
```

### LINKED ISSUE #801: Remember the original unicode escape sequence
Do we want to store the very original code, including the unicode escape sequences?
@cal101 provided a patch for that

https://github.com/ftomassetti/javaparser/pull/5/files

@matozoid is in a better position than me to comment on it

### LINKED ISSUE #1236: I think there is a bug with unicode string literals
Hi fellas,

I have this code:

```
static String unicode_rightPointingTriangle() {
  return "\u25B8";
}
```
and it seems the JavaParser mangles it into, ultimately, "?".

Has this been tested? I encode all my Java sources as UTF-8 - but it shouldn't matter anyways as it is ecaped as \u, as you see.

I think there really is a bug in the parser/pretty printer.

Greetings,
Stefan

matozoid:
Fixes #1236 
Fixes #801 
Fixes #600
Fixes #209 

@ftomassetti @cal101 @Gottox @digulla @cdietrich - @ptitjes tells us in the original #209 that unicode escape sequences should be gone before the lexer even starts. This allows for unicode escapes in every element of the source code. This is what we currently do, and this is what causes the escape sequences from the source code to be irretrievably lost to the user.

I would like to reverse this decision and stop preprocessing the escapes. This will be the result:
- good: the original escape sequences will be kept and can be printed out again.
- bad: it becomes impossible to use escape sequences in many positions. However, non-ASCII characters are only allowed in comments, string and char literals, and identifiers, so if we make sure they are allowed there this should be a very marginal issue. Intellij doesn't even recognize this case.

I think doing this will support a lot more use cases, and will break few to none.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
@@ -37,6 +37,7 @@ import java.util.function.Consumer;
  * A literal string.
  * <br/><code>"Hello World!"</code>
  * <br/><code>"\"\n"</code>
+ * <br/><code>"\u2122"</code>
  * <br/><code>"™"</code>
  * <br/><code>"💩"</code>
  *
--- a/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -31,16 +31,20 @@ import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.ast.stmt.Statement;
+import com.github.javaparser.ast.stmt.SwitchEntryStmt;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.ast.type.IntersectionType;
 import com.github.javaparser.ast.type.Type;
 import org.junit.Test;
 
+import java.io.IOException;
+import java.nio.file.Path;
 import java.util.Optional;
 
 import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
-import static com.github.javaparser.Providers.*;
+import static com.github.javaparser.Providers.provider;
 import static com.github.javaparser.Range.range;
+import static com.github.javaparser.utils.CodeGenerationUtils.mavenModuleRoot;
 import static com.github.javaparser.utils.TestUtils.assertInstanceOf;
 import static com.github.javaparser.utils.Utils.EOL;
 import static org.junit.Assert.assertEquals;
@@ -191,7 +195,7 @@ public class JavaParserTest {
         Statement lambdaBody = lambdaExpr.getBody();
         assertEquals(range(3, 68, 3, 101), lambdaBody.getRange().get());
     }
-    
+
     @Test
     public void testNotStoringTokens() {
         JavaParser javaParser = new JavaParser(new ParserConfiguration().setStoreTokens(false));
@@ -208,5 +212,16 @@ public class JavaParserTest {
     public void trailingWhitespaceIsIgnored() {
         BlockStmt blockStmt = JavaParser.parseBlock("{} // hello");
         assertEquals("{}", blockStmt.getTokenRange().get().toString());
+    }
+
+    @Test
+    public void everyTokenHasACategory() throws IOException {
+        final int tokenCount = GeneratedJavaParserConstants.tokenImage.length;
+        Path tokenTypesPath = mavenModuleRoot(JavaParserTest.class).resolve("../javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java");
+        CompilationUnit tokenTypesCu = JavaParser.parse(tokenTypesPath);
+        // -1 to take off the default: case.
+        int switchEntries = tokenTypesCu.findAll(SwitchEntryStmt.class).size()-1;
+        // The amount of "case XXX:" in TokenTypes.java should be equal to the amount of tokens JavaCC knows about:
+        assertEquals(tokenCount, switchEntries);
     }
 }
--- a/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -207,6 +207,6 @@ public class JavaParserTest {
     @Test
     public void trailingWhitespaceIsIgnored() {
         BlockStmt blockStmt = JavaParser.parseBlock("{} // hello");
-        assertEquals("\"}\"   <95>   (line 1,col 2)-(line 1,col 2)", blockStmt.getTokenRange().get().getEnd().toString());
+        assertEquals("{}", blockStmt.getTokenRange().get().toString());
     }
 }
--- a/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -207,6 +207,6 @@ public class JavaParserTest {
     @Test
     public void trailingWhitespaceIsIgnored() {
         BlockStmt blockStmt = JavaParser.parseBlock("{} // hello");
-        assertEquals("\"}\"   <94>   (line 1,col 2)-(line 1,col 2)", blockStmt.getTokenRange().get().getEnd().toString());
+        assertEquals("\"}\"   <95>   (line 1,col 2)-(line 1,col 2)", blockStmt.getTokenRange().get().getEnd().toString());
     }
 }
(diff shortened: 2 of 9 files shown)


matozoid:
The important part of this change is removing `JAVA_UNICODE_ESCAPE=true;`, by the way.

ptitjes:
I guess no one uses escapes inside keywords because keyboard letters... Great idea indeed!

ftomassetti:
I guess that for those very few cases in which they want to process unicode escapes in unexpected places we could have some preprocessing before JavaParser kicks in

javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
```diff
@@ -37,6 +37,7 @@
  * A literal string.
  * <br/><code>"Hello World!"</code>
  * <br/><code>"\"\n"</code>
+ * <br/><code>"\u2122"</code>
  * <br/><code>"™"</code>
  * <br/><code>"💩"</code>
```
ftomassetti, line 42: I am not sure what GitHub wants to tell me with this icon here :)

javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
```diff
@@ -207,6 +207,6 @@ public void trailingCodeIsAnError() {
     @Test
     public void trailingWhitespaceIsIgnored() {
         BlockStmt blockStmt = JavaParser.parseBlock("{} // hello");
-        assertEquals("\"}\"   <94>   (line 1,col 2)-(line 1,col 2)", blockStmt.getTokenRange().get().getEnd().toString());
+        assertEquals("\"}\"   <95>   (line 1,col 2)-(line 1,col 2)", blockStmt.getTokenRange().get().getEnd().toString());
```
ftomassetti: what this "95" represents here?

javaparser-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
```diff
@@ -58,4 +62,11 @@ public void cannotRemoveCommentNotUsedAnywhere() {
         Comment c = new LineComment("A comment");
         assertFalse(c.remove());
     }
+
+    @Test
+    public void mayContainUnicodeEscapes() {
```
ftomassetti: Maybe "unicodeEscapesArePreservedInComments"

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/SimpleNameTest.java
```diff
@@ -42,4 +43,9 @@ public void identifierMustNotBeNull() {
         new SimpleName(null);
     }
 
+    @Test
+    public void mayContainUnicodeEscapes() {
+        SimpleName name = parseSimpleName("xxx\\u2122xxx");
+        assertEquals("xxx\\u2122xxx", name.asString());
+    }
 }
```
ftomassetti: Newline :D

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/StringLiteralExprTest.java
```diff
@@ -0,0 +1,14 @@
...
+import static com.github.javaparser.JavaParser.parseExpression;
+import static org.junit.Assert.*;
+
+public class StringLiteralExprTest {
+    @Test
+    public void unicodeEscapesDoNotGetPreprocessed() {
+        StringLiteralExpr omega = parseExpression("\"xxx\\u03a9xxx\"");
+        assertEquals("xxx\\u03a9xxx", omega.getValue());
+    }
+}
```
ftomassetti: Newline :D

ftomassetti (approved):
The idea is good and should solve a significant issue and the implementation also looks good.

Thumbs up!

javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
matozoid: The omega or the "emoticon"? ;-)

javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
matozoid: It the token type. Annoying, but it did remind me to put the new token type in the big category switch.

javaparser-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
matozoid: Done

javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
ftomassetti: the emoticon :)

javaparser-testing/src/test/java/com/github/javaparser/JavaParserTest.java
ftomassetti: Maybe we can add a comment about that

javaparser-core/src/main/java/com/github/javaparser/ast/expr/StringLiteralExpr.java
matozoid: I knew someone was eventually going to notice ;-)

stefan-reich:
Cool

davidsusu:
IMO preprocessing is the right way. Its only drawback is that it is harder to reproduce the original source. Harder, but not impossible. For a solution to a similar problem see [javascript source maps](https://developers.google.com/web/tools/chrome-devtools/javascript/source-maps). JavaParser could optionally look up for original source for regenerating untouched parts of code.

davidsusu:
**Off.** I knew that there was a related thing from my online past, but I could not remember until now. I am not very active on SO, but about two years ago there was a question about text extracting from HTML. OP's problem was similar: search and extract from **HTML** by **text**. We can transform HTML to text (as a preprocessing), then search in the text for the wanted part, and finally: extract the original source of it. Also, I wrote a simplified example (in PHP). [My answer on SO](https://stackoverflow.com/questions/36690567/search-html-for-2-phrases-ignoring-all-tags-and-strip-everything-else/36694339#36694339).

matozoid:
Hey @davidsusu, JavaParser keeps the output of the lexer around, you can ask every node what it's "token range" is and you will get the original source code (if any.) The issue here is that if we preprocess, we don't have a mapping to the source code anymore, but a mapping to the output of the preprocessor. Nobody will ever know what the original was. I think @cal101's solution to this same problem was to keep all information around, and do preprocessing, so that would be the better solution - at a cost.

Anyway, it is true that preprocessing is what the JLS prescribes. The solution implemented here is a simple but non-conformant solution to a common problem with JP, with the hopes that the part it breaks is so obscure that nobody will notice. So I'm really curious how you did notice, because this assumption seems wrong!

matozoid:
@davidsusu - I'm still curious!

davidsusu:
@matozoid Unit tests just fail. However I personally went over this problem: changed the tests.

I remember to only one production code where this would have been a problem, an old fashioned database mapping system with code generator, which created unicode escape sequences for non-english letters in field names. No longer in use. Legacy code can contain some kind of strange (but, of course, standard compliant strange).

matozoid:
@davidsusu thanks - once you (or anyone else) runs into that kind of source I'll put a unicode preprocessor option in the next release.

tibcoktredget:
I've just run into a problem with JavaParser 3.7.1 parsing some code generated by JavaCC:

`        jjtn000.setValue(t.image.substring(t.image.indexOf('\u005c'') + 1, t.image.length() - 1));
`

\u005c is the escape sequence for the backslash character, so this means:

`        jjtn000.setValue(t.image.substring(t.image.indexOf('\'') + 1, t.image.length() - 1));
`

but JavaParser does not parse this

Keith

matozoid:
@tibcoktredget Interesting. Can you open this as a separate issue?

tibcoktredget:
Thanks @matozoid for your prompt response - I've raised the issue

davidsusu:
@tibcoktredget, @matozoid Isn't this the **same** problem? (The `char` literal parsed before JavaParser could recognize that the second single quote is escaped.)

matozoid:
@davidsusu it's a process problem: these are replies to a closed PR. If a new change is needed, we need an open issue.

PR #1246, opened 2017-11-13
make ResolvedEnumConstantDeclaration extends ResolvedValueDeclaration

ftomassetti:


ftomassetti:
Given it is such a simple change I would go on and merge before it get out-dated again (just waiting for tests to pass)

PR #1249, opened 2017-11-14
Issue 1247

### LINKED ISSUE #1247: IfStmt.hasElseBlock() does not work correctly
The `IfStmt.hasElseBlock()` method exhibits a somewhat surprising behavior. Consider:

```
IfStmt ifStmt = (IfStmt) JavaParser.parseStatement("if (cond) doSomething(); else doSomethingElse();");
System.out.println(ifStmt.hasElseBlock()); // prints: false
```
The above prints `false`. Granted, the else-branch is not a block, but a single statement. And indeed,
```
IfStmt ifStmt2 = (IfStmt) JavaParser.parseStatement("if (cond) doSomething(); else { doSomethingElse(); }");
System.out.println(ifStmt2.hasElseBlock()); // prints: true
```
...prints `true`. All of that would be fine. However, the following...
```
IfStmt ifStmt3 = (IfStmt) JavaParser.parseStatement("if (cond1) doSomething(); else if (cond2) doSomethingElse();");
System.out.println(ifStmt3.hasElseBlock()); // prints: true
```
...prints `true` too. But why? In the latter case, the else-branch consists of a single statement (namely, another if-statement), not a block.

I think that either *all* of the above should print `true` (in which case the method may be more aptly named `hasElseBranch()`), or *only* the second example should print `true`, while the two others should print `false` (since only the second example does have an else-*block*).

Is this a bug, or am I simply missing something?

ftomassetti:
Fix #1247

The CSM needs to know if we have an else block or a cascading if, however the method `hasElseBlock` should just tell us if the `IfStmt` has an else block or not.

I added:
* hasElseBranch
* hasCascadingIfStmt

And I corrected `hasElseBlock`


matozoid:
Can you add some Javadoc to these methods?

ftomassetti:
Javadoc added

ftomassetti:
Updating master

matozoid:
+1

PR #1252, opened 2017-11-16
Fix annotations on type scope getting on to all parts of type

### LINKED ISSUE #1251: Extra Generic Type Added
I'm having an issue where exporting duplicates a generic where one should not be.
I had to change all the file extensions to `.txt` from `.java` so I could attach them.

In particular in the input file: 

      public TypeUtilsTest<String>.Tester tester;

Becomes in the output file:

     public TypeUtilsTest<String>.Tester<String> tester;

The parser I'm using is v3.5.3 on JDK 1.8:

```
import com.github.javaparser.ast.NodeList;
import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
import com.github.javaparser.ast.expr.AnnotationExpr;

import java.io.File;
import java.io.FileWriter;
import java.io.IOException;
import java.util.HashSet;
import java.util.List;

public class JPMinimal {

    public static void extract(final String fileName, final String outputFile) {
        CompilationUnit cu;
        File file = new File(fileName);
        File outFile = new File(outputFile);
        try {
            cu = JavaParser.parse(file);
        } catch (FileNotFoundException f) {
            throw new RuntimeException("Could not find file: "+file);
        }
        try {
            FileWriter fooWriter = new FileWriter(outFile, false); // true to append
            // false to overwrite.
            fooWriter.write(cu.toString());
            fooWriter.close();
        } catch (IOException e) {
            throw new RuntimeException("EXCEPTION: Could not write: "+file);
        }
    }

        extract(
                "/Users/loren/Sites/pullTest/bertGithub/src/main/java/edu/gatech/cc/regressionFinder" +
                        "/TypeUtilsTest.java",
                "/Users/loren/Sites/pullTest/bertGithub/src/main/java/edu/gatech/cc/regressionFinder" +
                        "/TypeUtilsTestOut.java"
        );
    }
}
```

[TypeUtilsTest.txt](https://github.com/javaparser/javaparser/files/1480269/TypeUtilsTest.txt)
[TypeUtilsTestOut.txt](https://github.com/javaparser/javaparser/files/1480270/TypeUtilsTestOut.txt)

matozoid:
Fixes #1251


matozoid:
Do I get a prize for quickest bug fix? :-D

PR #1261, opened 2017-11-29
NameExpr & other stuff resolvable

ftomassetti:
Making other stuff resolvable.

By iterating we are improving integration with JSS and make things much easier for the user. The idea is that for most use cases the user could ignore JavaParserFacade and just resolve things from AST nodes


matozoid:
+1

PR #1276, opened 2017-12-08
Generate toXyz methods

### LINKED ISSUE #1267: Create mapXyz in addition to ifXyz
We have nice new methods to deal with subclasses of nodes, like "ifSwitchStmt" "asSwitchStmt" and so on. 

I'm missing one that works like `map`: it returns a result if the node is of some subtype, and empty if not.

Suggestions:
`T withSwitchStmt(Function<SwitchStmt, T> f)`
`T mapSwitchStmt(Function<SwitchStmt, T> f)`
or maybe one that simply returns an `Optional<SwitchStmt>` and you can do with it what you like.

matozoid:
Fixes #1267

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
@@ -36,6 +36,7 @@ import javax.annotation.Generated;
 import com.github.javaparser.TokenRange;
 import java.util.function.Consumer;
 import static com.github.javaparser.utils.CodeGenerationUtils.f;
+import java.util.Optional;
 
 /**
  * Any declaration that can appear between the { and } of a class, interface, or enum.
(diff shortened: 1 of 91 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
```diff
@@ -282,4 +283,48 @@ public void ifMethodDeclaration(Consumer<MethodDeclaration> action) {
     @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
     public void ifTypeDeclaration(Consumer<TypeDeclaration> action) {
     }
+
+    public Optional<AnnotationDeclaration> toAnnotationDeclaration() {
```
ftomassetti, line 287: In practice I think I would prefer to get directly the type, without the Optional wrapping it. Otherwise my code would start having a ton of `get()`

matozoid: I've been looking into JSS Navigator a little, and my advice is to look into map and ifPresent a lot more, since those two replace many use cases for the current ifXyz/asXyz. This PR doesn't affect your code by the way :-)

PR #1285, opened 2017-12-11
Issue 1277

### LINKED ISSUE #1277: ModifierVisitor and LexicalPreservingPrinter
Thanks for providing LexicalPreservingPrinter. I try to add a chained method call for each found method call in a ModifierVisitor but I receive a UnsupportedOperationException when using LexicalPreservingPrinter.

```java
	static class AddFooCallModifierVisitor extends ModifierVisitor<Void> {
		@Override
		public Visitable visit(MethodCallExpr n, Void arg) {
			// Add a call to foo() on every found method call
			return new MethodCallExpr(n, "foo");
		}
	}

	@Test
	public void testInvokeModifierVisitor() throws IOException {
		String code = "class A {" + EOL +
				      "  public String message = \"hello\";" + EOL +
				      "   void bar() {" + EOL + 
				      "     System.out.println(\"hello\");" + EOL + 
				      "   }" + EOL +
				      "}";

		CompilationUnit cu = JavaParser.parse(code);
		cu.accept(new AddFooCallModifierVisitor(), null);
	}
```

Givs me:
`
java.lang.UnsupportedOperationException: kept child(ClassOrInterfaceType) vs "String"   <89>   (line 2,col 10)-(line 2,col 15)
`
If I remove the _message_ attribute it works fine.
Any idea what I am doing wrong?

Is there a way to catch formatting errors like this and proceed the processing (with some minor formatting issues)?

ftomassetti:
Fix #1277 by preventing to replace a node with itself

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
@@ -132,6 +132,13 @@ public class NodeList<N extends Node> implements List<N>, Iterable<N>, HasParent
 
     @Override
     public N set(int index, N element) {
+        if (index < 0 || index >= innerList.size()) {
+            throw new IllegalArgumentException("Illegal index. The index should be between 0 and " + innerList.size()
+                    + " excluded. It is instead " + index);
+        }
+        if (element == innerList.get(index)) {
+            return element;
+        }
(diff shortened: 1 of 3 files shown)


matozoid:
Looks fine to me.

ftomassetti:
Then let's merge it

PR #1291, opened 2017-12-14
Fix nodelist replace bug

### LINKED ISSUE #1290: .replace() method sometimes doesn't properly replace AST nodes
Take following call statement: `callSomeFun(r -> r instanceof SomeType);`.

Let's call the lambda expression argLambdaExpr and create a new MethodCallExpr to lambdaWrapper() called methodCallLambdaReplacement. When I call `argLambdaExpr.replace(methodCallLambdaReplacement)` the resulting callSomeFun-expression should be:

`callSomeFun(lambdaWrapper());` 

In fact, in the `arguments` list the lambda expression in the MethodCallExpr is replaced with our new methodCallLambdaReplacement expression. However, the `childNodes` element of the MethodCallExpr still contains the lambda expression (in addition to the new method call expression). I believe the lambda expression is supposed to be be removed from the childNodes list by the .replace() method as well.

matozoid:
Fixes #1290

### DIFF
--- a/javaparser-testing/src/test/java/com/github/javaparser/ast/expr/MethodCallExprTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/ast/expr/MethodCallExprTest.java
@@ -1,0 +1,20 @@
+package com.github.javaparser.ast.expr;
+
+import org.junit.Test;
+
+import static com.github.javaparser.JavaParser.parseExpression;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class MethodCallExprTest {
+    
+    @Test
+    public void replaceLambdaIssue1290() {
+        MethodCallExpr methodCallExpr = parseExpression("callSomeFun(r -> r instanceof SomeType)").asMethodCallExpr();
+        LambdaExpr lambdaExpr = methodCallExpr.getArgument(0).asLambdaExpr();
+        MethodCallExpr lambdaWrapper = new MethodCallExpr("lambdaWrapper");
+        lambdaExpr.replace(lambdaWrapper);
+        
+        assertEquals(2, methodCallExpr.getChildNodes().size());
+    }
+
+}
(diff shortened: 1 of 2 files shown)


javaparser-testing/src/test/java/com/github/javaparser/ast/expr/MethodCallExprTest.java
```diff
@@ -0,0 +1,20 @@
...
+    public void replaceLambdaIssue1290() {
+        MethodCallExpr methodCallExpr = parseExpression("callSomeFun(r -> r instanceof SomeType)").asMethodCallExpr();
+        LambdaExpr lambdaExpr = methodCallExpr.getArgument(0).asLambdaExpr();
+        MethodCallExpr lambdaWrapper = new MethodCallExpr("lambdaWrapper");
+        lambdaExpr.replace(lambdaWrapper);
+        
+        assertEquals(2, methodCallExpr.getChildNodes().size());
+    }
+
+}
```
ftomassetti: newline :D

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/MethodCallExprTest.java
```diff
@@ -0,0 +1,20 @@
...
+public class MethodCallExprTest {
+    
+    @Test
+    public void replaceLambdaIssue1290() {
+        MethodCallExpr methodCallExpr = parseExpression("callSomeFun(r -> r instanceof SomeType)").asMethodCallExpr();
+        LambdaExpr lambdaExpr = methodCallExpr.getArgument(0).asLambdaExpr();
+        MethodCallExpr lambdaWrapper = new MethodCallExpr("lambdaWrapper");
+        lambdaExpr.replace(lambdaWrapper);
+        
+        assertEquals(2, methodCallExpr.getChildNodes().size());
```
ftomassetti, line 18: Could we add a check on the node being replaced having a null parent?

ftomassetti (approved):
Looks good!

javaparser-testing/src/test/java/com/github/javaparser/ast/expr/MethodCallExprTest.java
matozoid: No :-P

PR #1292, opened 2017-12-14
Make global cache thread safe

### LINKED ISSUE #1286: PhantomNodeLogic is not thread safe
Hi all!

I ran into the following problem:

The IdentityHashmap based logic in https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java#L38
fails if multiple threads call JavaParser.parse(code) at the same time, as the IdentityHashmap (accessed statically) is not thread-safe. Leading to weird, hard to debug errors like:

```
com.github.javaparser.ParseProblemException: Unknown error
Problem stacktrace : 
  com.github.javaparser.printer.lexicalpreservation.PhantomNodeLogic.isPhantomNode(PhantomNodeLogic.java:50)
  com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter.storeInitialTextForOneNode(LexicalPreservingPrinter.java:254)
  com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter.access$100(LexicalPreservingPrinter.java:63)
  com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter$4.process(LexicalPreservingPrinter.java:242)
  com.github.javaparser.ast.visitor.TreeVisitor.visitBreadthFirst(TreeVisitor.java:94)
  com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter.storeInitialText(LexicalPreservingPrinter.java:245)
  com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter.lambda$setup$0(LexicalPreservingPrinter.java:104)
```

A simple fix to this problem as suggested by https://docs.oracle.com/javase/7/docs/api/java/util/IdentityHashMap.html would by using the synchronized version: `https://docs.oracle.com/javase/7/docs/api/java/util/IdentityHashMap.html`

Thanks for checking this out!

Gyula

matozoid:
Fixes #1286

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
@@ -38,7 +38,7 @@ class PhantomNodeLogic {
 
     private static final int LEVELS_TO_EXPLORE = 3;
 
-    private static final Map<Object, Boolean> isPhantomNodeCache = synchronizedMap(new IdentityHashMap<>());
+    private static final Map<Node, Boolean> isPhantomNodeCache = synchronizedMap(new IdentityHashMap<>());
 
     private static final AstObserver cacheCleaner = new AstObserverAdapter() {
         @Override
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
@@ -27,6 +27,9 @@ import com.github.javaparser.ast.observer.AstObserverAdapter;
 import com.github.javaparser.ast.type.UnknownType;
 
 import java.util.IdentityHashMap;
+import java.util.Map;
+
+import static java.util.Collections.synchronizedMap;
 
 /**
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
```diff
@@ -35,7 +38,7 @@
 
     private static final int LEVELS_TO_EXPLORE = 3;
 
-    private static final IdentityHashMap<Node, Boolean> isPhantomNodeCache = new IdentityHashMap<>();
+    private static final Map<Object, Boolean> isPhantomNodeCache = synchronizedMap(new IdentityHashMap<>());
```
ftomassetti: Why Object instead of Node?

ftomassetti (approved):
Aside my only comment it looks good

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
matozoid: Ooooooops

matozoid:
Comment was fixed. I'll pull it.

PR #1298, opened 2017-12-21
Avoid SourcePrinter to calculate newline while updating cursor Position

tarilabs:
... by disallow passing strings with newlines to the sourceprinter.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -53,14 +53,33 @@ import static com.github.javaparser.utils.Utils.normalizeEolInTextBlock;
 public class PrettyPrintVisitor implements VoidVisitor<Void> {
     protected final PrettyPrinterConfiguration configuration;
     protected final SourcePrinter printer;
+    private Deque<Position> methodChainPositions = new LinkedList<>();
 
     public PrettyPrintVisitor(PrettyPrinterConfiguration prettyPrinterConfiguration) {
         configuration = prettyPrinterConfiguration;
         printer = new SourcePrinter(configuration.getIndent(), configuration.getEndOfLineCharacter());
+        pushMethodChainPosition(printer.getCursor()); // initialize a default position for methodChainPositions, it is expected by method #resetMethodChainPosition()
     }
 
     public String getSource() {
         return printer.getSource();
+    }
+
+    public void resetMethodChainPosition(Position position) {
+        this.methodChainPositions.pop();
+        this.methodChainPositions.push(position);
+    }
+
+    public void pushMethodChainPosition(Position position) {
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
```diff
@@ -59,68 +52,78 @@ public SourcePrinter unindent() {
...
-
-    public void pushMethodChainPosition(Position position) {
-        this.methodChainPositions.push(position);
-    }
-
-    public Position peekMethodChainPosition() {
-        return this.methodChainPositions.peek();
-    }
-
-    public Position popMethodChainPosition() {
```
tarilabs, line 120: All these method are actually just meaningful for PrettyPrinterVisitor, hence moved there.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1348,10 +1367,15 @@ public void visit(final BlockComment n, final Void arg) {
         if (configuration.isIgnoreComments()) {
             return;
         }
-        printer
-                .print("/*")
-                .print(normalizeEolInTextBlock(n.getContent(), configuration.getEndOfLineCharacter()))
-                .println("*/");
+        final String commentContent = normalizeEolInTextBlock(n.getContent(), configuration.getEndOfLineCharacter());
```
tarilabs, line 1370: > option 3: I think only blockcomment is still printing strings with newlines. What if we disallow passing strings with newlines to the sourceprinter, then do the line splitting in the blockcomment visit method?

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1348,10 +1367,15 @@ public void visit(final BlockComment n, final Void arg) {
...
+        String[] lines = commentContent.split("\\R", -1); // as BlockComment should not be formatted, -1 to preserve any trailing empty line if present
```
matozoid, line 1371: Oh, didn't know about \R. Thorough! :-)

matozoid:
Looks fine to me, and I heard it fixes the performance problems, so here it goes!

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
tarilabs: Borrowed from here https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java#L226
;)

matozoid: Hehehe, I guess I forgot ;-)

PR #1302, opened 2017-12-28
Move module name configuration to correct pom file

sormuras:
Prior to this PR all artifacts were configured to the report `com.github.javaparser.core` as their module name. Now only the **core** jar contains the manifest attribute.

### DIFF
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -168,6 +168,18 @@
                     </execution>
                 </executions>
             </plugin>
+            <!-- Set JPMS module name to `com.github.javaparser.core` -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>com.github.javaparser.core</Automatic-Module-Name>
+                        </manifestEntries>
+                    </archive>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -179,17 +179,6 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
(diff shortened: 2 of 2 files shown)


matozoid:
Some more questions:
- do you think it's useful to do this for the parent pom too? We release it on Maven central. (The other modules are not)
- @ftomassetti can you do this for the symbol solver too?

sormuras:
If it is

- a valid use-case to depend on the parent module
- and the packages exported by the parent and core (and other) modules are unique

then yes.

sormuras:
This article sums it up very well: http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html

> - Module names must be reverse-DNS, just like package names, e.g. org.joda.time.
> - Modules are a group of packages. As such, the module name must be related to the package names.
> - Module names are strongly recommended to be the same as the name of the super-package.
> - Creating a module with a particular name takes ownership of that package name and everything beneath it.
> - As the owner of that namespace, any sub-packages may be grouped into sub-modules as desired so long as no package is in two modules.

matozoid:
The parent module contains no code and results in a pom, so it seems the answer is that this PR is okay as it is?

sormuras:
Ja.

ftomassetti:
Regarding doing so for JSS: there we use gradle, so I will need to figure out how to achieve the same thing but in principle yes

sormuras:
We set it via https://github.com/junit-team/junit5/blob/master/junit-jupiter-engine/junit-jupiter-engine.gradle#L20

```groovy
jar {
	manifest {
		attributes(
			'Automatic-Module-Name': 'org.junit.jupiter.engine'
		)
	}
}
```

Or like copied from http://branchandbound.net/blog/java/2017/12/automatic-module-name/

```groovy
jar {
    manifest {
        name = "mylibrary"
        instruction "Automatic-Module-Name", "com.acme.mylibrary"
    }
}
```

matozoid:
Thanks!

Issue #1301, opened 2017-12-28
LexicalPreservingPrinter does not play well with core-generators.


matozoid:
If we enable it by uncommenting this: https://github.com/javaparser/javaparser/blob/master/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java#L23

then code generation fails with a unhelpful message.

@ftomassetti I think this is a good use case for a nice debugging session :-)

ftomassetti:
I uncommented those three lines and run `sh run_core_generators.sh`. The build succedeed, however I can see this error:
```
[INFO] --- exec-maven-plugin:1.6.0:java (generate-javaparser-core) @ javaparser-core-generators ---
New source root at "/Users/federico/repos/javaparser/javaparser-core/src/main/java"
New source root at "/Users/federico/repos/javaparser/javaparser-core/target/generated-sources/javacc"
Running TypeCastingGenerator
Parsing /Users/federico/repos/javaparser/javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationDeclaration.java
Parsing /Users/federico/repos/javaparser/javaparser-core/src/main/java/com/github/javaparser/ast/body/BodyDeclaration.java
[WARNING]
java.lang.NoClassDefFoundError: javax/annotation/Generated
	at com.github.javaparser.generator.Generator.annotateGenerated(Generator.java:31)
	at com.github.javaparser.generator.Generator.addMethod(Generator.java:90)
	at com.github.javaparser.generator.Generator.addOrReplaceWhenSameSignature(Generator.java:61)
	at com.github.javaparser.generator.core.node.TypeCastingGenerator.generateIsType(TypeCastingGenerator.java:94)
	at com.github.javaparser.generator.core.node.TypeCastingGenerator.generateNode(TypeCastingGenerator.java:57)
	at com.github.javaparser.generator.NodeGenerator.generate(NodeGenerator.java:30)
	at com.github.javaparser.generator.core.CoreGenerator.run(CoreGenerator.java:46)
	at com.github.javaparser.generator.core.CoreGenerator.main(CoreGenerator.java:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #1309, opened 2018-01-07
Generate a Java token kind enum with a mapping to the integer kinds.

matozoid:
@ftomassetti please check if this is what you need, and if the naming is good.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
@@ -2,6 +2,7 @@ package com.github.javaparser.generator.core;
 
 import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.generator.core.node.*;
+import com.github.javaparser.generator.core.other.TokenKindGenerator;
 import com.github.javaparser.generator.core.visitor.*;
 import com.github.javaparser.utils.Log;
 import com.github.javaparser.utils.SourceRoot;
@@ -11,28 +12,37 @@ import java.nio.file.Paths;
 
 /**
  * Generates all generated visitors in the javaparser-core module.
+ * Suggested usage is by running the run_core_generators.sh script.
+ * You may want to run_metamodel_generator.sh before that.
  */
 public class CoreGenerator {
+    private static final ParserConfiguration parserConfiguration = new ParserConfiguration()
+//                                .setStoreTokens(false)
+//                                .setAttributeComments(false)
+//                                .setLexicalPreservationEnabled(true)
+            ;
+
     public static void main(String[] args) throws Exception {
         if (args.length != 1) {
             throw new RuntimeException("Need 1 parameter: the JavaParser source checkout root directory.");
         }
         Log.setAdapter(new Log.StandardOutStandardErrorAdapter());
         final Path root = Paths.get(args[0], "..", "javaparser-core", "src", "main", "java");
-        final SourceRoot sourceRoot = new SourceRoot(root)
+        final SourceRoot sourceRoot = new SourceRoot(root, parserConfiguration)
 //                .setPrinter(LexicalPreservingPrinter::print)
-                .setParserConfiguration(new ParserConfiguration()
-//                                .setStoreTokens(false)
-//                                .setAttributeComments(false)
-//                                .setLexicalPreservationEnabled(true)
-                );
+                ;
 
-        new CoreGenerator().run(sourceRoot);
+        final Path generatedJavaCcRoot = Paths.get(args[0], "..", "javaparser-core", "target", "generated-sources", "javacc");
+        final SourceRoot generatedJavaCcSourceRoot = new SourceRoot(generatedJavaCcRoot, parserConfiguration)
+//                .setPrinter(LexicalPreservingPrinter::print)
+                ;
+
+        new CoreGenerator().run(sourceRoot, generatedJavaCcSourceRoot);
 
         sourceRoot.saveAll();
     }
 
-    private void run(SourceRoot sourceRoot) throws Exception {
+    private void run(SourceRoot sourceRoot, SourceRoot generatedJavaCcSourceRoot) throws Exception {
         new TypeCastingGenerator(sourceRoot).generate();
         new GenericListVisitorAdapterGenerator(sourceRoot).generate();
         new GenericVisitorAdapterGenerator(sourceRoot).generate();
@@ -58,5 +68,6 @@ public class CoreGenerator {
         new MainConstructorGenerator(sourceRoot).generate();
         new FinalGenerator(sourceRoot).generate();
         new AcceptGenerator(sourceRoot).generate();
+        new TokenKindGenerator(sourceRoot, generatedJavaCcSourceRoot).generate();
     }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
@@ -55,6 +55,34 @@ public class JavaToken {
     }
 
     JavaToken(Token token, List<JavaToken> tokens) {
+        // You could be puzzled by the following lines
+        // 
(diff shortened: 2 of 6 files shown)


javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
```diff
@@ -11,28 +12,37 @@
 
 /**
  * Generates all generated visitors in the javaparser-core module.
+ * Suggested usage is by running the run_core_generators.sh script.
+ * You may want to run_metamodel_generator.sh before that.
  */
 public class CoreGenerator {
+    private static final ParserConfiguration parserConfiguration = new ParserConfiguration()
+//                                .setStoreTokens(false)
```
ftomassetti, line 20: Why do we have stuff commented out in the file?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -51,41 +55,36 @@ public JavaToken(int kind, String text) {
...
-        String text = token.image;
-
         // You could be puzzled by the following lines
-        //
+        // 
         // The reason why these lines are necessary is the fact that Java is ambiguous. There are cases where the
         // sequence of characters ">>>" and ">>" should be recognized as the single tokens ">>>" and ">>". In other
         // cases however we want to split those characters in single GT tokens (">").
-        //
+        //
```
ftomassetti, line 63: Maybe we do not want these extra spaces. Not that it really matters :)

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -248,6 +238,315 @@ public boolean isOperator() {
...
..., REQUIRES(64), TO(65), WITH(66), OPEN(67), OPENS(68), USES(69), MODULE(70), EXPORTS(71), PROVIDES(72), TRANSITIVE(73), LONG_LITERAL(74), INTEGER_LITERAL(75), DECIMAL_LITERAL(76), HEX_LITERAL(77), OCTAL_LITERAL(78), BINARY_LITERAL(79), FLOATING_POINT_LITERAL(80), DECIMAL_FLOATING_POINT_LITERAL(81), DECIMAL_EXPONENT(82), HEXADECIMAL_FLOATING_POINT_LITERAL(83), HEXADECIMAL_EXPONENT(84), HEX_DIGITS(85), UNICODE_ESCAPE(86), CHARACTER_LITERAL(87), STRING_LITERAL(88), IDENTIFIER(89), LETTER(90), PART_LETTER(91), LPAREN(92), RPAREN(93), LBRACE(94), RBRACE(95), LBRACKET(96), RBRACKET(97), SEMICOLON(98), COMMA(99), DOT(100), AT(101), ASSIGN(102), LT(103), BANG(104), TILDE(105), HOOK(106), COLON(107), EQ(108), LE(109), GE(110), NE(111), SC_OR(112), SC_AND(113), INCR(114), DECR(115), PLUS(116), MINUS(117), STAR(118), SLASH(119), BIT_AND(120), BIT_OR(121), XOR(122), REM(123), LSHIFT(124), PLUSASSIGN(125), MINUSASSIGN(126), STARASSIGN(127), SLASHASSIGN(128), ANDASSIGN(129), ORASSIGN(130), XORASSIGN(131), REMASSIGN(132), LSHIFTASSIGN(133), RSIGNEDSHIFTASSIGN(134), RUNSIGNEDSHIFTASSIGN(135), ELLIPSIS(136), ARROW(137), DOUBLECOLON(138), RUNSIGNEDSHIFT(139), RSIGNEDSHIFT(140), GT(141), CTRL_Z(142);
```
ftomassetti, line 244: Could we generate it with one per line?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -248,6 +238,315 @@ public boolean isOperator() {
...
+
+        private final int kind;
+
+        Kind(int kind) {
+            this.kind = kind;
+        }
+
+        public Kind valueOf(int kind) {
+            switch(kind) {
```
ftomassetti: I do not think we need this switch we can just iterate on Kind.values and find the one matching

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -335,14 +633,15 @@ public int hashCode() {
...
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
         JavaToken javaToken = (JavaToken) o;
-
-        if (kind != javaToken.kind) return false;
-        if (!text.equals(javaToken.text)) return false;
-
+        if (kind != javaToken.kind)
```
ftomassetti, line 641: I personally always use braces with the if statement

javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -53,11 +53,7 @@
     private List<AstObserver> observers = new ArrayList<>();
 
     public NodeList() {
-        this((Node) null);
-    }
-
-    public NodeList(Node parent) {
-        setParentNode(parent);
+        parentNode = null;
```
ftomassetti, line 56: How is this related?

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -248,6 +238,315 @@ public boolean isOperator() {
...
+
+        private final int kind;
+
+        Kind(int kind) {
+            this.kind = kind;
+        }
+
+        public Kind valueOf(int kind) {
```
ftomassetti: should this be static?

ftomassetti (approved):
It looks good but I would drop the switch in valueof and make the method static

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
matozoid: See #1301

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: Everything gets reformatted because this file is now part of the code generation step. See also #1301 :-)

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: Everything gets reformatted because this file is now part of the code generation step. See also #1301 :-)

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: I think it should, good one!

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: I'm now thinking it could have been a single array access if we generate that array instead. Hmmm.

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: Nothing happened here but reformatting of IDE generated code.

javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
matozoid: Nothing happened here but reformatting of IDE generated code.

matozoid: I wanted to create a nodelist with a single item, but very confusingly, if you have only one item it becomes the parent instead of an item?!? This constructor wasn't used anywhere so I removed it. Deprecation would mean that it got in the way for longer. Hopefully nobody depends on this behaviour.

PR #1323, opened 2018-01-12
Speed up storing of initial text information.

sergerrr:
Speed up storing of initial text information. Use tree search instead of linear one.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
@@ -208,7 +208,9 @@ public class LexicalPreservingPrinter {
     private static void storeInitialText(Node root) {
         Map<Node, List<JavaToken>> tokensByNode = new IdentityHashMap<>();
 
-        // We go over tokens and find to which nodes they belong. Note that we choose most specific nodes.
+        // We go over tokens and find to which nodes they belong. Note that we do not traverse the tokens as they were
+        // on a list but as they were organized in a tree. At each time we select only the branch corresponding to the
+        // range of interest and ignore all other branches
         for (JavaToken token : root.getTokenRange().get()) {
             Range tokenRange = token.getRange().orElseThrow(() -> new RuntimeException("Token without range: " + token));
             Node owner = findNodeForToken(root, tokenRange);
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -208,26 +208,13 @@ public void concreteListReplacement(NodeList changedList, int index, Node oldVal
...
-                // we do not consider "phantom" nodes here, like the fake type of variable in FieldDeclaration
-                if (!PhantomNodeLogic.isPhantomNode(node)) {
-                    nodesDepthFirst.add(node);
-                }
-            }
-        }.visitLeavesFirst(root);
-
-        // We go over tokens and find to which nodes they belong. Note that we start from the most specific nodes
-        // and we move up to more general nodes
+        // We go over tokens and find to which nodes they belong. Note that we choose most specific nodes.
```
ftomassetti: Perhaps we can improve this comment, by mentioning the word "tree"

ftomassetti (approved):
This seems a very good idea.

Have you any measurements? Did it improve performance in your use case?

It certainly makes sense on a theoretical level, just wondering if it works well also in practice

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
sergerrr: In my case there is the service class of 11,5 kilolines of code where we ensure that method parameters are annotated with @WebParam(name="..."). There are 720 methods. When I tried javaparser (with lexical preserving to keep rest of code unchanged) for this task it takes more than 7 minutes to be done. And ~10 seconds after fix.

sergerrr: About mentioning word "tree"... My English is not so fluent. Maybe the following?
"// We go over tokens and find to which nodes they belong. Note that we choose most specific nodes traversing down the tree."

ftomassetti: 10 seconds seem << than 7 minutes, even if your cases is a limit case.

About English... probably you know it better than me :)

I think your message is good but I would be slightly more specific, with something like this:

"// We go over tokens and find to which nodes they belong. Note that we do not traverse the tokens as they were on a list but as they were organized in a tree. At each time we select only the branch corresponding to the range of interest and ignore all other branches"

Anyway this is a great contribution and we are very grateful for it!

ftomassetti:
Looks good, +1 on my side, let @matozoid merge it

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
sergerrr: I commited your version of comment. It looks more specific and clear.
Thank you!

matozoid:
Thanks everyone!

PR #1330, opened 2018-01-13
Stop ignoring jars under lib

### LINKED ISSUE #1325: unignore files
These files are ignored but shouldn't be:

git rm --cached --force "javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0/lib/javaslang-2.0.3.jar"
git rm --cached --force "javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0/lib/javassist-3.19.0-GA.jar"
git rm --cached --force "javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0/lib/javaparser-core-3.3.0.jar"
git rm --cached --force "javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0/lib/guava-21.0.jar"

ftomassetti:
Fix #1325

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-lib
+/lib
 bin
 .manager
 target


ftomassetti:
Mmm, maybe some IDEs download dependencies under lib? In that case it could make sense to avoid id

matozoid:
I put that there because some plugin was putting stuff there. I haven't seen it for a while, so let's just pull this.

PR #1331, opened 2018-01-13
Marking some tests as slow tests

### LINKED ISSUE #1326: Make AnalyseJavaParserTest optional
It's taking a huge amount of time in each build.

ftomassetti:
Fix #1326

Using JUnit test categories we can mark some tests as slow (or any other way we want).

Running from the root of the project `mvn test -P NonSlowTests` it takes 1:08 minutes, while running `mvn test` takes 03:24 minutes. 

@matozoid what do you think? No extra directory, just an annotation to use and an extra parameter when launching mvn to skip the slow tests


matozoid:
I'd prefer it the other way around, so that I don't have to go through the slow tests on every build, *unless* I add a parameter.

ftomassetti:
One possibility (that I do not like) would be to drop categories and instead add in each of those tests `assumeTrue("true".equals(System.getProperty("runSlowTests")));`

matozoid:
Well, if I would follow this suggestion with my slow tests, I would be downloading a huge zip on every build :-(

ftomassetti:
A few improvements:
* `mvn test` now skips the slow tests
* we can add the category to the whole test class
* adding the configuration to run tests in parallel reduce the time for running the only JSS non-slow tests from 29 to 20 seconds

Does it work? I am not sure I understand the issue with the huge zip.

Also, if you want we can set the configuration to run tests in parallel also for JP, reducing further the time.

matozoid:
Run `BulkParseTest` if you like, it's an interesting thing.

I'll look at this after I finish something.

matozoid:
I figured my point is bad: the slow tests in -core aren't tests. Well, manual tests really.

matozoid:
It's quite a decrease in test coverage... Hmmmmm....

matozoid:
And I guess your computer is twice as fast as mine :-)
```
[INFO] javaparser-parent .................................. SUCCESS [  0.997 s]
[INFO] javaparser-core .................................... SUCCESS [ 19.970 s]
[INFO] javaparser-testing ................................. SUCCESS [ 43.922 s]
[INFO] javaparser-core-generators ......................... SUCCESS [  0.357 s]
[INFO] javaparser-metamodel-generator ..................... SUCCESS [  0.185 s]
[INFO] javaparser-symbol-solver-model ..................... SUCCESS [  1.579 s]
[INFO] javaparser-symbol-solver-logic ..................... SUCCESS [  1.434 s]
[INFO] javaparser-symbol-solver-core ...................... SUCCESS [  3.150 s]
[INFO] javaparser-symbol-solver-testing ................... SUCCESS [01:00 min]
```

matozoid:
If you prefer running the slow tests by default, then revert to that. I can get used to adding a parameter I think.

ftomassetti:
I think it makes sense to disable them by default for new contributors

ftomassetti:
Now the question is: on Travis and AppVeyor should we run also the slow tests?

matozoid:
Hmmmm, it will make dealing with PR's terrible because pulling one and updating another will take a long time, but it will also increase coverage...

matozoid:
Is this PR mergeable? I'd like to have something in place so my `mvn clean install` goes a little faster :-)

ftomassetti:
For me it is good to go

matozoid:
@ftomassetti - I don't even know how to run the slow tests now :-( How do I do it?

Issue #1311, opened 2018-01-07
Let pretty printer put EOL's between enum constants


matozoid:
Even better: make a configuration option that says "when there are more than this amount of constants, align vertically."

PR #1332, opened 2018-01-13
Make enum constant alignment configurable

matozoid:
See #1311


javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
```diff
@@ -195,7 +195,14 @@ public boolean invalid() {
 
     public enum Category {
 
-        WHITESPACE_NO_EOL, EOL, COMMENT, IDENTIFIER, KEYWORD, LITERAL, SEPARATOR, OPERATOR;
+        WHITESPACE_NO_EOL,
```
ftomassetti, line 198: Nice!

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -150,4 +151,13 @@ public PrettyPrinterConfiguration setOrderImports(boolean orderImports) {
         this.orderImports = orderImports;
         return this;
     }
+
+    public int getMaxEnumConstantsToAlignHorizontally() {
+        return maxEnumConstantsToAlignHorizontally;
+    }
+
+    public PrettyPrinterConfiguration setMaxEnumConstantsToAlignHorizontally(int maxEnumConstantsToAlignHorizontally) {
```
ftomassetti: Do we have a number to indicate infinite? Like -1

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -38,6 +38,7 @@
     private String indent = "    ";
     private String endOfLineCharacter = EOL;
     private Function<PrettyPrinterConfiguration, PrettyPrintVisitor> visitorFactory = PrettyPrintVisitor::new;
+    private int maxEnumConstantsToAlignHorizontally = 5;
```
ftomassetti: We could add a public static constant named `DEFAULT_MAX_ENUM_CONSTANTS_TO_ALIGN_HORIZONTALLY-

ftomassetti (approved):
Good, much nicer code!

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
matozoid: Oh, could use some Javadoc, I guess.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
matozoid: I used to do that because it was standard advice from books and teachers, but I've noticed that in the case of a single use constant it is often clearer not to. In this case: the line says that the default value for `maxEnumConstantsToAlignHorizontally` is 5. Making a constant would only cause an unnecessary indirection in my opinion.

javaparser-core/src/main/java/com/github/javaparser/JavaToken.java
matozoid: Yes! Thanks for suggesting this :-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
ftomassetti: It could be useful if we make it public, so that users could check what is default value and use it in their logic somehow

matozoid: Ah well, here you go then.

PR #1335, opened 2018-01-14
Use Mockito instead of EasyMock

matozoid:
I saw you poor symbol solver people had to use horrible, horrible EasyMock, so I gave you the glory of Mockito instead.

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
@@ -45,8 +45,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.Collections;
 
-import static org.easymock.EasyMock.*;
 import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 public class ContextTest extends AbstractTest {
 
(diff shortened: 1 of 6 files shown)


matozoid:
Intellij was completely confused with all the test java files by the way, thinking they were real source code.

ftomassetti:
Regarding IntelliJ: I think you can mark the problematic subdirectories as "excluded" and it should stop presenting them. Note that IntelliJ is doing the wrong thing because it is considerign for compilation sources that are under in the resources directory

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
```diff
@@ -162,22 +159,20 @@ public void resolveReferenceToClassesInTheSamePackage() throws ParseException {
...
+        ResolvedClassDeclaration compilationUnitDecl = mock(ResolvedClassDeclaration.class);
+        when(compilationUnitDecl.getName()).thenReturn("CompilationUnit");
+        when(compilationUnitDecl.getQualifiedName()).thenReturn("my.packagez.CompilationUnit");
+        TypeSolver typeSolver = mock(TypeSolver.class);
+        when(typeSolver.getRoot()).thenReturn(typeSolver);
+        when(typeSolver.solveType("java.lang.Object")).thenReturn(new ReflectionClassDeclaration(Object.class, typeSolver));
+        when(typeSolver.tryToSolveType("my.packagez.CompilationUnit")).thenReturn(SymbolReference.solved(compilationUnitDecl));
         SymbolSolver symbolSolver = new SymbolSolver(typeSolver);
-        replay(typeSolver, compilationUnitDecl);
+
```
ftomassetti, line 170: No need to specify the replay here?

ftomassetti (approved):
I am convinced!

PR #1337, opened 2018-01-15
Add Intellij files that shouldn't have been ignored

matozoid:
See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

This is so we can mark directories as ignored etc. and share that setting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,24 +6,39 @@ target
 *~
 .project
 .classpath
-.idea/
-*.iml
 release.properties
 **/.DS_Store
-*.class
-
-# Mobile Tools for Java (J2ME)
-.mtj.tmp/
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
-target
 
 .gradle
-*.ipr
-*.iws
 build-*
 *.asc
 build
 out
 .java-version
+
+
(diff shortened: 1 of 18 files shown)


ftomassetti:
Ok, it makes sense. Normally I regenerate project configurations from Maven or Gradle so I typically do not include them but this makes sense and I read the explanation you shared

.gitignore
```diff
@@ -6,24 +6,39 @@ target
...
 release.properties
 **/.DS_Store
-*.class
-
-# Mobile Tools for Java (J2ME)
-.mtj.tmp/
 
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
-target
```
ftomassetti, line 20: is the target directory generated by maven? Could we want to exclude it for some reason?

.idea/codeStyleSettings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ftomassetti, line 1: Nice!

ftomassetti (approved):
Sounds good

.gitignore
matozoid: It was duplicated, so this is just deduplicating it. Target is still ignored :-)

.idea/codeStyleSettings.xml
matozoid: What? XML?

PR #1344, opened 2018-01-17
Link the operators from assign expressions and binary expressions.

### LINKED ISSUE #1342: Suggestion: make enums AssignExpr.Operator and BinaryExpr.Operator consistent
So I was trying to convert some compound assignment expressions (i.e., `+=`, `*=`, etc.) into simple assignment expressions with a binary expression on the right side. For instance, `a += 3` would become `a = a + 3` and so on.

For this, I need to infer, from the compound assignment expression's operator, the equivalent binary expression operator that I want to use on the right side.

I came up with this very simple piece of code, where `assignExpr` is the compound assignment expression I'm looking at:

```
try {
  BinaryExpr.Operator operator = BinaryExpr.Operator.valueOf(assignExpr.getOperator().toString());
  // create the desired BinaryExpr here
} catch (IllegalArgumentException e) {
  // handle the exception somehow, should not happen in practice
}
```

The trick here is to first convert `assignExpr.getOperator()` (which is an element of type `AssignExpr.Operator`) to a String, then convert it to an element of type `BinaryExpr.Operator`. Since there exists no element of type `AssignExpr.Operator` for which there is no equally named element in `BinaryExpr.Operator`, the exception is never raised in practice.

Unfortunately, this does not work for two very simple reasons:
-  `AssignExpr.Operator.AND` corresponds to `BinaryExpr.Operator.BINARY_AND` (and _not_ `BinaryExpr.Operator.AND`)
-  `AssignExpr.Operator.OR` corresponds to `BinaryExpr.Operator.BINARY_OR` (and _not_ `BinaryExpr.Operator.OR`)

So in order to do the conversion correctly I ended up having to do this:

```
try {
  BinaryExpr.Operator operator;
  // unfortunately, the operator AssignExpr.Operator.AND is called BINARY_AND for BinaryExpr, and the
  // operator AssignExpr.Operator.OR is called BINARY_OR for BinaryExpr; otherwise, the names are consistent
  if (assignExpr.getOperator() == AssignExpr.Operator.AND) {
    operator = BinaryExpr.Operator.BINARY_AND;
  } else if (assignExpr.getOperator() == AssignExpr.Operator.OR) {
    operator = BinaryExpr.Operator.BINARY_OR;
  } else {
    operator = BinaryExpr.Operator.valueOf(assignExpr.getOperator().toString());
  }
  // create the desired BinaryExpr here
} catch (IllegalArgumentException e) {
  // handle the exception somehow, should not happen in practice
}
```

Obviously, this is much uglier. :cry:

I suggest renaming `AssignExpr.Operator.AND` to `AssignExpr.Operator.BINARY_AND`, and `AssignExpr.Operator.OR` to `AssignExpr.Operator.BINARY_OR`. That way, the naming of the operators would be consistent between `AssignExpr` and `BinaryExpr`. And consistency is a good thing. :smiley: 

What do you think? Probably, this couldn't happen before JP 4.x to avoid breaking backwards compatibility within JP 3.x.

matozoid:
Fixes #1342

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
@@ -49,27 +49,33 @@ public final class AssignExpr extends Expression {
 
     public enum Operator implements Printable {
 
-        ASSIGN("="),
-        PLUS("+="),
-        MINUS("-="),
-        MULTIPLY("*="),
-        DIVIDE("/="),
-        AND("&="),
-        OR("|="),
-        XOR("^="),
-        REMAINDER("%="),
-        LEFT_SHIFT("<<="),
-        SIGNED_RIGHT_SHIFT(">>="),
-        UNSIGNED_RIGHT_SHIFT(">>>=");
+        ASSIGN("=", null),
+        PLUS("+=", BinaryExpr.Operator.PLUS),
+        MINUS("-=", BinaryExpr.Operator.MINUS),
+        MULTIPLY("*=", BinaryExpr.Operator.MULTIPLY),
+        DIVIDE("/=", BinaryExpr.Operator.DIVIDE),
+        BINARY_AND("&=", BinaryExpr.Operator.BINARY_AND),
+        BINARY_OR("|=", BinaryExpr.Operator.BINARY_OR),
+        XOR("^=", BinaryExpr.Operator.XOR),
+        REMAINDER("%=", BinaryExpr.Operator.REMAINDER),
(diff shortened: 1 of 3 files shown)


matozoid:
@malteskoruppa - if you could double check the mapping, that would be great.

(I could of course have used your name mapping trick, but this is well typed :-) )

malteskoruppa:
@matozoid I double-checked your commit and the mapping makes perfect sense, I think it is 100% correct. :smiley: 

(and of course, your solution is more elegant than the name mapping trick :-) )

matozoid:
@malteskoruppa thanks! Here it goes - you can find it in this weekend's release.

PR #1346, opened 2018-01-18
Clean poms & download test source

matozoid:
@ftomassetti to finish this I would need to know which versions "javaparser_src" and "javaparser_new_src" are...


ftomassetti:
Why do you want to remove these files?

matozoid:
Because
- intellij is still confused and I'm really annoyed with that
- there is no need to have them cluttering up git since they are available in the maven repo

javaparser-symbol-solver-testing/pom.xml
```diff
@@ -38,7 +39,6 @@
                         <configuration>
                             <excludedGroups>com.github.javaparser.SlowTest</excludedGroups>
                             <parallel>methods</parallel>
-                            <threadCount>4</threadCount>
```
ftomassetti: Why? Isn't this making the execution faster?

ftomassetti:
In my IntelliJ marking these directories as excluded solved the issue:
screen shot 2018-01-20 at 12 23 00

However if we can have a solution that reduce the size of the checkout and make easy to get the source (it is done from maven) why not. My only doubt is that we do not have changed any of those files manually...

Also, we include the source code generated by JavaCC which is contained in a separate directory. Can we get that layout?

Regarding the versions: good question... I tried find it by looking at the git history but I could not find any hint.

matozoid:
Won't excluding those directories mean that intellij won't copy those files to the output anymore? So that only the Maven build still works?

Sorry for the emotional bits by the way.

I'll go figure out which JP's are stored there.

ftomassetti:
> Won't excluding those directories mean that intellij won't copy those files to the output anymore? > So that only the Maven build still works?

I am not sure because I always build from Maven, from intellij I run tests and those work.

> Sorry for the emotional bits by the way.

No problem at all, this stuff is frustrating!

> I'll go figure out which JP's are stored there.

That could annoyingly take a while, I tried following the git history to understand at least when I included those files but I had to have moved them around a git does not help me solving the mistery.

matozoid:
I see now why the normal and generated sources are put in a different place. The only way to split that is with a lot of filtering. Ah well.

matozoid:
Okay, the JSS extraction seems to go without a problem, so I'll merge this part.

Issue #1294, opened 2017-12-15
[API] Allow retrieving parsed Javadoc elements


dhardtke:
Hey there,

**Use case**:
I intend to replace / merge existing Javadoc from one `MethodDeclaration` with Javadoc from another `MethodDeclaration`, e.g. when using `* {@inheritDoc}` on inherited methods.

**Problem**:
Right now, it is not possible to retrieve the `JavadocDescriptionElement`s of a `JavadocDescription` instance.
Likewise, I guess it would be nice to also be able to retrieve `JavadocBlockTag`s from `Javadoc` instances.

Is there a particular reason why there are methods like `com.github.javaparser.javadoc.description.JavadocDescription#addElement` or `com.github.javaparser.javadoc.Javadoc#addBlockTag(com.github.javaparser.javadoc.JavadocBlockTag)`, etc.?

 I would prefer to have one method like `getElements()` which would return the raw List so we can modify it based on our wishes.

I can create a PR if my idea is accepted :)

matozoid:
Let's ask our friend @ftomassetti who created this :-)

ftomassetti:
I do not see why not, I would say it is ok to add those methods, however I would suggest to _NOT_ modify the list, for that I would suggest using the API. Ideally I would return an immutable list.

dhardtke:
Then I would suggest to create an immutable List using `Collections.unmodifiableList(list)` and remove methods like `addElement`, etc. - or create a new immutable List every time `addElement` is called.

matozoid:
Hmmm, this API is so different from the rest of JavaParser... I'll make another task to refactor it someday.

matozoid:
Hey @dhardtke - are you still interested in creating that PR? Otherwise I'll do it.

dhardtke:
I can do it, sure. So, what's the plan? Should I add methods like "removeElement()", etc.?

matozoid:
Sure, go wild!

PR #1347, opened 2018-01-19
Improve Javadoc and JavadocDescription to allow modification

dhardtke:
see #1294 

I also did cleaned up the code a little bit (wrong assertEquals() parameter order for example).

It might be a breaking change for some people since I removed Javadoc#getBlockTags() so the List is no longer exposed directly.


matozoid:
Nice :-)

Could you mark getBlockTags deprecated instead of removing it? Add a little Javadoc telling users why.

dhardtke:
Done :)

matozoid:
Cool! Let's see what @ftomassetti thinks of it.

javaparser-core/src/main/java/com/github/javaparser/javadoc/Javadoc.java
```diff
@@ -106,8 +106,25 @@ public JavadocDescription getDescription() {
         return description;
     }
 
+    public JavadocBlockTag getBlockTag(int i) {
```
ftomassetti: Why we do not want to allow the direct modification of the list? Do we want to prevent certain changes?

javaparser-testing/src/test/java/com/github/javaparser/JavadocParserTest.java
```diff
@@ -103,7 +103,8 @@ public void parseBlockTagsAndProvideTagName() {
 
 
         assertEquals(underTest, JavadocParser.parse(expectedText));
-        assertEquals(underTest.getBlockTags().get(0).getTagName(), "unofficial");
+        assertEquals(1, underTest.getBlockTagCount());
```
ftomassetti: Nicer

javaparser-testing/src/test/java/com/github/javaparser/JavadocParserTest.java
```diff
@@ -103,7 +103,8 @@ public void parseBlockTagsAndProvideTagName() {
...
+        assertEquals("unofficial", underTest.getBlockTag(0).getTagName());
```
ftomassetti: Nicer

javaparser-testing/src/test/java/com/github/javaparser/javadoc/JavadocTest.java
```diff
@@ -75,7 +75,7 @@ public void toCommentForJavadocWithTwoLinesOfJustDescriptionAndOneBlockTag() thr
     public void descriptionAndBlockTagsAreRetrievable() {
         Javadoc javadoc = JavaParser.parseJavadoc("first line\nsecond line\n\n@param node a node\n@return result the result");
         assertEquals(javadoc.getDescription().toText(), "first line\nsecond line");
```
ftomassetti: We should also invert this one

ftomassetti (commented):
To me it looks good. I am just not sure about deprecating the method to get a list of blocks. It would made a lot of sense if we wanted to prevent modification, but we don't and so it seems that we are just making more difficult to iterate.

Perhaps we could leave it and make it return a copy of the list (possibly not modifiable)

javaparser-core/src/main/java/com/github/javaparser/javadoc/Javadoc.java
dhardtke: In my opinion direct modification wouldn't be bad, by introducing these additional methods (`getBlockTag(int)`, etc.) we are only cluttering the code without any added value.

We should probably then add a `getElements()` method in `JavadocDescription` for consistency as well.

Regarding modification:
Other classes, such as [SwitchStmt](https://github.com/matozoid/javaparser/blob/268ded9301f347f74267f57927a34e29a6bee7a1/javaparser-core/src/main/java/com/github/javaparser/ast/stmt/SwitchStmt.java#L89) follow the same approach with a visible getter to retrieve the underlying List...

matozoid:
@dhardtke @ftomassetti anybody still working on this? It's about the miss the release.

dhardtke:
I am "still" working on this, as in: I am still waiting for a reply whether it is okay to allow direct modification of the Lists, so that I can add another getter and remove all the other methods I added that clutter the code

JavaParser follows that "regime" at several other places, so why not here as well?! :)

matozoid:
At JavaParser we happily embrace mutability :-D But do wait until @ftomassetti responds.

matozoid:
Hi! What is the state of this?

dhardtke:
@ftomassetti did not answer regarding mutability of the internal lists, but I did implement it regardlessly in https://github.com/javaparser/javaparser/pull/1347/commits/755c9fb2d0135f52d76ec3ec742616806183e951

matozoid:
I'll merge it and people who don't like it can send another PR ;-)

matozoid:
Fixed the merge in another PR

PR #1360, opened 2018-01-24
Deal with no common type for variables being available.

### LINKED ISSUE #1359: Replacing type of a FieldDeclaration with multiple variables
I want to use JavaParser to parse a "template Java" source file and generate multiple source files from it mostly with simply changed field types.

The following self-contained code snippet fails for me with the exception:
`Exception in thread "main" java.lang.AssertionError: The variables do not have a common type.`

Somehow, JavaParser thinks that the PrimitiveType has array levels and the assertion in [NodeWithVariables$1Helper.toArrayLevel()](https://github.com/javaparser/javaparser/blob/b64b7a425769228e24ee8784c7f535fc7b49d32e/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java#L128) fails. Also, [the abort condition](https://github.com/javaparser/javaparser/blob/b64b7a425769228e24ee8784c7f535fc7b49d32e/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java#L124) does not fire since level = 0 and the reported array level of PrimitiveType is also 0.

I am using the Maven release version 3.5.11 of JavaParser.

```Java
import com.github.javaparser.ast.body.FieldDeclaration;
import com.github.javaparser.ast.type.PrimitiveType;
import com.github.javaparser.ast.type.Type;

public class Test {
    CompilationUnit template = JavaParser.parse("class A {float a, b, c, d;}");
    template.accept(new VoidVisitorAdapter<Void>() {
      Type intType = PrimitiveType.intType();

      public void visit(FieldDeclaration n, Void arg) {
        n.getVariables().forEach(vd -> {
          vd.getType().replace(intType);
        });
      }
    }, null);
  }
}
```

matozoid:
Fixes #1359


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
```diff
@@ -109,28 +110,28 @@ default Type getElementType() {
      * <br/>For <code>int[] a[][],b[],c[][];</code> this is int[][].
      */
     @DerivedProperty
-    default Type getMaximumCommonType() {
-        return calculateMaximumCommonType(this.getVariables().stream().map(v -> v.getType()).collect(Collectors.toList()));
+    default Optional<Type> getMaximumCommonType() {
```
matozoid, line 113: This is the main change: there is not always a maximum common type, don't fail hard on it.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
```diff
@@ -145,8 +146,8 @@ private Type toArrayLevel(Type type, int level) {
             // the pretty-printed string got for a node. We just check all them are the same and if they
             // are we just just is not null
             Object[] values = types.stream().map(v -> {
-                Type t = helper.toArrayLevel(v, currentLevel);
-                return t == null ? null : t.toString();
+                Optional<Type> t = helper.toArrayLevel(v, currentLevel);
+                return t.map(Node::toString).orElse(null);
```
matozoid, line 150: I'd say pretty printing is not something that should end up as input for any algorithm, but that's for another time.

matozoid, line 150: Possibly there's a nicer way than `orElse(null)` but I want to go to sleep now :-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -482,7 +482,7 @@ public void visit(final FieldDeclaration n, final Void arg) {
         printMemberAnnotations(n.getAnnotations(), arg);
         printModifiers(n.getModifiers());
         if (!n.getVariables().isEmpty()) {
-            n.getMaximumCommonType().accept(this, arg);
+            n.getMaximumCommonType().ifPresent(t -> t.accept(this, arg));
```
matozoid: In general I'm saying "if there is no common type I'm just going to skip some stuff and fail silently."

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -401,16 +401,17 @@ private static NodeText interpret(Node node, CsmElement csm, NodeText nodeText)
...
+            variableDeclarator.getParentNode().ifPresent(parent -> {
+                NodeWithVariables<?> nodeWithVariables = (NodeWithVariables) parent;
+                nodeWithVariables.getMaximumCommonType().ifPresent(mct -> {
+                    int extraArrayLevels = variableDeclarator.getType().getArrayLevel() - mct.getArrayLevel();
+                    for (int i = 0; i < extraArrayLevels; i++) {
+                        nodeText.addElement(new TokenTextElement(LBRACKET));
+                        nodeText.addElement(new TokenTextElement(RBRACKET));
+                    }
+                });
+            });
```
matozoid: I'm not sure what the impact of not executing this part of the code is, but all the tests still work.

matozoid:
I still need to create a test.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithVariables.java
```diff
@@ -109,28 +110,28 @@ default Type getElementType() {
...
         class Helper {
             // Conceptually: given a type we start from the Element Type and get as many array levels as indicated
             // From the implementation point of view we start from the actual type and we remove how many array
             // levels as needed to get the target level of arrays
             // It returns null if the type has less array levels then the desired target
-            private Type toArrayLevel(Type type, int level) {
+            private Optional<Type> toArrayLevel(Type type, int level) {
                 if (level > type.getArrayLevel()) {
-                    return null;
+                    return Optional.empty();
```
ftomassetti, line 126: Do we know of cases in which it hapens? Can it happen only if something is wrong? Should we throw an exception in that case?

ftomassetti (commented):
I have just one comment: I am not sure when in practice happen that we have a null common type. Perhaps it happen only if there is some AST built manually in a weird way. If that is the case I would think that throwing the exception would make the life of users simpler

matozoid:
I'll go write some test cases and maybe refactor things a bit. Maybe there is a good case for an exception.

javaparser-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -173,18 +174,33 @@ public void prettyAlignMethodCallChains_disabled() {
...
+
+    @Test
+    public void printingInconsistentVariables() {
+        FieldDeclaration fieldDeclaration = parseBodyDeclaration("int a, b;").asFieldDeclaration();
+
+        assertEquals("int a, b;", fieldDeclaration.toString());
+
+        fieldDeclaration.getVariable(0).setType(PrimitiveType.doubleType());
+
+        assertEquals("??? a, b;", fieldDeclaration.toString());
```
matozoid, line 200: @ftomassetti I made the pretty printer print ??? when the type is inconsistent.

javaparser-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/FieldDeclarationTransformationsTest.java
```diff
@@ -44,24 +44,33 @@ protected FieldDeclaration consider(String code) {
...
         it.setModifiers(EnumSet.of(Modifier.PROTECTED));
         assertTransformedToString("protected int A;", it);
     }
 
+    @Test
+    public void changingTypes() {
+        FieldDeclaration it = consider("int a, b;");
+        assertTransformedToString("int a, b;", it);
+        it.getVariable(0).setType("Xyz");
+        assertTransformedToString(" a, b;", it);
```
matozoid, line 72: Lex pres printer prints nothing when the type is inconsistent, and I don't know how to make it print something - I guess it's not really how it's supposed to work. Alternatively, we could throw an exception here. What do you think @ftomassetti ?

ftomassetti: I would say that throwing an exception makes sense. This is better than producing wrong code. Optionally we could make this configurable, but I would go for the exception

ftomassetti:
Now I understand that there are transformations from consistent state to consistent state that requires going through an inconsistent state. This is tricky I do not have a better solution than you do, so +1

javaparser-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/FieldDeclarationTransformationsTest.java
matozoid: I'm not sure how I would throw an exception for this - isn't this all the way in the CSM where we're printing an IDENTIFIER which is null?

PR #1372, opened 2018-01-29
Redo Javadoc formatting

### LINKED ISSUE #1371: Javadoc output bug (3.5.12)
Given the original javadoc:

```java
    /**
     * Gets one of the buttons used in the dialog. Returns null if the specified
     * button does not exist or the dialog has not yet been fully created (for
     * example, via {@link #show()} or {@link #create()}).
     *
     * @param whichButton The identifier of the button that should be returned.
     *            For example, this can be
     *            {@link DialogInterface#BUTTON_POSITIVE}.
     * @return The button from the dialog, or null if a button does not exist.
     */
```

After processing it by adding a tag (`javadoc.addBlockTag(String,String)`):

```java
    /**
     * Gets one of the buttons used in the dialog. Returns null if the specified
     * button does not exist or the dialog has not yet been fully created (for
     * example, via {@link #show()} or {@link #create()}).
     *
     * @param whichButton The identifier of the button that should be returned.            For example, this can be            {@link DialogInterface#BUTTON_POSITIVE}.
     * @return The button from the dialog, or null if a button does not exist.
     * @mytag 3
     */
```

<em><sup>*</sup> Scroll to the right</em>

I think #907 might be a culprit here. New lines are definitely removed, but indent is kept.

 Is there an option to _keep_ original javadoc?

matozoid:
Fixes #1371 

End of lines and indentation are now left alone. This changes current behaviour a bit, but seeing how little reports we get on this I guess it's not a big issue :-)

Anyway, it's a combination of the Javadoc parser and the pretty printer, they were both doing their best to mess up any formatting.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
@@ -25,13 +25,13 @@ import com.github.javaparser.ast.comments.JavadocComment;
 import com.github.javaparser.javadoc.Javadoc;
 import com.github.javaparser.javadoc.JavadocBlockTag;
 import com.github.javaparser.javadoc.description.JavadocDescription;
-import com.github.javaparser.utils.Utils;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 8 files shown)


matozoid:
Looks like I'll need to poke at the Windows build a little :-/

PR #1378, opened 2018-01-31
Avoid cyclic dependency breaking initialization

### LINKED ISSUE #1376: Converting from AssignExpr.Operator to BinaryExpr.Operator sometimes fails
@matozoid I tried to bump up the JavaParser version that I'm using so as to be able to use the operator conversion that you recently added (see https://github.com/javaparser/javaparser/issues/1342).

I stumbled upon the _weirdest_ bug. I have no clue what's causing it, and it took me ages to narrow it down to the following minimal example. This is happening with JavaParser 3.5.12.

Consider this piece of code:
```
@Test
public void testConvertOnly() {
	Optional<BinaryExpr.Operator> op = AssignExpr.Operator.PLUS.toBinaryOperator();
	System.out.println(op); // prints: Optional[PLUS] as expected
	assertTrue(op.isPresent()); // works
}
```

This is working perfectly fine.

Now consider this. I only try to parse an expression, then I do exactly the same thing:
```
@Test
public void testParseAndConvert() {
	JavaParser.parseExpression("i < j");

	Optional<BinaryExpr.Operator> op = AssignExpr.Operator.PLUS.toBinaryOperator();
	System.out.println(op); // prints: Optional.empty ... why ?!
	assertTrue(op.isPresent()); // fails :-(
}
```
I don't know why, but when I try to parse the above expression first, suddenly the operator conversion fails (it doesn't have to be the particular expression `i < j`... it appears parsing any expression that uses a binary operator will trigger this). These things should be completely unrelated, but for some reason they are not. :confused: 

It gets even weirder. When I do the conversion first, THEN parse, THEN do the conversion again, everything works fine again!
```
@Test
public void testConvertAndParseAndConvert() {
	Optional<BinaryExpr.Operator> op = AssignExpr.Operator.PLUS.toBinaryOperator();
	System.out.println(op); // prints: Optional[PLUS]
	assertTrue(op.isPresent()); // works

	JavaParser.parseExpression("i < j");

	op = AssignExpr.Operator.PLUS.toBinaryOperator();
	System.out.println(op); // prints: Optional[PLUS]
	assertTrue(op.isPresent()); // works
}
```

Can you reproduce this? (Please tell me you can, I'm losing my mind here. :rofl:). And if yes, any idea what's happening here?

This was very difficult to narrow down... the tests in my application suddenly started failing randomly, depending on what order they happened to be executed in... even though they are, of course, written to be independent of each other. But, since operator conversion seems to work or not work depending on some unrelated AST parsing that may or may not happen beforehand, they were suddenly not independent anymore. :wink:

matozoid:
Fixes #1376

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
@@ -24,14 +24,19 @@ import com.github.javaparser.ast.AllFieldsConstructor;
 import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
+
 import static com.github.javaparser.utils.Utils.assertNotNull;
+
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.visitor.CloneVisitor;
 import com.github.javaparser.metamodel.AssignExprMetaModel;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
 import com.github.javaparser.printer.Printable;
+
 import javax.annotation.Generated;
+
 import com.github.javaparser.TokenRange;
+
 import java.util.function.Consumer;
 import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
```diff
@@ -49,34 +54,56 @@
...
-            this.binaryOperatorEquivalent = binaryOperatorEquivalent;
         }
 
         public String asString() {
             return codeRepresentation;
         }
 
         public Optional<BinaryExpr.Operator> toBinaryOperator() {
-            return Optional.ofNullable(binaryOperatorEquivalent);
+            switch (this) {
```
ftomassetti, line 81: what about searching over BinaryExpr.values for an element with a matching name?

matozoid: - less type safe (http://wiki.c2.com/?StringlyTyped)
- less dull code
- less likely to break when new operators are introduced
- less efficient

Another alternative: consulting a `Map<AssignExpr.Operator, Optional<BinaryExpr.Operator>>` in that method.

I have a light preference for the current solution because I consider it clean :-)

ftomassetti: I am happy with your solution, not feeling strongly about this

matozoid: Good, then I don't have to change it again ;-)

PR #1382, opened 2018-02-02
Fix infinite recursion when solving java.lang.Object

### LINKED ISSUE #1364: StackoverflowError when using a custom TypeSolver reading JRE sources
I wrote a custom TypeSolver that reads .java source files directly from the "src.zip" distributed with the JDK.
In production, that particular TypeSolver provokes a StackoverflowError inside JavaParser / JavaParserSymbolSolver:
>at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.getAllFields(JavaParserClassDeclaration.java:105)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.lambda$getAllFields$1(JavaParserClassDeclaration.java:105)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1380)

Code is probably most useful, so here is my little reproduction class:

```java
public class Reproduction {
    public static void main(String[] args) throws IOException {
        final CompilationUnit compilationUnit = JavaParser.parse(String.join(System.lineSeparator(),
                "public class Reproduction {",
                "    public void printSomething() {",
                "        System.err.println();",
                "    }",
                "}"));

        final JavaParserFacade facade = JavaParserFacade.get(new ReproductionJreSourcesTypeSolver());
        compilationUnit.accept(new MethodCallVisitor(), facade);
    }

    private static class ReproductionJreSourcesTypeSolver implements TypeSolver {
        private ZipFile zipFile;

        ReproductionJreSourcesTypeSolver() throws IOException {
            final String javaHome = System.getProperty("java.home", null);
            this.zipFile = new ZipFile(Paths.get(javaHome, "lib", "src.zip").toFile(), ZipFile.OPEN_READ);
        }

        @Override
        public TypeSolver getParent() {
            return null;
        }

        @Override
        public void setParent(final TypeSolver parent) {
        }

        @Override
        public SymbolReference<ResolvedReferenceTypeDeclaration> tryToSolveType(final String name) {
            SymbolReference<ResolvedReferenceTypeDeclaration> result = SymbolReference.unsolved(ResolvedReferenceTypeDeclaration.class);

            final String filename = name.replaceAll(Pattern.quote("."), "/") + ".java";
            System.out.printf("Trying to solve %s from filename %s%n", name, filename);

            final ZipEntry entry = zipFile.getEntry(String.format("%s/%s", "java.base", filename));

            if (entry != null) {
                try {
                    final CompilationUnit compilationUnit = JavaParser.parse(zipFile.getInputStream(entry));

                    final String[] nameElements = name.split("\\.");

                    for (int i = nameElements.length; i > 0; i--) {
                        final StringBuilder typeName = new StringBuilder();

                        for (int j = i - 1; j < nameElements.length; j++) {
                            typeName.append(nameElements[j]);
                        }

                        final Optional<TypeDeclaration<?>> astTypeDeclaration = Navigator.findType(compilationUnit, typeName.toString());

                        if (astTypeDeclaration.isPresent()) {
                            result = SymbolReference.solved(JavaParserFacade.get(this).getTypeDeclaration(astTypeDeclaration.get()));
                        }
                    }
                } catch (IOException ignored) {
                }
            }

            return result;
        }
    }

    private static class MethodCallVisitor extends VoidVisitorAdapter<JavaParserFacade> {
        @Override
        public void visit(final MethodCallExpr n, final JavaParserFacade arg) {
            System.out.println(arg.solve(n));

            super.visit(n, arg);
        }

        public void visit(final ObjectCreationExpr n, final JavaParserFacade arg) {
            System.out.println(arg.solve(n));

            super.visit(n, arg);
        }
    }
}
```
I couldn't figure out why this happens, for some reason the Visitor leads to solving of "java.lang.Object" over and over again (see the output).

PS:
When the JreTypeSolver works flawlessly, I thought of sharing it with other users (maybe I'll open a PR). Also, I wrote another TypeSolver that is capable of resolving types from JAR files / any files on the classpath.

dhardtke:
This fixes #1364 by not recursing in the `getAllAncestors()` method when being the `java.lang.Object` class already.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -290,16 +290,21 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration {
     @Override
     public List<ResolvedReferenceType> getAncestors() {
         List<ResolvedReferenceType> ancestors = new ArrayList<>();
-        ResolvedReferenceType superclass = getSuperClass();
-        if (superclass != null) {
-            ancestors.add(superclass);
-        }
-        if (wrappedNode.getImplementedTypes() != null) {
-            for (ClassOrInterfaceType implemented : wrappedNode.getImplementedTypes()) {
-                ResolvedReferenceType ancestor = toReferenceType(implemented);
-                ancestors.add(ancestor);
+
+        // We want to avoid infinite recursion in case of Object having Object as ancestor
+        if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
+            ResolvedReferenceType superclass = getSuperClass();
+            if (superclass != null) {
+                ancestors.add(superclass);
+            }
+            if (wrappedNode.getImplementedTypes() != null) {
+                for (ClassOrInterfaceType implemented : wrappedNode.getImplementedTypes()) {
+                    ResolvedReferenceType ancestor = toReferenceType(implemented);
+                    ancestors.add(ancestor);
+                }
             }
         }
+
         return ancestors;
     }
 
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1364.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1364.java
@@ -1,0 +1,91 @@
+package com.github.javaparser.symbolsolver;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseResult;
+import com.github.javaparser.ParseStart;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.Providers;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.expr.MethodCallExpr;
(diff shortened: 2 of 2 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -290,16 +290,21 @@ public boolean isTypeParameter() {
...
         List<ResolvedReferenceType> ancestors = new ArrayList<>();
-        ResolvedReferenceType superclass = getSuperClass();
-        if (superclass != null) {
-            ancestors.add(superclass);
-        }
-        if (wrappedNode.getImplementedTypes() != null) {
-            for (ClassOrInterfaceType implemented : wrappedNode.getImplementedTypes()) {
-                ResolvedReferenceType ancestor = toReferenceType(implemented);
-                ancestors.add(ancestor);
+
```
ftomassetti, line 293: I agree with the solution, which is necessary if we want to support parsing Object itself. I never thought that was possible.

ftomassetti (approved):

matozoid:
Any chance for a unit test?

dhardtke:
Done, though it isn't too nice that the test is considered to be satisfied if the symbol solver throws an `UnsolvedSymbolException`.

Btw: why does the solver throw the deprecated `UnsolvedSymbolException` from `com.github.javaparser.symbolsolver.javaparsermodel.UnsolvedSymbolException`, and not the one from `com.github.javaparser.resolution.UnsolvedSymbolException`?

ftomassetti:
I think because they are in different modules (not at PC right now)

PR #1388, opened 2018-02-02
Remove TODO

### LINKED ISSUE #395: Inspect TODO verify it does not print comments connected to the type
This TODO is found in MethodDeclaration.java

matozoid:
Fixes #395 

It's rather obviously using the no comments configuration. You might get annotations printed there, though.

I still think this code shouldn't be in here, it's too much effort to keep in working order.


ftomassetti:
I honestly do not remember what I meant with that comment so... +1

PR #1390, opened 2018-02-02
[DRAFT]  fix comment attribution todo

### LINKED ISSUE #398: Inspect TODO < or <= ?
Found in PositionUtils.java

matozoid:
Fixes #398


javaparser-core/src/main/java/com/github/javaparser/utils/PositionUtils.java
```diff
@@ -93,50 +90,50 @@ private static int compare(Node a, Node b, boolean ignoringAnnotations) {
...
-            ClassOrInterfaceDeclaration casted = (ClassOrInterfaceDeclaration) node;
-            return casted.getName();
-        } else {
-            return node;
-        }
+    private static Position beginPositionWithoutConsideringAnnotations(Node node) {
+        return getLastAnnotation(node)
+                .flatMap(annotation -> annotation.getRange()
+                        .map(range -> range.end))
+                .orElse(node.getRange().get().begin);
```
matozoid, line 118: I found this weird method that says "I'll tell you where to find the end of the annotations" but does "here's a node that I consider after the annotations for some types".

I thought I'd better just find the position after the annotations (or should we be after the modifiers too, like `public` `abstract` etc.? In case of a class it even skips `class`!) but that breaks a test:

```java
class Issue40{
    @GET
    @Path("original")
    /**
    * Return the original user.
    */
    public User getOriginalUser(String userName) {
        return userService.getOriginalUser(userName);
    }
}
```
This no longer has a comment on the method, it has a comment on the `User` return type. Which is logical since my change moved the "begin position without considering annotations" from the start of `User` to the end of `@Path("original")`.

@ftomassetti - considering that this Javadoc is illegal (it must be before the annotations), where should the comment go?

ftomassetti: Good question, with no absolute answer, as it frequently the case with comments. I would still feel the comment should go on the method but I am not sure we want to make adapt  comments attributions to treat annotations specially

matozoid: Comment attribution has a setting for this already - I'm struggling with it here :-) Yeah, I think it should go on the method too, which means the whole setting is badly named. It's more like a "put comments on logical places" instead of "put comments on the first following thing."

ftomassetti: Well, the fact is that users perceive the method to start in a different place w.r.t. where the AST perceives it start. We have to reconciliate these views where they differ

ftomassetti:
What do we want to do with this PR?

matozoid:
Work on it :-) But I'll close it for now.

PR #1403, opened 2018-02-03
Issue 883 var type

matozoid:
Looks like there is going to be only one language change in Java 10 (which is out in a few months already!) and it's the `var` type.

@ftomassetti we'll need to be able to resolve these things. How do we do that? Do check https://bugs.openjdk.java.net/browse/JDK-8151454


matozoid:
Everything except symbol solving works. @ftomassetti how should I approach this? Or can you do it?

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -35,14 +40,73 @@
...
+        JAVA_1_1(new Java1_1Validator(), null),
+        JAVA_1_2(new Java1_2Validator(), null),
+        JAVA_1_3(new Java1_3Validator(), null),
+        JAVA_1_4(new Java1_4Validator(), null),
+        JAVA_5(new Java5Validator(), null),
+        JAVA_6(new Java6Validator(), null),
+        JAVA_7(new Java7Validator(), null),
+        JAVA_8(new Java8Validator(), null),
+        JAVA_9(new Java9Validator(), null),
+        JAVA_10(null, new Java10PostProcessor());
```
ftomassetti: Maybe we could call it JAVA_10_preview and tell that we are going to remove it later, when Java_10 is actually released

javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
```diff
@@ -78,7 +73,7 @@ public String asString() {
         }
 
         public Optional<BinaryExpr.Operator> toBinaryOperator() {
-            switch (this) {
+            switch(this) {
```
ftomassetti, line 76: I like the space before the parenthesis :)

ftomassetti (approved):
I do not see any issue with this

ftomassetti:
Resolving this thing should be easy, we have just to assign them the correct type using inference

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
matozoid: Hmmm, yes, that's an interesting suggestion.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/AssignExpr.java
matozoid: Thank you, it's done with Code Generation (c)

matozoid:
@ftomassetti are you going to build the resolver?

Be sure to read the JEP before doing so.

PR #1405, opened 2018-02-04
Add removeFirst and removeLast to NodeList.

stephenramthun:
Issue #1389

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
@@ -87,6 +87,27 @@ public class NodeList<N extends Node> implements List<N>, Iterable<N>, HasParent
         return innerList.remove(node);
     }
 
+    public boolean removeFirst() {
+        if (innerList.size() < 1) {
+            return false;
+        }
+        Node node = innerList.get(0);
+        notifyElementRemoved(0, node);
+        node.setParentNode(null);
+        return innerList.remove(0) != null;
+    }
+
+    public boolean removeLast() {
+        if (innerList.size() < 1) {
+            return false;
+        }
+        int index = innerList.size() - 1;
+        Node node = innerList.get(index);
+        notifyElementRemoved(index, node);
+        node.setParentNode(null);
+        return innerList.remove(index) != null;
+    }
+
     @SafeVarargs
     public static <X extends Node> NodeList<X> nodeList(X... nodes) {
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
```diff
@@ -87,6 +87,27 @@ public boolean remove(Node node) {
...
 
+    public boolean removeFirst() {
+        if (innerList.size() < 1) {
+            return false;
+        }
+        Node node = innerList.get(0);
+        notifyElementRemoved(0, node);
+        node.setParentNode(null);
+        return innerList.remove(0) != null;
+    }
```
matozoid: Isn't it easier to use the methods on NodeList instead of manipulating innerlist?

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeListTest.java
```diff
@@ -137,6 +139,48 @@ public void removeNode() {
...
+        assertEquals(Arrays.asList("'int a;' REMOVAL in list at 0"), changes);
+        assertEquals(cd.getMembers().size(), 4);
+
+        for (int i = 3; i >= 0; i--) {
+            assertTrue(cd.getMembers().removeFirst());
+            assertEquals(cd.getMembers().size(), i);
+        }
+
+        assertFalse(cd.getMembers().removeFirst());
+        assertEquals(cd.getMembers().size(), 0);
```
matozoid: Thorough!

javaparser-testing/src/test/java/com/github/javaparser/ast/NodeListTest.java
```diff
@@ -271,7 +315,7 @@ public void addBeforeFirst() {
         final NodeList<Name> list = nodeList(abc, new Name("bcd"), new Name("cde"));
 
         list.addBefore(new Name("xxx"), abc);
-        
+
```
matozoid, line 316: Oops, someone hit the reformat button ;-) (Don't worry about it)

matozoid:
Thanks!

javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
stephenramthun: You are absolutely right! I've simplified both removeFirst and removeLast so that they use NodeList's remove-method now.

PR #1419, opened 2018-02-14
Make a simple "var" inferrer.

### LINKED ISSUE #1408: Infer "var" type

matozoid:
Fixes #1408.

@ftomassetti - what is the correct way to deal with errors? What if we try to resolve `var` in a position where it was not allowed to be? What if there is no initializer to take the type from?

@ftomassetti - is the test written well?

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -474,8 +474,14 @@ public class JavaParserFacade {
             UnionType unionType = (UnionType) type;
             return new ResolvedUnionType(unionType.getElements().stream().map(el -> convertToUsage(el, context)).collect(Collectors.toList()));
         } else if (type instanceof VarType) {
-            final VariableDeclarator variableDeclarator = (VariableDeclarator)type.getParentNode().get();
-            return variableDeclarator.getInitializer().get().calculateResolvedType();
+            Node parent = type.getParentNode().get();
+            if (!(parent instanceof VariableDeclarator)) {
+                throw new IllegalStateException("Trying to resolve a `var` which is not in a variable declaration.");
+            }
+            final VariableDeclarator variableDeclarator = (VariableDeclarator) parent;
+            return variableDeclarator.getInitializer()
+                    .map(Expression::calculateResolvedType)
+                    .orElseThrow(() -> new IllegalStateException("Cannot resolve `var` which has no initializer."));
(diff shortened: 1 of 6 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
```diff
@@ -473,6 +473,9 @@ protected ResolvedType convertToUsage(com.github.javaparser.ast.type.Type type,
         } else if (type instanceof UnionType) {
             UnionType unionType = (UnionType) type;
             return new ResolvedUnionType(unionType.getElements().stream().map(el -> convertToUsage(el, context)).collect(Collectors.toList()));
+        } else if (type instanceof VarType) {
+            final VariableDeclarator variableDeclarator = (VariableDeclarator)type.getParentNode().get();
+            return variableDeclarator.getInitializer().get().calculateResolvedType();
```
ftomassetti: This is what I would do

ftomassetti (approved):
Wow, this seems surprisingly simple!

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/VarTypeTest.java
```diff
@@ -0,0 +1,46 @@
...
+            .setSymbolResolver(new JavaSymbolSolver(typeSolver)));
+
+    @Test
+    public void resolveAPrimitive() {
+        CompilationUnit ast = javaParser.parse(ParseStart.COMPILATION_UNIT, provider("class X{var abc = 1;}")).getResult().get();
+        VarType varType = ast.findFirst(VarType.class).get();
+
+        ResolvedType resolvedType = varType.resolve();
+
+        assertEquals(ResolvedPrimitiveType.byName("int"), resolvedType);
```
ftomassetti: I would use ResolvedPrimitiveType.INT

ftomassetti:
Tests look good.

Regarding the fact an initializer is not present: I would throw an IllegalStateException, but we could create also some hierarchy of exceptions to represent these semantic problems

matozoid:
@ftomassetti I've added the error handling. In my experience exception hierarchies don't really add much - people just want to know what went wrong and fix it, so a helpful message is better :-)

matozoid:
Oh, @ftomassetti - in which package should this test be?

ftomassetti:
I think under resolution is the best option. We have a typeinference package but it is about something different

matozoid:
Okay, so I guess this is good to go!

PR #1423, opened 2018-02-19
SourceRoot implementation for JSS

daanschipper:
Utility class to add all jars and source roots of java files to a `TypeSolver`. It traverses the file tree and searches for files with the extension `.jar` and `.java`, which are then added to the typesolver. It keeps track of which source roots have already been added to the typesolver and skips a directory if its a subpath of a known root.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
@@ -1,0 +1,147 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2018 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.symbolsolver.utils;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseProblemException;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
+import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.JarTypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
+import com.github.javaparser.utils.Log;
+
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.logging.Level;
+
+import static com.github.javaparser.utils.Utils.assertNotNull;
+import static java.nio.file.FileVisitResult.CONTINUE;
+import static java.nio.file.FileVisitResult.SKIP_SIBLINGS;
+
+public class SymbolSourceRoot {
+
+    private static java.util.logging.Logger logger = java.util.logging.Logger.getLogger(JavaParserFacade.class.getCanonicalName());
+
+    private final Path root;
+    private CombinedTypeSolver typeSolver = new CombinedTypeSolver(new ReflectionTypeSolver(false));
+
+    public SymbolSourceRoot(Path root) {
+        assertNotNull(root);
+        if (!Files.isDirectory(root)) {
+            throw new IllegalArgumentException("Only directories are allowed as root path!");
+        }
+        this.root = root.normalize();
+        Log.info("New symbol source root at \"%s\"", this.root);
+    }
+
+    public TypeSolver walk() throws IOException {
+        Files.walkFileTree(root, new JavaSymbolSolverWalker());
+        Files.walkFileTree(root, new JarVisitor());
+        return typeSolver;
+    }
+
+    public Optional<TypeSolver> tryToWalk() {
+        try {
+            Files.walkFileTree(root, new JavaSymbolSolverWalker());
+            Files.walkFileTree(root, new JarVisitor());
+            return Optional.of(typeSolver);
+        } catch (IOException e) {
+            return Optional.empty();
+        }
+    }
+
+    public TypeSolver getTypeSolver() {
+        return typeSolver;
+    }
+
+    /**
+     * The path that was passed in the constructor.
+     */
+    public Path getRoot() {
+        return root;
+    }
+
+    /**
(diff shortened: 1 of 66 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
```diff
@@ -0,0 +1,147 @@
...
+        return typeSolver;
+    }
+
+    public Optional<TypeSolver> tryToWalk() {
+        try {
+            Files.walkFileTree(root, new JavaSymbolSolverWalker());
+            Files.walkFileTree(root, new JarVisitor());
+            return Optional.of(typeSolver);
+        } catch (IOException e) {
+            return Optional.empty();
```
ftomassetti: We could log the error here

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public TypeSolver walk() throws IOException {
+        Files.walkFileTree(root, new JavaSymbolSolverWalker());
+        Files.walkFileTree(root, new JarVisitor());
+        return typeSolver;
+    }
+
+    public Optional<TypeSolver> tryToWalk() {
+        try {
+            Files.walkFileTree(root, new JavaSymbolSolverWalker());
```
ftomassetti: We could call `walk` here

ftomassetti (approved):
Thank you! This would be very useful.
It could be a bit tricky to test this one but if we could write a unit test it would be perfect

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Set;
+import java.util.logging.Level;
+
+import static com.github.javaparser.utils.Utils.assertNotNull;
+import static java.nio.file.FileVisitResult.CONTINUE;
+import static java.nio.file.FileVisitResult.SKIP_SIBLINGS;
+
+public class SymbolSourceRoot {
+
+    private static java.util.logging.Logger logger = java.util.logging.Logger.getLogger(JavaParserFacade.class.getCanonicalName());
```
matozoid: It's better to use only the Log class, no use making logging configuration more complex :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.HashSet;
+import java.util.Optional;
+import java.util.Set;
+import java.util.logging.Level;
+
+import static com.github.javaparser.utils.Utils.assertNotNull;
+import static java.nio.file.FileVisitResult.CONTINUE;
+import static java.nio.file.FileVisitResult.SKIP_SIBLINGS;
+
+public class SymbolSourceRoot {
```
matozoid: "Source root" is the name we use for a directory full of source files, with the contained directories corresponding to packages. I'm guessing this is more like a LibraryRoot or ProjectRoot or JarRoot or so?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
```diff
@@ -0,0 +1,147 @@
...
+
+        @Override
+        public FileVisitResult visitFile(Path file, BasicFileAttributes attr) throws FileNotFoundException {
+            if (attr.isRegularFile()) {
+                PathMatcher matcher = FileSystems.getDefault().getPathMatcher("glob:**.java");
+                if (matcher.matches(file)) {
+                    try {
+                        Optional<Path> root = JavaParser.parse(file.toFile()).getStorage()
+                                .map(CompilationUnit.Storage::getSourceRoot);
+                        if (root.isPresent()) {
```
matozoid: Try ifPresent :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
matozoid: This definitely needs clear documentation. The description of the PR is a good start.

The name is pretty confusing as it seems to imply it has something to do with SourceRoot, which it does not. It's more of a... SymbolSolverQuickSetup? ProjectConfigurationAnalyser?

matozoid:
My comments are disappearing :-O

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
daanschipper: I purposely did it this way, since I need to return `SKIP_SIBLINGS` for the FileVisitor

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
daanschipper: I agree, I was not sure on the naming of the class. I'll change it and add some documentation

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSourceRoot.java
matozoid: Ah, okay!

daanschipper:
I'm working on adding a junit test now!

matozoid:
Here, I wrote something on why that custom logger exists: https://matozoid.github.io/2018/01/19/javaparsers-logger.html

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/utils/SymbolSolverQuickSetupTest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.github.javaparser.utils.SourceRoot;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+
+/**
```
daanschipper: More types of classes can be tried to be resolved if desired

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/utils/SymbolSolverQuickSetupTest.java
```diff
@@ -0,0 +1,50 @@
...
+ * Try to resolve all the ClassOrInterfaceDeclaration and MethodCallExpr in some resources folder. If it fails to do
+ * so, an IllegalStateException is thrown.
+ */
+public class SymbolSolverQuickSetupTest {
```
daanschipper, line 20: More types of classes can be tried to be resolved if desired

matozoid:
+1

daanschipper:
Is there a better method for testing if a class can be resolved than testing for each class if 
`compilationUnit.findAll(SomeResolvable.class).forEach(SomeResolvable::resolve));` does not throw any errors?

matozoid:
That's an @ftomassetti question.

daanschipper:
I encountered some additional problems I need to resolve first, this version is not 100% working

matozoid:
Alright, instead of closing you can also edit the title to "[DRAFT] blablabla"

daanschipper:
Oke, I know how to fix it, but i didn't have access to a laptop to build the project properly. I plan on fixing it this week

daanschipper:
The problems I encountered where that when the main directory contained a hidden directory, the whole directory is skipped. Fixed this by changing `SKIP_SIBLINGS` to `SKIP_SUBTREE` in the `FileWalker`. Furthermore I added the option to add a directory filter. If the method within return true for some provided path the directory is skipped. I used this to skip the test folder for instance.

matozoid:
Oh, I noticed `SKIP_SIBLINGS` in your suggested code for `SourceRoot` but didn't mention it. Oops, sorry!

Anyway, thanks for the effort :-)

PR #1424, opened 2018-02-19
Create functional language levels. Fix a random superfluous interface.

matozoid:
Tears postprocessors and validators for language levels apart so advanced users can use them separately.

Creates new language levels for "the very latest features" and "the most common Java version."

@ftomassetti removes an "implements NodeWithDeclaration" that was already done in the parent class. It seems that the "getAllInterfaces" call now reports one less implemented interface. Is that correct? It implements the same interface twice, should it be reported twice?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
@@ -44,10 +44,12 @@ public class ParserConfiguration {
     public enum LanguageLevel {
         /** Does no post processing or validation. Only for people wanting the fastest parsing. */
         RAW(null, null),
-        /** Always points to the most popular Java version. */
+        /** The most used Java version. */
         POPULAR(new Java8Validator(), null),
-        /** Always points to the newest Java version supported. */
-        LATEST(new Java11Validator(), new Java11PostProcessor()),
+        /** The latest Java version that is available. */
+        CURRENT(new Java8Validator(), null),
+        /** The newest Java features supported. */
+        BLEEDING_EDGE(new Java11Validator(), new Java11PostProcessor()),
         /** Java 1.0 */
         JAVA_1_0(new Java1_0Validator(), null),
         /** Java 1.1 */
--- a/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
@@ -42,19 +42,36 @@ import static com.github.javaparser.utils.Utils.assertNotNull;
  */
 public class ParserConfiguration {
     public enum LanguageLevel {
+        /** Does no post processing or validation. Only for people wanting the fastest parsing. */
         RAW(null, null),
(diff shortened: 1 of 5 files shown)


javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -42,19 +42,36 @@
  */
 public class ParserConfiguration {
     public enum LanguageLevel {
+        /** Does no post processing or validation. Only for people wanting the fastest parsing. */
         RAW(null, null),
+        /** Always points to the most popular Java version. */
+        POPULAR(new Java8Validator(), null),
```
ftomassetti: CURRENT?

javaparser-core/src/main/java/com/github/javaparser/version/Java10PostProcessor.java
```diff
@@ -1,35 +1,24 @@
...
 
 /**
  * Processes the generic AST into a Java 10 AST and validates it.
  */
 public class Java10PostProcessor extends PostProcessors {
-    protected final PostProcessor java10Validator = new Java10Validator().postProcessor();
-    protected final PostProcessor varNodeCreator = new PostProcessor() {
-        @Override
-        public void process(ParseResult<? extends Node> result, ParserConfiguration configuration) {
+    protected final PostProcessor varNodeCreator = (result, configuration) ->
```
ftomassetti, line 12: Nice

ftomassetti (approved):
Looks good!

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
matozoid: You mean: add CURRENT? Or "this should be CURRENT"?

matozoid:
@ftomassetti - there's a possible bug in the symbol solver in my remarks in the description

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
ftomassetti: I mean this should be current

ftomassetti:
About the interface being reported twice: good question, I am not sure

matozoid:
I've made an issue to track it.

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
matozoid: Java 9 is current, but Java 8 is popular. I thought it would be useful to have a "most commonly used" option. I can add current?

ftomassetti:
What about calling it default?

matozoid:
@ftomassetti that gives raise to another question (but I think there's an issue for that?) - what should be the default level?

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
matozoid: CURRENT added :-)

PR #1437, opened 2018-02-22
Escape eol's for JSON output

### LINKED ISSUE #1421: JsonPrinter sometimes outputs malformed JSON (multi-line comments)
Hey there,

First off, thank you all for creating `javaparser`, it's awesome! Also, I'm a big fan of @ftomassetti's detailed blog posts on parsing languages (e.g. [Java](https://tomassetti.me/parsing-in-java/) or [Python](https://tomassetti.me/parsing-in-python/)).

I'm working on a tool that uses `javaparser` that I hope to open source in a few months.

## The Issue

I've been playing with `javaparser` and it's been great, but `JsonPrinter` seems to print out malformed JSON in certain cases, specifically with multi-line comments.

## Reproduction Steps

In my `pom.xml` I have:

~~~xml
<dependency>
    <artifactId>java-symbol-solver-core</artifactId>
    <version>0.6.0</version>
</dependency>
~~~

My `javaparser` code, based on @matozoid's [blog post](https://matozoid.github.io/2017/12/13/inspecting-an-ast.html), looks like:

~~~java
// JavaParserMain.java
...
CompilationUnit cu = JavaParser.parse(new FileInputStream("JsonPrintingExample.java"));
JsonPrinter printer = new JsonPrinter(true);
// Save result of printer.output(cu) to a file
~~~

A minimal example file that demonstrates the issue is:
~~~java
/*
 * Some comment
 */
public class JsonPrintingExample {
}
~~~

After running my `javaparser` code (listed above) on the example file, the resulting JSON is:
~~~json
{"type":"CompilationUnit","types":[{"type":"ClassOrInterfaceDeclaration","isInterface":"false","name":{"type":"SimpleName","identifier":"JsonPrintingExample"},"comment":{"type":"BlockComment","content":"
 * Some comment
 "}}]}
~~~

When you try to parse the file:
~~~bash
$ cat output.json | json_pp
invalid character encountered while parsing JSON string, at character offset 203 (before "\n * Some comment\n ...") at /usr/bin/json_pp5.18 line 45.
~~~

## Troubleshooting

I observe the same error when I remove the asterix before the comment text or when I move the comment onto just 2 lines, e.g.

~~~java
/*
 Some comment
 */
~~~
or
~~~java
/* Some comment
 */
~~~
However, the following results in valid JSON:
~~~java
/* Some comment */
~~~

## Potential Solutions

Would this be an easy fix in [JsonPrinter](https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/printer/JsonPrinter.java)?

I see #1353, which looks like you've run into some errors like this before- "so that we can use a real library and avoid silly escaping problems."

I've found [Google gson](https://github.com/google/gson) fairly easy to use in prior projects, but I'm not sure if it's a good fit here.

What do you all think?

Also, thanks in advance for your time! :)

matozoid:
Fixes #1421

### DIFF
--- a/javaparser-testing/src/test/java/com/github/javaparser/printer/JsonPrinterTest.java
+++ b/javaparser-testing/src/test/java/com/github/javaparser/printer/JsonPrinterTest.java
@@ -9,6 +9,12 @@ import static com.github.javaparser.JavaParser.*;
 import static com.github.javaparser.utils.Utils.EOL;
 import static org.junit.Assert.*;
 
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
+import java.io.BufferedWriter;
+
 public class JsonPrinterTest {
     @Test
     public void testWithType() {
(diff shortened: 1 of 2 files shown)


javaparser-testing/src/test/java/com/github/javaparser/printer/JsonPrinterTest.java
```diff
@@ -9,6 +9,12 @@
 import static com.github.javaparser.utils.Utils.EOL;
 import static org.junit.Assert.*;
 
+import java.io.FileOutputStream;
```
clintgibler, line 12: I was using this imports for writing the results to file to more easily check that it was valid JSON. You can probably remove these `java.io` imports.

PR #1458, opened 2018-03-06
Fix incorrect order in search for class.

### LINKED ISSUE #1457: SymbolSolver fails with StackOverflow because of incorrect ancestor search order
Offtop: these stack overflow bugs get funnier the deeper I go.
So in project (https://github.com/KNightWeng/TAkeMeHome) there's a situation that can be simplified like so: there're classes ``name1.A``, ``name1.B``, ``name2.B``. The hierachy is like so: ``name1.A`` extends ``name1.B``, ``name2.B`` extends ``name1.A``.
When solving ``name1.A`` its ancestor is solved, and at ComilationUnitContext:161 the imports are processed before the package, in which ``name1.A`` resides. And one of the imports is ``name2.*``. Because of incorrect processing order, ``name2.B`` is solved as ancestor of ``name1.A`` (instead of proper ``name1.B``).
It wouldn't be a big issue if ``name2.B`` didn't extend ``name1.A``. But it does. And so ``A`` must be solved when the ancestors of ``name2.B`` are solved. If you remember, ``A`` was our starting point.
So the code loops endlessly.

P.S. After writing this, I started to wonder if maybe the order is indeed ``imports > package``, but the IntellIJ Idea supports my opinion that ``package > imports``.

AwesomeLemon:
Fix #1457
Simple change of search order.
Was:
1) imports
2) current package

Now:
1) current package
2) imports

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -129,6 +129,22 @@ public class CompilationUnitContext extends AbstractJavaParserContext<Compilatio
             }
         }
 
+        // Look in current package
+        if (this.wrappedNode.getPackageDeclaration().isPresent()) {
+            String qName = this.wrappedNode.getPackageDeclaration().get().getName().toString() + "." + name;
+            SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
+            if (ref != null && ref.isSolved()) {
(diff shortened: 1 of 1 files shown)


ftomassetti:
Looks good!
With a test it would be perfect... :D

AwesomeLemon:
Unfortunately, I do not have time right now to get acquainted with a new for me test library.
I understand a wish for test, but cannot promise anything.
At least this fixed my problems in that repo I referenced, and all existing tests remained unbroken. :D

ftomassetti:
Then it will be good enough for now. Thank you for your contribution: you made JavaParser better!

AwesomeLemon:
Glad to help!

matozoid:
@AwesomeLemon if you were looking on the wiki where it mentions JBehave, then you'll be happy to hear that I just updated that page (after three years, oops) - https://github.com/javaparser/javaparser/wiki/Testing

PR #1466, opened 2018-03-13
add missed parameter

izebit:
Maybe I am wrong, but it seems to me that there is a mistake here


ftomassetti:
mmm... it seems suspicious indeed. Let's see what tests think about that

matozoid:
+1

We should document that this is now the way to do symbol solving: https://github.com/javaparser/javasymbolsolver-maven-sample/blob/master/src/main/java/com/yourorganization/maven_sample/MyAnalysis.java

izebit:
Sorry, I don't quite follow you. Did I understand you correctly, that I had to write tests for this case?

matozoid:
@izebit the test suite runs for every PR. It has passed as you can see from all the green :-)

izebit:
@matozoid To be honest, I don't understand what you are talking about. I don't have much knowledge about  `java parser`. To my mind, your example isn't related to my case, and what should I do?

matozoid:
@izebit we seem to be missing each other's points here. The PR is fine. Thank you for contributing!

PR #1471, opened 2018-03-17
add implementations for checking of existing annotations and getting them

izebit:
Hello, I found that several methods in javassist models didn't have implementations and I made a decision to implement them.    
When I was writing tests, I was puzzled because I needed to have jar artifacts with specific classes and interfaces. For example, I needed enum, which implemented an interface, and moreover, this interface should have an annotation. So, I added my jar file to the pr. I don't know what the best solution is in this case. I am ready to discuss it and try to find a better solution than I have now.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistInterfaceDeclaration.java
```diff
@@ -182,19 +182,7 @@ public boolean isAssignableBy(ResolvedReferenceTypeDeclaration other) {
...
-                }
-                if (Arrays.stream(annotationRaw.getClass().getInterfaces()).anyMatch(it -> it.getCanonicalName().equals(canonicalName))) {
-                    return true;
-                }
-            }
-        } catch (ClassNotFoundException e) {
-            throw new RuntimeException(e);
-        }
-        return false;
+        return ctClass.hasAnnotation(canonicalName);
```
matozoid, line 185: I'd like to see what @ftomassetti thinks of this before I merge.

ftomassetti: It seems correct to me. My only doubt is it this try tro load the class or not, I think not, so it should be ok

ftomassetti (approved):
Looks good to me

matozoid:
That, and the existence of tests is good enough for me :-) Thank you again, @izebit !

izebit:
You are welcome, I am glad to be useful for the project 😉

PR #1495, opened 2018-04-05
Added .iml file for javaparser-symbol-solver-model

malteskoruppa:
I cloned the repository since I intend to write a few PRs in the future. The command `mvn clean install` worked fine.

However, when I then imported the project into IntelliJ IDEA, the IDE complained because it refused to see the module `javaparser-symbol-solver-model`. Thus, any file throughout the project that depended on some class in the module `javaparser-symbol-solver-model` was marked as erroneous because the IDE could not find that dependency (even though it was there and I could see its files in the project window). Therefore, I could not sensibly work on the source code or run tests.

It turned out that the problem was that the module `javaparser-symbol-solver-model` is missing an `.iml` file in its root directory. *All* other modules (`javaparser-core`, `javaparser-symbol-solver-core`, etc.) come with an `.iml` file. Is there any reason why the module `javaparser-symbol-solver-model` is treated specially and is the *only* module that does not come with its own `.iml` file? Or has it simply been forgotten for this module?

Anyhow, I generated an `.iml` file for `javaparser-symbol-solver-model` from the POM and that solved all my problems. :-)

I'm creating a PR so that other users who want to import the project into IntelliJ IDEA in the future don't stumble upon the same problem.

### DIFF
--- a/javaparser-symbol-solver-model/.gitignore
+++ b/javaparser-symbol-solver-model/.gitignore
@@ -11,7 +11,6 @@
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 .idea
-*.iml
 target
 build
 /.classpath
--- a/javaparser-symbol-solver-model/javaparser-symbol-solver-model.iml
+++ b/javaparser-symbol-solver-model/javaparser-symbol-solver-model.iml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<module org.jetbrains.idea.maven.project.MavenProjectsManager.isMavenModule="true" type="JAVA_MODULE" version="4">
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_8">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <content url="file://$MODULE_DIR$">
+      <sourceFolder url="file://$MODULE_DIR$/src/main/java" isTestSource="false" />
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="module" module-name="javaparser-core" />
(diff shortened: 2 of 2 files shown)


ftomassetti:
Make sense. I personally would not commit iml files but it makes total sense to be consistent

+1

daanschipper:
`.iml` files should not be committed, see https://www.gitignore.io/api/intellij.

malteskoruppa:
@daanschipper That is certainly up for discussion. The JetBrains people say that `.iml` *should* be shared. See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

At any rate, things should be consistent. Either commit all the `.iml` files, or none. Here, it was the case that all `.iml` files were committed *except* one, and I think nobody wants that. :-)

daanschipper:
That link that you send specifically mentions that ".iml files for the Gradle or Maven based projects, since these files will be generated on import" can be considered to not be shared 😉 

But I do agree that it has to be consistent, it should however be created by Maven

malteskoruppa:
@daanschipper Touché! But, also, that link _you_ shared actually does _not_ ignore `.iml` files. :wink:  The entry for `*.iml` files in that link is specifically commented out with reference to this comment: https://github.com/joeblau/gitignore.io/issues/186#issuecomment-215987721
...which explains why the gitignore.io people ended up deciding _not_ to ignore `.iml` files.

matozoid:
Consider the attempt to share the intellij project files to be an experiment :-) I don't think it adds too much at the moment and if it keeps causing issues it will be out again. But for now the iml goes in for consistency :-)

malteskoruppa:
I actually like the idea of sharing the IntelliJ project files. At least, some of them. [That's what the JetBrains people say too](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems). It makes importing the project into IntelliJ a bit less painful in my opinion :-) (e.g., when IntelliJ doesn't automatically generate the `.iml` files as it happened to me for some reason). It also allows to share some configurations such as code style settings etc. to improve consistency among all committers. (Of course, only for those who use IntelliJ, but that's already better than nothing.)

But I also know from personal experience that there are *very* different views on this, and this is probably not the best place to discuss the matter. :wink: I just wanted to say: I'm happy they're there!

PR #1497, opened 2018-04-05
Resolving constructor declarations from object creation expressions

malteskoruppa:
Currently, the method `ObjectCreationExpr#invokeResolvedConstructor()` invariably throws an `UnsupportedOperationException`.

This is because the latter method calls `JavaSymbolSolver#resolveDeclaration(Node, Class)`, passing the `ObjectCreationExpr` as the first argument. This method in turn checks if the given node is an instance of one of several supported types, and throws an `UnsupportedOperationException` otherwise. `ObjectCreationExpr` is not among those supported types, thus the exception is thrown.

This PR adds the missing piece of logic to the method `JavaSymbolSolver#resolveDeclaration(Node, Class)`, such that `ObjectCreationExpr#invokeResolvedConstructor()` now works.

Additionally, I made the class `DefaultConstructorDeclaration` public. Otherwise, it is not possible for users to check if a `ResolvedConstructorDeclaration` returned by `ObjectCreationExpr#invokeResolvedConstructor()` is an instance of `DefaultConstructorDeclaration`.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
@@ -3,10 +3,7 @@ package com.github.javaparser.symbolsolver;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.*;
-import com.github.javaparser.ast.expr.Expression;
-import com.github.javaparser.ast.expr.MethodCallExpr;
-import com.github.javaparser.ast.expr.NameExpr;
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
```diff
@@ -116,6 +113,16 @@ public void inject(CompilationUnit destination) {
                 throw new UnsolvedSymbolException("We are unable to find the method declaration corresponding to " + node);
             }
         }
+        if (node instanceof ObjectCreationExpr) {
```
ftomassetti, line 116: Nice

ftomassetti (approved):
Good work, thanks!

PR #1503, opened 2018-04-07
Issue 1498 failing tests

### LINKED ISSUE #1498: Failing tests in javaparser-symbol-solver-testing
There are 20 failing tests in module `javaparser-symbol-solver-testing`. More precisely, the failing tests are all 20 tests in the class `AnalyseJavaSymbolSolver060Test`.

This does not become apparent when running `mvn clean test`, because the class `AnalyseJavaSymbolSolver060Test` belongs to the category `SlowTest` and such tests are excluded in the pom.xml of the `javaparser-symbol-solver-testing` module.

The reason why the tests fail is simple and easily fixed: This class attempts to parse & run some tests on the source code of JSS 0.6.0. The source code of JSS 0.6.0 should be included in the directory `javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0/`, but it is missing.

As you can see, in the JSS repository, there are two sub-directories `lib` and `src` here:
https://github.com/javaparser/javasymbolsolver/tree/master/java-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0

In the same location of the JP repository, these two sub-directories are missing:
https://github.com/javaparser/javaparser/tree/master/javaparser-symbol-solver-testing/src/test/resources/javasymbolsolver_0_6_0

Copying these two directories to this location in the JavaParser repository fixes the failing tests.

matozoid:
Fixes #1498


matozoid:
Aw, have to delete the maven config still.

PR #1504, opened 2018-04-07
Slow tests oddness

matozoid:


matozoid:
@ftomassetti - can you say by looking at the changes what could be wrong with the tests?

ftomassetti:
Tests should not write file. I uncommented the lines to write to initially generate the files and then if there were big changes. The idea is to verify that we keep obtaining the same results

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/AnalyseNewJavaParserTest.java
```diff
@@ -79,25 +78,24 @@ private void parse(String fileName) throws IOException {
...
-        assertTrue("No UnsupportedOperationException expected when analyzing " + path, 0 == sourceFileInfoExtractor.getUnsupported());
+        assertEquals("No failures expected when analyzing " + path, 0, sourceFileInfoExtractor.getKo());
+        assertEquals("No UnsupportedOperationException expected when analyzing " + path, 0, sourceFileInfoExtractor.getUnsupported());
 
         if (!Files.exists(dstFile)) {
             // If we need to update the file uncomment these lines
-            PrintWriter writer = new PrintWriter(dstFile.toAbsolutePath().toFile(), "UTF-8");
-            writer.print(output);
-            writer.close();
+//        try(PrintWriter writer = new PrintWriter(dstFile.toFile(), "UTF-8")) {
```
ftomassetti, line 96: Yes, that should have been commented :)

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_CommentsInserter.txt
```diff
@@ -1,49 +1,10 @@
...
+  Line 38) modifiers.contains(Modifier.PROTECTED) ==> java.util.AbstractCollection.contains(java.lang.Object)
+  Line 40) modifiers.contains(Modifier.PRIVATE) ==> java.util.AbstractCollection.contains(java.lang.Object)
+  Line 46) sortByBeginPosition(nodes, false) ==> com.github.javaparser.utils.PositionUtils.sortByBeginPosition(java.util.List<T>, boolean)
+  Line 53) EnumSet.noneOf(Modifier.class) ==> java.util.EnumSet.noneOf(java.lang.Class<E>)
+  Line 114) cloneNodes(_n.getPackage(), _arg) ==> com.github.javaparser.ast.visitor.CloneVisitor.cloneNodes(T, java.lang.Object)
+  Line 46) setCheck(check) ==> com.github.javaparser.ast.stmt.AssertStmt.setCheck(com.github.javaparser.ast.expr.Expression)
+  Line 49) ((Node) this).tryAddImportToParentCompilationUnit(typeClass) ==> com.github.javaparser.ast.Node.tryAddImportToParentCompilationUnit(java.lang.Class<?>)
+  Line 86) buf.append(indentation) ==> java.lang.StringBuilder.append(java.lang.String)
   Line 52) comments.isEmpty() ==> java.util.TreeSet.isEmpty()
-  Line 63) cu.getChildrenNodes() ==> com.github.javaparser.ast.Node.getChildrenNodes()
```
ftomassetti, line 2: The fact the generated files has less lines does not seem very good

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_ParseProblemException.txt
```diff
@@ -1,10 +1 @@
   Line 21) createMessage(assertNotNull(problems)) ==> com.github.javaparser.ParseProblemException.createMessage(java.util.List<com.github.javaparser.Problem>)
-  Line 21) assertNotNull(problems) ==> com.github.javaparser.utils.Utils.assertNotNull(T)
```
ftomassetti, line 2: Same here. Well, I suspect the other hundreds of files are the generated ones so I will not look further

ftomassetti (commented):
The code does not seem suspicious.
I just wonder if we introduced the issue some time ago and we did not realize untile now because your changes does not seem strange.

matozoid:
@ftomassetti I made it write to the files so you could see the difference here :-) it's disabled in the next commit

matozoid:
@ftomassetti there are two suspects: me moving the test code around, and the slow tests not having run for a while. If you think these results are fine, you can merge them.

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_CommentsInserter.txt
matozoid: Most of them have (a lot) more, and some dont. It might be the wrong version of JP that we're testing on? But I thought I just copied the source, nothing more.

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_ParseProblemException.txt
matozoid: What does it mean to have more or less results here? The tests take ages, and for some reason they all work fine from the IDE. It's hard to debug.

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_CommentsInserter.txt
ftomassetti: Copied the source? What do you mean?

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_ParseProblemException.txt
ftomassetti: If from the IDE the results are different could it be that it uses different paths or a different parallelism I suppose. Definitely very weird.

javaparser-symbol-solver-testing/src/test/resources/javaparser_methodcalls_expected_output/com_github_javaparser_CommentsInserter.txt
matozoid: When I moved the source from `test/resource` to `test/testsource` to stop it from confusing the IDE.

ftomassetti: Mmm... we are not trying to load them as resources, right?

matozoid: Nah, I fixed all the path handling. And we don't have a problem with not finding, but a problem with seeing something different. Worth a double check though.

matozoid:
Closed because the problem encountered here has been solved in another PR

PR #1512, opened 2018-04-12
Make ExplicitConstructorInvocationStmt implement NodeWithArguments<ExplicitConstructorInvocationStmt>

malteskoruppa:
This was previously missing.

Besides being clearly sensible in terms of the semantic meaning of the `NodeWithArguments` abstraction, this modification also allows simplifying the implementation of `ExplicitConstructorInvocationStmt` a little bit by removing some methods which already have a default implementation in the interface and adds a new method `addArgument(String arg)` to `ExplicitConstructorInvocationStmt` which it did not have before.


matozoid:
Whoops! Yes, that is absolutely correct, thanks!

malteskoruppa:
These AppVeyor builds tend to fail often in the last few PRs because of a timeout. :-/

matozoid:
Yes, something in one of the last PR's made build times sky rocket. Will make an issue.

PR #1519, opened 2018-04-22
Fix problem with slow tests

matozoid:
Some tests are not concurrent, so I disabled the multithreaded test execution for now.


ftomassetti (commented):
The first files seem reasonable to me but I did not checked all of them

matozoid:
@ftomassetti none of the expected_result files were touched. It was the concurrency that was mixing up all the results. Now it's fine again.

ftomassetti:
Good to hear that!

Issue #1522, opened 2018-04-25
Cannot resolve constructor declaration from object creation expression with scope


malteskoruppa:
Consider the following code:

```
class ConstructorCalls {

    void testNormalConstructor() {
        new OtherClass();
    }

    void testInnerClassConstructor() {
        OtherClass oc = new OtherClass();
        oc.new InnerClass();
    }
}

class OtherClass {

    OtherClass() {}

    class InnerClass {

        InnerClass() {}
    }
}
```

Calling `resolveInvokedConstructor()` on the object creation expression `new OtherClass()` works great! :smiley: 

However, calling `resolvedInvokedConstructor()` on the object creation expression `oc.new InnerClass()` throws an `UnsolvedSymbolException`. :disappointed: 

It would appear that the symbol solver cannot cope with object creation expressions that have a scope. Is that correct?

A pull request with corresponding tests is incoming.

PR #1523, opened 2018-04-25
Tests for resolving constructor declarations

malteskoruppa:
Here are two tests for resolving constructors using the symbol solver. :-)

The first one tests resolving a constructor declaration from a normal object creation expression without scope. This one works!

The second one tests resolving a constructor declaration from an object creation expression with scope. This one fails, unfortunately.

Please see #1522 for details.

Of course this PR is not (yet) meant to be merged, since the second test fails. But I am submitting this PR in the hope that it will help in solving this problem.

@ftomassetti Do you think it would be difficult to fix the symbol solver so that the second test passes? Unfortunately I don't know how... if you could add the corresponding code to this PR, it would help me understand how the symbol solver works (besides fixing a bug). I am willing to help out in the future as I can, but I do need some help to get started. :-)

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparser/Navigator.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparser/Navigator.java
@@ -121,6 +121,33 @@ public final class Navigator {
         return found;
     }
 
+    /**
+     * Returns the {@code (i+1)}'th constructor of the given type declaration, in textual order. The constructor that
+     * appears first has the index 0, the second one the index 1, and so on.
+     *
+     * @param td    The type declaration to search in. Note that only classes and enums have constructors.
+     * @param index The index of the desired constructor.
+     * @return The desired ConstructorDeclaration if it was found, and {@code null} otherwise.
+     */
+    public static ConstructorDeclaration demandConstructor(TypeDeclaration<?> td, int index) {
+        ConstructorDeclaration found = null;
+        int i = 0;
+        for (BodyDeclaration<?> bd : td.getMembers()) {
+            if (bd instanceof ConstructorDeclaration) {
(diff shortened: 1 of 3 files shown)


ftomassetti:
I had no idea you could do that in Java...

```
oc.new InnerClass()
```

Really?!

malteskoruppa:
Sure, that's the normal way to instantiate an inner class from an outside class. When you instantiate an inner class, you always need to specify its enclosing instance. Otherwise, how could an inner class instance access instance variables from its enclosing instance?

Of course, when you instantiate an inner class from the enclosing class itself, you don't need it, because then, `new InnerClass()` means the same as `this.new InnerClass()` and so the enclosing instance is already specified implicitly. But, when you want to instantiate an inner class from another class which is not the enclosing class, you need to specify the enclosing instance.

Please see [JLS 8.1.3: Inner Classes and Enclosing Instances](https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.1.3) as well as [JLS 15.9: Class Instance Creation Expressions](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.9).

Quotation from the latter section:

> _Qualified class instance creation expressions_ begin with a _Primary_ expression or an _ExpressionName_.
> 
> A qualified class instance creation expression enables the creation of instances of inner member classes and their anonymous subclasses.

ftomassetti:
Ok, this one was tricky because of the way qualified access works but I added a commit to make the second test pass. I took the commit from @malteskoruppa and added a new one creating PR #1524

Normally stuff is easier than this in the JavaSymbolSolver, you ended up finding a very complex case :)

Thank you for your work. I realize that JSS has not a great design and it is quite complex to work with.

malteskoruppa:
Awesome! :-)

I indeed seem to have found a complex case. ;-)

I think this PR can be rejected now, as it's gotten obsolete with #1524.

matozoid:
Thanks for giving @ftomassetti a good start!

PR #1524, opened 2018-04-26
Malteskoruppa constructor resolution

### LINKED ISSUE #1522: Cannot resolve constructor declaration from object creation expression with scope

ftomassetti:
Fixes #1522

### DIFF
--- a/javaparser-core-testing/javaparser-core-testing.iml
+++ b/javaparser-core-testing/javaparser-core-testing.iml
@@ -10,6 +10,11 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.jbehave:jbehave-core:4.3" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: commons-io:commons-io:2.6" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.apache.commons:commons-lang3:3.7" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.apache.commons:commons-text:1.2" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: com.github.valfirst:jbehave-junit-runner:2.1.3" level="project" />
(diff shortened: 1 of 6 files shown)


malteskoruppa:
Wow, that was fast! :-)

Ok, I think I understand what you did there. You essentially created an `ObjectCreationContext` which is able to resolve qualified object creation expressions by resolving the type of the scope, checking that this resolved type is a reference type, and then searching for the inner class within that reference type. It gives me an idea how resolving expressions works and what contexts are good for. Thank you very much!

One of the main problems with starting to understand the symbol solver is that the source code does not have a lot of documentation, unfortunately, so looking at individual commits where you know what problem they solve helps.

Side note: It looks like you updated the `javaparser-core-testing.iml` with some duplicate entries. It probably doesn't hurt but it's redundant. Not sure why your IDE did that. If you're using IntelliJ, you can probably cause IntelliJ to remove the duplicate entries again by right-clicking on `javaparser [javaparser-parent]` folder at the very top of the project window, then selecting _Maven -> Reimport_.

matozoid:
I'll merge when the intellij config is fixed (or until release time - then I'll fix it myself ;-) )

I'm on holiday for the weekend by the way, so release will be next week.

malteskoruppa:
> I'm on holiday for the weekend by the way, so release will be next week.

Have a good time! :-)

matozoid:
Thanks @malteskoruppa !

matozoid:
Well, there's still time for a weekend release.

PR #1530, opened 2018-05-06
Fix solving interfaces extending interfaces

### LINKED ISSUE #1528: Getting all extended interfaces does not work
The following Groovy script shows the problem:

```groovy
@Grab('com.github.javaparser:javaparser-symbol-solver-core:3.6.3')
import com.github.javaparser.symbolsolver.JavaSymbolSolver
import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade
import com.github.javaparser.symbolsolver.model.resolution.TypeSolver
import com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver
import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver
import com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver

def typeSolver = new ReflectionTypeSolver()
JavaParser.staticConfiguration.symbolResolver = new JavaSymbolSolver(typeSolver)
def javaParserFacade = JavaParserFacade.get typeSolver
def compilationUnit = JavaParser.parse 'public interface Foo extends Comparable { }'
println javaParserFacade.getTypeDeclaration(compilationUnit.types[0]).allInterfacesExtended
println compilationUnit.types[0].implementedTypes
println compilationUnit.types[0].extendedTypes
```

I think the problem is that in https://github.com/javaparser/javaparser/blob/9a169f50ad23330d19d44a1553710fcdf7e4aad8/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java#L124-L125 `implementedTypes` is used instead of `extendedTypes`, as of course an interface does not implement, but extend other interfaces.

matozoid:
Fixes #1528 

Some slow tests are failing, something seems wrong on master.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java
@@ -121,10 +121,8 @@ public class JavaParserInterfaceDeclaration extends AbstractTypeDeclaration impl
     @Override
     public List<ResolvedReferenceType> getInterfacesExtended() {
         List<ResolvedReferenceType> interfaces = new ArrayList<>();
-        if (wrappedNode.getImplementedTypes() != null) {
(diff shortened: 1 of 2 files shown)


ftomassetti (approved):
+1

PR #1537, opened 2018-05-08
Include curly braces in inline tags

ftomassetti:
Include curly braces in inline tags


matozoid:
Seems better than before. +1

PR #1540, opened 2018-05-08
Fix typo

### LINKED ISSUE #1538: Typo in JavaDoc
http://static.javadoc.io/com.github.javaparser/javaparser-core/3.6.3/com/github/javaparser/ast/nodeTypes/NodeWithStatements.html#addStatement-java.lang.String-

`It will use JavaParser.parseStatement(String) inside, so it should end with a semi column`

should probably be

`It will use JavaParser.parseStatement(String) inside, so it should end with a semicolon`

:-)

matozoid:
Fixes #1538


ftomassetti:
+1 even if it is surprisingly failing...

matozoid:
It was a problem with Travis. That's what you get when you use free stuff! :-)

PR #1545, opened 2018-05-09
prettyAlignMethodCallChainsIndentsArgumentsWithBlocksCorrectly

### LINKED ISSUE #1543: Wrong indent in code blocks within broken down method call chains
The following Groovy code

```groovy
@Grab('com.github.javaparser:javaparser-symbol-solver-core:3.6.4')
import com.github.javaparser.printer.PrettyPrinter
import com.github.javaparser.printer.PrettyPrinterConfiguration

println new PrettyPrinter(new PrettyPrinterConfiguration().setColumnAlignFirstMethodChain(true)).print(JavaParser.parse('class Foo { void bar() { foo().bar().baz(() -> { boo().baa().bee(); }).bam(); } }'))
```

produces

```java
class Foo {

    void bar() {
        foo().bar()
             .baz(() -> {
            boo().baa()
                 .bee();
        })
             .bam();
    }
}
```

but I'd expect it to produce

```java
class Foo {

    void bar() {
        foo().bar()
             .baz(() -> {
                 boo().baa()
                      .bee();
             })
             .bam();
    }
}
```

So block indent = method call chain indent + configuration.indent instead of current indent + configuration.indent.
Basically the indent of the line where the block started + the configured indent.

matozoid:
Fixes #1543 

@Vampire can you review this?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -1447,7 +1447,7 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
             //noinspection unchecked
             NodeList<ImportDeclaration> modifiableList = new NodeList<>(n);
             modifiableList.sort(
-                    comparingInt((ImportDeclaration left) -> left.isStatic() ? 0 : 1)
+                    comparingInt((ImportDeclaration i) -> i.isStatic() ? 0 : 1)
                             .thenComparing(NodeWithName::getNameAsString));
             for (Object node : modifiableList) {
                 ((Node) node).accept(this, arg);
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -21,7 +21,6 @@
 
 package com.github.javaparser.printer;
 
-import com.github.javaparser.Position;
 import com.github.javaparser.ast.*;
 import com.github.javaparser.ast.body.*;
 import com.github.javaparser.ast.comments.BlockComment;
@@ -30,6 +29,7 @@ import com.github.javaparser.ast.comments.JavadocComment;
 import com.github.javaparser.ast.comments.LineComment;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.modules.*;
+import com.github.javaparser.ast.nodeTypes.NodeWithName;
 import com.github.javaparser.ast.nodeTypes.NodeWithTypeArguments;
 import com.github.javaparser.ast.nodeTypes.NodeWithVariables;
 import com.github.javaparser.ast.stmt.*;
@@ -42,9 +42,8 @@ import java.util.stream.Collectors;
 
 import static com.github.javaparser.ast.Node.Parsedness.UNPARSABLE;
 import static com.github.javaparser.utils.PositionUtils.sortByBeginPosition;
-import static com.github.javaparser.utils.Utils.isNullOrEmpty;
-import static com.github.javaparser.utils.Utils.normalizeEolInTextBlock;
-import static com.github.javaparser.utils.Utils.trimTrailingSpaces;
+import static com.github.javaparser.utils.Utils.*;
+import static java.util.Comparator.comparingInt;
 
 /**
  * Outputs the AST as formatted Java source code.
@@ -54,33 +53,14 @@ import static com.github.javaparser.utils.Utils.trimTrailingSpaces;
 public class PrettyPrintVisitor implements VoidVisitor<Void> {
     protected final PrettyPrinterConfiguration configuration;
     protected final SourcePrinter printer;
-    private Deque<Position> methodChainPositions = new LinkedList<>();
 
     public PrettyPrintVisitor(PrettyPrinterConfiguration prettyPrinterConfiguration) {
         configuration = prettyPrinterConfiguration;
         printer = new SourcePrinter(configuration.getIndent(), configuration.getEndOfLineCharacter());
-        pushMethodChainPosition(printer.getCursor()); // initialize a default position for methodChainPositions, it is expected by method #resetMethodChainPosition()
     }
 
     public String getSource() {
         return printer.getSource();
-    }
-
-    public void resetMethodChainPosition(Position position) {
-        this.methodChainPositions.pop();
-        this.methodChainPositions.push(position);
-    }
-
-    public void pushMethodChainPosition(Position position) {
-        this.methodChainPositions.push(position);
-    }
-
-    public Position peekMethodChainPosition() {
-        return this.methodChainPositions.peek();
-    }
-
-    public Position popMethodChainPosition() {
-        return this.methodChainPositions.pop();
     }
 
     private void printModifiers(final EnumSet<Modifier> modifiers) {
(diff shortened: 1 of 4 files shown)


Vampire:
The actual code, or do you mean trying it out and verifying the result?

matozoid:
The code. Or don't if you think I always produce perfect code :-)

Vampire:
Surely not, noone does :-D
I'm just not sure I'm suitable for that as a foreigner to the codebase, but I can surely have a look later.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1460,13 +1446,9 @@ public void visit(NodeList n, Void arg) {
...
             NodeList<ImportDeclaration> modifiableList = new NodeList<>(n);
-            modifiableList.sort((left, right) -> {
-                int sort = Integer.compare(left.isStatic() ? 0 : 1, right.isStatic() ? 0 : 1);
-                if (sort == 0) {
-                    sort = left.getNameAsString().compareTo(right.getNameAsString());
-                }
-                return sort;
-            });
+            modifiableList.sort(
+                    comparingInt((ImportDeclaration left) -> left.isStatic() ? 0 : 1)
```
Vampire: shouldn't be called `left` anymore

javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
```diff
@@ -24,51 +24,65 @@
...
-            cursor = Position.pos(cursor.line, cursor.column + indentationLength);
-        }
+        append(indents.peek());
     }
 
     /**
      * Append the source string passed as argument to the buffer.
-     * If this is being appended at the beginning of a line, performs indentation first.
-     * <p> 
+     * If this is being appended at the beginning of a line, performs indentSize first.
```
Vampire: Too many replacements I'd say

javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
```diff
@@ -82,12 +96,12 @@ public SourcePrinter print(final String arg) {
 
     /**
      * Append the source string passed as argument to the buffer, then append a newline.
-     * If this is being appended at the beginning of a line, performs indentation first.
-     * <p>  
+     * If this is being appended at the beginning of a line, performs indentSize first.
+     * <p>
```
Vampire: Here also too many replacements I'd say

javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
```diff
@@ -24,51 +24,65 @@
...
         return this;
     }
 
     public SourcePrinter unindent() {
-        level--;
+        indents.pop();
         return this;
     }
 
+    private void append(String arg) {
```
Vampire: Should also be used in `print()` and `println()`

Vampire (changes requested):
lgtm besides the cosmetics I commented
And of course that this is a breaking change, but I guess this is ok for you?
I don't know your policy on this.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -37,19 +37,23 @@
...
+        return indentString.toString();
     }
 
     /**
-     * Set the string to use for indenting. For example: "\t", "    ", "".
+     * Set the size of the indent in spaces.
      */
-    public PrettyPrinterConfiguration setIndent(String indent) {
+    public PrettyPrinterConfiguration setIndent(int indent) {
         this.indent = assertNotNull(indent);
```
Vampire: Additionally, `assertNotNull` is not necessary anymore it just uselessly tests that the auto-boxing works

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -37,19 +37,23 @@
...
+    private int indent = 4;
     private String endOfLineCharacter = EOL;
     private Function<PrettyPrinterConfiguration, PrettyPrintVisitor> visitorFactory = PrettyPrintVisitor::new;
     private int maxEnumConstantsToAlignHorizontally = DEFAULT_MAX_ENUM_CONSTANTS_TO_ALIGN_HORIZONTALLY;
 
     public String getIndent() {
-        return indent;
+        StringBuilder indentString = new StringBuilder();
+        for(int i=0; i<indent; i++){
+            indentString.append(" ");
```
Vampire: `'  '` should be enough, no need for a `String` object

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
Vampire: Or rather it should `assertNotNegative`

Vampire (changes requested):
~~And there is probably a logic error if you combine method and call-chain aligning.
With `class Foo { void bar() { foo().bar().baz(() -> { boo().baa().bee(a, b, c); }).bam(); } }` and both options enabled you get~~
```java
...
```
~~instead of the expected~~
```java
...
```
Scratch that, wrong expectation.

javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
matozoid: Yes, except println because that does funny cursor logic.

matozoid:
I wanted to write something about API breaking for a while, so here: http://javaparser.org/pragmatic-versioning/ :-)

javaparser-core/src/main/java/com/github/javaparser/printer/SourcePrinter.java
Vampire: Ah, right, didn't look closely enough, then only for `print`

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
```diff
@@ -31,26 +33,49 @@
...
     private Function<PrettyPrinterConfiguration, PrettyPrintVisitor> visitorFactory = PrettyPrintVisitor::new;
     private int maxEnumConstantsToAlignHorizontally = DEFAULT_MAX_ENUM_CONSTANTS_TO_ALIGN_HORIZONTALLY;
 
+    /**
+     * Set the string to use for indenting. For example: "\t", "    ", "".
+     * @deprecated use setIndentSize
+     */
+    @Deprecated
+    public PrettyPrinterConfiguration setIndent(String indent) {
+        throw new NotImplementedException();
```
Vampire: How about instead the following?
- warn about deprecation if you do deprecation warnings
- if argument is spaces only, set `indentSize` to the length
- if argument contains anything else, throw exception (probably most users have couple of spaces or tabs, latter will maybe be added by my pending PR if you accept it, everything else is probably OK if it is not supported anymore)

Would make exisiting code simply continue to work like before besides a deprecation warning.

Vampire (changes requested):

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
matozoid: Well, because we have little manpower and 194 issues open :-)

matozoid: Anyway, I assumed you were going to build a PR to make this stuff work again - but maybe I misunderstood. (Could be, happens more than I would like)

Vampire: I'll get tabs working properly with my PR, yes, as soon as you merged this, I'll probably adapt to the latest changes and submit it for your review. :-)

Vampire (approved):
lgtm 👍

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinterConfiguration.java
matozoid: Here we go!

PR #1546, opened 2018-05-09
Fix Javadoc formatting

### LINKED ISSUE #1544: Please ensure that there is at least one space at the start of a Javadoc line
If you run the Groovy code
```groovy
@Grab('com.github.javaparser:javaparser-symbol-solver-core:3.6.4')
import com.github.javaparser.printer.PrettyPrinter

def clazz = JavaParser.parse('class Foo { }')
clazz.types[0].setJavadocComment('''
This is

a test!
''')
println new PrettyPrinter().print(clazz)
```
You get as result
```java
/**
 *This is
 *
 *a test!
 */
class Foo {
}
```
Which is not too pretty.
Please ensure that there is at least one space whitespace between the leading asterisk and the Javadoc line, but not if there is no text following the asterisk, so the the desired output would be
```java
/**
 * This is
 *
 * a test!
 */
class Foo {
}
```
without a trailing space at the middle line

matozoid:
Fixes #1544 

Look @ftomassetti, this removes that odd " "+ in the Javadoc setter that you didn't like :-)

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/visitor/CloneVisitorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/visitor/CloneVisitorTest.java
@@ -87,12 +87,12 @@ public class CloneVisitorTest {
         while (typeItr.hasNext()) {
             typeDeclaration = typeItr.next();
             if (typeDeclaration.getMembers() == null) {
-                assertEquals(typeDeclaration.getComment().get().getContent(), " javadoc");
+                assertEquals("javadoc", typeDeclaration.getComment().get().getContent());
             } else {
                 Iterator<BodyDeclaration<?>> bodyItr = typeDeclaration.getMembers().iterator();
                 while (bodyItr.hasNext()) {
                     BodyDeclaration<?> bodyDeclaration = bodyItr.next();
-                    assertEquals(bodyDeclaration.getComment().get().getContent(), " javadoc");
+                    assertEquals("javadoc", bodyDeclaration.getComment().get().getContent());
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


javaparser-core-testing/src/test/java/com/github/javaparser/ast/visitor/CloneVisitorTest.java
```diff
@@ -87,12 +87,12 @@ public void cloneJavaDocTest() {
         while (typeItr.hasNext()) {
             typeDeclaration = typeItr.next();
             if (typeDeclaration.getMembers() == null) {
-                assertEquals(typeDeclaration.getComment().get().getContent(), " javadoc");
+                assertEquals("javadoc", typeDeclaration.getComment().get().getContent());
```
ftomassetti, line 90: right...

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/AnnotationMemberDeclarationTransformationsTest.java
```diff
@@ -154,22 +154,22 @@ public void replacingAnnotation() {
     public void addingJavadoc() {
         AnnotationMemberDeclaration it = consider("int foo();");
         it.setJavadocComment("Cool this annotation!");
-        assertTransformedToString("@interface AD { /** Cool this annotation!*/" + EOL +
+        assertTransformedToString("@interface AD { /**Cool this annotation!*/" + EOL +
```
ftomassetti, line 157: Right, we want to permit to have comments without extra spaces

ftomassetti (approved):
+1

Issue #1461, opened 2018-03-09
Wrong indentation when adding a new annotation


nguyentruongtho:
I am using java-parser for adding a new annotation:
```
unitTestCase.addSingleMemberAnnotation(
                    IGNORE_ANNOTATION,
                    new StringLiteralExpr("flaky test"))
```
and I also use `LexicalPreservingPrinter` to preserve the code that doesn't change.

This ended up giving me a wrong format code as following:

before:
``` 
  @Test
  public void testCase() {
...
```

after:
``` 
  @Test
 @org.junit.Ignore("flaky test")
 public void testCase() {
...
```

expect:
``` 
  @Test
  @org.junit.Ignore("flaky test")
  public void testCase() {
...
```

Javaparser version: `3.5.12`

matozoid:
Interesting!

matozoid:
I've put together the test case: https://github.com/matozoid/javaparser/commit/82a96e2c5db9571911966113d31b2a256964418f

ThLeu:
As I have also noticed this particular issue and I wanted to have a closer look at the lexical preservation printing anyway I invested some time in this issue.
The following I have found out so far.

Debugging with your test @matozoid it seems to have to do with the "reshuffling" in the difference calculation when the CsmMix element is used. Because the issue seems to be gone when you remove the mix wrapper method from the MethodDeclaration in ConcreteSyntaxModel
```
       concreteSyntaxModelByClass.put(MethodDeclaration.class, sequence(
                orphanCommentsBeforeThis(),
                comment(),
                mix(memberAnnotations(), modifiers()),
                typeParameters(),
                ...
        ));
```

```
       concreteSyntaxModelByClass.put(MethodDeclaration.class, sequence(
                orphanCommentsBeforeThis(),
                comment(),
                memberAnnotations(), 
                modifiers(),
                typeParameters(),
                ...
        ));
```
In Difference.apply() the nodeText get changed at the wrong index position
```
[...]

PR #1549, opened 2018-05-11
Fix annotation indenting in lexical preserving printer

ThLeu:
Ok I threw together a fix for #1461 which hopefully also makes the algorhitm a bit more robust for such cases. Please take a thorough look at it and tell me what you think and if you have any concerns or anything.

The main difference is that the corresponding elements that are searched in a `mix()` wrapper method when calculating the difference are prioritized by how good the match actually is. Priorization is done by checking if the previous and / or the next element are still the same (which would count as the best match) and so on up to the "worst" match, which is if just the elements themselves match (without their neighbours). That is also how it previously worked. 

@matozoid I have used your testcase for this, I hope you don't mind :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ out
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# Project-specific stuff:
+.idea/*
+
 # User-specific stuff:
 .idea/**/workspace.xml
 .idea/**/tasks.xml
--- a/.gitignore
+++ b/.gitignore
@@ -23,9 +23,6 @@ out
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
-# Project-specific stuff:
-.idea/*
-
 # User-specific stuff:
 .idea/**/workspace.xml
 .idea/**/tasks.xml
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
@@ -25,13 +25,17 @@ import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.Parameter;
 import com.github.javaparser.ast.expr.SimpleName;
+import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.ast.type.ArrayType;
 import com.github.javaparser.ast.type.PrimitiveType;
 import com.github.javaparser.printer.lexicalpreservation.AbstractLexicalPreservingTest;
+import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;
 import org.junit.Test;
 
-import java.io.IOException;
 import java.util.EnumSet;
+
+import static com.github.javaparser.utils.TestUtils.assertEqualsNoEol;
+import static com.github.javaparser.utils.Utils.EOL;
 
 /**
  * Transforming MethodDeclaration and verifying the LexicalPreservation works as expected.
@@ -46,7 +50,7 @@ public class MethodDeclarationTransformationsTest extends AbstractLexicalPreserv
     // Name
 
     @Test
-    public void settingName() throws IOException {
+    public void settingName() {
         MethodDeclaration it = consider("void A(){}");
         it.setName("B");
         assertTransformedToString("void B(){}", it);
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -23,6 +23,9 @@ out
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio and Webstorm
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
+# Project-specific stuff:
+.idea/*
```
ThLeu: I don't know if this would cause anyone trouble, if so I will of course remove it. I just added it because git wanted to commit the "javaparser.iml" file from the IntelliJ IDEA project.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -795,34 +793,99 @@ void apply(NodeText nodeText, Node node) {
...
+            int nextCsmElementIndex = csmElementListIterator.nextIndex();
+
+            Map<MatchClassification, Integer> potentialMatches = new EnumMap(MatchClassification.class);
+            for (int i = startIndex; i<nodeText.getElements().size(); i++){
+                if (!correspondingIndices.contains(i)) {
+                    TextElement textElement = nodeText.getTextElement(i);
+
+                    boolean isCorresponding = isCorrespondingElement(textElement, csmElement, node);
+
+                    if (isCorresponding) {
```
ThLeu, line 811: Everything within this if statement accounts for the main change.
1. checking of the the previous and next element
2. keeping the potential matches

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -795,34 +793,99 @@ void apply(NodeText nodeText, Node node) {
...
+                            potentialMatches.putIfAbsent(MatchClassification.SAME_ONLY, i);
                         }
                     }
-                } else {
-                    throw new UnsupportedOperationException();
                 }
             }
+
+            // Prioritize the matches from best to worst
+            Optional<MatchClassification> bestMatchKey = potentialMatches.keySet().stream()
```
ThLeu, line 840: sorting the potential matches according to their priorization and keeping the best match.

.gitignore
matozoid: Yeah, we have the project files checked in, so the best approach is to check why intellij modified the iml file, and revert it if makes no sense.

ThLeu: Probably had to do with how I imported the project, did it another way now and it seems to be fine now. I reverted the commit.

matozoid: Thanks!

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
```diff
@@ -119,4 +123,26 @@ public void replacingOnlyParameter() throws IOException {
...
+                        "}" + EOL
+        );
+        cu.getType(0).getMethods().get(0).addSingleMemberAnnotation(
+                "org.junit.Ignore",
+                new StringLiteralExpr("flaky test"));
+
+        String result = LexicalPreservingPrinter.print(cu.findCompilationUnit().get());
+        assertEqualsNoEol("class X {\n" +
+                "  @Test\n" +
+                "  @org.junit.Ignore(\"flaky test\")\n" +
```
ftomassetti, line 199: Given how complex the code is tests are vital. Thank you for adding it

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -795,34 +793,97 @@ void apply(NodeText nodeText, Node node) {
...
-            if (!usedIndexes.contains(i)) {
-                TextElement textElement = nodeText.getTextElement(i);
-                if (csmElement instanceof CsmToken) {
-                    CsmToken csmToken = (CsmToken)csmElement;
-                    if (textElement instanceof TokenTextElement) {
-                        TokenTextElement tokenTextElement = (TokenTextElement)textElement;
-                        if (tokenTextElement.getTokenKind() == csmToken.getTokenType() && tokenTextElement.getText().equals(csmToken.getContent(node))) {
-                            usedIndexes.add(i);
-                            return i;
+    private List<Integer> findIndexOfCorrespondingNodeTextElement(List<CsmElement> elements, NodeText nodeText, int startIndex, Node node) {
```
ftomassetti, line 796: I really wonder how I ended up to write such complex code without comments and how someone managed to figure it out and improve it. Kudos!

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -795,34 +793,97 @@ void apply(NodeText nodeText, Node node) {
...
+
+            if (bestMatchKey.isPresent()) {
+                correspondingIndices.add(potentialMatches.get(bestMatchKey.get()));
+            }
+        }
+
+        return correspondingIndices;
+    }
+
+    private enum MatchClassification {
```
ftomassetti, line 854: Very nice addition

ftomassetti (approved):
The only improvement I can suggest is more tests because we will eventually need to refactor the code and the best investment we can make is in tests that will survive any refactoring.

I sincerely appreciate your work on this part of the codebase, which is very tricky, as you probably figured it out by now :)

ThLeu:
Yes it is quite a tricky part, but very interesting, so I liked the challenge :)
Kudos to you for implementing the lexical preservation printing in the first place :) and for everyone's work on the javaparser.

I totally agree about more tests, I will add some more.
I might also do some minor refactorings if you don't mind, just so some parts of that code will become a little easier to understand. As you mentioned it is quite complex without comments but we all have been there :)

ThLeu:
Unfortunately this does not also solve cases for adding modifiers to a method declaration with existing annotations (it acts the same as before the change). The spaces will be added at awkward positions, I will open another issue for that with the corresponding test cases.

ThLeu:
Ok the whole adding / removing / replacing modifier and removing annotations seems to not be fixed yet unfortunately, but I will open new issues for these cases.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -843,14 +843,16 @@ void apply(NodeText nodeText, Node node) {
 
             if (bestMatchKey.isPresent()) {
                 correspondingIndices.add(potentialMatches.get(bestMatchKey.get()));
+            } else {
+                correspondingIndices.add(-1);
```
ThLeu: With this necessary addition, some unit tests for modifiers work now aswell. This is necessary so that the code depending on this List can work in the same way as before.

ThLeu:
Added some more tests and a minor fix. Now some modifications on modifiers work aswell.
What still does not work as expected is:
1. Adding modifiers to methods with annotations
2. Removing annotations

I will open new issues for these cases.

If you are ok with it I will finish here with the changes on this PR.

matozoid:
Thanks @ThLeu! We really appreciate the work on this part of the library.

Issue #1559, opened 2018-05-12
Module descriptor generation is not very fluently


Vampire:
Most generation API is pretty fluently and able to use with Strings.
But to generate a simple module descriptor, you have to use

```java
ModuleDeclaration module = new ModuleDeclaration()
        .addSingleMemberAnnotation(Generated.class, "generator")
        .addSingleMemberAnnotation(SuppressWarnings.class, "\"module\"")
        .setName("org.javacord.api");

ModuleExportsStmt internalExports = new ModuleExportsStmt().setName("org.javacord.api.entity.channel.internal");
internalExports.getModuleNames().add(new Name("org.javacord.core"));

module
        .getModuleStmts()
        .addLast(new ModuleRequiresStmt().addModifier(Modifier.TRANSITIVE).setName("java.desktop"))
        .addLast(new ModuleExportsStmt().setName("org.javacord.api.entity.channel"))
        .addLast(internalExports)
        .addLast(new ModuleUsesStmt().setType("org.javacord.api.util.internal.DelegateFactoryDelegate"));

new CompilationUnit()
        .setModule(module);
```

if I didn't do anything stupid.
Would be nice if we could instead do something like

```java
new CompilationUnit()
        .addModule("org.javacord.api")
        .addSingleMemberAnnotation(Generated.class, "generator")
        .addSingleMemberAnnotation(SuppressWarnings.class, "\"module\"")
        .addStatement("requires transitive java.desktop;")
        .addStatement("exports org.javacord.api.entity.channel;")
        .addStatement("exports org.javacord.api.entity.channel.internal to org.javacord.core;")
        .addStatement("uses org.javacord.api.util.internal.DelegateFactoryDelegate;");
```

PR #1561, opened 2018-05-12
Parse modules and their directives

### LINKED ISSUE #1558: Parsing a Jigsaw module statement does not work
Should `JavaParser.parseStatement("exports org.javacord.api.entity.channel.internal to org.javacord.core;");` not work?
Actually `ModuleStmt` does not implement `Statement`, so I guess no, but why?
If there is no reason, I'd suggest to add this, or how are you supposed to parse a single module statement?

matozoid:
Add parse methods and clear up some documentation.

Fixes #1558 

Any comments, @Vampire ?

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -32,6 +32,8 @@ import com.github.javaparser.ast.stmt.*;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.ast.type.IntersectionType;
 import com.github.javaparser.ast.type.Type;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
@@ -48,6 +50,16 @@ import static com.github.javaparser.utils.Utils.EOL;
 import static org.junit.Assert.*;
 
 public class JavaParserTest {
+
+    @Before
+    public void setToLatestJava() {
+        JavaParser.getStaticConfiguration().setLanguageLevel(BLEEDING_EDGE);
+    }
+
+    @After
+    public void resetJavaLevel() {
+        JavaParser.getStaticConfiguration().setLanguageLevel(CURRENT);
+    }
 
     @Test
     public void rangeOfAnnotationMemberDeclarationIsCorrect() {
@@ -249,7 +261,6 @@ public class JavaParserTest {
 
     @Test
     public void parseModuleDeclaration() {
-        JavaParser.getStaticConfiguration().setLanguageLevel(JAVA_9);
         JavaParser.parseModuleDeclaration("module X {}");
     }
 
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -39,13 +39,13 @@ import java.nio.file.Path;
 import java.util.Optional;
 
 import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
+import static com.github.javaparser.ParserConfiguration.LanguageLevel.*;
 import static com.github.javaparser.Providers.provider;
 import static com.github.javaparser.Range.range;
 import static com.github.javaparser.utils.CodeGenerationUtils.mavenModuleRoot;
 import static com.github.javaparser.utils.TestUtils.assertInstanceOf;
 import static com.github.javaparser.utils.Utils.EOL;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class JavaParserTest {
 
@@ -54,7 +54,7 @@ public class JavaParserTest {
         String code = "@interface AD { String foo(); }";
         CompilationUnit cu = JavaParser.parse(code);
         AnnotationMemberDeclaration memberDeclaration = cu.getAnnotationDeclarationByName("AD").get().getMember(0).asAnnotationMemberDeclaration();
-        assertEquals(true, memberDeclaration.getRange().isPresent());
+        assertTrue(memberDeclaration.getRange().isPresent());
         assertEquals(new Range(new Position(1, 17), new Position(1, 29)), memberDeclaration.getRange().get());
     }
 
@@ -63,7 +63,7 @@ public class JavaParserTest {
         String code = "@interface AD { String[] foo(); }";
         CompilationUnit cu = JavaParser.parse(code);
         AnnotationMemberDeclaration memberDeclaration = cu.getAnnotationDeclarationByName("AD").get().getMember(0).asAnnotationMemberDeclaration();
-        assertEquals(true, memberDeclaration.getRange().isPresent());
+        assertTrue(memberDeclaration.getRange().isPresent());
         assertEquals(new Range(new Position(1, 17), new Position(1, 31)), memberDeclaration.getRange().get());
     }
 
@@ -74,11 +74,11 @@ public class JavaParserTest {
         Optional<Range> range;
 
         range = expression.getLevels().get(0).getRange();
-        assertEquals(true, range.isPresent());
+        assertTrue(range.isPresent());
         assertEquals(new Range(new Position(1, 8), new Position(1, 12)), range.get());
 
         range = expression.getLevels().get(1).getRange();
-        assertEquals(true, range.isPresent());
+        assertTrue(range.isPresent());
         assertEquals(new Range(new Position(1, 13), new Position(1, 17)), range.get());
     }
 
@@ -89,11 +89,11 @@ public class JavaParserTest {
         Optional<Range> range;
 
         range = expression.getLevels().get(0).getRange();
-        assertEquals(true, range.isPresent());
+        assertTrue(range.isPresent());
         assertEquals(new Range(new Position(1, 8), new Position(1, 9)), range.get());
 
         range = expression.getLevels().get(1).getRange();
-        assertEquals(true, range.isPresent());
+        assertTrue(range.isPresent());
         assertEquals(new Range(new Position(1, 10), new Position(1, 11)), range.get());
     }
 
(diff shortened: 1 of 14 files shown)


matozoid:
Also fixes #1559

javaparser-core-testing/src/test/java/com/github/javaparser/modules/ModuleDeclarationTest.java
```diff
@@ -178,4 +181,24 @@ public void testCsmPrinting() {
...
+                .addDirective("uses org.javacord.api.util.internal.DelegateFactoryDelegate;");
+
+        assertEqualsNoEol("@Generated(generator) @SuppressWarnings(\"module\") \n" +
+                "module org.javacord.api {\n" +
+                "    requires transitive java.desktop;\n" +
+                "    exports org.javacord.api.entity.channel;\n" +
+                "    exports org.javacord.api.entity.channel.internal to org.javacord.core;\n" +
+                "    uses org.javacord.api.util.internal.DelegateFactoryDelegate;\n" +
+                "}\n", moduleDeclaration.toString());
+    }
```
Vampire: Would you mind replacing by dummy names?
I'm not sure, it being a good practice to contain other projects code in your tests.
And remove the `Generated`, unfortunately the compiler complains that Generated (neither the old, nor the new one) target modules.

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleDeclaration.java
```diff
@@ -197,4 +202,12 @@ public boolean replace(Node node, Node replacementNode) {
         }
         return super.replace(node, replacementNode);
     }
+
+    /**
+     * Add a directive to the module, like "
+     */
```
Vampire: like what?

Vampire (changes requested):
Maybe also add `ModuleExportsStmt ModuleExportsStmt#addModuleName(String)`, and `ModuleDeclaration#addDirective(ModuleStmt)` for
```java
moduleDirectives
    .addLast(new ModuleExportsStmt()
        .setName("foo.bar")
        .addModuleName("other.foo")
        .addModuleName("other.bar")
    )
```
and
```java
moduleDeclaration
    .addDirective(new ModuleExportsStmt()
        .setName("foo.bar")
        .addModuleName("other.foo")
        .addModuleName("other.bar")
    )
```

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
```diff
@@ -226,24 +226,35 @@ public void everyTokenHasACategory() throws IOException {
...
         assertEquals("j", forStmt.getInitialization().get(0).asVariableDeclarationExpr().getVariables().get(1).getNameAsString());
-        assertEquals(false, forStmt.getInitialization().get(0).asVariableDeclarationExpr().getVariables().get(0).getInitializer().isPresent());
-        assertEquals(true, forStmt.getInitialization().get(0).asVariableDeclarationExpr().getVariables().get(1).getInitializer().isPresent());
+        assertFalse(forStmt.getInitialization().get(0).asVariableDeclarationExpr().getVariables().get(0).getInitializer().isPresent());
+        assertTrue(forStmt.getInitialization().get(0).asVariableDeclarationExpr().getVariables().get(1).getInitializer().isPresent());
+    }
+
+    @Test
+    public void parseModuleDeclaration() {
+        JavaParser.getStaticConfiguration().setLanguageLevel(JAVA_9);
```
Vampire: Is it good to change the static configuration in a test?

javaparser-core-testing/src/test/java/com/github/javaparser/modules/ModuleDeclarationTest.java
matozoid: Oh sure!

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
matozoid: Hey, this review stuff really works :-)

matozoid:
Okay, comments processed!

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
```diff
@@ -32,29 +32,41 @@
...
 import static com.github.javaparser.utils.TestUtils.assertInstanceOf;
 import static com.github.javaparser.utils.Utils.EOL;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.*;
 
 public class JavaParserTest {
 
+    @Before
+    public void setToLatestJava() {
```
Vampire, line 55: Actually I did the same in the pretty printer test in the PR you just merged. Can you please change it there too?
It thought there might be some way to create a custom JavaParser with custom configuration, so that you can have differently configured parsers (is there not?) and was too lazy to find out how as it was not instantly clear, so I thought I wait until you solve my comment here. :-D

Vampire: Also, if it is only needed for one of the tests in the class, it might be better to do it at start and end of the test method, not in `@Before` and `@After`, as that will do it for all methods in the test class.

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleExportsStmt.java
```diff
@@ -161,4 +164,9 @@ public void ifModuleExportsStmt(Consumer<ModuleExportsStmt> action) {
     public Optional<ModuleExportsStmt> toModuleExportsStmt() {
         return Optional.of(this);
     }
+
+    public ModuleExportsStmt addModuleName(String name) {
```
Vampire, line 168: Was he naughty that he doesn't get some docs? :-)

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleDeclaration.java
```diff
@@ -197,4 +202,16 @@ public boolean replace(Node node, Node replacementNode) {
...
     }
+
+    /**
+     * Add a directive to the module, like "exports R.S to T1.U1, T2.U2;"
+     */
+    public ModuleDeclaration addDirective(String directive) {
+        return addDirective(JavaParser.parseModuleDirective(directive));
+    }
+
+    public ModuleDeclaration addDirective(ModuleStmt directive) {
```
Vampire, line 213: also naughty?

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
matozoid: Oh, yes you can simply do new JavaParser() and get an instance (see http://javaparser.org/the-quick-and-the-full-api-of-javaparser/) but the static methods don't know about that instance :-/

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleExportsStmt.java
matozoid: I can't think of a doc that would add anything, so I let it speak for itself :-)

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleDeclaration.java
matozoid: Indeed!

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
matozoid: Looks like your PP tests are fine - am I missing something?

Vampire: I also did the static configuration thing witout resetting.
Look at #1565 what I mean, I fixed it for my test.
I'd always perfer local configuration for tests instead of static configuration or the tests influence each other probably.

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleExportsStmt.java
Vampire: Ok, but while we are at self-explaining, I personally don't find the `moduleName` name pretty self-explanatory per-se. What is the module name(s) of an export statement? I know it now as I just tried it in absence of any other possibility to set the `to` part, but it was more trial & error than a self-explanatory name. ;-)
That mandates for a rename though, not for added JavaDoc. :-D

javaparser-core-testing/src/test/java/com/github/javaparser/modules/ModuleDeclarationTest.java
```diff
@@ -180,4 +181,39 @@ public void testCsmPrinting() {
...
+                );
+
+        moduleDeclaration
+                .addDirective(new ModuleExportsStmt()
+                        .setName("foo.bar.x")
+                        .addModuleName("other.foo")
+                        .addModuleName("other.bar")
+                );
+
+        assertEqualsNoEol("@SuppressWarnings(\"module\") \n" +
```
Vampire: With my annotation prettifying PR in, you need to remove the space before the newline.

Vampire (changes requested):

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleDeclaration.java
Vampire: Ok, just thought because its sibling got docs, but that is probably because it is a string that has to be properly formatted. :-)

javaparser-core/src/main/java/com/github/javaparser/ast/modules/ModuleExportsStmt.java
Vampire: Ok, just thought because `ModuleDeclaration#addDirective(String)'  got docs, but that is probably because it is a string that has to be properly formatted. :-)

matozoid: Yes, here the type gives it away.

PR #1563, opened 2018-05-13
Pretty Annotations And Receiver Parameter

Vampire:
- Only print a comma after the receiver parameter if there will follow other parameters
- Print annotations at pretty places and remove code duplication


ftomassetti:
+1

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -911,7 +907,11 @@ public void visit(final ExplicitConstructorInvocationStmt n, final Void arg) {
     @Override
     public void visit(final VariableDeclarationExpr n, final Void arg) {
         printComment(n.getComment(), arg);
-        printAnnotations(n.getAnnotations(), false, arg);
+        if (n.getParentNode().map(ExpressionStmt.class::isInstance).orElse(false)) {
```
matozoid, line 910: Nice!

matozoid:
I shouldn't forget to put this in the release docs.

PR #1575, opened 2018-05-14
Add more utility stuff

### LINKED ISSUE #1569: Parsing addTypeParameter method
A `NodeWithTypeParameters.addTypeParameter(String)` method would be nice, so that you can do

```java
nodeWithTypeParameters.addTypeParameter('T extends MessageAttachableListener & ObjectAttachableListener')
```

instead of

```java
def messageAttachableListener = ...
def objectAttachableListener = ...
nodeWithTypeParameters.addTypeParameter(new TypeParameter('T', NodeList.nodeList(messageAttachableListener, objectAttachableListener)))
```

### LINKED ISSUE #1572: CompilationUnit#addModule does not actually set the module
The new `com.github.javaparser.ast.CompilationUnit#addModule` does create a new module and returns it, but it does not actually set it to the CU, so it never gets printed.

matozoid:
Fixes #1572 
Fixes #1569


Vampire (approved):
lgtm

PR #1582, opened 2018-05-16
More alignment corrections

### LINKED ISSUE #1570: Indentation of code block in wrapped method call chains still wrong
The indentation of code blocks in method call chains is still wrong.
Now it is indented too far.
If you change `com.github.javaparser.printer.PrettyPrinterTest#prettyAlignMethodCallChainsIndentsArgumentsWithBlocksCorrectly` to

```java
        CompilationUnit cu = JavaParser.parse("class Foo { void bar() { foo().bar().baz(() -> { if (Object.class.isAssignableFrom(this)) { boo().baa().bee(); } boo().baa().bee(); }).bam(); } }");
```

and

```java
        assertEqualsNoEol("class Foo {\n" +
                "\n" +
                "    void bar() {\n" +
                "        foo().bar()\n" +
                "             .baz(() -> {\n" +
                "                 if (Object.class.isAssignableFrom(this)) {\n" +
                "                     boo().baa()\n" +
                "                          .bee();\n" +
                "                 }\n" +
                "                 boo().baa()\n" +
                "                      .bee();\n" +
                "             })\n" +
                "             .bam();\n" +
                "    }\n" +
                "}\n", printed);
```

You will see the failure :-(

matozoid:
Fixes #1570 
@Vampire ?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -714,7 +714,11 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
         printComment(n.getComment(), arg);
         n.getScope().ifPresent(scope -> {
             scope.accept(this, arg);
-            if (configuration.isColumnAlignFirstMethodChain()) {
+            // pick the kind of expressions where vertically aligning method calls is okay.
+            boolean mayAlignColumns = n.findParent(Statement.class)
+                    .map(p -> p.isReturnStmt() || p.isExpressionStmt())
+                    .orElse(false);
+            if (mayAlignColumns && configuration.isColumnAlignFirstMethodChain()) {
                 if (!(scope instanceof MethodCallExpr) || !((MethodCallExpr) scope).getScope().isPresent()) {
                     printer.reindentWithAlignToCursor();
                 } else {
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -714,7 +714,11 @@ public void visit(final MethodCallExpr n, final Void arg) {
         printComment(n.getComment(), arg);
         n.getScope().ifPresent(scope -> {
             scope.accept(this, arg);
-            if (configuration.isColumnAlignFirstMethodChain()) {
+            // pick the kind of expressions where vertically aligning method calls is okay.
+            boolean mayAlignColumns = n.findParent(Statement.class)
+                    .map(p -> p.isReturnStmt() || p.isExpressionStmt())
```
Vampire: How about `ThrowStmt` and `AssertStmt`?
Shouldn't those also apply?

This calculation is not really costly, is it?
Because if it is, it should be done inside the `if (configuration.isColumnAlignFirstMethodChain())` block instead of in front of it.

Vampire (changes requested):
Nice, thanks :-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: Oops, caught me cutting corners :-) I'll go through all statements tonight and see if they should be in the list.

matozoid: `AssertStmt` `LabeledStmt` `ReturnStmt` `ThrowStmt` I would guess.

Vampire: Doesn't a `LabeledStmt` always wrap another `Statement`, so will never be returned at this place? 

`AssertStmt` maybe too, yes.

Don't forget `ExpressionStmt`.

PR #1584, opened 2018-05-17
Minor Refactorings on Difference Class

ThLeu:
I did some minor refactorings on the Difference Class, mainly the apply method.
This is just the first step and I wanted to verify this with you first, before I start changing more.
What is your opinion on this kind of refactoring? Is it going into the right direction?

The main goal for me is to get a better understanding of the difference application by making it more obvious in code and eventually be able to solve the rest of the bugs with lexical preserving more easily :-).

The diff looks actually worse than it should because many things are just renaming or shifting stuff around a bit.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
@@ -485,13 +485,17 @@ public class Difference {
         }
         boolean addedIndentation = false;
         List<TokenTextElement> indentation = LexicalPreservingPrinter.findIndentation(node);
+
+        List<TextElement> givenElements = nodeText.getElements();
+        int givenIndex = 0;
+
+        List<DifferenceElement> diffElements = this.elements;
         int diffIndex = 0;
-        int nodeTextIndex = 0;
         do {
-            if (diffIndex < this.elements.size() && nodeTextIndex >= nodeText.getElements().size()) {
-                DifferenceElement diffEl = elements.get(diffIndex);
-                if (diffEl instanceof Kept) {
-                    Kept kept = (Kept) diffEl;
+            if (diffIndex < diffElements.size() && givenIndex >= givenElements.size()) {
+                DifferenceElement diffElement = diffElements.get(diffIndex);
+                if (diffElement instanceof Kept) {
+                    Kept kept = (Kept) diffElement;
                     if (kept.element instanceof CsmToken) {
                         CsmToken csmToken = (CsmToken) kept.element;
                         if (TokenTypes.isWhitespaceOrComment(csmToken.getTokenType())) {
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -85,6 +85,20 @@ public CsmElement getElement() {
         public boolean isAdded() {
             return true;
         }
+
+        public boolean isIndent() { return element instanceof CsmIndent; }
```
ftomassetti, line 89: this one and the following could be final perhaps

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -433,7 +485,7 @@ private TextElement toTextElement(CsmElement csmElement) {
     }
 
     private int considerCleaningTheLine(NodeText nodeText, int nodeTextIndex) {
-        while (nodeTextIndex >=1 && nodeText.getElements().get(nodeTextIndex - 1).isWhiteSpace() && !nodeText.getElements().get(nodeTextIndex - 1).isNewline()) {
+        while (nodeTextIndex >=1 && nodeText.getElements().get(nodeTextIndex - 1).isSpaceOrTab()) {
```
ftomassetti, line 488: cleaner, nice

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -485,78 +537,80 @@ void apply(NodeText nodeText, Node node) {
...
-                        } else {
-                            throw new IllegalStateException("Cannot keep element because we reached the end of nodetext: "
-                                    + nodeText + ". Difference: " + this);
-                        }
+            if (diffIndex < diffElements.size() && givenIndex >= givenElements.size()) {
+                DifferenceElement diffElement = diffElements.get(diffIndex);
+                if (diffElement instanceof Kept) {
+                    Kept kept = (Kept) diffElement;
+
+                    if (kept.isWhiteSpaceOrComment()) {
```
ftomassetti, line 552: Nice

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -485,78 +537,80 @@ void apply(NodeText nodeText, Node node) {
         }
         boolean addedIndentation = false;
         List<TokenTextElement> indentation = LexicalPreservingPrinter.findIndentation(node);
+
+        List<TextElement> givenElements = nodeText.getElements();
```
ftomassetti, line 541: Given is not super-clear to me as a term in this context

ftomassetti (approved):
To me it looks good. Damn, there is work to do to simplify this thing, I am happy you are doing the first steps in the right direction. Thank you!

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ThLeu: Do you have any suggestion? :)
I first called it "oldElements" but that seemed a but unclear aswell.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ThLeu: Ye I will probably go over all of these again because I have a feeling that it might be good to move some of these methods to the superclass when all is moved.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ftomassetti: perhaps "originalElements"?

ThLeu:
If you are ok with it, you can merge the PR and I will continue with another one. I just wanted to double check with you.

ftomassetti:
Works for me. @matozoid anything against me merging this one?

matozoid:
Here you go

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ThLeu: of course 🤦‍♂️ thanks

PR #1588, opened 2018-05-19
Difference class refactoring

ThLeu:
OK what did I do so far. 

- Split the Difference class into Difference and DifferenceElementCalculator
- DifferenceElementCalculator does not anymore return an instance of Difference but the elements list
- Moved DifferenceElement interface and its implementing classes to their own files
- Made the local variables from Difference.apply to fields
- Extracted some methods in Difference.apply
- Renamed given variables to original ;-)
- Added sonarlint to gitignore


ftomassetti (approved):
This was a big diff!
I understand it makes sense in this case. I appreciated the fact we have smaller classes and Difference got smaller, which is great (almost half the size it had before!).

Keep up the great work!

Hopefully this will make things more manageable for everyone else looking at this piece of code

matozoid:
I hear a +1, merging!

matozoid:
Oh oops, one thing I should have added: @ThLeu - every time you figure out how something complex work, or every time you see an undocumented class, can you add some Javadoc about it?

ThLeu:
I will try to add Javadoc where I can and some Unit Tests for the now smaller methods.
The biggest chunks are now broken apart so the next steps will probably be more in the details and therefore the understanding is even more important :-)

matozoid:
Thanks! That will help future contributors to understand the code too :-)

Issue #1593, opened 2018-05-22
Method call chain indentation is not reset on expression end


Vampire:
This test should succeed:

```java
        CompilationUnit cu = JavaParser.parse("class Foo { void bar() { foo().bar(); foo().bar().baz(() -> { boo().baa().bee(); }).bam(); } }");
        String printed = new PrettyPrinter(new PrettyPrinterConfiguration().setColumnAlignFirstMethodChain(true))
                .print(cu);

        assertEqualsNoEol("class Foo {\n" +
                "\n" +
                "    void bar() {\n" +
                "        foo().bar();\n" +
                "        foo().bar()\n" +
                "             .baz(() -> {\n" +
                "                 boo().baa()\n" +
                "                      .bee();\n" +
                "             })\n" +
                "             .bam();\n" +
                "    }\n" +
                "}\n", printed);
```
but it does not produce
```java
class Foo {

    void bar() {
        foo().bar();
        foo().bar()
             .baz(() -> {
                 boo().baa()
                      .bee();
             })
             .bam();
    }
}
```
but
```java
class Foo {

    void bar() {
        foo().bar();
             foo().bar()
                  .baz(() -> {
                      boo().baa()
                           .bee();
                  })
                  .bam();
    }
}
```
is generated

PR #1594, opened 2018-05-22
Fix indentation realignment when wrapping call-chains

Vampire:
Fix for #1593

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -30,6 +30,7 @@ import com.github.javaparser.ast.comments.LineComment;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.modules.*;
 import com.github.javaparser.ast.nodeTypes.NodeWithName;
+import com.github.javaparser.ast.nodeTypes.NodeWithTraversableScope;
 import com.github.javaparser.ast.nodeTypes.NodeWithTypeArguments;
 import com.github.javaparser.ast.nodeTypes.NodeWithVariables;
 import com.github.javaparser.ast.stmt.*;
@@ -38,6 +39,7 @@ import com.github.javaparser.ast.visitor.Visitable;
 import com.github.javaparser.ast.visitor.VoidVisitor;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import static com.github.javaparser.ast.Node.Parsedness.UNPARSABLE;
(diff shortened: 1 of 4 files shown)


matozoid:
This makes me think the whole approach to this feature is wrong, since it's spreading out over unrelated code and becoming a maintenance nightmare :-/

Vampire:
Yep, #1595 actually is the same.
alignment re-indent not reverted in-time.
It must be reverted after `Long.valueOf(foo)`, but even with this change is only reverted after the full statement is finished.

Maybe it was not the worst idea to keep indentation and alignment separate?

Or what solution do you have in mind?

Vampire:
Updated with new algorithm and no unrelated code touched, also fixes #1595 now.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -712,37 +715,100 @@ public void visit(final SuperExpr n, final Void arg) {
...
+        while (s.filter(NodeWithTraversableScope.class::isInstance).isPresent()) {
+            Optional<Expression> parentScope = s.map(NodeWithTraversableScope.class::cast)
+                    .flatMap(NodeWithTraversableScope::traverseScope);
+            if (s.filter(MethodCallExpr.class::isInstance).isPresent() && parentScope.isPresent()) {
+                methodCallWithScopeInScope.set(true);
+                break;
+            }
+            s = parentScope;
+        }
+
```
matozoid: Should we do all this stuff even when users usually do not use the alignment settings?

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -712,37 +715,100 @@ public void visit(final SuperExpr n, final Void arg) {
     @Override
     public void visit(final MethodCallExpr n, final Void arg) {
         printComment(n.getComment(), arg);
+
+        // we are at the last method call of a call chain
+        // this means we do not start reindenting for alignment or we undo it
+        AtomicBoolean lastMethodInCallChain = new AtomicBoolean(true);
```
matozoid: JavaParser only promises thread safety when you don't reuse object instances, so this could be a boolean, I think.

matozoid:
Be prepared to be asked for explanations on this code for as long as Java exists :-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
Vampire: You are probably right, I changed the order, so that `columnAlignFirstMethodChain` is calculated first and the other two are only evaluated if it is `true`.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
Vampire: Nope, this is not about thread safety, but about finalness. The booleans are used inside the `n.getScope().ifPresent` lambda and for this they have to be effectively final.
An alternative is to copy the booleans into another boolean fields that then is effectively final, or using a one-element array, but I usually prefer using `AtomicBoolean` or `AtomicReference` for this as it is more object oriented unless they cause a serious performance issue.

Vampire:
`Be prepared to be asked for explanations on this code for as long as Java exists :-)`
That's ok, but I hoped I commented enough for it to be understandable enough. :-)

Vampire:
Update pushed

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: Oh hey, I learned a new trick :-)

matozoid:
Seems fine! Thanks!

PR #1600, opened 2018-05-24
Adding resolveAccessedValue() to FieldAccessExpr

malteskoruppa:
This PR adds a method `resolveAccessedValue()` to `FieldAccessExpr` which allows the user to easily get from a field access expression to the corresponding field declaration. This functionality is similar in spirit to `MethodCallExpr#resolveInvokedMethod()` and `ObjectCreationExpr#resolveInvokedConstructor()`.

In passing, this PR also fixes a so-far unreported bug in `JavaSymbolSolver` where the case of a `FieldAccessExpr` was not handled in the method `resolveDeclaration(Node,Class)`, leading to an `UnsolvedSymbolException` when trying to call this method for a `FieldAccessExpr` node.

In addition, this PR adds two tests to test the new method.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
@@ -38,6 +38,8 @@ import com.github.javaparser.metamodel.JavaParserMetaModel;
 import javax.annotation.Generated;
 import com.github.javaparser.TokenRange;
 import com.github.javaparser.metamodel.OptionalProperty;
+import com.github.javaparser.resolution.declarations.ResolvedValueDeclaration;
+
 import java.util.function.Consumer;
 
 /**
@@ -252,6 +254,10 @@ public final class FieldAccessExpr extends Expression implements NodeWithSimpleN
     @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
     public void ifFieldAccessExpr(Consumer<FieldAccessExpr> action) {
         action.accept(this);
+    }
+
+    public ResolvedValueDeclaration resolveAccessedValue() {
+        return getSymbolResolver().resolveDeclaration(this, ResolvedValueDeclaration.class);
     }
 
     @Override
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
@@ -29,15 +29,21 @@ import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.visitor.GenericVisitor;
 import com.github.javaparser.ast.visitor.VoidVisitor;
+
 import java.util.Optional;
+
 import static com.github.javaparser.utils.Utils.assertNotNull;
+
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.visitor.CloneVisitor;
 import com.github.javaparser.metamodel.FieldAccessExprMetaModel;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
+
 import javax.annotation.Generated;
+
 import com.github.javaparser.TokenRange;
 import com.github.javaparser.metamodel.OptionalProperty;
+import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedValueDeclaration;
 
 import java.util.function.Consumer;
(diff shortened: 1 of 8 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
```diff
@@ -254,6 +256,10 @@ public void ifFieldAccessExpr(Consumer<FieldAccessExpr> action) {
         action.accept(this);
     }
 
+    public ResolvedValueDeclaration resolveAccessedValue() {
```
ftomassetti: Good idea

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
```diff
@@ -133,6 +133,16 @@ public void inject(CompilationUnit destination) {
                 throw new UnsolvedSymbolException("We are unable to find the value declaration corresponding to " + node);
             }
         }
+        if (node instanceof FieldAccessExpr) {
+            SymbolReference<? extends ResolvedValueDeclaration> result = JavaParserFacade.get(typeSolver).solve((FieldAccessExpr) node);
+            if (result.isSolved()) {
+                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
+                    return resultClass.cast(result.getCorrespondingDeclaration());
+                }
```
ftomassetti, line 141: Do we need an else here? Perhaps throwing an exception

ftomassetti (approved):
Looks good to me.
Thank you!

javaparser-core/src/main/java/com/github/javaparser/ast/expr/FieldAccessExpr.java
matozoid: Could it use a little Javadoc?

malteskoruppa: Sure, I'll add a JavaDoc. :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
malteskoruppa: I followed the general design of the method `JavaSymbolSolver#resolveDeclaration(Node,Class)`. It makes a lot of case distinctions, and an `else` is used nowhere.

Personally I wouldn't go for a case distinction based on if/else at all. Instead, I'd use a local visitor to perform that case distinction. This way, we can be sure that no type of node is forgotten (respectively, that there is some kind of default handler for unexpected nodes). That would be the subject of another PR, though.

malteskoruppa:
I just noticed that `NameExpr` has a similar method already which is simply called `resolve()`. I would suggest renaming it to `resolveAccessValue()` for consistency reasons. That way,

* `MethodCallExpr` has a method `resolveInvokedMethod()` which returns a `ResolvedMethodDeclaration`.
* `ObjectCreationExpr` has a method `resolveInvokedConstructor()` which returns a `ResolvedConstructorDeclaration`.
* `FieldAccessExpr` has a method `resolveAccessedValue()` which returns a `ResolvedValueDeclaration`.
* `NameExpr` has a method `resolveAccessedValue()` which returns a `ResolvedValueDeclaration`.

I'm adding another commit to do this in a few minutes. If you prefer the method in `NameExpr` to be simply called `resolve()`, you can discard that upcoming commit. :-)

malteskoruppa:
Actually, it may be even better to call of these methods simply `resolve()`, since there is an interface `Resolvable<T>` that could be implemented by all of them (`NameExpr` already implements that interface). That would probably be even better.

malteskoruppa:
There we go. Now all of the classes `NameExpr`, `FieldAccessExpr`, `MethodCallExpr`, `ObjectCreationExpr`, and `ExplicitConstructorInvocationStmt` implement the interface `Resolvable<T>` for the respectively appropriate type `T` (well, `NameExpr` already implemented it, but it was missing in all the other classes). All corresponding methods have been renamed to simply `resolve()`. All five `resolve()` methods in these classes got a Javadoc. In my opinion, this is now very consistent and well-documented. :-) I hope you'll agree.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
malteskoruppa: Oh sorry, I misread this and I thought you were referring to another if-clause (the `if (node instanceof NameExpr)` one, just before). But now I see that you meant the `if (result.isSolved())` one.

Well there is an `else` clause! :-) So no worries. As you can see, there is an `else` clause that does throw an exception:

```
    if (result.isSolved()) {
        if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
            return resultClass.cast(result.getCorrespondingDeclaration());
        }
    } else {
        throw new UnsolvedSymbolException("We are unable to find the value declaration corresponding to " + node);
    }
```

matozoid:
@malteskoruppa if you're renaming existing methods, please make an effort to create deprecated methods with the original name that simply forward to the new name, and put a note for users how to solve the deprecation ("call xxx instead")

malteskoruppa:
Good idea! Like this? :-)

matozoid:
Cool! @ftomassetti also happy with these changes?

PR #1604, opened 2018-05-28
Add getters to JavadocInlineTag

tommythorsen:
This change adds getType(), getContent() and getName() methods to
JavadocInlineTag.


tommythorsen:
For context, I'm currently looking into a bug with the impsort-maven-plugin which uses javaparser to sort, and now remove unused imports. Here's the bug: https://github.com/revelc/impsort-maven-plugin/issues/7

I need the ability to parse Javadoc comments and extract referenced classnames. Javaparser has all of the functionality I need, except that there is no way to get the necessary information out of the JavadocInlineTag class.

ftomassetti:
It looks good to me. Thank you for your PR and for using JavaParser in your project!

PR #1606, opened 2018-05-28
Save files within callback of sourceroot.parse

### LINKED ISSUE #1605: Unable to save files with callback in parse method
The following test

```
public void saveInCallback() throws IOException {
    sourceRoot.parse("", sourceRoot.getParserConfiguration(), (localPath, absolutePath, result) -> SourceRoot.Callback.Result.SAVE);
}
```
throws a IOException with the message that the location of the source root is a directory

daanschipper:
Fixes #1605 and added two unit tests.

The path given to `save(CompilationUnit, Path)` was the wrong one, I changed it to the path of compilation unit instead of the whole package (which was the directory).


matozoid:
And thanks again :-)

javaparser-core-testing/javaparser-core-testing.iml
```diff
@@ -10,6 +10,11 @@
     </content>
     <orderEntry type="inheritedJdk" />
     <orderEntry type="sourceFolder" forTests="false" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.jbehave:jbehave-core:4.3" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: commons-io:commons-io:2.6" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.apache.commons:commons-lang3:3.7" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: org.apache.commons:commons-text:1.2" level="project" />
+    <orderEntry type="library" scope="TEST" name="Maven: com.github.valfirst:jbehave-junit-runner:2.1.3" level="project" />
```
matozoid, line 17: This looks odd.

daanschipper: I have no idea what this exactly means. I guess this was changed by one of the maven tasks. And as a heads up, there are two .iml files in the javaparser-core-testing module, `javaparser-core-testing.iml` and `javaparser-testing.iml`

matozoid: Hmmm, I think I'll just remove these files from the repo again. There's nothing but nonsensical merge problems with them.

malteskoruppa: Just FYI, the presence of two .iml files in the javaparser-core-testing module was caused by merging the symbol solver into JavaParser. Back then, the file was called `javaparser-testing.iml`, now it is `javaparser-core-testing.iml`. In other words, `javaparser-testing.iml` is a historical obsolete artifact that can be removed. :-)

matozoid: Correct :-) And if these thing had been on the ignore list, this mistake wouldn't have propagated over the world ;-)

PR #1608, opened 2018-05-28
Fix binary operator transformations.

jooyunghan:
java.jj 's operator names used to find BinaryExpr.Operator using reflection.
Therefore their names should be same with each others.


jooyunghan:
I should have split this commit because it contains another change. I will make a new PR.

matozoid:
Huh? What's this about? We can't go changing operator names just like that, that would mean forcing all users to change their code. Can we start with opening an issue?

jooyunghan:
I agree with you.

matozoid:
Okay, do open an issue so we can investigate and see what to do.

jooyunghan:
Opened #1610

Issue #1610, opened 2018-05-29
Transforming BinaryExpr fails 


jooyunghan:
Following test case fails

```
/**
 * Transforming BinaryExpr and verifying the LexicalPreservation works as expected.
 */
public class BinaryOperatorTransformationsTest extends AbstractLexicalPreservingTest {

    @Test
    public void instanceToStatic() throws IOException {
        considerExpression("a && b");
        expression.asBinaryExpr().setRight(new NameExpr("c"));
        assertTransformedToString("a && c", expression);
    }

}
```

with 

```
in com.github.javaparser.printer.lexicalpreservation.transformations.ast.body.BinaryOperatorTransformationsTest
instanceToStatic(com.github.javaparser.printer.lexicalpreservation.transformations.ast.body.BinaryOperatorTransformationsTest)  Time elapsed: 0.005 s  <<< ERROR!
java.lang.RuntimeException: Attribute 'operator' does not corresponding to any expected value. Text: AND
	at com.github.javaparser.printer.lexicalpreservation.transformations.ast.body.BinaryOperatorTransformationsTest.instanceToStatic(BinaryOperatorTransformationsTest.java:39)
Caused by: java.lang.NoSuchFieldException: AND
	at com.github.javaparser.printer.lexicalpreservation.transformations.ast.body.BinaryOperatorTransformationsTest.instanceToStatic(BinaryOperatorTransformationsTest.java:39)
```

matozoid:
Hi! Thanks for the report. I was afraid something like this would happen because the CSM only looks at names of properties, and some of them are ambiguous. This is actually one of the reasons I made the metamodel, to have a reference to every property in the AST. The correct solution would be to use that instead of the current enum with property names.

PR #1611, opened 2018-05-29
Fix CsmAttribute(OPERATOR) problem 

jooyunghan:
Binary/Unary/AssignExpr uses CsmAttribute(OPERATOR). It tries to get a token type according to the operator name. However (BinaryExp|UnaryExpr|AssignExpr).Operator uses same name for different operators, and uses different name with GeneratedToken's.

That's why code transformation related to operators fails.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
@@ -1,0 +1,53 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2016 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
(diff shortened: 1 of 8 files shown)


jooyunghan:
#1610

matozoid:
For as far as I understand this code, it seems to be a good way to fix this issue right now.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    public void ifStmtTransformation() throws IOException {
+        Statement stmt = consider("if (a) {} else {}");
+        stmt.asIfStmt().setCondition(new NameExpr("b"));
+        assertTransformedToString("if (b) {} else {}", stmt);
+    }
+
+    @Test
+    public void switchEntryCsmHasTrailingUnindent() throws IOException {
```
ftomassetti, line 57: Very nice tests! Thank you!

ftomassetti (approved):
To me it seems nicely tested. Which is good. I also like the presence of comments.
Nice work!

PR #1614, opened 2018-05-29
using string.replace with char instead of string

daanschipper:
`CodeGenerationUtils.packageToPath(String package)` only replace one single char in the package, using the appropriate method circumvents a call to `Pattern.compile(regex)`, which is considered expensive.


ftomassetti:
Nice idea!

matozoid:
Can't argue with that :-)

PR #1618, opened 2018-05-29
Lots of documentation and a little refactoring of ProjectRoot

matozoid:
@Vampire @daanschipper - what do you think?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
@@ -11,7 +11,8 @@ import java.nio.file.PathMatcher;
 import java.util.Optional;
 
 /**
- * Defines a strategy to collect files for the ProjectRoot.
+ * A strategy for discovering the structure of a project.
+ * Implementations could read a pom.xml, a Gradle build file, a makefile...
  */
 public interface CollectionStrategy {
 
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/ParserCollectionStrategy.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/ParserCollectionStrategy.java
@@ -9,7 +9,12 @@ import java.nio.file.attribute.BasicFileAttributes;
 import static java.nio.file.FileVisitResult.*;
 
 /**
- * Strategy which collects all SourceRoots and returns them in a ProjectRoot object.
+ * A brute force strategy for discovering a project structure.
+ * It will search through the given project root path for Java files,
+ * look at their package declarations, and figure out the root directories for those files.
+ * No project definition files like pom.xml or build.gradle are used.
+ * This strategy is crude, but can work for many cases.
+ * Note that any build artifacts will also be detected: jar files in target directories and so on.
  */
 public class ParserCollectionStrategy implements CollectionStrategy {
 
@@ -28,23 +33,26 @@ public class ParserCollectionStrategy implements CollectionStrategy {
         ProjectRoot projectRoot = new ProjectRoot(path, parserConfiguration);
         try {
             Files.walkFileTree(path, new SimpleFileVisitor<Path>() {
-                Path current_root;
-                PathMatcher javaMatcher = getPathMatcher("glob:**.java");
+                private Path currentRoot;
+                private PathMatcher javaMatcher = getPathMatcher("glob:**.java");
+                private PathMatcher jarMatcher = getPathMatcher("glob:**.jar");
 
                 @Override
                 public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
                     if (javaMatcher.matches(file)) {
-                        current_root = getRoot(file).orElse(null);
-                        if (current_root != null) {
+                        currentRoot = getRoot(file).orElse(null);
+                        if (currentRoot != null) {
                             return SKIP_SIBLINGS;
                         }
+                    } else if (jarMatcher.matches(file)) {
+                        projectRoot.addJarFile(file);
                     }
                     return CONTINUE;
                 }
 
                 @Override
                 public FileVisitResult preVisitDirectory(Path dir, BasicFileAttributes attrs) throws IOException {
-                    if (Files.isHidden(dir) || (current_root != null && dir.startsWith(current_root))) {
+                    if (Files.isHidden(dir) || (currentRoot != null && dir.startsWith(currentRoot))) {
                         return SKIP_SUBTREE;
                     }
                     return CONTINUE;
@@ -52,9 +60,9 @@ public class ParserCollectionStrategy implements CollectionStrategy {
 
                 @Override
                 public FileVisitResult postVisitDirectory(Path dir, IOException e) {
-                    if (dir.equals(current_root)) {
+                    if (dir.equals(currentRoot)) {
                         projectRoot.addSourceRoot(dir);
-                        current_root = null;
+                        currentRoot = null;
                     }
                     return CONTINUE;
                 }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/AarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/AarTypeSolver.java
@@ -25,6 +25,9 @@ import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 
 /**
+ * Will let the symbol solver look inside an Android aar file while solving types.
+ * (It will look inside the contained classes.jar)
+ *
  * @author Federico Tomassetti
  */
 public class AarTypeSolver implements TypeSolver {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
@@ -25,6 +25,9 @@ import java.util.ArrayList;
 import java.util.List;
 
 /**
+ * A container for type solvers. All solving is done by the contained type solvers.
+ * This helps you when an API asks for a single type solver, but you need several.
+ *
  * @author Federico Tomassetti
  */
 public class CombinedTypeSolver implements TypeSolver {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
@@ -32,6 +32,8 @@ import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 
 /**
+ * Will let the symbol solver look inside a jar file while solving types.
+ *
  * @author Federico Tomassetti
  */
 public class JarTypeSolver implements TypeSolver {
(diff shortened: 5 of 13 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategy.java
```diff
@@ -6,22 +6,16 @@
...
-import static java.nio.file.FileVisitResult.SKIP_SUBTREE;
+import java.nio.file.Path;
 
 /**
- * Strategy which collects all SourceRoots and initialises the TypeSolver and returns the SourceRoots configured
- * with the TypeSolver in a ProjectRoot object.
+ * Attempts to configure the symbol solver to look at the source code and jar files found in a project directory.
  */
-public class SymbolSolverCollectionStrategy implements CollectionStrategy {
+public class SymbolSolverCollectionStrategy {
```
matozoid, line 18: This thing has become really badly named now.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategyTest.java
```diff
@@ -1,29 +1,30 @@
...
 import java.nio.file.Path;
-import java.util.Optional;
 
 import static org.junit.Assert.assertTrue;
 
 public class SymbolSolverCollectionStrategyTest {
 
-    private final Path root = CodeGenerationUtils.mavenModuleRoot(SymbolSolverCollectionStrategyTest.class).resolve("").getParent();
-    private final ProjectRoot projectRoot = new CollectionContext(new SymbolSolverCollectionStrategy()).collect(root);
+    private final Path root = CodeGenerationUtils.mavenModuleRoot(JavaParser.class);
```
matozoid, line 20: This test took ages because it had tons of source roots, and many of them double. It's now contained to just javaparser-core.

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
```diff
@@ -11,7 +11,8 @@
 import java.util.Optional;
 
 /**
- * Defines a strategy to collect files for the ProjectRoot.
+ * A strategy for discovering the structure of a project.
+ * Implementations could read a pom.xml, a Gradle build file, a makefile...
```
ftomassetti, line 15: Nice

javaparser-core/src/main/java/com/github/javaparser/utils/ParserCollectionStrategy.java
```diff
@@ -28,33 +33,36 @@ public ProjectRoot collect(Path path) {
...
+                private Path currentRoot;
+                private PathMatcher javaMatcher = getPathMatcher("glob:**.java");
+                private PathMatcher jarMatcher = getPathMatcher("glob:**.jar");
 
                 @Override
                 public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
                     if (javaMatcher.matches(file)) {
-                        current_root = getRoot(file).orElse(null);
-                        if (current_root != null) {
+                        currentRoot = getRoot(file).orElse(null);
```
ftomassetti, line 43: much better!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/AarTypeSolver.java
```diff
@@ -25,6 +25,9 @@
 import java.util.zip.ZipEntry;
 
 /**
+ * Will let the symbol solver look inside an Android aar file while solving types.
```
ftomassetti, line 28: Nice

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
```diff
@@ -25,6 +25,9 @@
 import java.util.List;
 
 /**
+ * A container for type solvers. All solving is done by the contained type solvers.
```
ftomassetti, line 28: Agreed

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
```diff
@@ -74,14 +76,11 @@ private File dumpToTempFile(InputStream inputStream) throws IOException {
         byte[] buffer = new byte[8 * 1024];
 
         try {
-            OutputStream output = new FileOutputStream(tempFile);
-            try {
+            try (OutputStream output = new FileOutputStream(tempFile)) {
```
ftomassetti, line 79: can the try with variables have a finally? If that is the case we could combine the two try

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
```diff
@@ -159,12 +158,10 @@ public ResolvedReferenceTypeDeclaration solveType(String name) throws UnsolvedSy
     private class ClasspathElement {
         private JarFile jarFile;
         private JarEntry entry;
-        private String path;
 
-        ClasspathElement(JarFile jarFile, JarEntry entry, String path) {
+        ClasspathElement(JarFile jarFile, JarEntry entry) {
```
ftomassetti, line 162: Ok, it was not used anymore...

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
```diff
@@ -24,12 +24,21 @@
 import java.util.Optional;
 
 /**
+ * Uses reflection to resolve types.
+ * All classes on the classpath used to run your application will be found.
```
ftomassetti, line 28: Well, not all classes if the flag jreOnly is set.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategy.java
```diff
@@ -34,48 +28,16 @@ public SymbolSolverCollectionStrategy(ParserConfiguration parserConfiguration) {
...
-                        typeSolver.add(new JavaParserTypeSolver(current_root.toFile()));
-                        current_root = null;
-                    }
-                    return CONTINUE;
-                }
-            });
-        } catch (IOException e) {
-            Log.error(e, "Unable to walk %s", path);
-        }
+        ProjectRoot projectRoot = new ParserCollectionStrategy(parserConfiguration).collect(path);
```
ftomassetti, line 32: This look simpler

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategy.java
```diff
@@ -34,48 +28,16 @@ public SymbolSolverCollectionStrategy(ParserConfiguration parserConfiguration) {
...
+        projectRoot.getSourceRoots().forEach(sourceRoot -> typeSolver.add(new JavaParserTypeSolver(sourceRoot.getRoot())));
+        projectRoot.getJarFiles().forEach(jarFile -> {
+            try {
+                typeSolver.add(new JarTypeSolver(jarFile));
+            } catch (IOException e) {
+                throw new RuntimeException("Something went wrong while configuring the symbol solver.", e);
```
ftomassetti, line 38: Should we add the name of the Jar in the exception message?

ftomassetti (approved):
Much better! Thank you for figuring out and documenting the goal of so many classes!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
Vampire: Yes, try-with-resources can have finally and catch blocks as usual, so yes, would definitely make sense to combine the two try blocks.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategy.java
Vampire: Well, then rename it, removing the interface is already a backwards-incompatible change. :-)

Vampire (commented):
I think `com.github.javaparser.utils.CollectionStrategy#getPathMatcher` should be moved to `com.github.javaparser.utils.ParserCollectionStrategy` or inlined, as it is only used there and not ultimately useful for potential other implementations and more importantly it is totally uninteresting for consumers of the interface. It could even be made static.

It is also questionable whether `com.github.javaparser.utils.CollectionStrategy#getRoot` is interesting gor consumers of the interface. I guess it is also more a utility for implementations and in that it is maybe also just of limited usefulness. So maybe it should also be moved to `com.github.javaparser.utils.ParserCollectionStrategy` completely.
Interfaces should define an interface to consumers, not provide utility methods for their implementations.

Furthermore I'd suggest to change `CollectionStrategy` and `collect(Path path)` to `CollectionStrategy<T>` and `collect(T whatever)`. Who says a project root can only be calculated from a path? You could make one from a Gradle `Project` instance or from a `SourceSet` instance.

Of course you don't have to implement that interface, but I think it would be worth generalizing this interface.

While I think about it, what's the use for the interface anyway?
Why not just have classes that provide project roots, do you consume them somewhere, or are they expected to be consumed by someone else?
Actually it is a `Function`, currently a `Function<Path, ProjectRoot>` and possibly a `Function<T, ProjectRoot>` if the method would be renamed.
While it is still questionable maybe whether this interface is actually needed.

They are maybe more a `Supplier<ProjectRoot>`, or a `Function<T, ProjectRoot>` that would also fit to the `SymbolSolverCollectionStrategy`. If you actually need a generic way to give those to some common place which currently is not the case.

daanschipper:
This looks good! Its nice to see that you renamed the collection strategies and the documentation made everything a lot more clear. When I implemented `ProjectRoot` the bigger picture wasn't completely clear.

However, I purposely separated `ParserCollectionStrategy` and `SymbolSolverCollectionStrategy` and have them implement `CollectionStrategy` since the latter is _much_ slower. 
- `ParserCollectionStrategy` collects all source roots of a project with the intention to use it only with javaparser-core. To do this, you only need to visit _one_ .java file per source root (to parse it with javaparser and get the source root). Since the source root of this file is the same as all other .java files within the source root, we can exit this directory completely.
- `SymbolSolverCollectionStrategy` collects all source roots of a project _and_ all .jar files present in the root of the project with the intention to use it with java-symbol-solver-core. Besides finding the source root (the same way as `ParserCollectionStrategy`, we traverse _each_ file in the root of the project and if it's a .jar file we add it to our typesolver.
I had the idea to extend this with a `GradleCollectionStrategy` and a `MavenCollectionStrategy` that extends `ParserCollectionStrategy` and get the jars from the build.gradle and pom.xml.

This is also the reason that `com.github.javaparser.utils.CollectionStrategy#getPathMatcher` is within `CollectionStrategy` since I had to use this method in both implementing classes, at the time this worked easily and I didn't bother to look further.

daanschipper:
The test `com.github.javaparser.symbolsolver.utils.SymbolSolverCollectionStrategyTest` is failing because of the difference in walking the file tree

matozoid:
Thanks for all the feedback!

PR #1620, opened 2018-05-30
Add and use AutoCloseable subclasses of generated Providers to be able to use try-with-resources

Vampire:


javaparser-core/src/main/javacc-support/com/github/javaparser/AutoCloseableProvider.java
```diff
@@ -0,0 +1,4 @@
+package com.github.javaparser;
+
+public interface AutoCloseableProvider extends Provider, AutoCloseable {
```
matozoid, line 3: I know why you're doing it this way, but damn, now we have "AutoCloseable" all over the API :-/

matozoid:
In my opinion this is a lot of effort if the only goal is to use another way of closing the resource in a private method somewhere.

Vampire:
Well, you said it makes sense. :grinning:
I don't know whether others can use `Provider`?
Or whether you can make the JavaCC generated `Provider` extend `Autocloseable`, that would be the cleanest solution, but I have no experience with JavaCC.
If you don't like it, just close it, would be ok for me.

javaparser-core/src/main/javacc-support/com/github/javaparser/AutoCloseableProvider.java
Vampire: Yep, tell me how to modify the JavaCC generated file instead. :-)
With Gradle it would be easy as pie to just post-process the generated file, e. g. with JavaParser.

matozoid:
JavaCC doesn't support this case I think. One way would be to copy the whole Provider java file and make our own, but when possible I like to use the generated things. So thanks for trying this out, I guess there's no simple way to do it :-(

PR #1623, opened 2018-05-31
Exit sourceroot callback

daanschipper:
You can now exit parsing the source root if you return `TERMINATE`. This might be useful when looking for something specific, you don't have to traverse the whole directory.

I also came across a test class with a lot of warnings, I just cleaned it up.I can make a separate pr if desired.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
@@ -25,8 +25,7 @@ import java.util.stream.Collectors;
 
 import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
 import static com.github.javaparser.Providers.provider;
-import static com.github.javaparser.utils.CodeGenerationUtils.fileInPackageRelativePath;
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -259,6 +256,52 @@ public CompilationUnit parse(String startPackage, String filename) {
         }
     }
 
+    private FileVisitResult callback(Path absolutePath, ParserConfiguration configuration, Callback callback) throws IOException {
+        if (!Files.exists(absolutePath)) {
```
ftomassetti: Should this happen?

ftomassetti (approved):

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
daanschipper: Yes that can happen. But I see now that the error is either thrown again or properly catched by the invoking methods, so I'll remove this check.

PR #1626, opened 2018-05-31
Lots of documentation

matozoid:
Only the docs this time.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
@@ -25,7 +25,7 @@ import java.util.Optional;
 
 /**
  * Uses reflection to resolve types.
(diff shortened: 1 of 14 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
```diff
@@ -24,12 +24,21 @@
 import java.util.Optional;
 
 /**
+ * Uses reflection to resolve types.
+ * All classes on the classpath used to run your application will be found.
```
Vampire: Still only true if the `boolean` constructor is used and `false` given.
If `true` given or no parameter, then only JRE classes are found.

Vampire (changes requested):

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
matozoid: Fixed

PR #1635, opened 2018-06-07
NormalAnnotationExpr: addPair should allow any expression as a value

marcus-nl:
NormalAnnotationExpr.addPair should allow any expression as a value, instead of just NameExpr.


matozoid:
Makes sense. Thanks!

PR #1637, opened 2018-06-07
Refactoring of Difference application and fix for Issue #1567

### LINKED ISSUE #1567: Wrong indentation when removing annotations
As promised, one of the issues that currently do not work yet with the LexicalPreservingPrinter.

The result of the test below is currently the following
```
class X {
    public void testCase() {
  }
}
```
besides correctly removing the annotation two extra spaces are added before the modifier
**The problem is identical when there is no modifier at all.**

MethodDeclarationTransformationsTest
```
    @Test
    public void removingAnnotations() {
        considerCode(
                "class X {" + EOL +
                        "  @Override" + EOL +
                        "  public void testCase() {" + EOL +
                        "  }" + EOL +
                        "}" + EOL
        );

        cu.getType(0).getMethods().get(0).getAnnotationByName("Override").get().remove();

        String result = LexicalPreservingPrinter.print(cu.findCompilationUnit().get());
        assertEqualsNoEol(
                "class X {\n" +
                        "  public void testCase() {\n" +
                        "  }\n" +
                        "}\n", result);
    }
```

ThLeu:
fixes #1567 

I tried to make the correction of spaces clearer with this refactoring while fixing the issue.
As mentioned I introduced a "RemovedGroup" which holds all consecutive Removed elements. This makes it easier to handle removal of a complete line for now.
This can hopefully be merged with the rest of the "whitespace correction logic" to become even clearer.

I also added some tests and JavaDoc.
There is one test I added that will not work yet (I will add an issue for that).

The logic is also still problematic with CsmToken elements (CsmChild works pretty well) as there is currently no way to navigate from CsmToken to the previous or next one, maybe it could be added as with the TokenRange in CsmChild (any idea @ftomassetti or @matozoid? :-))


javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
```diff
@@ -252,4 +253,45 @@ public void addingAnnotationsShort() {
...
+        assertEqualsNoEol(
+                "class X {\n" +
+                        "  public void testCase() {\n" +
+                        "  }\n" +
+                        "}\n", result);
+    }
+
+    @Ignore
+    @Test
+    public void removingAnnotationsWithSpaces() {
```
ThLeu, line 279: Test mentioned in #1638

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmMix.java
```diff
@@ -49,6 +49,6 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "CsmMix[" + String.join(", ", elements.stream().map(e -> elements.toString()).collect(Collectors.toList())) + "]";
+        return elements.stream().map(e -> e.toString()).collect(Collectors.joining(",", "CsmMix[", "]"));
```
ThLeu, line 52: Small additional bugfix, the toString Method printed too much.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -121,37 +113,14 @@ private int considerEnforcingIndentation(NodeText nodeText, int nodeTextIndex) {
      * to the difference (adding and removing the elements provided).
      */
     void apply() {
-        do {
-            if (diffIndex < diffElements.size() && originalIndex >= originalElements.size()) {
-                DifferenceElement diffElement = diffElements.get(diffIndex);
-                if (diffElement instanceof Kept) {
-                    Kept kept = (Kept) diffElement;
+        extractReshuffledDiffElements(diffElements);
```
ThLeu, line 116: I moved the logic for the Reshuffled element out of the do while loop because it made it more complicated since the Reshuffled elements are "extracted" into additional Addition, Kept and Removed elements which are traversed in the do while loop.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -121,37 +113,14 @@ private int considerEnforcingIndentation(NodeText nodeText, int nodeTextIndex) {
...
-                    nodeText.addElement(originalIndex, addedElement.toTextElement());
-                    originalIndex++;
-                    diffIndex++;
-                } else {
-                    throw new UnsupportedOperationException(diffElement.getClass().getSimpleName());
-                }
-            } else if (diffIndex >= diffElements.size() && originalIndex < originalElements.size()) {
-                TextElement originalElement = originalElements.get(originalIndex);
+        do {
+            boolean isLeftOverDiffElement = applyLeftOverDiffElements();
```
ThLeu, line 120: These correspond to the first if and else if in the do while loop.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -175,129 +143,233 @@ void apply() {
...
+     * Example:
+     * <pre>
+     * Elements: Kept|Removed1|Removed2|Kept|Removed3|Added|Removed4
+     * Groups:        <----Group1---->       Group2         Group3
+     * Keys:          Removed1+Removed2      Removed3       Removed4
+     * </pre>
+     *
+     * @return Map with all Removed elements as keys to their corresponding RemovedGroup
+     */
+    private Map<Removed, RemovedGroup> combineRemovedElementsToRemovedGroups() {
```
ThLeu, line 301: This is where the RemovedGroups get created.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -175,129 +143,233 @@ void apply() {
...
+                Comment comment = (Comment) originalElementChild.getChild();
                 if (!comment.isOrphan() && comment.getCommentedNode().isPresent() && comment.getCommentedNode().get().equals(removed.getChild())) {
                     nodeText.removeElement(originalIndex);
                 } else {
                     originalIndex++;
                 }
             } else {
                 nodeText.removeElement(originalIndex);
-                if (originalIndex < originalElements.size() && originalElements.get(originalIndex).isNewline()) {
-                    originalIndex = considerCleaningTheLine(nodeText, originalIndex);
```
ThLeu, line 281: considerCleaningTheLine was integrated with the logic in cleanTheLineOfLeftOverSpace (line 398)

matozoid:
Thanks! About prev/next token: I vaguely remember trying to unify CsmToken and JavaToken and running into some problems - that would have been the simplest solution. Feel free to add it - I fear it will be some copy/pasting from JavaToken.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
```diff
@@ -252,4 +253,45 @@ public void addingAnnotationsShort() {
...
+        cu.getType(0).getMethods().get(0).getAnnotationByName("Override").get().remove();
+
+        String result = LexicalPreservingPrinter.print(cu.findCompilationUnit().get());
+        assertEqualsNoEol(
+                "class X {\n" +
+                        "  public void testCase() {\n" +
+                        "  }\n" +
+                        "}\n", result);
+    }
 }
```
ftomassetti: nitpick: newline at the end of the file

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/RemovedGroup.java
```diff
@@ -0,0 +1,215 @@
...
+import com.github.javaparser.printer.concretesyntaxmodel.CsmToken;
+
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+class RemovedGroup implements Iterable<Removed> {
```
ftomassetti: What this class represent?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/RemovedGroup.java
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            @Override
+            public Removed next() {
+                return removedList.get(currentIndex++);
+            }
+
+        };
+    }
+}
```
ftomassetti: Newline here :)

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ftomassetti: It sounds like a smart thing to do

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -175,129 +143,233 @@ void apply() {
...
-                // element that has a mapping
-                int originalCsmIndex = -1;
-                for (int nj = ni + 1; nj< nextOrderElements.size() && originalCsmIndex==-1; nj++) {
-                    if (correspondanceBetweenNextOrderAndPreviousOrder.containsKey(nj)) {
-                        originalCsmIndex = correspondanceBetweenNextOrderAndPreviousOrder.get(nj);
-                        if (!elementsToAddBeforeGivenOriginalCSMElement.containsKey(originalCsmIndex)){
-                            elementsToAddBeforeGivenOriginalCSMElement.put(originalCsmIndex, new LinkedList<>());
-                        }
-                        elementsToAddBeforeGivenOriginalCSMElement.get(originalCsmIndex).add(nextOrderElements.get(ni));
+    private void extractReshuffledDiffElements(List<DifferenceElement> diffElements) {
```
ftomassetti, line 192: This seems as big as the methods I use to write :D

ftomassetti (approved):
What I can say is that I see:

1. Meaningful tests being added: VERY GOOD
2. Code being split up in smaller components: VERY GOOD

So I think we are going in the right direction. Kudos for the work on this one!

P.S. This stuff is not easy, is it :D ?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
ThLeu: Well it was just a renaming for the most part :smile:
I didn't really change much here as far as I remember, so....I didn't write it 😝

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/RemovedGroup.java
ThLeu: Thanks, I will add a JavaDoc when I get home.

ThLeu:
Thanks for your comments.
@matozoid Ok, thank you, maybe I will have a look at CsmToken / JavaToken topic with another issue or so :)
@ftomassetti Oh no, it is definitely not easy :smile: whitespace and indentation in general seems like a topic for itself :) this is also a reason why I tried to handle it a bit separately

ThLeu:
I will fix the issues you mentioned when I get home, seems like it shouldn't take long :) ... if you do not find any others.

matozoid:
Good work! @ThLeu, any chance we can get you to look at these issues? https://github.com/javaparser/javaparser/labels/Lexical%20Preservation

ThLeu:
Sure thing, I will continue to look at some of the lexical preservation issues.

PR #1644, opened 2018-06-12
Throw an UnsolvedSymbolException instead of a RuntimeException

malteskoruppa:
Consider the following code:

```
package testcase;

import java.io.IOException;
import javax.servlet.http.*;

public abstract class UnknownMethods extends HttpServlet {
	public void test2(HttpServletRequest request, HttpServletResponse response) throws IOException {
		String data = "world!";
		response.getWriter().println("hello " + data);
	}
}
```

Trying to resolve the method call expression `response.getWriter().println("hello " + data)` correctly throws an `UnsolvedSymbolException`, since the symbol solver obviously cannot resolve the method call expression whose corresponding method declaration is buried in the javax package (the reflection type solver only supports resolution for classes that are in the classloader, if I understand correctly).

Now consider instead the following code:

```
package testcase;

import java.io.IOException;
import javax.servlet.http.*;

public abstract class UnknownMethods extends HttpServlet {
	public void test1(HttpServletRequest request, HttpServletResponse response) throws IOException {
		response.getWriter().println("hello world!");
	}
}
```

Now again try to resolve the method call expression `response.getWriter().println("hello world!")`.

**Expected behavior:** The symbol solver should throw an `UnsolvedSymbolException`, as above.
**Actual behavior:** The symbol solver throws an unspecific `RuntimeException`.

This PR fixes this bug and adds corresponding tests.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -166,7 +166,7 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
             try {
                 typeOfScope = JavaParserFacade.get(typeSolver).getType(scope);
             } catch (Exception e) {
-                throw new UnsolvedSymbolException("Issue calculating the type of the scope of " + this);
+                throw new UnsolvedSymbolException(scope.toString(), wrappedNode.toString(), e);
             }
             if (typeOfScope.isWildcard()) {
                 if (typeOfScope.asWildcard().isExtends() || typeOfScope.asWildcard().isSuper()) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -166,7 +166,7 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
             try {
                 typeOfScope = JavaParserFacade.get(typeSolver).getType(scope);
             } catch (Exception e) {
-                throw new RuntimeException("Issue calculating the type of the scope of " + this, e);
+                throw new UnsolvedSymbolException("Issue calculating the type of the scope of " + this);
             }
             if (typeOfScope.isWildcard()) {
                 if (typeOfScope.asWildcard().isExtends() || typeOfScope.asWildcard().isSuper()) {
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/UnsolvedSymbolException.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/UnsolvedSymbolException.java
@@ -28,26 +28,48 @@ package com.github.javaparser.resolution;
  */
 public class UnsolvedSymbolException extends RuntimeException {
 
-    private String context;
+    /**
+     * The name of the symbol that could not be resolved.
+     */
     private String name;
 
-    public UnsolvedSymbolException(String name, String context) {
-        super("Unsolved symbol in " + context + " : " + name);
-        this.context = context;
-        this.name = name;
-    }
+    /**
+     * Additional information that provides more details on the context that a symbol could not be resolved in, or
+     * {@code null} if there is no contextual information, or if the contextual information is unknown.
+     */
+    private String context;
+
+    /**
+     * The throwable that caused this {@code UnsolvedSymbolException} to get thrown, or {@code null} if this
+     * {@code UnsolvedSymbolException} was not caused by another throwable, or if the causative throwable is unknown.
+     */
+    private Throwable cause;
 
     public UnsolvedSymbolException(String name) {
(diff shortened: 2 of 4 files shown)


ThLeu:
+1 for explicit exceptions

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -166,7 +166,7 @@ public final Context getParent() {
             try {
                 typeOfScope = JavaParserFacade.get(typeSolver).getType(scope);
             } catch (Exception e) {
-                throw new RuntimeException("Issue calculating the type of the scope of " + this, e);
+                throw new UnsolvedSymbolException("Issue calculating the type of the scope of " + this);
```
ThLeu: Just a note, not a review... :smile:
Maybe to keep the causing exception `e` UnsolvedSymbolException should be extended with another constructor, what do you think? Otherwise you lose some valuable information.

malteskoruppa: I agree it could help, let's see what @ftomassetti has to say about it. :-)

ftomassetti: I would say this depends on the type of the exception. What about having a specific catch for meaningful resolution exceptions and a generic one that throws a RuntimeException?

ftomassetti (approved):

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
malteskoruppa: I would argue that throwing a generic `RuntimeException` is *never* a good idea. Ideally, every thrown exception should be meaningful. The code needs to figure out what went wrong and throw a sensible exception accordingly. Throwing a generic `RuntimeException` is a bit like a segfault or something... the application just crashes and does not give any specific information to the user.

In addition, generic catches (`catch(Exception e)`) are generally not a good idea in my opinion, because it completely obscures what kind of exceptions the programmer expects to be thrown. In case several different types of exceptions are expected to be potentially thrown that need to be handled in the same way, one should use a multi catch (`catch(XException | YException | ZException e)`).

So in this case, I am not particularly fond of the `catch(Exception e)` (but that was already there before and I do not know what types of exceptions you expect may happen), but I think it is fair to handle all exceptions at this point by rethrowing an `UnsolvedSymbolException`, since that is effectively what it means: the type of the scope could not be solved for some reason. :-)

ftomassetti: My understanding is that we should handle only exceptions to which we can react meaningfully in that context, so I would agree we could capture some exceptions and throw UnsolvedSymbolException in reaction.

As you said throwing a RuntimeException is suboptimal but the reason for doing that is to decorate it with additional information. It helped debugging because the code is complex and when there was an exception I could follow what was happening by looking at the information added into the exception ("oh, I got this issue while I was trying to solve this symbol in this context and that was caused by this other problem..."). So for me throwing a RuntimeException was almost as I did not catch it: I just saw it passing, added some debugging infor and resend it on its way.

matozoid:
I'm not putting this in the release because there still seems to be discussion about this.

matozoid:
@malteskoruppa @ftomassetti - can you come to a conclusion on this PR please? :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
ftomassetti: Why not going back to what @ThLeu suggested and add a constructor to UnsolvedSymbolException which preserve the root cause (i.e., the other exception)? So you get the UnsolvedSymbolException and I get a way to debug

ftomassetti:
Comment added

matozoid:
@malteskoruppa - this PR is missing a second release now...

malteskoruppa:
Fine by me! I'll take care of it, but I'm kind of busy right now. It may take me a while before I get to it, but I'll try to do it asap. :)

matozoid:
Okay @malteskoruppa, I've invented a new tag for this state :-)

malteskoruppa:
@ftomassetti @matozoid Does this look ok?

An `UnsolvedSymbolException` can now have an optional `cause`, which is a wrapped exception as discussed (well, I made it general enough so it can be any `Throwable`, not only an exception).

On that occasion I also added some Javadoc comments to `UnsolvedSymbolException` and removed some duplicated code in that class.

The `AbstractJavaParserContext` now instantiates the `UnsolvedSymbolException` correctly when it could not resolve the type of the scope symbol, providing the name of the symbol and the wrapped exception, and even the context. So now the `UnsolvedSymbolException` thrown at that point provides even more information than the `RuntimeException` ever did. :wink:

ftomassetti:
It looks good to me. If Travis is happy with that for me it can be merged

ftomassetti:
And thank you for your work on this one!

PR #1645, opened 2018-06-12
Solve fields of classes with unresolvable ancestors

malteskoruppa:
Consider the following code:

```
import com.third.party.library.UnknownClass;

public class ClassExtendingUnknownClass extends UnknownClass {

    private int foo;

    public int getFoo() {
        return foo;
    }

    public int getFoo2() {
        return this.foo;
    }
}
```

The point here is that while we do have the source code of `ClassExtendingUnknownClass`, we do not have the source code of its parent class `UnknownClass`, which comes from some external third-party library that is unknown.

Nevertheless, we should be able to resolve the field declaration corresponding to the `NameExpr` `foo` in method `getFoo()`, as well as the field declaration corresponding to the `FieldAccessExpr` `this.foo` in method `getFoo2()`. We don't need to know the source code of the parent class for that. However, when we try to instruct the symbol solver to resolve the field declaration of either the `NameExpr` or the `FieldAccessExpr`, an `UnsolvedSymbolException` is thrown.

This is because, essentially, the symbol solver attempts to resolve the parent class and the implemented interfaces *before* it even looks at the fields of the current class. In the above example, an `UnsolvedSymbolException` is thrown while trying to resolve the parent class (since the parent class is unknown), which propagates backwards to the original call to `resolve()` of the `NameExpr` or the `FieldAccessExpr`.

This PR fixes that bug, thus making it possible to resolve a field declaration in scenarios such as the above.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -166,7 +166,7 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
             try {
                 typeOfScope = JavaParserFacade.get(typeSolver).getType(scope);
             } catch (Exception e) {
-                throw new RuntimeException("Issue calculating the type of the scope of " + this, e);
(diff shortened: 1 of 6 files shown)


malteskoruppa:
I accidentally branched off this PR from the branch I created for my previous PR #1644, sorry about that. I'm closing this PR and reopening a new one so that the two PRs are separate.

PR #1646, opened 2018-06-12
Solve fields of classes with unresolvable ancestors

malteskoruppa:
Consider the following code:

```java
import com.third.party.library.UnknownClass;

public class ClassExtendingUnknownClass extends UnknownClass {

    private int foo;

    public int getFoo() {
        return foo;
    }

    public int getFoo2() {
        return this.foo;
    }
}
```

The point here is that while we do have the source code of `ClassExtendingUnknownClass`, we do not have the source code of its parent class `UnknownClass`, which comes from some external third-party library that is unknown.

Nevertheless, we should be able to resolve the field declaration corresponding to the `NameExpr` `foo` in method `getFoo()`, as well as the field declaration corresponding to the `FieldAccessExpr` `this.foo` in method `getFoo2()`. We don't need to know the source code of the parent class for that. However, when we try to instruct the symbol solver to resolve the field declaration of either the `NameExpr` or the `FieldAccessExpr`, an `UnsolvedSymbolException` is thrown.

This is because, essentially, the symbol solver attempts to resolve the parent class and the implemented interfaces *before* it even looks at the fields of the current class. In the above example, an `UnsolvedSymbolException` is thrown while trying to resolve the parent class (since the parent class is unknown), which propagates backwards to the original call to `resolve()` of the `NameExpr` or the `FieldAccessExpr`.

This PR fixes that bug, thus making it possible to resolve a field declaration in scenarios such as the above.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -293,14 +293,31 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration {
 
         // We want to avoid infinite recursion in case of Object having Object as ancestor
         if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
-            ResolvedReferenceType superclass = getSuperClass();
+            ResolvedReferenceType superclass;
+            try {
+                superclass = getSuperClass();
+            } catch (UnsolvedSymbolException e) {
+                // in case we could not resolve the super class, we may still be able to resolve (some of) the
+                // implemented interfaces and so we continue gracefully with an (incomplete) list of ancestors
+                superclass = null;
+            }
             if (superclass != null) {
                 ancestors.add(superclass);
             }
             if (wrappedNode.getImplementedTypes() != null) {
                 for (ClassOrInterfaceType implemented : wrappedNode.getImplementedTypes()) {
-                    ResolvedReferenceType ancestor = toReferenceType(implemented);
-                    ancestors.add(ancestor);
+                    ResolvedReferenceType ancestor;
+                    try {
+                        ancestor = toReferenceType(implemented);
+                    } catch (UnsolvedSymbolException e) {
+                        // in case we could not resolve some implemented interface, we may still be able to resolve the
+                        // extended class or (some of) the other implemented interfaces and so we continue gracefully
+                        // with an (incomplete) list of ancestors
+                        ancestor = null;
+                    }
+                    if (ancestor != null) {
+                        ancestors.add(ancestor);
+                    }
                 }
             }
         }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -90,8 +90,8 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration {
     @Override
     public String toString() {
         return "JavaParserClassDeclaration{" +
-                "wrappedNode=" + wrappedNode +
-                '}';
+               "wrappedNode=" + wrappedNode +
+               '}';
     }
 
     ///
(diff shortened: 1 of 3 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -293,14 +293,31 @@ public boolean isTypeParameter() {
 
         // We want to avoid infinite recursion in case of Object having Object as ancestor
         if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
-            ResolvedReferenceType superclass = getSuperClass();
+            ResolvedReferenceType superclass;
+            try {
+                superclass = getSuperClass();
+            } catch (UnsolvedSymbolException e) {
+                // in case we could not resolve the super class, we may still be able to resolve (some of) the
```
ftomassetti: I think it could be useful to somehow track this. Callers of `getAncestors` have no way to know that some ancestors where skipped. This could lead to confusing results later on

ftomassetti: Perhaps we could add a flag to the method to specify if we are ok obtaining an incomplete list of ancestors. Otherwise we could return some stubs representing unresolved ancestors.

ftomassetti (commented):
I think the goal makes total sense and this is something we should do, however I am afraid that just ignoring ancestors that cannot be retrieven could cause confusion

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
malteskoruppa: Ok, I'll add a commit to do that. :-)

malteskoruppa: There we go. What do you think? :-)

malteskoruppa:
There we go. What do you think? :-)

ftomassetti:
To me it seems reasonable: @matozoid ? A better idea for this one?

malteskoruppa:
In general, what do you think about the following idea? However, note that it would be a huge paradigm shift for the entire symbol solver.

How about an interface `UnresolvedType` (as opposed to the existing interface `ResolvedType`) that represents a type that could not be resolved and contains only the meta-data that is available (e.g., the simple name of the type). Whenever the symbol solver cannot resolve a type, instead of throwing an `UnsolvedSymbolException`, it returns an `UnresolvedType` instead. That way, in this particular case for instance, the list of ancestors could contain some `ResolvedType` and some `UnresolvedType` elements (for this reason, an additional common super interface `PossiblyResolvedType` or similar seems sensible).

Basically, for every interface or class `Resolved*` in the symbol solver, there would be an equivalent interface or class `Unresolved*`, e.g.,: `UnresolvedReferenceType`, `UnresolvedPrimitiveType`, and so on, all of which inherit from `UnresolvedType`, basically mirroring the class hierarchy below `ResolvedType`.

Then, the `UnsolvedSymbolException` could be made completely obsolete and would never be thrown anywhere again.

Pro:
- It would make the symbol solver much more robust, because users wouldn't need to worry about unchecked exceptions being thrown all the time.
- It would allow to handle cases such as the above (with the ancestor list) somewhat more elegantly.

Contra:
- It would be a significant amount of work, as it would require rewriting large parts of the symbol solver.

I don't think this is something that can be done "ad hoc", so for now, I completely agree that the solution with the flag is reasonable. However, this is something that we could think about in the long run. If you intend to rewrite the symbol solver at some point, maybe this could go into the rewriting as well. It's just a thought. :-)

ftomassetti:
Would it be simpler to have an implementation of ResolvedType that represents an "unsolved type" and basically throw an exception when calling some of its methods?

Now, the terminology could be an issue because in this case by "ResolvedType" I would mean a type I tried to resolve, or the type I obtained from the resolving process: it could either be a valid response (i.e., a valid type) or something representing the fact the type could not be found.

Alternatively we could renamed ResolvedType in PossiblyResolvedType and derive the two sub interfaces.

Still what I like about exceptions is that I can follow them (especially considering I wrap them and recast them) so that it is possible to find the root cause of an issue. I think this is useful for debugging ("oh, you could not solve this expression because it involves a call to a method on an object of this class which inherit from that other class and we cannot solve that other class" -> this is more useful that knowing just that I could not solve a certain expression)

matozoid:
+1 for the PR itself.

malteskoruppa:
Yes, maybe we should take this discussion elsewhere, as far as I can see everyone agrees on the PR itself. :-)

Anyway, to reply to @ftomassetti :

> Would it be simpler to have an implementation of ResolvedType that represents an "unsolved type" and basically throw an exception when calling some of its methods?
> 
> Now, the terminology could be an issue because in this case by "ResolvedType" I would mean a type I tried to resolve, or the type I obtained from the resolving process: it could either be a valid response (i.e., a valid type) or something representing the fact the type could not be found.

I agree that the terminology would be very unintuitive here. For me "ResolvedType" suggests it was successfully resolved (and not just some result of the resolving process).

> Alternatively we could renamed ResolvedType in PossiblyResolvedType and derive the two sub interfaces.

I like this much more. :-)

> Still what I like about exceptions is that I can follow them (especially considering I wrap them and recast them) so that it is possible to find the root cause of an issue. I think this is useful for debugging ("oh, you could not solve this expression because it involves a call to a method on an object of this class which inherit from that other class and we cannot solve that other class" -> this is more useful that knowing just that I could not solve a certain expression)

You got a point there. How about continuing to throw `UnsolvedSymbolException`s internally as it is now, but never let them reach the user? Instead, `UnsolvedSymbolException`s are caught by the symbol solver internally, and an `UnresolvedType` is generated in this case and returned. The `UnresolvedType` could simply store the `UnsolvedSymbolException` that led to its generation as an internal field.

That way we have the best of both worlds:
* Users don't have to worry about unchecked exceptions.
* Scenarios such as the one with the ancestors can be handled using `UnresolvedType`.
* It's still possible to debug just as easily. As soon as we encounter an `UnresolvedType`, we can have a look at its stored exception to understand what happened.

Again, just a thought... I'm not sure whether this is the best approach. Also, it would still be a lot of work to do this, and I'm not sure if it would be worth it.

Issue #1490, opened 2018-04-02
Cannot edit/remove method comment if it's duplicated in the class


pierremorel-enactor:
I'm trying the following code

```
MethodDeclaration methodDeclaration = ...
// Cannot set a new javadoc comment
Javadoc javadoc = new Javadoc(JavadocDescription.parseText("Change Javadoc"));
methodDeclaration.setJavadocComment("", javadoc);

// Cannot remove Comment
methodDeclaration.removeComment();

// Cannot removeJavaDocComment
methodDeclaration.removeJavaDocComment();
```

on the following class:

```
public class MyClass {

  /**
   * Comment A
   */
  public void oneMethod() {
  }

  /**
   * Comment A
   */
  public void anotherMethod() {
  }

}
```

The issue seems to be that the javadoc comment is exactly the same for both methods.

Stack trace:
```
java.lang.IllegalStateException
	at com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter$Observer.concretePropertyChange(LexicalPreservingPrinter.java:179)
	at com.github.javaparser.ast.observer.PropagatingAstObserver.propertyChange(PropagatingAstObserver.java:62)
	at com.github.javaparser.ast.Node.lambda$notifyPropertyChange$3(Node.java:539)
	at java.util.ArrayList.forEach(ArrayList.java:1249)
	at com.github.javaparser.ast.Node.notifyPropertyChange(Node.java:539)
	at com.github.javaparser.ast.Node.setComment(Node.java:248)
	at com.github.javaparser.ast.nodeTypes.NodeWithJavadoc.setJavadocComment(NodeWithJavadoc.java:71)
	at com.github.javaparser.ast.nodeTypes.NodeWithJavadoc.setJavadocComment(NodeWithJavadoc.java:76)
	at com.enactor.refactor.visitor.ReplaceInheritDocsAnnotationVisitor.visit(ReplaceInheritDocsAnnotationVisitor.java:58)
	at com.enactor.refactor.visitor.ReplaceInheritDocsAnnotationVisitor.visit(ReplaceInheritDocsAnnotationVisitor.java:1)
	at com.github.javaparser.ast.body.MethodDeclaration.accept(MethodDeclaration.java:120)
	at com.github.javaparser.ast.visitor.VoidVisitorAdapter.lambda$visit$29(VoidVisitorAdapter.java:170)
	at java.util.ArrayList.forEach(ArrayList.java:1249)
[...]

PR #1649, opened 2018-06-12
duplicate comment removal fix and tests for issue #1490

ThLeu:
tests for issue #1490 

It seems that this issue is fixed already, the tests work without a problem.

If you do not like my coding style for tests (mainly the duplication of test code and the AAA (Arrange Act Assert) style) please tell me and I will change that 😄 . Although I recommend that style 😛 for several reasons.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
@@ -21,12 +21,10 @@
 
 package com.github.javaparser.ast.comments;
 
-import com.github.javaparser.JavaParser;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
-import com.github.javaparser.ast.expr.Name;
 import com.github.javaparser.javadoc.Javadoc;
 import com.github.javaparser.javadoc.description.JavadocDescription;
 import com.github.javaparser.printer.PrettyPrinterConfiguration;
@@ -35,7 +33,6 @@ import org.junit.Test;
 import java.util.EnumSet;
 
 import static com.github.javaparser.JavaParser.parse;
-import static com.github.javaparser.JavaParser.parseName;
 import static com.github.javaparser.utils.Utils.EOL;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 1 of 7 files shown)


ftomassetti (approved):
+1

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
```diff
@@ -69,4 +77,125 @@ public void unicodeEscapesArePreservedInComments() {
...
+                EOL);
+
+        MethodDeclaration methodDeclaration = cu.findFirst(MethodDeclaration.class).get();
+
+        // Act
+        Javadoc javadoc = new Javadoc(JavadocDescription.parseText("Change Javadoc"));
+        methodDeclaration.setJavadocComment("", javadoc);
+
+        // Assert
+        assertEquals("public class MyClass {" + EOL +
```
matozoid: Take a look at assertEqualsNoEol (or so) - building on unix and windows is painful, and has everything to do with comparing eol's :-)

ThLeu: To be honest, I don't see much of a difference between using EOL and \n everywhere :)
But I guess assertEqualsNoEol gives the freedom to copy pase in Intellij IDEA, is that why you use it?
I will change it if you like.

ThLeu:
Ok, As i guessed, the bug isn't fixed and just happens when using the LexicalPreservingPrinter.
My last commits will fix the issue #1490 for javadocs BUT....
1. we have the same indentation problems here as we had / have in the Difference class of lexical preservation. It keeps too many spaces for now.
2. normal comments are not supported yet (as they were before), but this is a different issue.

I added unit tests that are ignored for now (with the reason why they are ignored).
So this PR could be merged if you like and the other issues can be handled separately.

fixes #1490

ThLeu:
As you can see in the commits I added the Range to the comparison of the TokenTextElements if there is more than one. 
The same problem happened afterwards with the NodeText.findElement method which uses a TextElementMatcher to find the correct element. However the default behaviour always found the first comment of the two, even when changing the second one. I refined that by enabling the combination of different TextElementMatchers with the `and()` Method on the interface and adding a first one `matchByRange()` to TextElement.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
matozoid: Do change it - EOL is the *platform specific* end of line character, and `assertEqualsNoEol` replaces each and every eol character to \n so you can safely compare to strings with \n in it. These tests fail on Windows.

ThLeu: I changed it now. Maybe I misunderstand something but I can safely say that the test was running on Windows before, as I programmed it on Windows.
Also `Utils.EOL` already uses `System.getProperty("line.separator")` which works on all platforms and `assertEqualsNoEol`, as I understand it, basically replaces the "\n" back to `Utils.EOL`, which is why I didn't really see a difference between before and after, maybe I am missing something.
Anyway, I changed it and I didn't really want to start a big discussion :)

matozoid:
Thanks @ThLeu !

PR #1656, opened 2018-06-17
Issue1441: NodeWithMembers

kotari4u:
Added below methods
addFieldWithInitializer(Class<?> type, String name, Expression initializer, Modifier... modifiers)
addFieldWithInitializer(String type, String name, Expression initializer, Modifier... modifiers)
addPrivateField(Type type, String name)
addPublicField(Type type, String name)
addProtectedField(Type type, String name)


ftomassetti:
To me it seems reasonable @matozoid  ?

matozoid:
It's unit tested and everything, cool!

Thanks @kotari4u ! I'll merge this when we're done with the discussion in the issue.

PR #1657, opened 2018-06-18
Appveyor build fix

matozoid:


matozoid:
So it was the insane memory settings we used that broke everything :-D

ftomassetti:
Interesting! I tend to be generous with memory :D

Issue #1668, opened 2018-06-26
Resolving VariableDeclaration throws IllegalStateException


daanschipper:
When trying to resolving the stringArray in the following example, JSS throws an `java.lang.IllegalStateException: com.github.javaparser.ast.expr.VariableDeclarationExpr`

Example:
```
public class X {
  public static void main(String[] args) {
    String s = "a,b,c,d,e";
    String[] stringArray = s.split(',');
  }
}
```

Stacktrace:
```
java.lang.IllegalStateException: com.github.javaparser.ast.expr.VariableDeclarationExpr

	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserFieldDeclaration.<init>(JavaParserFieldDeclaration.java:53)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:101)
	at com.github.javaparser.ast.body.VariableDeclarator.resolve(VariableDeclarator.java:279)
```

PR #1669, opened 2018-06-26
[TESTCASE] Resolving VariableDeclarator

daanschipper:
Providing test case for #1668


javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
```diff
@@ -0,0 +1,62 @@
...
+        assertTrue(parseResult.isSuccessful());
+        assertTrue(parseResult.getResult().isPresent());
+
+        CompilationUnit compilationUnit = parseResult.getResult().get();
+        Optional<VariableDeclarator> stringArray = compilationUnit.findFirst(VariableDeclarator.class, variableDeclarator ->
+                variableDeclarator.getNameAsString()
+                        .equals("stringArray"));
+        assertTrue(stringArray.isPresent());
+        VariableDeclarator variableDeclarator = stringArray.get();
+        variableDeclarator.getParentNode().ifPresent(parent -> {
```
ftomassetti: Don't we expect it to be present? Should we assert it is present?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
```diff
@@ -0,0 +1,62 @@
...
+            if (parent instanceof VariableDeclarationExpr) {
+                VariableDeclarationExpr variableDeclarationExpr = (VariableDeclarationExpr) parent;
+                ResolvedType resolvedType = variableDeclarationExpr.calculateResolvedType();
+                assertEquals("java.lang.String[]", resolvedType.describe());
+            }
+        });
+        ResolvedFieldDeclaration resolve = variableDeclarator.resolve();
+        assertEquals("java.lang.String[]", resolve.declaringType().getQualifiedName());
```
ftomassetti: mmm... can the qualified name corresponds to an array type?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
```diff
@@ -0,0 +1,62 @@
...
+    @Before
+    public void setUp() {
+        TypeSolver typeSolver = new ReflectionTypeSolver();
+        ParserConfiguration config = new ParserConfiguration();
+        config.setSymbolResolver(new JavaSymbolSolver(typeSolver));
+        javaParser = new JavaParser(config);
+    }
+
+    @Test
+    public void test() {
```
ftomassetti: I think we could use a different name here

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
matozoid: A qualified name is identifier.identifier.identifier etc. How should an array type be represented?

ftomassetti: Anyway the declaring type of a field cannot be an array type, it must be a class, an interface or an enum

matozoid: So java.lang.String in this case?

ftomassetti: In this case it does not make much sense becose `stringArray` is not a field... so I would not expect this to work.

ftomassetti: But in this case:
```
class A {
   int foo;
}
```

The declaring type of `foo` is `A`.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
daanschipper: It should indeed be present, imo the rest of the test shouldn't even continue if for some case the VariableDeclarator has not been found. But on the other hand this is not what the test is for, so this assert could be dropped

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
daanschipper: The reason I used an array type as qualified name is because that is what is returned by `variableDeclarationExpr.calculateResolvedType().describe`. I don't know what it should return, but I had to fill in something.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
matozoid: Oh, if it's a test, just `get()` that thing.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1668.java
matozoid: @ftomassetti @daanschipper can you come to a resolution on this?

ftomassetti: I would expect this to fail. If it does not let leave it as it is

matozoid: Double check: you would expect this test to fail, so the whole testcase and the connected issue are to be closed?

matozoid:
@daanschipper @ftomassetti - this seems to be passing now? Is it fixed?

daanschipper:
That is interesting, I'll take a look

daanschipper:
No, this is still failing, don't know what why travis is showing it as passing. I have no time or intention to solve this issue right away, I'll close the pr for now.

ftomassetti:
I will reopen it as the test is still useful and I could use it later to fix the issue

Issue #1675, opened 2018-07-05
UnsolvedSymbolException when trying to resolve FieldDeclaration that is an inner class of an extending type of an extending type


daanschipper:
When trying to resolve the following `FieldDeclaration` I'm getting an UnsolvedSymbolException

```
public class X extends Y {
   private Foo foo;
}

public class Y extends Z { }

public class Z {
    public static class Foo {}
}
```

`JavaParserTypeDeclarationAdapter.solveType()` only looks at the internal classes of its direct ancestor, but fails to do this iteratively. So the internal classes of `Y` are checked for a `TypeDeclaration` with name `Foo`, but of course this isn't found. However, the internal types of ancestor of `Y` aren't evaluated at all.

The stacktrace:
```
Exception in thread "main" UnsolvedSymbolException{context='unknown', name='LengthConfig'}
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:438)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convert(JavaParserFacade.java:496)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convert(JavaParserFacade.java:492)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserFieldDeclaration.getType(JavaParserFieldDeclaration.java:72)
```

jlerbsc:
This issue seems to be resolved.

    public class Issue1675Test extends AbstractSymbolResolutionTest {

    @Test
    public void test() {
        final Path testRoot = adaptPath("src/test/resources/issue1675");
        TypeSolver reflectionTypeSolver = new ReflectionTypeSolver();
        JavaParserTypeSolver javaParserTypeSolver = new JavaParserTypeSolver(testRoot);
        CombinedTypeSolver combinedTypeSolver = new CombinedTypeSolver(reflectionTypeSolver, javaParserTypeSolver);
        ParserConfiguration configuration = new ParserConfiguration()
                .setSymbolResolver(new JavaSymbolSolver(combinedTypeSolver));
        StaticJavaParser.setConfiguration(configuration);

        String src = "public class X extends Y {\n" +
                "   private Foo foo;\n" +
                "}";

        CompilationUnit cu = StaticJavaParser.parse(src);

        List<FieldDeclaration> fds = cu.findAll(FieldDeclaration.class);
        fds.forEach(fd-> {
[...]

PR #1676, opened 2018-07-05
Recursive checking ancestors for TypeDeclaration with provided name

daanschipper:
As stated in #1675, only the direct ancestors internal types are checked, but not its ancestors. This aims to resolve that problem.

I was unable to make the method return an `Optional` instead of null since its using a recursive call within a for loop.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
@@ -68,19 +68,33 @@ public class JavaParserTypeDeclarationAdapter {
         }
 
         // Look into extended classes and implemented interfaces
-        for (ResolvedReferenceType ancestor : this.typeDeclaration.getAncestors()) {
-        	try {
-	            for (ResolvedTypeDeclaration internalTypeDeclaration : ancestor.getTypeDeclaration().internalTypes()) {
-	                if (internalTypeDeclaration.getName().equals(name)) {
-	                    return SymbolReference.solved(internalTypeDeclaration);
-	                }
-	            }
-        	} catch (UnsupportedOperationException e) {
-	            // just continue using the next ancestor
+        ResolvedTypeDeclaration type = checkAncestorsForType(name, this.typeDeclaration);
+        return ((type != null) ? SymbolReference.solved(type) : context.getParent().solveType(name, typeSolver));
+    }
+
+    /**
+     * Recursively checks the ancestors of the {@param declaration} if an internal type is declared with a name equal
+     * to {@param name}.
+     * @return A ResolvedTypeDeclaration matching the {@param name}, null otherwise
+     */
+    private ResolvedTypeDeclaration checkAncestorsForType(String name, ResolvedReferenceTypeDeclaration declaration) {
+        for (ResolvedReferenceType ancestor : declaration.getAncestors()) {
+            try {
+                for (ResolvedTypeDeclaration internalTypeDeclaration : ancestor.getTypeDeclaration().internalTypes()) {
+                    if (internalTypeDeclaration.getName().equals(name)) {
+                        return internalTypeDeclaration;
+                    }
+                }
+                // check recursively the ancestors of this ancestor
+                ResolvedTypeDeclaration ancestorDeclaration = checkAncestorsForType(name, ancestor.getTypeDeclaration());
+                if (ancestorDeclaration != null) {
+                    return ancestorDeclaration;
+                }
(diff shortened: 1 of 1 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
```diff
@@ -68,19 +68,33 @@ public JavaParserTypeDeclarationAdapter(com.github.javaparser.ast.body.TypeDecla
...
+        return ((type != null) ? SymbolReference.solved(type) : context.getParent().solveType(name, typeSolver));
+    }
+
+    /**
+     * Recursively checks the ancestors of the {@param declaration} if an internal type is declared with a name equal
+     * to {@param name}.
+     * @return A ResolvedTypeDeclaration matching the {@param name}, null otherwise
+     */
+    private ResolvedTypeDeclaration checkAncestorsForType(String name, ResolvedReferenceTypeDeclaration declaration) {
+        for (ResolvedReferenceType ancestor : declaration.getAncestors()) {
```
ftomassetti, line 81: Should we consider the visibility of internal types?

ftomassetti:
A test would be a great addition to this PR :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
daanschipper: That check wasn't done before.
But this could be added to as method to `ResolvedReferenceTypeDeclaration` such as `getAncestors(boolean visibility)` and that method would be called here.

daanschipper:
I was lazy regarding the tests, I packaged the source and tested it in my project and it resolved the problem. 
Regarding tests, are there some util functions present to load some files and inject them with an already configured typesolver?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
```diff
@@ -68,19 +68,33 @@ public JavaParserTypeDeclarationAdapter(com.github.javaparser.ast.body.TypeDecla
...
+                        return internalTypeDeclaration;
+                    }
+                }
+                // check recursively the ancestors of this ancestor
+                ResolvedTypeDeclaration ancestorDeclaration = checkAncestorsForType(name, ancestor.getTypeDeclaration());
+                if (ancestorDeclaration != null) {
+                    return ancestorDeclaration;
+                }
+            } catch (UnsupportedOperationException e) {
+                // just continue using the next ancestor
```
matozoid, line 94: Well, that seems to need at least a big //FIXME

daanschipper: The try catch was already present in the previous version, I just copied this part

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
ftomassetti: well, maybe because the internal types declared in the current type are all visible, while only the non private are visible among the inherited.
Or alternatively because I was sloppy writing the original code :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
ftomassetti: I would blame the original author (who is probably me) and see if we can improve it

matozoid:
Oh, @daanschipper, by the way:
> I was unable to make the method return an Optional instead of null since its using a recursive call within a for loop.

if this is not part of the public API it's not a big deal. Otherwise, if you can't figure out a way to do it, open an issue for it. The guarantee that JP & JSS never return null to the user is a very important one. Forget it in one place and users will be null-checking all over the place again.

matozoid:
Okay, @daanschipper - is this ready?

daanschipper:
I would think this is ready. However it is not yet perfect, but I don't really have the time atm to improve it further.

matozoid:
Then thank you very much, @daanschipper :-)

PR #1678, opened 2018-07-06
Fixing signatures of ReflectionConstructorDeclaration

malteskoruppa:
This PR fixes the following unreported bug.

Consider this code:

```
import java.io.File;

class MethodLikeSignaturesTest {

	public void foo() {
		File file = new File("somepath");
	}
}
```

Now when you try to resolve the constructor declaration that corresponds to the object creation expression `new File("somepath")` using a `ReflectionTypeSolver`, you get a `ReflectionConstructorDeclaration`. So far so good. If you now call `getQualifiedSignature` on this `ReflectionConstructorDeclaration`, you get the following String:

`java.io.File.java.io.File(java.lang.String)
`

I think we all agree that this is incorrect. Essentially, the problem is that the method `ResolvedDeclaration#getName()` is defined such that it should return the *simple* name of a constructor or method, but in the case of a `ReflectionConstructorDeclaration`, it returns the *qualified* name instead. This then leads to further bugs such as the above.

I fixed `ReflectionConstructorDeclaration#getName()` to return the simple name, as it should be. Therefore, in the above example, `getQualifiedSignature()` now returns:

`java.io.File.File(java.lang.String)
`

...as one would expect. :) (essentially the same as you would use in a Javadoc comment: `java.io.File#File(java.lang.String)`, except that JSS uses a dot (`.`) where Javadoc comments use a sharp (`#`)).

This was a quickie, hoping that this will still make it into this week-end's release so I can use it on Monday... :wink:

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionConstructorDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionConstructorDeclaration.java
@@ -69,7 +69,7 @@ public class ReflectionConstructorDeclaration implements ResolvedConstructorDecl
 
     @Override
     public String getName() {
-        return constructor.getName();
+        return constructor.getDeclaringClass().getSimpleName();
     }
 
     @Override
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodLikeSignaturesTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodLikeSignaturesTest.java
@@ -1,0 +1,51 @@
+package com.github.javaparser.symbolsolver.resolution;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import com.github.javaparser.resolution.declarations.ResolvedConstructorDeclaration;
+import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
+import com.github.javaparser.symbolsolver.javaparser.Navigator;
+import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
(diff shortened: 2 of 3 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionConstructorDeclaration.java
```diff
@@ -69,7 +69,7 @@ public ResolvedParameterDeclaration getParam(int i) {
 
     @Override
     public String getName() {
-        return constructor.getName();
+        return constructor.getDeclaringClass().getSimpleName();
```
ftomassetti, line 72: Right

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodLikeSignaturesTest.java
```diff
@@ -0,0 +1,51 @@
...
+        ObjectCreationExpr objectCreationExpr = method.getBody().get().getStatements().get(0)
+                                                        .asExpressionStmt().getExpression().asVariableDeclarationExpr()
+                                                        .getVariable(0).getInitializer().get().asObjectCreationExpr();
+
+        ResolvedConstructorDeclaration resolvedConstructorDeclaration =
+                JavaParserFacade.get(new ReflectionTypeSolver()).solve(objectCreationExpr).getCorrespondingDeclaration();
+
+        assertEquals("File", resolvedConstructorDeclaration.getName());
+        assertEquals("File(java.lang.String)", resolvedConstructorDeclaration.getSignature());
+        assertEquals("java.io.File.File(java.lang.String)", resolvedConstructorDeclaration.getQualifiedSignature());
```
ftomassetti, line 33: And the test, great!

ftomassetti (approved):
To me it looks good. At a later time we could think about introducing the sharp instead of the dot as the separator

Thanks for your work

malteskoruppa:
I completely agree about the sharp instead of the dot, but that's a story for another time... :)

Thank you for the super fast review and have a great week-end!

ftomassetti:
AppVeyor is failing but that seems not a problem. I will go ahead and merge it

matozoid:
@malteskoruppa - the issue tracker is our memory, stuff the sharp issue on there :-)

malteskoruppa:
@matozoid I will remember in the future. Should I still open (& close) an issue for this one?

matozoid:
@malteskoruppa oh no, the PR is enough. I meant an issue about the `#` or the `.`. If we don't create an issue it will be forgotten for who knows how long :-)

malteskoruppa:
Oh, ok, I completely misunderstood, sorry! :-)

I'm going to open an issue about the sharp sign.

PR #1684, opened 2018-07-08
Addressing test failure due to package name/class name conflict

ftomassetti:
I have this failure on my mac due to the fact that we are trying to load classes to figure out which parts of a qualified name corresponds to classes. At some point we find a package name (`com/github/javaparser/printer/concretesyntaxmodel`) which corresponds to a class name but for the case ('com/github/javaparser/printer/ConcreteSyntaxModel`). This on mac causes a NoClassDefFound error to be thrown (`java.lang.NoClassDefFoundError: com/github/javaparser/printer/ConcreteSyntaxModel (wrong name: com/github/javaparser/printer/concretesyntaxmodel)`). I think this should be ignored.


matozoid:
I'm taking your word for it.

PR #1693, opened 2018-07-15
add FieldAccessExpr isInternal/isTopLevel

ftomassetti:
Add these methods to distinguish between different types of field accesses. Typically we are interested only in top level field accesses, because others could refer to the path through packages or through internal types. For example, we should resolve only top level FieldAccessExpr unless we know what we are doing.


matozoid:
"Sorry, this diff is temporarily unavailable due to heavy server load."

Wow... I blame Microsoft.

PR #1696, opened 2018-07-15
add possibility to resolve array.length 

### LINKED ISSUE #1695: Create a virtual ValueDeclaration for array.length
Currently we can resolve a FieldAccessExpr and get a ValueDeclaration, however that does not work for array.length because there is no actual declaration to point to.

ftomassetti:
Fix #1695


matozoid:
Looks fine to me!

PR #1715, opened 2018-07-29
Implement Naming logic - Last part

ftomassetti:
This is now ready to be merged


matozoid:
I'm afraid you'll need someone else to check this code.

matozoid:
@ftomassetti if you're splitting up this PR, will you close this one?

ftomassetti:
I would keep it for now as a reminder. Eventually this should become empty as all the differences would end up in master and at that point I would close it

ftomassetti:
@matozoid I removed the "DRAFT" from the title because I think this is now ready to be merged

ftomassetti:
Since I derived other PRs from this one now this one is small enough (just 5 files changed: 3 are tests)

matozoid:
Well, @daanschipper was happy to review last time, so let's see if he does it again :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/naming/NameLogic.java
```diff
@@ -240,8 +276,683 @@ public static NameRole classifyRole(Node name) {
...
+
+        // * If the name to the left of the "." is reclassified as an ExpressionName, then this AmbiguousName is
+        //   reclassified as an ExpressionName. A later step determines whether or not a member with the name Identifier
+        //   actually exists.
+
+        if (leftNameCategory == NameCategory.EXPRESSION_NAME) {
+            return NameCategory.EXPRESSION_NAME;
+        }
+
+        throw new UnsupportedOperationException();
```
daanschipper: What is left unsupported? Could you specify why this exception is thrown?

daanschipper (approved):
I'm not familiar with the exact language specifications, so it's hard to review this. Code wise it looks all good.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/naming/NameLogic.java
ftomassetti: All cases should be covered, however if the language evolves and there are new usages of names then this exception could be thrown

daanschipper: The message of the exception should describe that reason in my opinion. Now the user will receive a random exception

ftomassetti:
Added explanation to all UnsupportedOperationException. I am going to merge as soon as tests complete

ftomassetti:
The tests passed on the server but for some reason it did not update here. Merging anyway

PR #1717, opened 2018-07-30
Delegate solving type to the TypeSolver if a NotFoundException is thrown

### LINKED ISSUE #1716: Resolving reference outside current class
When trying to get for instance a superclass that resides in another 'source' (read not same class pool of the current typesolver), a `NotFoundException` is thrown and forwarded as `RuntimeException`.

```
java.lang.RuntimeException: javassist.NotFoundException: javax.servlet.http.HttpServlet
    at com.github.javaparser.symbolsolver.javassistmodel.JavassistClassDeclaration.getSuperClass(JavassistClassDeclaration.java:305) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javassistmodel.JavassistClassDeclaration.getAncestors(JavassistClassDeclaration.java:177) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.checkAncestorsForType(JavaParserTypeDeclarationAdapter.java:81) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.checkAncestorsForType(JavaParserTypeDeclarationAdapter.java:89) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveType(JavaParserTypeDeclarationAdapter.java:71) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveType(ClassOrInterfaceDeclarationContext.java:90) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:436) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:413) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.symbolsolver.JavaSymbolSolver.toResolvedType(JavaSymbolSolver.java:191) ~[javaparser-symbol-solver-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:294) ~[javaparser-core-3.6.13-SNAPSHOT.jar:?]
    at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:60) ~[javaparser-core-3.6.13-SNAPSHOT.jar:?]
```

daanschipper:
Fixes #1716 

When a `NotFoundException` is thrown, try to solve it with the typeSolver instead of forwarding the exception. Only when it's also not found by the typeSolver, then throw an exception.


daanschipper:
Although this doesn't throw an exception while using `JavassistClassDeclaration.getSuperClass()` anymore, the `JavassistClassDeclaration.solveMethod` is now throwing an `UnsolvedException`

@ftomassetti, I see that `solveMethod` is deprecated, which method should be used instead? I see your comment in `MethodResolutionLogic` about replacing `TypeDeclaration.solveMethod`, but that's all there is.

The stacktrace:
```
com.github.javaparser.resolution.UnsolvedSymbolException: Unsolved symbol : java.lang.String[]
	at com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver.solveType(CombinedTypeSolver.java:76) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.lambda$objectTypeArgumentToType$3(JavassistUtils.java:185) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at java.util.Optional.orElseGet(Optional.java:267) ~[?:1.8.0_131]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.objectTypeArgumentToType(JavassistUtils.java:184) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.typeArgumentToType(JavassistUtils.java:206) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.lambda$signatureTypeToType$2(JavassistUtils.java:138) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193) ~[?:1.8.0_131]
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948) ~[?:1.8.0_131]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481) ~[?:1.8.0_131]
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471) ~[?:1.8.0_131]
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708) ~[?:1.8.0_131]
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:1.8.0_131]
	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499) ~[?:1.8.0_131]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.signatureTypeToType(JavassistUtils.java:138) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistMethodDeclaration.getReturnType(JavassistMethodDeclaration.java:104) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.resolution.MethodUsage.<init>(MethodUsage.java:56) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.model.typesystem.ReferenceTypeImpl.getDeclaredMethods(ReferenceTypeImpl.java:149) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.resolution.types.ResolvedReferenceType.getAllMethods(ResolvedReferenceType.java:375) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.resolution.types.ResolvedReferenceType.lambda$getAllMethods$4(ResolvedReferenceType.java:377) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at java.util.ArrayList.forEach(ArrayList.java:1249) ~[?:1.8.0_131]
	at com.github.javaparser.resolution.types.ResolvedReferenceType.getAllMethods(ResolvedReferenceType.java:376) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistClassDeclaration.solveMethod(JavassistClassDeclaration.java:201) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.resolution.MethodResolutionLogic.solveMethodInType(MethodResolutionLogic.java:632) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveMethod(JavaParserTypeDeclarationAdapter.java:111) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveMethod(ClassOrInterfaceDeclarationContext.java:95) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.core.resolution.Context.solveMethodAsUsage(Context.java:89) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ContextHelper.solveMethodAsUsage(ContextHelper.java:63) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:171) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:393) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:113) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:552) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:276) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:46) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:113) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:416) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:313) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:224) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:243) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:128) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:129) ~[javaparser-symbol-solver-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:311) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:56) ~[javaparser-core-3.6.16-SNAPSHOT.jar:?]
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -202,7 +198,7 @@ private String getSuperclassFQN() {
                 }
             }
         } catch (NotFoundException e) {
-            throw new RuntimeException(e);
+            candidates.addAll(solveSuperClass().getAllMethods());
```
ftomassetti: mmm I am not convinced by this. Why is the "NotFoundException" thrown? I do not think we should ignore it.

Regarding delegating to the super class I would do that later, after the catch clause and do something similiar to what we do in ReflectionClassDeclaration:

```
        if (getSuperClass() != null) {
            ResolvedClassDeclaration superClass = (ResolvedClassDeclaration) getSuperClass().getTypeDeclaration();
            SymbolReference<ResolvedMethodDeclaration> ref = MethodResolutionLogic.solveMethodInType(superClass, name, argumentsTypes, staticOnly, typeSolver);
            if (ref.isSolved()) {
            }
        }
        for (ResolvedReferenceType interfaceDeclaration : getInterfaces()) {
            SymbolReference<ResolvedMethodDeclaration> ref = MethodResolutionLogic.solveMethodInType(interfaceDeclaration.getTypeDeclaration(), name, argumentsTypes, staticOnly, typeSolver);
            if (ref.isSolved()) {
            }
        }
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -302,12 +298,24 @@ public ResolvedReferenceType getSuperClass() {
             SignatureAttribute.ClassSignature classSignature = SignatureAttribute.toClassSignature(ctClass.getGenericSignature());
             return JavassistUtils.signatureTypeToType(classSignature.getSuperClass(), typeSolver, this).asReferenceType();
         } catch (NotFoundException e) {
-            throw new RuntimeException(e);
+            return solveSuperClass();
```
ftomassetti: mmm, still not sure why we are getting this NotFoundException

ftomassetti:
To me it seems that we should not get the NotFoundException in the first place and we are hiding the real issue (but I could be missing something)

daanschipper:
The structure that I have is as follows, the source of class A and B are loaded in with two different `JarTypeSolver`s and have a relation of A extends B. When asking for the superclass, you'll end up with asking `JavassistClassDeclaration` for its superclass, which is all loaded in with javassist. But this just can't find the class declared in B, so it throws a `NotFoundException`

daanschipper:
I think the `NotFoundException` is a valid exception to be thrown by the method `CtClass.getSuperClass`. It simply can't find it's super class since it is in another resource (another .jar).

I'm not sure on how to correctly deal with the exception. The following piece of code just ask it's parent 'Do you have a method with this name and these argumentTypes'? It _needs_ that information to continue and return a `SymbolReference` of that method. At least in the way the flow currently is written.
```
try {
    CtClass superClass = ctClass.getSuperclass();
    if (superClass != null) {
        SymbolReference<ResolvedMethodDeclaration> ref = new JavassistClassDeclaration(superClass, typeSolver).solveMethod(name, argumentsTypes, staticOnly);
        if (ref.isSolved()) {
            candidates.add(ref.getCorrespondingDeclaration());
        }
    }
} catch (NotFoundException e) {
    candidates.addAll(solveSuperClass().getAllMethods());
}
```

I think there are two options
- On a `NotFoundException`, delegate the resolving to the typeSolver. This probably means that the super class resides in another source. 
- Rewrite the flow. It is currently the task of `JavassistClassDeclaration.solveMethod()` to collect all the candidate methods with a similar signature, even that of its super class.

ftomassetti:
> I think the NotFoundException is a valid exception to be thrown by the method CtClass.getSuperClass. It simply can't find it's super class since it is in another resource (another .jar).

well, but JarTypeSolver share the same classpool behind the scenes exactly for that reason, so if both jars are added to the typesolver this should not happen I think

In any case I would refactor the code in a way that we do not expect NotFoundException to be thrown. Perhaps we can ask directly the TypeSolver to solve it for us using the super class qualified name. Something like `typeSolver.solveType(ctClass.getSuperClass().getName())`

daanschipper:
> well, but JarTypeSolver share the same classpool behind the scenes exactly for that reason, so if both jars are added to the typesolver this should not happen I think

This does happen if I'm right, so that does mean that there is an underlying problem. Let me check this.

Either way, I'll refactor the code to not expect a `NotFoundException`.

matozoid:
Hey @daanschipper - how is this one going?

daanschipper:
Sorry took a bit longer than expected..

> well, but JarTypeSolver share the same classpool behind the scenes exactly for that reason, so if both jars are added to the typesolver this should not happen I think

I don't understand how this is happening. In my setup I'm adding the individual jar files with 
```
ParseUtils.walkFiles(Paths.get(root.toString(), "/lib"), ".jar").forEach(jar -> {
    try {
        combinedTypeSolver.add(new JarTypeSolver(jar));
    } catch (IOException e) {
        logger.warn("Unable to add jar " + jar);
    }
});
```
that creates a new `JarTypeSolver` with each its own `ClassPool`.
```
   public JarTypeSolver(String pathToJar) throws IOException {
        this.classpathElements = new HashMap();
        this.classPool = new ClassPool(false);
        this.addPathToJar(pathToJar);
    }
```
`addPathToJar()` basically indexes all entries from the jar file.

As far as I'm seeing there is no shared ClassPool for the JarTypeSolvers, each just has its own. The pools are not linked together at all.

So when asking in `JavassistClassDeclaration.getSuperClass` for `this.ctClass.getSuperclass()`, it checks for 
```
public CtClass getSuperclass() throws NotFoundException {
        String supername = this.getClassFile2().getSuperclass();
        return supername == null ? null : this.classPool.get(supername);
    }
```
it ends up asking the classPool for the name of the superClass, but alas it can't find it and returns a NotFoundException.


So again two questions for you @ftomassetti:
- Did you intend to link all classpools together somehow and that failed? I see a static instance defined in `JarTypeSolver`, but apparently it doesn't work as intended.
- Did I do something wrong with the way I set the `CombinedTypeSolver` up?

ftomassetti:
My mistake, I was looking at getJarTypeSolver which creates a global instance and add all the jars to the same classpool.

Let me see if this can be solved by delegating in some place to the parent typesolver or it requires to have one single classpool knowing about all existing jars.

daanschipper:
I would suggest using a single instance of `JarTypeSolver` and adding all jars to the same classpool. That way the entries in the jar are properly indexed instead of iterating over every jar and asking for a name.

But you could indeed delegate it to the parent typesolver, that was why I initially suggested this function.

daanschipper:
So @ftomassetti, what's the plan for the JarTypeSolver, shall I make it so it uses only one classpool?

daanschipper:
Adding the jars with `combinedTypeSolver.add(JarTypeSolver.getJarTypeSolver(jar.toString()));` resolves the issue completely

ftomassetti:
Yes, but what if the dependecy is not to a class contained in another Jar but to a class provided in some other ways? For example through reflection, through source code, through the AAR TypeSolver or other types of TypeSolver we can have in the future? This is the only doubt I have: using the `JarTypeSolver.getJarTypeSolver` is a nice workaround for now because of the common classpool trick but I am afraid this could bite us later

daanschipper:
I propose the following to get the superclass within `JavassistClassDeclaration`. 
Get the superclass with `ctClass.getSuperclass()` which has 3 possible outcomes
  1. returns the class as found in the classpool and redirect solveMethod call
  2. returns null, the class has no superclass, just continue
  3. nfe, there exists a superclass but it is not found in the classpool. Ask the typesolver to try to solve the superclass' fqn. If this is successful, redirect solveMethod call. If not successful, throw runtime exception.

ftomassetti:
The whole issue is that Javassist try to resolve the super class while ideally it should return just the name and let us handle that, so that we can solve it at an higher level, possibly involving stuff in other jars or in stuff that is not a jar at all. To do this I would use ASM which can gives us the superclass name. It would also work if it it was not for this weird error I am getting:
```
[INFO] --- maven-enforcer-plugin:3.0.0-M1:enforce (enforce-versions) @ javaparser-symbol-solver-core ---
[INFO] Restricted to JDK 1.8 yet org.ow2.asm:asm:jar:6.2.1:compile contains module-info.class targeted to JDK 1.9
[WARNING] Rule 1: org.apache.maven.plugins.enforcer.EnforceBytecodeVersion failed with message:
Found Banned Dependency: org.ow2.asm:asm:jar:6.2.1
Use 'mvn dependency:tree' to locate the source of the banned dependencies.
```

daanschipper:
I used the methods `getSuperclassFQN` and `getInterfaceFQNs` to request the fqns and delegate that to the typesolver. 

I noticed that `JavassistInterfaceDeclaration` follows the exact same logic. Can I move this function to 
a new abstract class `JavassistTypeDeclaration`, let `JavassistInterfaceDeclaration` and `JavassistClassDeclaration` extend that class and put that function in there?

ftomassetti:
Right! I could not find anymore that method. That makes sense.  

You cannot add a common ancestor as JavassistClassDeclaration extends AbstractClassDeclaration. Perhaps some utility class with package-level visibility...

ftomassetti:
...oh and also JavassistEnumDeclaration would need the same approach

ftomassetti:
I could later add a test with two jars I have built

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -187,33 +183,26 @@ private String getSuperclassFQN() {
...
-            if (superClass != null) {
-                SymbolReference<ResolvedMethodDeclaration> ref = new JavassistClassDeclaration(superClass, typeSolver).solveMethod(name, argumentsTypes, staticOnly);
-                if (ref.isSolved()) {
-                    candidates.add(ref.getCorrespondingDeclaration());
-                }
-            }
-        } catch (NotFoundException e) {
-            throw new RuntimeException(e);
+        // add the method declaration of the superclass to the candidates, if present
+        SymbolReference<ResolvedMethodDeclaration> superClassMethodRef = MethodResolutionLogic.solveMethodInFQN
```
ftomassetti: Should we check if the method is null?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
```diff
@@ -637,6 +637,16 @@ private static boolean areOverride(MethodUsage winningCandidate, MethodUsage oth
         throw new UnsupportedOperationException(typeDeclaration.getClass().getCanonicalName());
     }
 
+    public static SymbolReference<ResolvedMethodDeclaration> solveMethodInFQN(String fqn, String name,
+                                                                             List<ResolvedType> argumentsTypes, boolean staticOnly, TypeSolver typeSolver) {
+        SymbolReference<ResolvedReferenceTypeDeclaration> typeRef = typeSolver.tryToSolveType(fqn);
+        if (typeRef.isSolved()) {
+            return solveMethodInType(typeRef.getCorrespondingDeclaration(), name, argumentsTypes,
+                    staticOnly, typeSolver);
+        }
```
ftomassetti, line 646: I would throw an exception if the type is not solved

ftomassetti (commented):
I approve the approach. However I think we should also fix getAncestors(), getSuperclass() and getInterfaces(). This could be done as part of a different PR though, as you wish

ftomassetti:
I fixed getSuperclass & getInterfaces and added a test in #1729

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
daanschipper: The method can't be null right? Either we find it or we don't, and we wrap that in a symbol reference

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
daanschipper: Why would you throw an exception and not an unsolved symbol reference?

ftomassetti: Well, because we have not the possibility to check it properly: we are not saying that the method could not be found, we are saying that we are unable to check it. Or worse, we could solve it in the wrong way. For example, if a given method can be solved in a superclass and in an interface but we cannot find the superclass, so we solve it in an interface that would lead to the wrong result.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
ftomassetti: Yes, I am not sure what I was implying here... let's disregard my comment. Sorry about that

daanschipper:
I got one final failing test, `AnalyseJavaSymbolSolver060Test.parseCoreJavassistmodel()` is failing on the line `parse("java-symbol-solver-core", "com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeParameter");`

It tries to resolve `wrapped.getClassBound().toString()`, so this ends up in `JavassistClassDeclaration.solveMethod()`. The line that is failing is `SymbolReference<ResolvedMethodDeclaration> superClassMethodRef = MethodResolutionLogic.solveMethodInFQN(superclassFQN, name, argumentsTypes, staticOnly, typeSolver);` because: 
- `typeSolver.solveType("javassist.bytecode.SignatureAttribute$Type")` returns an `UnsolvedSymbolException`, returning an empty SymbolReference, thus stopping the recursive call up the chain and not finding the actual method
- whereas the old code (`new JavassistClassDeclaration(ctClass.getSuperclass(), typeSolver).solveMethod(name, argumentsTypes, staticOnly);`) successfully returns a SymbolReference

So what I did is calling the functions `getSuperClass` and `getInterfaces`, which are properly implemented in #1729.

ftomassetti:
Ok, so you think that we should merge first #1729, so that later we can restart tests on this PR and they will pass?

ftomassetti:
Merged, tests should now run

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
daanschipper: This method is not used anymore, should I still keep it around and add the exception? Or just delete it entirely?

ftomassetti: Let's delete it. Thank you for your patience and your work on this one!

PR #1719, opened 2018-07-31
Use Name only in modules, no type

### LINKED ISSUE #1710: Uses and Provides should have a Name and not a type
According to [JLS 7.7.1](https://docs.oracle.com/javase/specs/jls/se10/html/jls-7.html#jls-7.7.1) the uses _TypeName_:

```
uses TypeName ; 
provides TypeName with TypeName {, TypeName} ;
```

a `TypeName` is defined as:
```
TypeName:
TypeIdentifier 
PackageOrTypeName . TypeIdentifier
PackageOrTypeName:
Identifier 
PackageOrTypeName . Identifier
```

We instead are using nodes of type `Type` which can be very different things (arrays, wildcards, etc.)

matozoid:
Fixes #1710. Sadly breaks the API, but it does fix a bug.

Please take note, @gunnarmorling :-)


ftomassetti:
Looks good to me!

Issue #1709, opened 2018-07-23
Improvements on names


ftomassetti:
I think we could consider some use cases regarding names in a way that does not require API changes in V3 and could pave the way to changes to be performed in V4.

These use cases could be immediately useful and in the longer run be needed to implement resolution of generic methods according to the specifications. Some changes in how the JLS describes names could be also due to the introduction of module names.

Basically I would say we could benefit from knowing:

* all names in the AST: Currently we have SimpleName and Name as two kinds of nodes that are unrelated. In V4 we could have a common ancestor. In V3 we could introduce a common interface of just having an external utility: `boolean NodeLogic.isName(Node)`
* names can be used in two ways: in declarations to define new entities, in references to point to existing entities. It could be useful to have a method to distinguish between declarations and references. For example to find all declarations or all references in an AST
* names used as references can refer to different things: packages, modules, types, expressions etc. What they can refer to in certain cases can be determined by their syntactical role. In a few cases they can be ambiguous and to solve the ambiguity we need to look at available symbols in a certain context. There is ch. 6 of the JLS that mainly discuss about that. I would add a method like `NodeCategory NodeLogic.getSyntacticCategory(Node name)` and `NodeCategory NodeLogic.getCategory(Node name, TypeSolver typeSolver)`.

PR #1721, opened 2018-08-04
Issue1709 name logic

ftomassetti:
First part of #1709 

This is the first step at implementing Names as define in Ch 6 of JLS. We could introduce a new interface but to avoid confusion I prefered to create a utility class named NameLogic and put the code there. Why? Because there are already classes named "Name" and "SimpleName" and according to Ch 6 some FieldAccessExpr would be name and some would be not:

```
a.b.c // FieldAccessExpr which is a name
a.b().c .. // FieldAccessExpr which is not a name
```

We could revisit that in JP 4.0

NameLogic has a few methods, all taking a Node:

* isAName: is the node a name or not?
* isSimpleName: is the name simple or qualified?
* isQualifiedName: is the name simple or qualified?
* classifyRole: is the name part of a declaration or a reference?
* nameAsString: the string representing the name

69 Tests are included in this PR


matozoid:
I believe you.

ftomassetti:
Thanks, then I will go on and merge it and prepare the next step.

By the way splitting the thing in smaller bits helped me review stuff and add more tests.

PR #1730, opened 2018-08-07
Clean up of MethodResolutionLogic

daanschipper:
Removed unused function that was added in #1717 and cleaned up some warnings.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
@@ -342,11 +342,16 @@ public class MethodResolutionLogic {
      * Filters out duplicate {@param methods} by their signature.
      */
     private static List<ResolvedMethodDeclaration> getMethodsWithoutDuplicates(List<ResolvedMethodDeclaration> methods) {
-        HashMap<String, ResolvedMethodDeclaration> res = new HashMap<>(methods.size());
+        List<ResolvedMethodDeclaration> res = new ArrayList<>();
+        Set<String> usedSignatures = new HashSet<>();
         for (ResolvedMethodDeclaration md : methods) {
-            res.putIfAbsent(md.getQualifiedSignature(), md);
+            String signature = md.getQualifiedSignature();
+            if (!usedSignatures.contains(signature)) {
+                usedSignatures.add(signature);
+                res.add(md);
+            }
         }
-        return new ArrayList<>(res.values());
+        return res;
     }
 
     /**
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
@@ -637,16 +637,6 @@ public class MethodResolutionLogic {
         throw new UnsupportedOperationException(typeDeclaration.getClass().getCanonicalName());
     }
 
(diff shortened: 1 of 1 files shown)


daanschipper:
There are still some mentions of duplicate code present (with `MethodCallExprContext` and `ConstructorResolutionLogic), but I'm unsure where to refactor those parts. 

Furthermore, there are some unused method parameters and access specifiers that can be restricted, but again, I'm unsure whether to change this.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
```diff
@@ -613,43 +584,31 @@ private static boolean areOverride(MethodUsage winningCandidate, MethodUsage oth
...
+            return ((JavassistClassDeclaration) typeDeclaration).solveMethod(name, argumentsTypes, staticOnly);
         }
-          if (typeDeclaration instanceof JavassistEnumDeclaration) {
+        if (typeDeclaration instanceof JavassistEnumDeclaration) {
             return ((JavassistEnumDeclaration) typeDeclaration).solveMethod(name, argumentsTypes, staticOnly);
         }
         throw new UnsupportedOperationException(typeDeclaration.getClass().getCanonicalName());
     }
 
-    public static SymbolReference<ResolvedMethodDeclaration> solveMethodInFQN(String fqn, String name,
```
ftomassetti, line 640: This is rather internal method, I am not expecting users to invoke this method directly so I think we could remove it.

However this means technically breking the API. What @matozoid think about that? Should we mark it as deprecated instead?

ftomassetti (approved):
To me it looks fine, I have just one comment for @matozoid

ftomassetti:
> Furthermore, there are some unused method parameters and access specifiers that can be restricted, but again, I'm unsure whether to change this.

I would mark the original methods as deprecated and add a copy of these methods without the extra params. Then I would ensure we always call the stripped down versions of the methods. I would also have the original methods delegating to the stripped down versions of the methods

ftomassetti:
Thanks for cleaning up this part: we really needed help in this area!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
matozoid: If you consider the symbol solver instable: delete it.
If you think nobody is using it: delete it.
Otherwise, deprecate with annotation and javadoc explaining what to use instead.

matozoid:
@daanschipper whenever you're working on this stuff and you find out how a method works, consider adding a few lines of Javadoc for future contributors :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
daanschipper: I only introduced this method in #1717, I think we can safely assume that nobody is using it

daanschipper:
There are still some warnings left (empty if/else statements, duplicate code, usage of deprecated functions, assesspecifiers), but I'm not sure on the specifics of those functions so I would rather leave it as it currently is.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
matozoid: Go ahead :-)

ftomassetti:
Nice, I would suggest we merge it as soon as tests pass. We can look into other warnings in separate PRs

Issue #1589, opened 2018-05-21
Incorrect resolved type of binary expression


Shoreray:
It seems that the method com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getBinaryTypeConcrete(Node left, Node right, boolean solveLambdas) does not handle type promotion correctly.

For example, when the resolved types of the left and right nodes are byte and char correspondingly, the returned type of the binary expression is byte, instead of the correct type int.

Shoreray:
The issue is not limited to promotion of primitive types. When the operator is "+", the right node of the binary expression is a string but the left node is not, the calculated type of the expression is whatever the type of the left node is.

ftomassetti:
Thanks for reporting this one!

matozoid:
@ftomassetti any hints for where we should look to solve this?

ftomassetti:
The primitive types know about the values to which they can promoted but this is probably not good enough and we should probably change the code that calculates the type of BinaryExpr. There should be a visitor that given an expression returns the type.

ftomassetti:
We should probably change this method in JavaParserFacade:
```java
    protected ResolvedType getBinaryTypeConcrete(Node left, Node right, boolean solveLambdas) {
        ResolvedType leftType = getTypeConcrete(left, solveLambdas);
        ResolvedType rightType = getTypeConcrete(right, solveLambdas);
        if (rightType.isAssignableBy(leftType)) {
            return rightType;
        }
        return leftType;
    }
```

ftomassetti:
Closed via #1732

PR #1732, opened 2018-08-08
Fix binary operator resolution

### LINKED ISSUE #1731: Error on retrieving type of String concatenation operator +
Type Solver is returning a wrong type for binary expressions when the right operand is a `String` and the left operand is a primitive. I'm using the latest version: **3.6.16**.
For example, the code below creates a compilation unit in which all binary expressions should get `String` as resolved type.

```
imports here...

public class StringConcatOpIssue {

		TypeSolver ts = new ReflectionTypeSolver();

		CompilationUnit compilationUnit = JavaParser.parse(
				"public class Class1 {"
				+ " public void method1() {"
				+ "  String s1 = \"string1\";"
				+ "  System.out.println( 1.0 + \"a_text\" );"
				+ "  System.out.println( 1 + s1 );"
				+ "  System.out.println( s1.charAt(2) + s1 );"
				+ "  System.out.println( \"a_text\" + 1.0 );"
				+ "  System.out.println( s1 + 1 );"
				+ "  System.out.println( s1 + s1.charAt(2) );"
				+ " }"
				+ "}");

		List<BinaryExpr> bExprs = compilationUnit.findAll(BinaryExpr.class);

		for (BinaryExpr be : bExprs) {
			ResolvedType rt = JavaParserFacade.get(ts).getType(be);
			System.out.println(rt);
		}
	}
}
```

But I got the output below, which is not the expected for the first 3 subjects:
```
PrimitiveTypeUsage{name='double'}
PrimitiveTypeUsage{name='int'}
PrimitiveTypeUsage{name='char'}
ReferenceType{java.lang.String, typeParametersMap=TypeParametersMap{nameToValue={}}}
ReferenceType{java.lang.String, typeParametersMap=TypeParametersMap{nameToValue={}}}
ReferenceType{java.lang.String, typeParametersMap=TypeParametersMap{nameToValue={}}}
```

Thank you in advance and I'm sorry if I did not follow a correct standardization for opening this issue.

ftomassetti:
Fix #1731 and #1589

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedPrimitiveType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedPrimitiveType.java
@@ -123,4 +123,7 @@ public enum ResolvedPrimitiveType implements ResolvedType {
         return boxTypeQName;
     }
 
+    public boolean isNumeric() {
+        return this != BOOLEAN;
+    }
 }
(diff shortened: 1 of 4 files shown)


matozoid:
+1

PR #1733, opened 2018-08-08
Name logic helper methods

ftomassetti:
Adding a few helper methods I added while working on the Name logic PR

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
@@ -239,10 +239,16 @@ public final class Name extends Node implements NodeWithIdentifier<Name>, NodeWi
         return super.replace(node, replacementNode);
     }
 
+    /**
+     * A top level name is a name that is not contained in a larger Name instance.
+     */
     public boolean isTopLevel() {
         return !isInternal();
     }
 
+    /**
+     * An internal name is a name that constitutes a part of a larger Name instance.
+     */
     public boolean isInternal() {
         if (this.getParentNode().isPresent() && this.getParentNode().get() instanceof Name) {
             Name parent = (Name)this.getParentNode().get();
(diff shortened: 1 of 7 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
```diff
@@ -238,4 +238,23 @@ public boolean replace(Node node, Node replacementNode) {
...
+    public boolean isTopLevel() {
+        return !isInternal();
+    }
+
+    /**
+     * An internal name is a name that constitutes a part of a larger Name instance.
+     */
+    public boolean isInternal() {
+        return getParentNode()
+                .filter(parent -> parent instanceof Name)
```
matozoid, line 254: I used my powers to push a rewritten version of this method :-)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
```diff
@@ -238,4 +238,23 @@ public boolean replace(Node node, Node replacementNode) {
...
+                .map(parent -> (Name) parent)
+                .flatMap(Name::getQualifier)
+                .map(parentNameQualifier -> parentNameQualifier == this)
```
matozoid: Is this check needed? Isn't this always the case?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
ftomassetti: Good idea!

javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
ftomassetti: Well, currently Name can have just one position where the child can be another Name (the qualifier position). However it could possibly change in the future. But yes, we could simplify it and just check if the parent is a Name

matozoid: Oh, yes, `Name` can only be qualified by another `Name`. I think it's a safe assumption to make :-)

ftomassetti:
Made isInternal simpler, will merge if tests pass on build server

PR #1735, opened 2018-08-09
Adding a resolve() method to ArrayAccessExpr

malteskoruppa:
I was thinking that it would be nice, similar to `FieldAccessExpr` which has a `resolve()` method that returns the resolved field declaration corresponding to the accessed field, to analogously implement a `resolve()` method for `ArrayAccessExpr` that returns the resolved array declaration corresponding to the accessed array. Of course, the array declaration could be either a `ResolvedFieldDeclaration` (if the array is a field) or a `JavaParserSymbolDeclaration` (if the array is declared locally). But, in any case, it is always a `ResolvedValueDeclaration`. :-)

Well to be honest I need that functionality, and hacking around it because it's missing is ugly. :grin: 

What do you think?


javaparser-core/src/main/java/com/github/javaparser/ast/expr/ArrayAccessExpr.java
```diff
@@ -169,6 +174,22 @@ public void ifArrayAccessExpr(Consumer<ArrayAccessExpr> action) {
...
+     * {@code ArrayAccessExpr} is returned. Otherwise, an {@link UnsolvedSymbolException} is thrown.
+     *
+     * @return a {@link ResolvedValueDeclaration} representing the declaration of the accessed array.
+     * @throws UnsolvedSymbolException if the declaration corresponding to the array access expression could not be
+     *                                 resolved.
+     * @see NameExpr#resolve()
+     * @see FieldAccessExpr#resolve()
+     */
+    @Override
+    public ResolvedValueDeclaration resolve() {
```
ftomassetti: My first doubt is on the name: we are not resolving the access itself but the array being accessed, so I would change the name here.

Note also that the array being accessed could be anything, including the resulting value of a method.

`foo()[2]`

What should we return in such cases?

malteskoruppa: Ah, yes, those are two very good points!

> My first doubt is on the name: we are not resolving the access itself but the array being accessed, so I would change the name here.

I was actually aware of that, but my thinking was that it would be nice if `ArrayAccessExpr` implements `Resolvable<ResolvedValueDeclaration>` by implementing the method `resolve()`, just like `FieldAccessExpr` does. That way, the naming scheme is consistent and both `*Access` expressions implement the same interface. Of course, it's true that there's a difference, as you mentioned, but my I was thinking that in the case of an `ArrayAccessExpr`, it only makes sense to resolve the array declaration, since there is no such thing as an array access declaration, and so that seemed like an intuitive behavior to me.

But since it was the very first thing you noticed, I changed my mind: I guess it's no so intuitive after all. So I agree we should change the name. What about `resolveAccessedArray()`?

> Note also that the array being accessed could be anything, including the resulting value of a method.

Oh of course, I forgot about that! :sweat_smile:  You are quite right. Hmm... how about having the method `resolveAccessedArray()` return an `Optional<ResolvedValueDeclaration>`, and have it return an empty Optional in such cases?

malteskoruppa: Now that I think about it, the same problem you mentioned also exists for `FieldAccessExpr`. The accessed field's object could be anything, including the resulting value of a method.

`foo().prop`

I am not sure what `FieldAccessExpr#resolve()` returns in such cases. I will check what happens when I have the time.

matozoid: Just to butt in with my lacking knowledge: what does this do exactly?
Resolving `a[0]` in the following...
`class X{ int[]a; void b(){ a[0];} }` gives... `int`? Or `int[]`? Or `X.a`?
`class X{ int[][]a; void b(){ a[0];} }` gives... `int`? Or `int[]`? Or `int[][]`? Or `X.a`?
`class X{ void b(){ int[][]a; a[0];} }` gives... `int`? Or `int[]`? Or `int[][]`?

ftomassetti: > But since it was the very first thing you noticed, I changed my mind: I guess it's no so intuitive after all. So I agree we should change the name. What about resolveAccessedArray()?

That works for me

> Oh of course, I forgot about that! 😅 You are quite right. Hmm... how about having the method resolveAccessedArray() return an Optional<ResolvedValueDeclaration>, and have it return an empty Optional in such cases?

Acceptable for me, we should specify it in the Javadoc

> Just to butt in with my lacking knowledge: what does this do exactly?
Resolving a[0] in the following...
class X{ int[]a; void b(){ a[0];} } gives... int? Or int[]? Or X.a?
class X{ int[][]a; void b(){ a[0];} } gives... int? Or int[]? Or int[][]? Or X.a?
class X{ void b(){ int[][]a; a[0];} } gives... int? Or int[]? Or int[][]?

I would expect it to return the declaration of `a`, from which we can find the type

malteskoruppa: @ftomassetti Cool! :-) I will try to extend this PR when I get around to it. Unfortunately, my time is rather limited at the moment, and I have to prioritize, but eventually I will do it. ;-)

@matozoid Yes, as Federico said, it would return the declaration of `a` in all situations, i.e., all of this is about resolving the actual declaration, and not about resolving the type of an array access. So in your first two examples, we would get a `ResolvedFieldDeclaration`, and your your last example, we would get a `JavaParserSymbolDeclaration`.

matozoid: Alright, would it be a good idea to work the term "element type" in there then?
--> https://docs.oracle.com/javase/specs/jls/se10/html/jls-10.html

malteskoruppa: Yes, I do like that idea! Thanks. :-)

matozoid:
This is about to miss the release.

matozoid:
@malteskoruppa it will miss the release again :-)

malteskoruppa:
Sorry about that. I will try to do it when I can, but in the worst case it might be that I can only get around to this one in mid-october. :(

malteskoruppa:
What do you think, is this better? :-)

Note: If you don't think this method is generally useful enough, we can just close this PR. It's nothing that a user couldn't by themselves really. It's only a small convenience method.

matozoid:
The Abstract Syntax Christmas Tree principle says the more utility functions the better. +1

matozoid:
@malteskoruppa @ftomassetti Merge or close? Last chance :-)

ftomassetti:
well, I would be more in favor of closing than merging. Sorry :(

malteskoruppa:
I agree, all things considered it was not such a good idea. Please feel free to close it. :)

PR #1736, opened 2018-08-09
Fix how Variable Declarators, Fields, and Enum constants are being represented

### LINKED ISSUE #1668: Resolving VariableDeclaration throws IllegalStateException

ftomassetti:
Fix #1668 and a related issue with JavaParserFieldDeclaration still used to represent enum constants.

Inherited test case provided in #1669


matozoid:
Can you improve the PR name?

matozoid:
+1

ftomassetti:
For some reasons the tests pass on JDK9 and they are running forever on JDK8...
screen shot 2018-08-10 at 10 12 03

ftomassetti:
`The job exceeded the maximum time limit for jobs, and has been terminated.`

Restarting

matozoid:
Maybe we should put some of the book money into renting a real CI cloud.

ftomassetti:
That could be a good idea. However we would need to find something that does not require too much effort to be managed. Perhaps the paid version of Travis?

matozoid:
@ftomassetti looks like the only paid thing at Travis is the enterprise self-hosted version?

Anyway, we should just improve the build times :-)

PR #1737, opened 2018-08-09
Solve method calls for methods declared in enums (when loaded from jar)

kgeilmann:
I discovered a problem when solving calls to methods declared in enums. If the enum is coming from a jar file, solving the call is rejected with 
`java.lang.IllegalArgumentException: Trying to instantiate a JavassistClassDeclaration with something which is not a class: javassist.CtClassType@e2cbe0[changed final class MyEnum extends java.lang.Enum `...

This is solved here.


ftomassetti:
+1
Thank you @kgeilmann !

PR #1740, opened 2018-08-12
Improving Context with new features needed for NameLogic

ftomassetti:
As part of the work to improve NameLogic I added a few more methods to Context to find the vars, params, and fields exposed to children.

This is derived from #1715


matozoid:
+1. Maybe you want someone else to look at this though.

ftomassetti:
Thanks. Let's wait a few days to see if someone wants to comment. I will ask on gitter if there are any takers.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/TryWithResourceContext.java
```diff
@@ -85,4 +89,26 @@ public TryWithResourceContext(TryStmt wrappedNode, TypeSolver typeSolver) {
                                                                   boolean staticOnly, TypeSolver typeSolver) {
         return getParent().solveMethod(name, argumentsTypes, false, typeSolver);
     }
+
+    @Override
+    public List<VariableDeclarator> localVariablesExposedToChild(Node child) {
+        for (int i=0;i<wrappedNode.getResources().size();i++) {
```
daanschipper: I'm a bit confused by this loop. You're getting all VariableDeclarationExprs that are declared _before_ the given argument child, but they way you're performing operations to get this list seems quite excessive. I count 4 iterations over the same list.

ftomassetti: Oh, right, I should save the resources in a list somewhere. Good catch!

daanschipper (approved):

ftomassetti:
Approved by two reviewevers, merging

Issue #1379, opened 2018-02-01
JSON ouput has multiple keys ("type") under the same dict


wwt17:
I found this part of JSON when I was trying to parse a method.
```javascript
{
  "parameters": [
    {
      "type": "Parameter",
      "isVarArgs": "false",
      "name": {
        "type": "SimpleName",
        "identifier": "db"
      },
      "type": {
        "type": "ClassOrInterfaceType",
        "name": {
          "type": "SimpleName",
          "identifier": "HighlyAvailableGraphDatabase"
        }
      }
    }
  ]
}
```
You can see there are two identical keys called "type". While this is not allowed according to JSON standard. If I parse this JSON, it will get only the last one.

ftomassetti:
Right! Thank you for reporting this

matozoid:
Any ideas on what the output should be then?

wwt17:
I think just change the string "type" to some other name (for example, "node") in the method output of class JSONPrinter will solve this issue.

matozoid:
Ah, I see now, I thought it was a different problem.

MysterAitch:
+1 

I have come up against this recently (though I don't see it present in the snippet quoted above). Solved it locally by changing "type" to "_type" to avoid the key collision on the assumption that there is unlikely to be an attribute/key of that name naturally occurring. Seems to be working well enough here so far.

matozoid:
Note that this doesn't have a high priority, because the JSON is meant to be *looked at* to examine the AST, not to be processed.

matozoid:
@MysterAitch - there is now a Json serialization module.

matozoid:
... and I've renamed type to _type.

PR #1741, opened 2018-08-12
Add a true JSON AST serializer

matozoid:
We've had a bunch of issues with the JsonPrinter and its friends, which were never meant to do serialization. These new classes are.

### DIFF
--- a/javaparser-core-serialization/src/main/java/com/github/javaparser/serialization/JavaParserJsonDeserializer.java
+++ b/javaparser-core-serialization/src/main/java/com/github/javaparser/serialization/JavaParserJsonDeserializer.java
@@ -1,0 +1,81 @@
+package com.github.javaparser.serialization;
+
+import com.github.javaparser.ast.Modifier;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.metamodel.BaseNodeMetaModel;
+import com.github.javaparser.metamodel.PropertyMetaModel;
+import com.github.javaparser.utils.Log;
+
+import javax.json.JsonArray;
+import javax.json.JsonObject;
+import javax.json.JsonReader;
+import javax.json.JsonString;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static com.github.javaparser.ast.NodeList.toNodeList;
+import static com.github.javaparser.metamodel.JavaParserMetaModel.getNodeMetaModel;
+
+/**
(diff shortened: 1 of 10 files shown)


javaparser-core-serialization/src/main/java/com/github/javaparser/serialization/JavaParserJsonDeserializer.java
```diff
@@ -0,0 +1,81 @@
+package com.github.javaparser.serialization;
```
ftomassetti, line 21: Do we need copyright statements on new files?

javaparser-core-serialization/src/main/java/com/github/javaparser/serialization/JavaParserJsonDeserializer.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    private Node deserializeObject(JsonObject nodeJson) {
+        try {
+            String serializedNodeType = nodeJson.getString("!");
+            BaseNodeMetaModel nodeMetaModel = getNodeMetaModel(Class.forName(serializedNodeType))
+                    .orElseThrow(() -> new IllegalStateException("Trying to deserialize an unknown node type: " + serializedNodeType));
+            Map<String, Object> parameters = new HashMap<>();
+            for (String name : nodeJson.keySet()) {
+                if (name.equals("!")) {
```
ftomassetti: What this special name mean? Do we need a constant for that?

javaparser-core-serialization/src/main/java/com/github/javaparser/serialization/JavaParserJsonSerializer.java
```diff
@@ -0,0 +1,62 @@
...
+    private void serialize(String nodeName, Node node, JsonGenerator generator) {
+        requireNonNull(node);
+        BaseNodeMetaModel nodeMetaModel = JavaParserMetaModel.getNodeMetaModel(node.getClass()).orElseThrow(() -> new IllegalStateException("Unknown Node: " + node.getClass()));
+
+        if (nodeName == null) {
+            generator.writeStartObject();
+        } else {
+            generator.writeStartObject(nodeName);
+        }
+        generator.write("!", node.getClass().getName());
```
ftomassetti: Perhaps we could use a constant for that or maybe give a way to set this name in some sort of configuration

ftomassetti (approved):
It seems ok to me.
I think that some way of processing nodes during serialization could become handy later. But let's users ask for that :D

matozoid:
Fixes #1379

PR #1747, opened 2018-08-14
Deprecate duplicated parent finding code

### LINKED ISSUE #1727: Duplicate methods: findParent(Class) and getAncestorOfType(Class)
It seems to me that there are two methods in the JavaParser API which do exactly the same thing:

* `Node#findParent(Class)`
* `HasParentNode#getAncestorOfType(Class)`

Looking at the implementation of these two methods also reveals that they do exactly the same thing.

In the spirit of a minimal and well-defined API with no duplicate methods, would it make sense to remove either of these methods?

I think having the method in `HasParentNode` (rather than `Node`) makes sense, since then, both `Node` and `NodeList` have it. I have no strong feelings about the method's name. :-)

matozoid:
Fixes #1727


javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
```diff
@@ -51,15 +51,28 @@
...
-            if (classType.isAssignableFrom(parent.getClass())) {
-                return Optional.of(classType.cast(parent));
+        return findParent(classType);
+    }
+
+    /**
+     * Walks the parents of this node, returning the first node of type "type" or empty() if none is found.
+     * Also works for other types, like the With... interfaces.
+     */
+    default <N> Optional<N> findParent(Class<N> type) {
```
ftomassetti: would be "ancestor" more correct here? The parent is only the immediate containing node, right?

ftomassetti (approved):
The code looks good, only doubt is on the name of the method

javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
matozoid: Yeah, I guess we can do that.

malteskoruppa: I agree, "findAncestor()" would probably be the best name. :-)

Except that this gives us *two* deprecated methods now, "findParent()" and "getAncestorOfType()", but, c'est la vie!

malteskoruppa (changes requested):
I agree we should rename it to "findAncestor()", but otherwise, good improvement!

malteskoruppa:
Cool! :)

malteskoruppa (approved):

Issue #1356, opened 2018-01-22
Integrate AccessLevel and AccessSpecifier


matozoid:
Three ways to represent modifiers is at least one too many :-)

PR #1754, opened 2018-08-16
Create a Modifier node

### LINKED ISSUE #912: FieldDeclaration or MethodDeclaration should not auto-reorder the nodes (or provide a way to get the original raw text if necessary)
I was trying to use this parser to validate that annotations are added in the correct order. The validation function is provided an instance of either FieldDeclaration or MethodDeclaration, and needs to determine if the annotations are in the correct order. The following, for example, would be valid:

@Nullable private int mVariable = 1;

But the following, would not:

private @Nullable int mVariable = 1;

However, the NodeList provided to FieldDeclaration (or MethodDeclaration) has already separated the annotations from the other modifiers, putting the annotations first, regardless of how they appeared in the raw Java code. Consequently, you can't identify the second example from an instance of FieldDeclaration.

One way to fix it would be to preserve the order in the NodeList that elements appear in the raw Java (or having a second NodeList to maintain this). A less good way, would be to expose the raw string.

matozoid:
`Modifier` is no longer an enum, it's a real `Node` now, so it's officially part of the AST. The enum is still there, it's called `Modifier.Keyword` now. `Modifier` has a field of that type.
Modifiers are no longer stored in an `EnumSet`, but in a `NodeList`.

Fixes #912


matozoid:
This also integrates the modifier enums. Fixes #1356

ftomassetti:
Should we work on this for JP 3.8?

matozoid:
@ftomassetti yes, I'll fix the merge trouble tonight.

matozoid:
@ftomassetti this is now in the same state as it was.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
```diff
@@ -10,13 +11,16 @@
...
+import static com.github.javaparser.ParserConfiguration.LanguageLevel.RAW;
+
 /**
  * Generates all generated visitors in the javaparser-core module.
  * Suggested usage is by running the run_core_generators.sh script.
  * You may want to run_metamodel_generator.sh before that.
  */
 public class CoreGenerator {
     private static final ParserConfiguration parserConfiguration = new ParserConfiguration()
+            .setLanguageLevel(RAW)
```
ftomassetti, line 23: Why have we stuff commented out? What `RAW` means in this context?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/TransformationsTest.java
```diff
@@ -36,7 +36,7 @@ public void unchangedComplexFile() throws IOException {
     @Test
     public void example1() throws IOException {
         considerExample("Example1_original");
-        cu.getClassByName("A").get().getFieldByName("a").get().setModifiers(EnumSet.of(Modifier.STATIC));
+        cu.getClassByName("A").get().getFieldByName("a").get().addModifier(STATIC);
```
ftomassetti: add seems semantically different from set but I will assume that in this case this does not matter

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -366,8 +366,8 @@ public ClassOrInterfaceDeclaration addClass(String name) {
      * @param modifiers the modifiers (like Modifier.PUBLIC)
      * @return the newly created class
      */
-    public ClassOrInterfaceDeclaration addClass(String name, Modifier... modifiers) {
-        ClassOrInterfaceDeclaration classOrInterfaceDeclaration = new ClassOrInterfaceDeclaration(Arrays.stream(modifiers).collect(Collectors.toCollection(() -> EnumSet.noneOf(Modifier.class))), false, name);
+    public ClassOrInterfaceDeclaration addClass(String name, Modifier.Keyword... modifiers) {
```
ftomassetti, line 369: much nicer :D

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
```diff
@@ -1,73 +1,217 @@
...
+
+    @Generated("com.github.javaparser.generator.core.node.PropertyGenerator")
+    public Keyword getKeyword() {
+        return keyword;
+    }
+
+    @Generated("com.github.javaparser.generator.core.node.PropertyGenerator")
+    public Modifier setKeyword(final Keyword keyword) {
+        assertNotNull(keyword);
+        if (keyword == this.keyword) {
```
ftomassetti, line 174: sometimes we use the space before the parenthesis (which I prefer) sometimes we don't.

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
```diff
@@ -1,73 +1,217 @@
...
+     * and puts them in a NodeList.
+     */
+    public static NodeList<Modifier> createModifierList(Modifier.Keyword... modifiers) {
+        return Arrays.stream(modifiers).map(Modifier::new).collect(toNodeList());
+    }
+
+    @Override
+    @Generated("com.github.javaparser.generator.core.node.RemoveMethodGenerator")
+    public boolean remove(Node node) {
+        if (node == null)
```
ftomassetti, line 193: I would always put braces with `if`, even when we have one statement. We do that in `setKeyword`

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
```diff
@@ -1,73 +1,217 @@
...
+            return false;
+        return super.remove(node);
+    }
+
+    @Override
+    @Generated("com.github.javaparser.generator.core.node.ReplaceMethodGenerator")
+    public boolean replace(Node node, Node replacementNode) {
+        if (node == null)
```
ftomassetti, line 201: Same here

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -575,7 +577,7 @@ public void register(AstObserver observer, ObserverRegistrationMode mode) {
         if (mode == null) {
             throw new IllegalArgumentException("Mode should be not null");
         }
-        switch (mode) {
+        switch(mode) {
```
ftomassetti: space or not space? This is the question :D

javaparser-core/src/main/javacc-support/com/github/javaparser/ModifierHolder.java
```diff
@@ -12,11 +12,11 @@
  * Helper class for {@link GeneratedJavaParser}
  */
 class ModifierHolder {
```
ftomassetti, line 14: Is this class still relevant?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistConstructorDeclaration.java
```diff
@@ -123,7 +123,7 @@ public ResolvedParameterDeclaration getParam(int i) {
     }
 
     @Override
-    public AccessSpecifier accessSpecifier() {
+    public Modifier.Keyword accessSpecifier() {
```
ftomassetti, line 126: Mmm, AccessSpecifier includes the default package access for which there is no corresponding modifier

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java
```diff
@@ -186,7 +186,7 @@ public boolean isTypeParameter() {
             fields.add(new ResolvedFieldDeclaration() {
                 
                 @Override
-                public AccessSpecifier accessSpecifier() {
+                public Modifier.Keyword accessSpecifier() {
```
ftomassetti, line 194: Mmm, AccessSpecifier includes the default package access for which there is no corresponding modifier

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/AstResolutionUtils.java
```diff
@@ -41,10 +41,6 @@
  */
 class AstResolutionUtils {
 
-    static AccessSpecifier toAccessLevel(EnumSet<Modifier> modifiers) {
```
ftomassetti, line 44: We may still need this as the AccessSpecifier is different that the modifiers. Some modifiers have nothing to do with access and the default access is not represented by modifiers

ftomassetti:
I think the code under `javaparser-symbol-solver-testing/src/test/test_sourcecode/javaparser_new_src/` should not be modified

ftomassetti:
Fixing some issues in https://github.com/ftomassetti/javaparser/tree/matozoid-issue_912_modifier_node

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
matozoid: This comes from the code generator which is always right, so the other thing is wrong :-)

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: Code generator :-)

javaparser-core/src/main/java/com/github/javaparser/ast/Modifier.java
matozoid: Code generator :-) If you want the code generator changed, open an issue.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistConstructorDeclaration.java
matozoid: But there *is* a corresponding `Modifier.Keyword`.

matozoid:
> I think the code under `javaparser-symbol-solver-testing/src/test/test_sourcecode/javaparser_new_src/` should not be modified

Ah, that could be one reason for the failures...

ftomassetti:
Yes, those and other files from test resources should stay unchanged. Fixing them and a few other things all tests seem ok running locally and I am now running those changes on Travis

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java
matozoid: I've been trying to refactor AccessSpecifier away because it overlapped with modifiers (and I think there's even a third one) - but if it is too much it can be reverted.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
matozoid: In time I would like to use the lexical preserving printer, but it's not stable enough yet.

ftomassetti:
The build passed on travis, so I created a PR on your fork against this branch: https://github.com/matozoid/javaparser/pull/2

matozoid:
Here it goes!

matozoid:
Before merging: remove all usages of EnumSet, I think this is all it was used for.

javaparser-core/src/main/javacc-support/com/github/javaparser/ModifierHolder.java
matozoid: It is produced when parsing modifiers and annotations. It holds both. This is where we need to have the discussion whether we want annotations and modifiers stored separately, or if we want to make them the same thing... Maybe it's better to keep them apart for now, it would make the API very "unspecific."

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
matozoid: RAW is "just parse, don't do semantic validation"

matozoid:
@ftomassetti - I need your veto on removing or restoring AccessSpecifier.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java
ftomassetti: Ok

ftomassetti:
I am in no position to veto: to me it seems something different from Modifier and I would prefer to keep it, but if you are convinced we do not need it let's get rid of it

matozoid:
@ftomassetti here is the current code:
```java

    /**
     * @return Modifier.Keyword.PUBLIC, Modifier.Keyword.PROTECTED,
     * Modifier.Keyword.PRIVATE, or Modifier.Keyword.DEFAULT when none of the others exists.
     */
    default Modifier.Keyword getAccessSpecifier() {
        for (Modifier modifier : getModifiers()) {
            switch (modifier.getKeyword()) {
                case PUBLIC:
                case PROTECTED:
                case PRIVATE:
                    return modifier.getKeyword();
            }
        }
        return Modifier.Keyword.DEFAULT;
    }
```
so the algorithm is preserved. It just reuses Modifier.Keyword instead of having a seperate enum.

matozoid:
@ftomassetti I just broke all the lex pres stuff... Can you check if I added Modifier correctly to the CSM?

ftomassetti:
mmm the tests seem to be passing, no?

matozoid:
That's really odd. Will give it another try then.

matozoid:
@ftomassetti if you would like access specifier back then we can do that another time.

ftomassetti:
ok

PR #1764, opened 2018-08-17
Rerun unicode code generator with JDK11

### LINKED ISSUE #1758: Rerun unicode character generation

matozoid:
Fixes #1758


ftomassetti:
I do not understand this, so I guess it is ok :D

matozoid:
It's interesting really. JDK 11 will supports Unicode... 10 or so. That means there are more unicode characters that can be used for identifiers and such.

PR #1765, opened 2018-08-17
EnumDeclaration.entries -> constants

### LINKED ISSUE #1386: Rename EnumDeclaration.entries to EnumDeclaration.constants
It's the JLS terminology.

matozoid:
Fixes #1386


ftomassetti:
I am not sure about this because constants could also mean public static final fields in my opinion. However it is true that JLS calls them constants...

DeepSnowNeeL:
Because it actually compiles to constants and the enum becomes a class :
https://stackoverflow.com/a/31040243/1281193 :)

ftomassetti:
Yes, that is true but there are different levels of abstractions between what developers see and what ends up in bytecode

matozoid:
Also not that we already have a `EnumConstantDeclaration`.

matozoid:
I'll let this go because nobody ever complained about the naming here.

PR #1767, opened 2018-08-17
Remove static JP API

matozoid:
This takes all the static methods from `JavaParser`, puts them in the new interface `JavaParserSugar` as default methods, then lets `JavaParser` implement that.

The advantage: no more confusion between static and instance methods on JavaParser, and no more confusion about that "static configuration"
The disadvantage: no more quick parse methods.
A way to relieve that a little: any class can implement `JavaParserSugar` and a `parse` method, and all the short methods are available.

JP itself regularly called the static parse methods. Now it calls the methods on `JavaParser.internalParser`, which I hope will not be abused by users.


matozoid:
Does anyone have a better idea than that JavaParserSugar class?

stefanleh:
>Does anyone have a better idea than that JavaParserSugar class?

Why not remove all static methods and only let people use the Parser via an instance?
The benefit of  `JavaParser.parse(...)` vs `new JavaParser().parse(...)` is not that big.
I would even prefer the builder-pattern for putting together the usual stuff (Symbolsolver config ...).

> ...which I hope will not be abused by users ...

If one overwrites (abuses) the parse method(s) its ones fault, right?

ftomassetti:
I think it is a good thing avoiding the confusion between static and instance methods as a lot of users were doing so and also me from time to time :D

matozoid:
@stefanleh this PR is about removing the static methods, so that wish has been fulfilled :-) I'm just not sure if that sugar class is the right approach. It's pretty novel, that's something.

stefanleh:
To me the addition of the the default methods to Interfaces in Java was more or less a concession to the many changes (implement it in every class using the interface) you had to made when adding something to an Interface which is used all over the place. Im not sure if that feature should be used upfront when creating new Interfaces. Id rather use an abstract class for that (eg. JavaParser extends AbstractJavaParser implements Parser).

matozoid:
The abstract class I'd like to avoid since there is only one "extends" slot, and spending that on a bunch of utility methods is not too great.... I think I'll just move everything back into JavaParser itself and leave the next step for later when the requirement is clearer (if required at all.)

stefanleh:
I agree to that one-slot argument, but whatever you might need in future to extend from should be placed well in that AbstractJavaParser class, shouldnt it?

However, removing the static methods would be nice.

matozoid:
The idea was to avoid having to put `javaparser.` in front of all the calls that are now static (like `parserStatement(...)` turns into `javaparser.parseStatement(...)`. So by extending that sugar interface you got the same short syntax. It's really that old ugly "extend an interface with only constants" pattern from before static imports. There's no other reason to extend JavaParser, so that's easy :-)

stefanleh:
>The idea was to avoid having to put javaparser. in front of all the calls that are now static (like parserStatement(...) turns into javaparser.parseStatement(...). So by extending that sugar interface you got the same short syntax. It's really that old ugly "extend an interface with only constants" pattern from before static imports. There's no other reason to extend JavaParser, so that's easy :-)

Funny enough im using JavaParser for replacing all the static calls in my project by referenceName.call() to gain some extensibility. 
To some it might be boilerplate but i like to see directly when reading the code on which reference/class i call the method.

matozoid:
Yes, the funny thing is that those static methods are part of the "easy" part of JavaParser. But people regularly get confused with the instance and static methods, the two configurations for one object, the one parse method that is not static, that it makes things easy only for the most basic of usages. So out it goes :-)

stefanleh:
>So out it goes :-)

Didnt wanted to keep from merging the change. Just wanted to discuss that topic a bit. Nevermind.

matozoid:
No problem, I won't have time to do this in the coming days anyway :-)

PR #1773, opened 2018-08-19
Completion of JavassistParameterDeclaration

kgeilmann:
This PR completes the implementation of JavassistParameterDeclaration. 
For the isXYZ methods it copies the behaviour of ReflectionParameterDeclaration and if available the parameter name gets read from the jar.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistUtils.java
```diff
@@ -206,4 +203,19 @@ private static ResolvedType typeArgumentToType(SignatureAttribute.TypeArgument t
...
+        CodeAttribute codeAttribute = methodInfo.getCodeAttribute();
+        if (codeAttribute != null) {
+            LocalVariableAttribute attr = (LocalVariableAttribute) codeAttribute.getAttribute(LocalVariableAttribute
+                    .tag);
+            if (attr != null) {
+                int pos = Modifier.isStatic(method.getModifiers()) ? 0 : 1;
+                return attr.variableName(paramNumber + pos);
+            }
+        }
+        return null;
```
ftomassetti: We should explain while this return null

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistParameterDeclarationTest.java
```diff
@@ -0,0 +1,168 @@
...
+    private TypeSolver typeSolver;
+
+    @Before
+    public void setup() throws IOException {
+        Path pathToJar = adaptPath("src/test/resources/javaparser-core-3.0.0-alpha.2.jar");
+        typeSolver = new CombinedTypeSolver(new JarTypeSolver(pathToJar), new ReflectionTypeSolver());
+    }
+
+    @Test
+    public void noNamesAvailable_inInterfaceMethods() {
```
ftomassetti: I think we should not use underscores in names

ftomassetti (approved):
To me it seems good and a very valuable contribution: thanks. I have just minor comments

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistUtils.java
matozoid: Nothing in JavaParser should return null really. Use Optional.

kgeilmann:
Good points. I've added a commit addressing your comments.

matozoid:
Thanks!

PR #1779, opened 2018-08-20
Fixed randomly failing tests.

kgeilmann:
I had some test failing in around every third build, so I fixed them.

The tests failed, because the surefire plugin executes up to 4 test methods in parallel (configured in pom.xml), and in these test classes all test methods modify the static configuration. Depending on the interleaving of the execution, this makes the test fail, because the type solver needed to proceed in one method has already been changed by another method.

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/EnumLiteralsInAnnotatedClassTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/EnumLiteralsInAnnotatedClassTest.java
@@ -1,7 +1,9 @@
 package com.github.javaparser.symbolsolver.resolution;
 
 import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseStart;
 import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StreamProvider;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.expr.FieldAccessExpr;
 import com.github.javaparser.symbolsolver.JavaSymbolSolver;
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/EnumLiteralsInAnnotatedClassTest.java
```diff
@@ -20,19 +23,20 @@
 public class EnumLiteralsInAnnotatedClassTest {
 
     @Test
-    public void resolveFieldOfEnumAsInternalClassOfClassUnqualifiedSamePackage() throws FileNotFoundException {
+    public void resolveFieldOfEnumAsInternalClassOfClassUnqualifiedSamePackage() throws IOException {
```
ftomassetti, line 26: why throwing a more generic exception?

ftomassetti (approved):
I have just a minor comments but I am ok with it

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/EnumLiteralsInAnnotatedClassTest.java
kgeilmann: Because the StreamProvider throws it.

ftomassetti: Right, thanks

ftomassetti:
+1

matozoid:
Thanks a lot!

Issue #1784, opened 2018-08-23
Under certain circumstances types are resolved incorrectly (modifiers are ignored)


jjbr:
Here is an interesting edge-case which the symbol solver doesn't handle quite right at the moment. 

Consider following code: 
```
public class Main {
	public class ResolveMe {
		public String s = "correct";
	}

	class Child extends ParentClass {
		void foo() {
			ResolveMe x = new ResolveMe();
		}
	}
}

class ParentClass {
	private class ResolveMe {
		public String s = "incorrect";
	}
}
```

Now try to resolve the type of `ResolveMe` in `foo()`, it will resolve this to `ParentClass.ResolveMe` where `Main.ResolveMe` would be correct. The underlying issue here is that the symbol solver (correctly) first checks ParentClass for any applicable types but then ignores the `private` modifier of `ParentClass.ResolveMe`. Also see com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter#solveType(String, TypeSolver) line 71: https://github.com/javaparser/javaparser/blob/244d595d430ca4f03096ec36cf4d5bf48ef962bf/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java#L71 

Pull request with corresponding failing test: #1785

ftomassetti:
This was fixed by #1789

PR #1785, opened 2018-08-23
Added testcase for #1784

jjbr:
See #1784


ftomassetti:
Thank you @jjbr . I have used your test case in #1789 (importing the commit, so it will be attributed to you). When we merge #1789 we can close this PR.

PR #1787, opened 2018-08-23
Update ReflectionClassDeclaration.java

stefanleh:
Add a small optimization when there is no method to to check for being applicable.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
@@ -154,6 +154,10 @@ public class ReflectionClassDeclaration extends AbstractClassDeclaration {
                 methods.add(ref.getCorrespondingDeclaration());
             }
         }
+        // when empty there is no sense in trying to find the most applicable
+        if (methods.isEmpty()) {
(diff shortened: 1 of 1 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
```diff
@@ -154,6 +154,10 @@ public String getQualifiedName() {
                 methods.add(ref.getCorrespondingDeclaration());
             }
         }
+        // when empty there is no sense in trying to find the most applicable
```
ftomassetti, line 157: Yes, that is correct but I wonder if this optimization is useful

stefanleh: I was debugging my other resultion issue and came across that part which got entered a lot even if there was an empty list. Maybe its not a big performance gain but it help me with debugging ;)

ftomassetti: Ok then :) Merging as tests complete

ftomassetti:
+1

ftomassetti:
Superseded by #1788 therefore I am closing this one

PR #1788, opened 2018-08-23
Feature/applicable methods filtering

stefanleh:
This is just readability change. Functionality should be untouched.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
@@ -154,6 +154,10 @@ public class ReflectionClassDeclaration extends AbstractClassDeclaration {
                 methods.add(ref.getCorrespondingDeclaration());
             }
         }
+        // when empty there is no sense in trying to find the most applicable
+        if (methods.isEmpty()) {
+            return SymbolReference.unsolved(ResolvedMethodDeclaration.class);
+        }
         return MethodResolutionLogic.findMostApplicable(methods, name, argumentsTypes, typeSolver);
     }
 
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
@@ -154,7 +154,10 @@ public class ReflectionClassDeclaration extends AbstractClassDeclaration {
                 methods.add(ref.getCorrespondingDeclaration());
             }
         }
-        // when empty there is no sense in trying to find the most applicable
+        // When empty there is no sense in trying to find the most applicable.
+        // This is useful for debugging. Performance is not affected as 
+        // MethodResolutionLogic.findMostApplicable method returns very early 
+        // when methods is empty.
         if (methods.isEmpty()) {
             return SymbolReference.unsolved(ResolvedMethodDeclaration.class);
         }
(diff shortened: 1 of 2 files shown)


stefanleh:
Sorry, there is also my last change (ReflectionClassDeclaration) in this branch. But it should disappear once it is merged.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java
```diff
@@ -154,6 +154,10 @@ public String getQualifiedName() {
                 methods.add(ref.getCorrespondingDeclaration());
             }
         }
+        // when empty there is no sense in trying to find the most applicable
```
ftomassetti: I would add also the comment that it helps debugging, because the performance gain should be negligible

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
```diff
@@ -368,8 +363,18 @@ public static boolean isApplicable(MethodUsage method, String name, List<Resolve
...
                                                                                 String name, List<ResolvedType> argumentsTypes, TypeSolver typeSolver, boolean wildcardTolerance) {
-        List<ResolvedMethodDeclaration> methodsWithMatchingName = methods.stream().filter(m -> m.getName().equals(name)).collect(Collectors.toList());
-        List<ResolvedMethodDeclaration> applicableMethods = getMethodsWithoutDuplicates(methodsWithMatchingName).stream().filter((m) -> isApplicable(m, name, argumentsTypes, typeSolver, wildcardTolerance)).collect(Collectors.toList());
+        
+        List<ResolvedMethodDeclaration> methodsWithMatchingName = methods.stream()
+                .filter(m -> m.getName().equals(name))
+                .collect(Collectors.toList());
+        
+        List<ResolvedMethodDeclaration> applicableMethods = methodsWithMatchingName.stream()
+                // Filters out duplicate {@param methods} by their signature.
```
ftomassetti: Here `{@param methods}` does not make sense

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
```diff
@@ -337,21 +340,13 @@ public static boolean isApplicable(MethodUsage method, String name, List<Resolve
         }
         return true;
     }
-
+    
     /**
-     * Filters out duplicate {@param methods} by their signature.
+     * filters by given function
```
ftomassetti: Please rephrase and capitalize

ftomassetti (commented):
Some minor changes requested

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
```diff
@@ -337,21 +340,13 @@ public static boolean isApplicable(MethodUsage method, String name, List<Resolve
         }
         return true;
     }
-
+    
     /**
-     * Filters out duplicate {@param methods} by their signature.
+     * Filters by given function {@param keyExtractor}.
```
ftomassetti: It is still not clear what this function does.
Why should I want to add something to a set then I then throw away?

stefanleh: Please check https://stackoverflow.com/questions/23699371/java-8-distinct-by-property. Its a stateful filter mechanism.

ftomassetti: Ok, but that is not very obvious. I would add a comment or even a link to that explanation. Or just sayins "This function uses a stateful filter mechanism"

ftomassetti:
Thank you so much @stefanleh for working with me on this one!
Your contribution is very much appreciated and you have also taught me something.

matozoid:
+1

PR #1790, opened 2018-08-23
Fix mistake with var and array initializers.

### LINKED ISSUE #1786: JAVA_10 parse problem: "var" cannot infer array types
The following code is parsed fine with both `LanguageLevel.JAVA_8` and `LanguageLevel.JAVA_9`, but it fails with `LanguageLevel.JAVA_10`:
```Java
public class Run {
        var arr = new String[]{"a"};
    }
}
```

```
com.github.javaparser.ParseProblemException: (line 3,col 17) "var" cannot infer array types.
    at com.github.javaparser.JavaParser.simplifiedParse(JavaParser.java:333)
    at com.github.javaparser.JavaParser.parse(JavaParser.java:301)
```

Test code that invokes parser(Kotlin):
```Kotlin
@Test
fun parseJava10() { // fail
    JavaParser.getStaticConfiguration().languageLevel = ParserConfiguration.LanguageLevel.JAVA_10
    JavaParser.parse("""public class Run {
                var arr = new String[]{ "a" };
            }
        }""")
}

@Test
fun parseJava9() { // success
    JavaParser.getStaticConfiguration().languageLevel = ParserConfiguration.LanguageLevel.JAVA_9
    JavaParser.parse("""public class Run {
                var arr = new String[]{ "a" };
            }
        }""")
}

@Test
fun parseJava8() { // success
    JavaParser.getStaticConfiguration().languageLevel = ParserConfiguration.LanguageLevel.JAVA_8
    JavaParser.parse("""public class Run {
                var arr = new String[]{ "a" };
            }
        }""")
}
```

matozoid:
Fixes #1786 

Also validate `var[] =...`

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/validator/chunks/VarValidator.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/validator/chunks/VarValidator.java
@@ -42,7 +42,7 @@ public class VarValidator implements TypedValidator<VarType> {
             return;
         }
         variableDeclarator.ifPresent(vd -> {
-            if(vd.getType().isArrayType()){
+            if (vd.getType().isArrayType()) {
                 reporter.report(vd, "\"var\" cannot have extra array brackets.");
             }
             Optional<Node> variableDeclarationExpr = vd.getParentNode();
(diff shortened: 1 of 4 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/validator/chunks/VarValidator.java
```diff
@@ -42,6 +42,9 @@ public void accept(VarType node, ProblemReporter reporter) {
             return;
         }
         variableDeclarator.ifPresent(vd -> {
+            if(vd.getType().isArrayType()){
```
ftomassetti: nitpick: I would add a space after the `if`

ftomassetti (approved):
+1

Issue #1783, opened 2018-08-23
UnsolvedSymbolException for variadic parameter


stefanleh:
When trying to resolve the method call 

`rtvElief(getMnudta(MNFIRM), kopf.get(KOBLINR), kopf.get(KOBFINR), fcod, "KOBLINR", LIBLAG)` 

i get an 

`UnsolvedSymbolException{context='null', name='We are unable to find the method declaration corresponding to rtvElief(getMnudta(MNFIRM), kopf.get(KOBLINR), kopf.get(KOBFINR), fcod, "KOBLINR", LIBLAG)', cause='null'}`.

Does the symbol solver support array parameters like in 

`rtvElief(Field p_firm, Field p_linr, Field p_finr, Fcod fcod, String errorField, FieldDescription... fields)`?

ftomassetti:
Do you mean variadic parameters? Yes, it supports them. If you share all the relevant code we can look into that. If you woul dreduce it to a failing test that would make our life even easier and we could get a solution more quickly

stefanleh:
Yes, its a variadic parameter, just didnt knew the english name .
Maybe its worth mentioning that the method referenced exists in several versions (overloading).
I will try to put together a testcase for you.

ftomassetti:
I also did not know the name before I started working on JavaParser :)
I think the issue could be related to overloading as variadic parameters in general are supported but picking the right overloaded method when variadic parameters are involved could be tricky.

On I side node I would love to learn more about what you are doing with JavaParser.

stefanleh:
**Update**: The bug is within `com.github.javaparser.symbolsolver.resolution.MethodResolutionLogic.isApplicable(ResolvedMethodDeclaration, String, List<ResolvedType>, TypeSolver, boolean)` and, of course, hard to find. I would appreciate some comments on what exactly that method checks and why.Unfortunately there arent many useful comments in the code.

stefanleh:
>On I side node I would love to learn more about what you are doing with JavaParser.

I have a project which is a 1:1 conversion from RPG language to java. It was implemented without using the OOP patterns possible with java. Therefore all classes use static methods to call each other. Now i use javaparser to find the classes to the methodcalls and generate spring (@Autowired) references in the classes and change the methodcalls to use them. This enables changing all static methods to dynamic ones. Why this cant be done manually? Its more than 40k classes.

ftomassetti:
[...]

PR #1791, opened 2018-08-23
[REFACTOR] Refactored logic of MethodResolutionLogic.isApplicable

stefanleh:
Refactored MethodResolutionLogic.isApplicable into DefaultApplicabilityCheckStrategy and made it configurable.

Please consider this as an idea to discuss.


ftomassetti:
First question: is it strictly necessary to change package to two classes? This would cause an API change and leads to file changes in 336 files

stefanleh:
I had to put the `ApplicabilityCheckStrategy` interface in the core project for being able to inject the strategy into `ParserConfiguration`. For the `isApplicable` method parameters i needed to move some other interfaces to the core project as well. This is the main cause for the many changes. I wish it would work without, but without loosing type safety i dont see another way. Do you?

ftomassetti:
Also, I saw your comment in #1783 but could you explain me in more detail what is the goal of this change?

ftomassetti:
Another thing: I saw you changed files under `javasymbolsolver_0_6_0/src`: you should not do that. That is a copy of the source code of an old version of a library later merged in JP. We use it to test JP and it should never be modified.

stefanleh:
>Also, I saw your comment in #1783 but could you explain me in more detail what is the goal of this change?

I want to write a more simple implementation of the Strategy so that it would fit my needs (be more generous about the applicability).
As i stated i dont think i can find the bug in the current implementation (i am sure the applicable method exists as the code compiles and i have the sources - but unfortunately cant share them).

stefanleh:
>Another thing: I saw you changed files under javasymbolsolver_0_6_0/src: you should not do that. That is a copy of the source code of an old version of a library later merged in JP. We use it to test JP and it should never be modified.

I will revert those changes. Tnx.

ftomassetti:
In my opinion there should be just one way to solve methods: the correct one. I understand your problem and I agree that fixing method resolution is hard (it is very hard for me who wrote most of that stuff, I cannot imagine for others!). However I do not think it would be useful for users to be able to inject their own method resolution logic.
@matozoid what do you think?

matozoid:
Agree with @ftomassetti - an alternative approach here would be to write your own custom logic to resolve methods to work around the current issue (without changing the library code,) but fixing the issue should definitely be the real goal.

stefanleh:
Ok then. I will invest some more hours trying to debug and then go with the approach mentioned by @matozoid. Btw, I also tried to write a testcase to show the issue but without any luck. Guess its cause i used simple versions of the parameter types involved.

PR #1810, opened 2018-08-29
Be more consistent in type arguments defaults

matozoid:
@lucamolteni - like this?


lucamolteni:
Yes @matozoid thank you very much

PR #1817, opened 2018-08-31
Bugfix: Solve member types of other types

malteskoruppa:
This PR fixes a few related bugs. Since they are related, this is a single PR composed of two commits:

- Commit a2e4d9e fixes problem 1.
- Commit cadc6c5 fixes problems 2 and 3.

None of these bugs has been reported so far as an issue, as far as I am aware.

**Problem 1**

Consider:

```
class A extends X.Y {
	void foo() {
		X.Y xy = null;
		xy.bar();
	}
}

class X {
	static class Y {
		void bar() {}
	}
}
```

Try to solve the method call expression `xy.bar()` (Note: Obviously, ignore the fact that this code would throw a null pointer exception at runtime if it was actually run. This is irrelevant for the symbol solver. I just wanted to make the example as simple and concise as possible.)

_Expected behavior / behavior after commit a2e4d9e:_ Solving the call expression should succeed and return the resolved method declaration `X.Y#bar()`.

_Actual behavior / behavior before commit a2e4d9e:_ An `UnsolvedSymbolException` is thrown.

**Problem 2:**

This is similar to the previous example, but uses interfaces instead of classes. Consider:

```
interface A extends X.Y {
	default void foo() {
		X.Y xy = null;
		xy.bar();
	}
}

interface X {
	interface Y {
		void bar();
	}
}
```

As above, try to solve the method call expression `xy.bar()`.

_Expected behavior / behavior after commit cadc6c5:_ Solving the call expression should succeed and return the resolved method declaration `X.Y#bar()`.

_Actual behavior / behavior before commit cadc6c5:_ An `UnsolvedSymbolException` is thrown.

**Problem 3:**

This is similar to the previous example, but the member interface which contains the called method has the same simple name as the interface from which the call originates (i.e., it is called `A` instead of `Y`).

```
interface A extends X.A {
	default void foo() {
		X.A xa = null;
		xa.bar();
	}
}

interface X {
	interface A {
		void bar();
	}
}
```

As above, try to solve the method call expression `xa.bar()`.

_Expected behavior / behavior after commit cadc6c5:_ Solving the call expression should succeed and return the resolved method declaration `X.A#bar()`.

_Actual behavior / behavior before commit cadc6c5:_ JSS gets completely confused because of the same names and ends up in an infinite recursion within the method `JavaParserTypeDeclarationAdapter#checkAncestorsForType(String, ResolvedReferenceTypeDeclaration)`, until the JVM crashes with a `StackOverflowError`.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparser/Navigator.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparser/Navigator.java
@@ -102,6 +102,14 @@ public final class Navigator {
         return cd;
     }
 
+    public static ClassOrInterfaceDeclaration demandInterface(CompilationUnit cu, String qualifiedName) {
+        ClassOrInterfaceDeclaration cd = demandClassOrInterface(cu, qualifiedName);
+        if (!cd.isInterface()) {
+            throw new IllegalStateException("Type is not an interface");
+        }
+        return cd;
+    }
+
     public static EnumDeclaration demandEnum(CompilationUnit cu, String qualifiedName) {
         Optional<TypeDeclaration<?>> res = findType(cu, qualifiedName);
         if (!res.isPresent()) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -116,7 +116,10 @@ public class CompilationUnitContext extends AbstractJavaParserContext<Compilatio
 
     @Override
     public SymbolReference<ResolvedTypeDeclaration> solveType(String name, TypeSolver typeSolver) {
+
         if (wrappedNode.getTypes() != null) {
+            // Look for types in this compilation unit. For instance, if the given name is "A", there may be a class or
+            // interface in this compilation unit called "A".
             for (TypeDeclaration<?> type : wrappedNode.getTypes()) {
                 if (type.getName().getId().equals(name)) {
                     if (type instanceof ClassOrInterfaceDeclaration) {
@@ -130,7 +133,11 @@ public class CompilationUnitContext extends AbstractJavaParserContext<Compilatio
                     }
                 }
             }
-            // look for member classes/interfaces of types in this compilation unit
+            // Look for member classes/interfaces of types in this compilation unit. For instance, if the given name is
+            // "A.B", there may be a class or interface in this compilation unit called "A" which has another member
+            // class or interface called "B". Since the type that we're looking for can be nested arbitrarily deeply
+            // ("A.B.C.D"), we look for the outermost type ("A" in the previous example) first, then recursively invoke
+            // this method for the remaining part of the given name.
             if (name.indexOf('.') > -1) {
                 SymbolReference<ResolvedTypeDeclaration> ref = null;
                 SymbolReference<ResolvedTypeDeclaration> outerMostRef =
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -32,6 +32,7 @@ import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserAnnotationDeclaration;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserEnumDeclaration;
+import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration;
 import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
 import com.github.javaparser.symbolsolver.resolution.MethodResolutionLogic;
@@ -134,9 +135,13 @@ public class CompilationUnitContext extends AbstractJavaParserContext<Compilatio
                 SymbolReference<ResolvedTypeDeclaration> ref = null;
                 SymbolReference<ResolvedTypeDeclaration> outerMostRef =
                         solveType(name.substring(0, name.indexOf(".")), typeSolver);
-                if (outerMostRef.isSolved() &&
+                if (outerMostRef != null && outerMostRef.isSolved() &&
                     outerMostRef.getCorrespondingDeclaration() instanceof JavaParserClassDeclaration) {
                     ref = ((JavaParserClassDeclaration) outerMostRef.getCorrespondingDeclaration())
+                                  .solveType(name.substring(name.indexOf(".") + 1), typeSolver);
+                } else if (outerMostRef != null && outerMostRef.isSolved() &&
+                           outerMostRef.getCorrespondingDeclaration() instanceof JavaParserInterfaceDeclaration) {
+                    ref = ((JavaParserInterfaceDeclaration) outerMostRef.getCorrespondingDeclaration())
                                   .solveType(name.substring(name.indexOf(".") + 1), typeSolver);
                 }
                 if (ref != null && ref.isSolved()) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -30,6 +30,7 @@ import com.github.javaparser.resolution.declarations.*;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserAnnotationDeclaration;
+import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserEnumDeclaration;
 import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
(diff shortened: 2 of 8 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparser/Navigator.java
```diff
@@ -102,6 +102,14 @@ public static ClassOrInterfaceDeclaration demandClass(CompilationUnit cu, String
         return cd;
     }
 
+    public static ClassOrInterfaceDeclaration demandInterface(CompilationUnit cu, String qualifiedName) {
```
ftomassetti, line 105: Ok, make sense

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -128,6 +130,24 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
                     }
                 }
             }
+            // look for member classes/interfaces of types in this compilation unit
```
ftomassetti: Ok, this part is not super-obvious. Could you add a comment to this and the previous part to explain what were we looking for before and what are we looking for here?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -168,7 +188,7 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
                     }
                     if (found) {
                         SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
-                        if (ref.isSolved()) {
+                        if (ref != null && ref.isSolved()) {
```
ftomassetti, line 198: why should that be null? I think it should not

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -179,7 +199,7 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
                 if (importDecl.isAsterisk()) {
                     String qName = importDecl.getNameAsString() + "." + name;
                     SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
-                    if (ref.isSolved()) {
+                    if (ref != null && ref.isSolved()) {
```
ftomassetti, line 209: why should that be null? I think it should not

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -188,7 +208,7 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
 
         // Look in the java.lang package
         SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType("java.lang." + name);
-        if (ref.isSolved()) {
+        if (ref != null && ref.isSolved()) {
```
ftomassetti, line 218: why should that be null? I think it should not

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
```diff
@@ -397,4 +397,58 @@ public void complexTypeSolving() {
 
         assertEquals(mainClass, resolvedTypeDeclaration.getParentNode().get());
     }
+
+    @Test
+    public void resolveMethodCallOfMethodInMemberClassOfAnotherClass() {
+
```
ftomassetti: nitpick: please remove this blank line

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
```diff
@@ -397,4 +397,58 @@ public void complexTypeSolving() {
...
+        SymbolReference<ResolvedMethodDeclaration> reference = JavaParserFacade.get(new ReflectionTypeSolver())
+                                                                       .solve(callExpr);
+
+        assertTrue(reference.isSolved());
+        assertEquals("X.Y.bar()", reference.getCorrespondingDeclaration().getQualifiedSignature());
+    }
+
+    @Test
+    public void resolveMethodCallOfMethodInMemberInterfaceOfAnotherInterface() {
+
```
ftomassetti: nitpick: please remove this blank line

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
```diff
@@ -397,4 +397,58 @@ public void complexTypeSolving() {
...
+        SymbolReference<ResolvedMethodDeclaration> reference = JavaParserFacade.get(new ReflectionTypeSolver())
+                                                                       .solve(callExpr);
+
+        assertTrue(reference.isSolved());
+        assertEquals("X.Y.bar()", reference.getCorrespondingDeclaration().getQualifiedSignature());
+    }
+
+    @Test
+    public void resolveMethodCallOfMethodInMemberInterfaceWithIdenticalNameOfAnotherInterface() {
+
```
ftomassetti: nitpick: please remove this blank line

ftomassetti (approved):
It looks good, I have just minor comments.

Thank you a lot for tackling this one!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
malteskoruppa: Sure.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
malteskoruppa: See below.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
malteskoruppa: Ok, my take is this:

The method `TypeSolver#tryToSolveType(String)` makes no guarantees that it never returns `null`. It is not annotated with an annotation such as `@NotNull`, nor is there a Javadoc comment explaining that this method never returns `null`. I didn't write this method, so for me there is no way to know if, perhaps, this method indeed never returns `null`, but this guarantee is simply undocumented. By default, lacking any documentation, I have to assume that it *could* return `null`.

Generally speaking, I think that if a method makes no guarantees that it does not return `null`, then we cannot simply assume that it will not return `null`. It's better to be safe than sorry.

I just tried to remove the `ref != null` checks that I added, but then there is a failing test because of a null pointer excepton: `ContextTest.resolveReferenceUsingQualifiedName()`. So indeed, it seems like the method  `TypeSolver#tryToSolveType(String)` sometimes returns `null`. Also, if you look around in the same method, there are `ref != null && ref.isSolved()` checks in various places that you wrote yourself. ;-) So it appears to be a valid paradigm. I simply added the `ref != null` check in those places where it was missing.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
malteskoruppa: idem

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
malteskoruppa: A matter of taste, but ok. :-)

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionTest.java
malteskoruppa: idem

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
malteskoruppa: See below.

malteskoruppa:
Good? :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
ftomassetti: I understand your position, let's merge this.

However no method in Javaparser is supposed to return null, so from my point of view if something return null is a bug. I will open a separate ticket to verify `TypeSolver#tryToSolveType(String)` does not return null and correct where it does so. Thank you for pointing out this

ftomassetti:
It looks good to me. Thank you for working on my comments.
Now the only issue is that tests failed but probably due to an hiccup in Travis. I am going to restart the build.

matozoid:
Looks good!

jlerbsc:
This issue seems to be resolved now.

    public class Issue1817Test extends AbstractSymbolResolutionTest {

    @Test
    public void test() {
        
        Path testResources= adaptPath("src/test/resources/issue1817");
        
        CombinedTypeSolver typeSolver = new CombinedTypeSolver();
        typeSolver.add(new ReflectionTypeSolver());
        typeSolver.add(new JavaParserTypeSolver(testResources));

        StaticJavaParser.setConfiguration(new ParserConfiguration().setSymbolResolver(new JavaSymbolSolver(typeSolver)));

        String s = 
                "interface A extends X.A {\n" + 
                "    default void foo() {\n" + 
                "        X.A xa = null;\n" + 
                "        xa.bar();\n" + 
                "    }\n" + 
                "}";
        
        CompilationUnit cu = StaticJavaParser.parse(s);
        
        MethodCallExpr mce = cu.findFirst(MethodCallExpr.class).get();
        
        assertEquals("X.A.bar",mce.resolve().getQualifiedName());
    }

}

PR #1825, opened 2018-09-04
Improve build speed

IlyaFinkelshteyn:
this build process does not work OK with dynamic memory on Hyper-V VMs, e.g. does not increase it as needed. Moving to GCE where VMs are with static memory.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,7 @@
 version: '{build}'
 os: Windows Server 2012
+environment:
+  appveyor_build_worker_cloud: gce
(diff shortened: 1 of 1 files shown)


ftomassetti:
Wow, a huge improvement!
Thank you so much for your work at AppVeyor and for taking the time to prepare this PR. It is highly appreciated

matozoid:
Thank you very much @IlyaFinkelshteyn, this helps a lot with this project, since it has to deal with end of line characters extensively :-)

PR #1831, opened 2018-09-06
wrap strings in quotes when printing yaml

clusterrun:
Ran this against our massive Java code base and all yaml files are now valid.
First ever open-source PR ... looking forward to learn from mistakes :pray:

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/YamlPrinterTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/YamlPrinterTest.java
@@ -26,6 +26,7 @@ import static org.junit.Assert.assertEquals;
 import org.junit.Test;
 
 import com.github.javaparser.JavaParser;
+import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.expr.Expression;
 
 public class YamlPrinterTest {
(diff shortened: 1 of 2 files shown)


matozoid:
Thank you @clusterrun! Only one thing: can you put the escaping code in a method, like "escapeValue" or so?

Oh, another thing ;-) If you want, can you supply a unit test?

Can I ask how you are using the Yaml files? Just to get some insight in what our users are doing :-)

clusterrun:
Thanks! I added a test and moved the escaping code in a method.

I am using it to generate a swagger-like API description of our custom API and objects. But I am doing that In two steps. 
1. First, I am converting all `.java` files to `.yaml` using javaparser and 
2. then, using python I go through the yaml files and generate the swagger-like API description.

Then a potential third step could be to generate code for client libraries.

I am actually converting yaml to json at the end of step one because python (step 2) parses json a lot faster than yaml. I would have liked to use `JsonPrinter` but it has deeper problems that I didn't have time to look into. (was getting duplicate keys :disappointed:  )

matozoid:
Hmmm, I think I fixed the duplicate key issue a few versions back... Anyway, you can investigate the serialization module, which is a more serious approach to JSON serialization: https://github.com/javaparser/javaparser/tree/master/javaparser-core-serialization

It's at least clear that these printers are being used for much more than just visualization, which they were designed for.

Thanks for this fix and for the feedback, you have made a lot of future users happy :-)

matozoid:
(It will be in next Sunday's release, by the way)

PR #1833, opened 2018-09-07
Resolving annotation expressions

malteskoruppa:
This PR heavily improves support for resolving annotation expressions in JSS. Resolving annotation expressions is very important for static analysis tools that wish to parse code and exhibit a certain behavior when annotated expressions are encountered.

* Commit 451b09e implements the method `AnnotationExpr#resolve()` to enable resolution of annotations directly from an annotation expression node. This is in the spirit of moving symbol resolution away from `JavaParserFacade` and into the AST nodes.
* Commit a4ce98e adds the new classes `ReflectionAnnotationDeclaration` and `ReflectionAnnotationMemberDeclaration` as well as the logic to enable resolution of annotation expressions whose declaration lives in the standard library (such as `@Override`, `@SuppressWarnings`, etc.). Previously, it was only possible to resolve declarations of annotation expressions when the annotation declaration had been parsed using JavaParser.
* Commit 11d97f0 adds the new classes `JavassistAnnotationDeclaration` and `JavassistAnnotationMemberDeclaration` as well as the logic to enable resolution of annotation expressions whose declaration lives in a provided jar file (e.g., when providing a JUnit 4 jar file, declarations such as `@Test`, `@Ignore`, `@Before`, etc.). Previously, it was only possible to resolve declarations of annotation expressions when the annotation declaration had been parsed using JavaParser or when the annotation declaration lived in the standard library.
* Commits dc99432 and 0471fe4 only make some minor improvements to the code.

Note that support for annotations in JSS is still not 100% complete. For example, it is not possible to get the default value of a `ReflectionAnnotationMemberDeclaration` or a `JavassistAnnotationMemberDeclaration` (instead, an `UnsupportedOperationException` is thrown). However, it is also not possible to get the default value of a `JavaParserAnnotationMemberDeclaration` either (here too, an `UnsupportedOperationException` is thrown), and that class has been around for much longer.

So support for annotations is still not complete, but definitely improved. If support for annotations in JSS was at 70% before, it is now maybe at 80%. :-) (random numbers of course, but you get the point)

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
@@ -63,11 +63,12 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
 
     @Override
     public String getClassName() {
-        String className = ctClass.getName().replace('$', '.');
-        if (getPackageName() != null) {
-            return className.substring(getPackageName().length() + 1, className.length());
+        String qualifiedName = getQualifiedName();
+        if(qualifiedName.contains(".")) {
+            return qualifiedName.substring(qualifiedName.lastIndexOf("."), qualifiedName.length());
+        } else {
+            return qualifiedName;
         }
-        return className;
     }
 
     @Override
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
@@ -78,11 +78,13 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
 
     @Override
     public boolean isAssignableBy(ResolvedType type) {
+        // TODO #1836
         throw new UnsupportedOperationException();
     }
 
     @Override
     public List<ResolvedFieldDeclaration> getAllFields() {
+        // TODO #1837
         throw new UnsupportedOperationException();
     }
 
@@ -98,11 +100,13 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
 
     @Override
     public Set<ResolvedMethodDeclaration> getDeclaredMethods() {
+        // TODO #1838
         throw new UnsupportedOperationException();
     }
 
     @Override
     public boolean hasDirectlyAnnotation(String canonicalName) {
+        // TODO #1839
         throw new UnsupportedOperationException();
     }
 
@@ -113,11 +117,13 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
 
     @Override
     public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
+        // TODO #1840
         throw new UnsupportedOperationException();
     }
 
     @Override
     public Optional<ResolvedReferenceTypeDeclaration> containerType() {
+        // TODO #1841
         throw new UnsupportedOperationException("containerType() is not supported for " + this.getClass().getCanonicalName());
     }
 
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
@@ -65,7 +65,7 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
     public String getClassName() {
         String qualifiedName = getQualifiedName();
         if(qualifiedName.contains(".")) {
-            return qualifiedName.substring(qualifiedName.lastIndexOf("."), qualifiedName.length());
+            return qualifiedName.substring(qualifiedName.lastIndexOf(".") + 1, qualifiedName.length());
         } else {
             return qualifiedName;
         }
@@ -108,8 +108,7 @@ public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration impl
 
     @Override
     public String getName() {
-        String[] nameElements = ctClass.getSimpleName().replace('$', '.').split("\\.");
-        return nameElements[nameElements.length - 1];
+        return getClassName();
     }
 
     @Override
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2016 Federico Tomassetti
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.javaparser.symbolsolver.javassistmodel;
+
+import com.github.javaparser.resolution.declarations.*;
+import com.github.javaparser.resolution.types.ResolvedReferenceType;
+import com.github.javaparser.resolution.types.ResolvedType;
+import com.github.javaparser.symbolsolver.logic.AbstractTypeDeclaration;
+import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
+import javassist.CtClass;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * @author Malte Skoruppa
+ */
+public class JavassistAnnotationDeclaration extends AbstractTypeDeclaration implements ResolvedAnnotationDeclaration {
+
+    private CtClass ctClass;
+    private TypeSolver typeSolver;
+    private JavassistTypeDeclarationAdapter javassistTypeDeclarationAdapter;
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "{" +
+               "ctClass=" + ctClass.getName() +
+               ", typeSolver=" + typeSolver +
+               '}';
+    }
+
+    public JavassistAnnotationDeclaration(CtClass ctClass, TypeSolver typeSolver) {
+        if (!ctClass.isAnnotation()) {
+            throw new IllegalArgumentException("Not an annotation: " + ctClass.getName());
+        }
+        this.ctClass = ctClass;
+        this.typeSolver = typeSolver;
+        this.javassistTypeDeclarationAdapter = new JavassistTypeDeclarationAdapter(ctClass, typeSolver);
+    }
+
+    @Override
+    public String getPackageName() {
+        return ctClass.getPackageName();
+    }
+
+    @Override
+    public String getClassName() {
+        String className = ctClass.getName().replace('$', '.');
+        if (getPackageName() != null) {
+            return className.substring(getPackageName().length() + 1, className.length());
+        }
+        return className;
+    }
+
+    @Override
+    public String getQualifiedName() {
+        return ctClass.getName().replace('$', '.');
+    }
+
+    @Override
+    public boolean isAssignableBy(ResolvedType type) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ResolvedFieldDeclaration> getAllFields() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean isAssignableBy(ResolvedReferenceTypeDeclaration other) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public List<ResolvedReferenceType> getAncestors() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Set<ResolvedMethodDeclaration> getDeclaredMethods() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean hasDirectlyAnnotation(String canonicalName) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public String getName() {
+        String[] nameElements = ctClass.getSimpleName().replace('$', '.').split("\\.");
+        return nameElements[nameElements.length - 1];
+    }
+
+    @Override
+    public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Optional<ResolvedReferenceTypeDeclaration> containerType() {
+        throw new UnsupportedOperationException("containerType() is not supported for " + this.getClass().getCanonicalName());
+    }
+
+    @Override
+    public List<ResolvedConstructorDeclaration> getConstructors() {
+        return Collections.emptyList();
+    }
+
+    @Override
+    public List<ResolvedAnnotationMemberDeclaration> getAnnotationMembers() {
+        return Stream.of(ctClass.getDeclaredMethods())
+                       .map(m -> new JavassistAnnotationMemberDeclaration(m, typeSolver))
+                       .collect(Collectors.toList());
+    }
+}
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
@@ -1,28 +1,31 @@
-package com.github.javaparser.symbolsolver.reflectionmodel;
+package com.github.javaparser.symbolsolver.javassistmodel;
 
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.resolution.declarations.ResolvedAnnotationMemberDeclaration;
 import com.github.javaparser.resolution.types.ResolvedType;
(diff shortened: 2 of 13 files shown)


matozoid:
Thanks!

Can you make some issues for what's still missing?

malteskoruppa:
I only know `getDefaultValue()` is not properly implemented yet. I think other things *may* be missing, but I'm not quite sure. ;-) Anyway, I will gladly make an issue for the default values of annotation members.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
+/*
+ * Copyright 2016 Federico Tomassetti
```
ftomassetti, line 2: Perhaps we should come up with a new Copyright notice, we keep stating that and then forget about it :)

We can change it after this PR anyway

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public String getPackageName() {
+        return ctClass.getPackageName();
+    }
+
+    @Override
+    public String getClassName() {
+        String className = ctClass.getName().replace('$', '.');
```
ftomassetti: Do we need a utility for this? I get this logic happen in different places.
We could start making a call to `getQualifiedName` here

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public String getQualifiedName() {
+        return ctClass.getName().replace('$', '.');
+    }
+
+    @Override
+    public boolean isAssignableBy(ResolvedType type) {
+        throw new UnsupportedOperationException();
```
ftomassetti: Can we reuse the logic we have in the other Javassist classes?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public List<ResolvedFieldDeclaration> getAllFields() {
+        throw new UnsupportedOperationException();
```
ftomassetti: Should this be empty? Or we consider the annotation fields as fields?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public List<ResolvedReferenceType> getAncestors() {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public Set<ResolvedMethodDeclaration> getDeclaredMethods() {
+        throw new UnsupportedOperationException();
```
ftomassetti: This can be an empty set I guess

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public boolean hasDirectlyAnnotation(String canonicalName) {
+        throw new UnsupportedOperationException();
```
ftomassetti: An annotation can be annotated I think. Maybe we can copy what we do for the other Javassist classes

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public String getName() {
+        String[] nameElements = ctClass.getSimpleName().replace('$', '.').split("\\.");
```
ftomassetti: can we call getClassName here?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+        return nameElements[nameElements.length - 1];
+    }
+
+    @Override
+    public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
+        throw new UnsupportedOperationException();
```
ftomassetti: Do annotations support type parameters?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    @Override
+    public Optional<ResolvedReferenceTypeDeclaration> containerType() {
+        throw new UnsupportedOperationException("containerType() is not supported for " + this.getClass().getCanonicalName());
```
ftomassetti: Perhaps we can copy what we do for other javassist classes

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
```diff
@@ -0,0 +1,40 @@
...
+    private TypeSolver typeSolver;
+
+    public JavassistAnnotationMemberDeclaration(CtMethod annotationMember, TypeSolver typeSolver) {
+        this.annotationMember = annotationMember;
+        this.typeSolver = typeSolver;
+    }
+
+    @Override
+    public Expression getDefaultValue() {
+        // TODO we should do something like this:
```
ftomassetti, line 24: Good, thanks for the comment

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
ftomassetti: Unless we treat the annotation fields as methods... confusing

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// Public methods
+    ///
+
+    @Override
+    public String getPackageName() {
+        if (clazz.getPackage() != null) {
+            return clazz.getPackage().getName();
+        }
+        return null;
```
ftomassetti: Can this method return null? Or should it return empty string?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Override
+    public String getClassName() {
+        String canonicalName = clazz.getCanonicalName();
+        if (canonicalName != null && getPackageName() != null) {
+            return canonicalName.substring(getPackageName().length() + 1, canonicalName.length());
+        }
+        return null;
```
ftomassetti: I am not sure if we want to return null here

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
```diff
@@ -0,0 +1,163 @@
...
+        if (!(o instanceof ReflectionAnnotationDeclaration)) return false;
+
+        ReflectionAnnotationDeclaration that = (ReflectionAnnotationDeclaration) o;
+
+        return clazz.getCanonicalName().equals(that.clazz.getCanonicalName());
+    }
+
+    @Override
+    public int hashCode() {
+        return clazz.hashCode();
```
ftomassetti: I would use the canonical name here, to avoid having two classes marked as equal but have a different hashCode

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @Override
+    public boolean isAssignableBy(ResolvedType type) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean isAssignableBy(ResolvedReferenceTypeDeclaration other) {
+        return false;
```
ftomassetti: It is safer to throw an exception here

ftomassetti (commented):
It looks good!
I added some comments, only few of them are things we need to work on before approving it

ftomassetti:
Thanks for the big contribution!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: I agree, let's do this after the PR. It needs to be done for many classes anyway. :-) I will add an issue about it.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: This is a problem that already exists in `JavaParserAnnotationDeclaration`. There too, this method simply throws an `UnsupportedOperationException`. I would suggest that, as @matozoid asked, I make an issue for each comment here where `JavaParserAnnotationDeclaration` also simply throws an `UnsupportedOperationException`, because that means that it is not a new issue, but something that is already a missing functionality now. That way, we can think about what we want to do in each case, and then this logic must be implemented for all *three* classes `JavaParserAnnotationDeclaration`, `ReflectionAnnotationDeclaration`, and `JavassistAnnotationDeclaration`, not only the two new ones which I added. Is that ok?

So... I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Again: Same thing in `JavaParserAnnotationDeclaration`, so I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Again: Same thing in `JavaParserAnnotationDeclaration`, so I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Again: Same thing in `JavaParserAnnotationDeclaration`, so I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Again: Same thing in `JavaParserAnnotationDeclaration`, so I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: Again: Same thing in `JavaParserAnnotationDeclaration`, so I'll make an issue about this one.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
malteskoruppa: Sure! I already made an issue about this one. :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
malteskoruppa: Done, commit coming in a sec.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See issue #1836.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See #1837.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See #1838.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See #1839.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See #1840.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
malteskoruppa: See #1841.

malteskoruppa:
You're welcome! :-)

@matozoid Ok, with the help of @ftomassetti 's comments I managed to summarize everything that is still missing as far as annotations are concerned in JSS in a few issues. ;-)

I think I worked on all comments except those those that concern functionality that is also already missing in `JavaParserAnnotationDeclaration`. Therefore I would consider these new problems and I made an issue about each one. That way we can work on them separately. Some of them are easy, while others may be more tricky.

matozoid:
You're awesome, @malteskoruppa !

ftomassetti:
On my side we are good to go, I will leave the honor to merge to @matozoid 

Thank you again!

matozoid:
Here it goes, thanks!

PR #1858, opened 2018-09-17
Fix #1857

### LINKED ISSUE #1857: A primitive type as parameter for generic type and JavassistTypeSolver
When we try to get fields for classes from jar archive we will have `UnsolvedSymbolException`. It happens if a field is parametrized and its generic type is primitive. 
For example:  

We have the class:
```
public class ClassWithFields {
    private List<byte[]> genericPrimitiveArrayField = new ArrayList<>();
   ...
}
```

and we want to get fields from this class:
```java
((JavassistClassDeclaration) anotherTypeSolver.solveType("com.github.javaparser.test.ClassWithFields")).getAllFields().forEach(e->e.getType());
```

After execute the code, which is above, we will see the next stacktrace:

>Unsolved symbol : byte[]
>UnsolvedSymbolException{context='null', name='byte[]', cause='null'} 
>	at com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver.solveType(CombinedTypeSolver.java:76)
>	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.lambda$objectTypeArgumentToType$3(JavassistUtils.java:182)

izebit:
It fixes the issue #1857

For testing I had to add to the archive `test-artifact-1.0.0.jar` a class with name `ClassWithFields` and with next content:

```java
public class ClassWithFields {
    private List<List<Integer>> genericParamObjectField = new ArrayList<>();
    private List<byte[]> genericPrimitiveArrayField = new ArrayList<>();
    private List<Integer[]> genericObjectArrayField = new ArrayList<>();
    private List<Integer> genericField = new ArrayList<>();
    private int primitiveField = 10;
    private Integer objectField = 10;
}
```


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistUtils.java
```diff
@@ -176,18 +177,26 @@ static String internalNameToCanonicalName(String typeName) {
     }
 
     private static ResolvedType objectTypeArgumentToType(SignatureAttribute.ObjectType typeArgument, TypeSolver typeSolver, ResolvedTypeParametrizable typeParametrizable) {
-        String typeName = typeArgument.jvmTypeName();
-        Optional<ResolvedType> type = getGenericParameterByName(typeName, typeParametrizable);
+        if(typeArgument instanceof SignatureAttribute.ArrayType){
```
ftomassetti, line 179: COuld we reuse https://github.com/javaparser/javaparser/blob/e201ce166190f2566633e7e285fde1f884888e0c/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistUtils.java#L132 ?

ftomassetti:
It seems fine in principle, just wondering if we can reuse a utility method we already have.
But thanks for the nice contribution!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistUtils.java
izebit: Yes, It looks like we could. Thank you for your remark, I will check it.

matozoid:
+1

Issue #1703, opened 2018-07-17
resolving MethodCallExpr in anonymous classes


Yonka:
Hi!
I have an issue while resolving type of MethodCallExpr scope variable in anonymous class.
Example:
```java
import org.springframework.transaction.TransactionStatus;
import org.springframework.transaction.support.TransactionCallbackWithoutResult;

public class AnonymousClassExample {
    private static final Integer f = 1;
    private void sentNotificationResultEventInTransaction() {
        new TransactionCallbackWithoutResult() {
            @Override
            protected void doInTransactionWithoutResult(TransactionStatus status) {
                f.toString();
            }
        };
    }
}
```
And when I try to resolve type of `f` while calling `m.getScope().get().asNameExpr().resolve()` from `public void visit(MethodCallExpr m, Void a)`, I get: 
java.lang.UnsupportedOperationException: CorrespondingDeclaration not available for unsolved symbol.

And one more interesting thing. SymbolSolver resolved type of `f` when I placed this example code to the same package with my visitor implementation.
What am I doing wrong?

matozoid:
@ftomassetti - we missed this question :-(

matozoid:
@Yonka - thanks for reporting this one :-)

PR #1860, opened 2018-09-20
Working on a test for #1703

ftomassetti:
Adding it as a reminder to complete this PR (now I have to board the plane :D)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -61,6 +61,9 @@ public abstract class ResolvedReferenceType implements ResolvedType,
     }
 
     public ResolvedReferenceType(ResolvedReferenceTypeDeclaration typeDeclaration, List<ResolvedType> typeArguments) {
+        if (typeDeclaration == null) {
+            throw new IllegalArgumentException("TypeDeclaration is not expected to be null");
+        }
         if (typeDeclaration.isTypeParameter()) {
             throw new IllegalArgumentException("You should use only Classes, Interfaces and enums");
         }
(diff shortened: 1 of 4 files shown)


ftomassetti:
Adding reviewers to ping about the fact that this is not a draft anymore

matozoid (approved):
I have no opinion :-(

malteskoruppa:
Sorry for being late to the party. :-)

I know that this is already merged, but I just spent some time trying to understand this PR now, and I am a bit confused. I first assumed that this PR was meant to fix an issue, but after spending some time looking at it, I think it is only meant to introduce a test for something that is working fine already?

I think this because:

After investigating issue #1703, I think that the user's problem was not a bug in the symbol solver, but only that they did not configure their symbol solver correctly. All the user needed to do was add a `JarTypeSolver` for the `spring-tx-<VERSION>.jar` ([from Maven central](https://mvnrepository.com/artifact/org.springframework/spring-tx/5.1.1.RELEASE) for example). This jar contains all the declarations needed so that the symbol solver can resolve the method call mentioned in that issue. When I use a combined `ReflectionTypeSolver`/`JarTypeSolver("spring-tx.jar")`, I can resolve the method call mentioned in issue #1703, _even when I revert all the commits of this PR_. In other words, adding such a `JarTypeSolver` is all that's needed. However, when I _don't_ add this JAR but only uses a pure `ReflectionTypeSolver`, then I get exactly the user's problem: `java.lang.UnsupportedOperationException: CorrespondingDeclaration not available for unsolved symbol.`, _in both cases, with or without the commits of this PR_.

Now looking at the modifications to the `src/main` code of this PR, if I am not mistaken, all it does is introduce a few null pointer checks and throw runtime exceptions with messages when certain variables are null. So, we get a `RuntimeException` with a message instead of a `NullPointerException` internally when we try to resolve the method call expression and the symbol solver does not have the needed information (from the jar). But it doesn't otherwise change anything.

Looking at the test case, I see that it uses a `MemoryTypeSolver` with a mock object to simulate the needed declaration instead of a `JarTypeSolver` of spring-tx. That works too, of course. But the only reason that the test passes now and it didn't before is this line, which was added along with the modifications to the main code in the last commit:
```java
when(cd.asReferenceType()).thenReturn(cd);
```
In other words, the test now correctly configures a return value in the mock object. So basically, the only reason that the last commit makes this test pass is that the test now configures the symbol solver correctly; the modifications to the main code make no difference here. It's the same effect as if we had used a `JarTypeSolver` instead (then no mock object would have been needed in the first place).

Sorry, this was longer than I thought it would be. :-)

So is it correct that this PR is meant to introduce a new test for something, but doesn't actually do any substantital fixes in the main code? If so, nice one, it's certainly a good test to have. :+1:

ftomassetti:
Yes, it is just a test to ensure that this thing that now work will keep working in the future

ftomassetti:
Thank you for reviewing it!

Issue #1866, opened 2018-09-28
How to find out if a node is inside a finally block?


malteskoruppa:
I found it curiously difficult to determine, for a node, whether that node is inside a `finally` block. It feels like this is something that should be fairly easy, yet I ended up with rather complex code. Maybe I'm just overworked and ready for the week-end? :-)

Given a node `node` for which I want to find out whether it is inside a `finally` block, I have this:

```java
boolean inFinallyBlock = false;

// check whether the node is inside a try statement
Optional<TryStmt> tryParent = node.findAncestor(TryStmt.class);

// if so, check whether that try statement has a finally block and if that block contains our node
if (tryParent.isPresent()) {
	Optional<BlockStmt> finallyBlock = tryParent.get().getFinallyBlock();
	if (finallyBlock.isPresent()) {
		Optional<Node> sameNode = finallyBlock.get().findFirst(Node.class, n -> n == node);
		if (sameNode.isPresent()) {
			inFinallyBlock = true;
		}
	}
}
```

It's really lengthy and rather inefficient. I look upwards for a `try` node first, then downwards (from the `finally` block) to my original node again.

1. Am I just not seeing the wood for the trees? Is there a conceptually more elegant way to do this?

2. If not, how about a findAncestor(...) method that takes a predicate? Just like there are variants of findFirst(...) and findAll(...) with and without predicates, a findAncestor(...) with predicate could be useful to do the above much more elegantly. Instead of the above I could do _something like_:

    ```java
    node.findAncestor(BlockStmt.class, bs -> bs.getParent().isTryStmt() &&
                                             ((TryStmt) bs.getParent()).getFinallyBlock().isPresent() &&
                                             ((TryStmt) bs.getParent()).getFinallyBlock().get() == bs);
    ```

    Ok, it's not the embodiment of elegance, but at least it's shorter and more efficient. And one can probably massage this predicate a bit to make it more beautiful. :-)

malteskoruppa:
I just noticed that the first variant which I wrote above (the one that first searches upwards, then downwards again) wouldn't even work properly in some contrived scenarios, e.g., when parsing code such as

```java
try {
  // ...
} catch (Exception e) {
  // ...
} finally {
  try {
    // ...
  } catch (Exception e) {
    foo(); // this is the node for which we want to find out whether it is inside a finally block; in this case, it is!
  }
}
```

The second variant (which would require a `findAncestor` method with predicate) would work, though.

malteskoruppa:
[...]

PR #1867, opened 2018-09-28
Implemented findAncestor(Class,Predicate)

malteskoruppa:
See #1866 :-)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
@@ -25,32 +25,37 @@ import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.observer.Observable;
 
 import java.util.Optional;
+import java.util.function.Predicate;
 
 /**
- * An object that has a parent node.
+ * An object that can have a parent node.
  */
 public interface HasParentNode<T> extends Observable {
 
     /**
-     * Return the parent node or null, if no parent is set.
+     * Returns the parent node, or {@code null} if no parent is set.
      */
     Optional<Node> getParentNode();
 
     /**
-     * Set the parent node.
+     * Sets the parent node.
      *
-     * @param parentNode the parent node or null, to set no parent
-     * @return return <i>this</i>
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
```diff
@@ -25,55 +25,69 @@
...
 
 /**
- * An object that has a parent node.
+ * An object that can have a parent node.
  */
 public interface HasParentNode<T> extends Observable {
 
     /**
-     * Return the parent node or null, if no parent is set.
+     * Returns the parent node, or {@code null} if no parent is set.
```
matozoid, line 36: Oh hey, this is my mistake probably: that should not say "null", but Optional.empty. Do you mind fixing that?

javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
```diff
@@ -25,55 +25,69 @@
...
     @Deprecated
     default <N> Optional<N> findParent(Class<N> type) {
         return findAncestor(type);
     }
 
     /**
-     * Walks the parents of this node, returning the first node of type "type" or empty() if none is found.
-     * Also works for other types, like the With... interfaces.
+     * Walks the parents of this node and returns the first node of type {@code type}, or {@code empty()} if none is
+     * found. The given type may also be an interface type, such as one of the {@code NodeWith...} interface types.
```
matozoid, line 75: Well, thanks for fixing that embarrassingly bad Javadoc :-D

javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
matozoid: Aw, I'll pull it since I don't want this to miss the release for such a silly reason.

malteskoruppa: You're right of course! I just prettified the Javadocs in that class in the process, but I didn't notice that the original Javadoc had a slight mistake there. I see you already did it, cool! :-)

PR #1869, opened 2018-10-03
Fix bug : CompilationUnit#addImport can handle array class of java.lang package.

teletha:


teletha:
Sry, code format was also changed...

PR #1870, opened 2018-10-03
CompilationUnit#addImport can handle array class of java.lang package.

teletha:


ftomassetti:
I am not sure it does make sense: you cannot really import array types

teletha:
There is no problem if it is an array type other than java.lang package.
Currently, multidimensional arrays (e.g. java.io.File[][][].class)  will also generate invalid code because creating a ImportDeclaration instance with name "[[Ljava.io.File;".

matozoid:
Thanks! @Teletha - do you have any idea why specifically java.lang classes give a problem?

@ftomassetti - the functionality was already there, it simply strips off the array brackets.

teletha:
The committed code will solve all problems related to (multidimensional) array types.

Simply the following code is wrong.
`else if (clazz.isArray() && **!ClassUtils.isPrimitiveOrWrapper(clazz.getComponentType()) && !clazz.getComponentType().getName().startsWith("java.lang")**)`

matozoid:
Ahhh, I see what the problem is now: the component type of an array is the type you get when you removed one pair of array brackets. But if there are more, you have to unwrap all of them!

I fixed it like this:
```java
    public CompilationUnit addImport(Class<?> clazz) {
        if(clazz.isArray()){
            return addImport(clazz.getComponentType());
        }
        if (ClassUtils.isPrimitiveOrWrapper(clazz) || clazz.getName().startsWith("java.lang"))
            return this;
        else if (clazz.isMemberClass())
            return addImport(clazz.getName().replace("$", "."));
        else if (clazz.isAnonymousClass() || clazz.isLocalClass())
            throw new RuntimeException(clazz.getName() + " is an anonymous or local class therefore it can't be added with addImport");
        return addImport(clazz.getName());
    }
```

Because I think your approach will fail on this:
```java
    @Test
    public void testAddImportArrayTypes() {
        cu.addImport(CompilationUnit[][][].class);
        cu.addImport(int[][][].class);
        cu.addImport(Integer[][][].class);
        cu.addImport(List[][][].class);
        assertEquals(2, cu.getImports().size());
        assertEquals("com.github.javaparser.ast.CompilationUnit", cu.getImport(0).getNameAsString());
        assertEquals("java.util.List", cu.getImport(1).getNameAsString());
    }
```

teletha:
It is very nice!

matozoid:
@Teletha if you would like your name on it you can make it a PR :-) (Otherwise I'll do it)

teletha:
I would appreciate it if you merge the fixed code on yourself, thanks.

matozoid:
I'll close this then.

PR #1884, opened 2018-10-12
Use data for resolution

matozoid:
Let's see if this makes things faster or slower.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -39,6 +39,7 @@ import com.github.javaparser.metamodel.*;
 import com.github.javaparser.printer.PrettyPrinter;
 import com.github.javaparser.printer.PrettyPrinterConfiguration;
 import com.github.javaparser.resolution.SymbolResolver;
+import com.github.javaparser.resolution.types.ResolvedType;
 
 import java.util.*;
 import java.util.function.Consumer;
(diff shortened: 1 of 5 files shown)


matozoid:
If I look at the AppVeyor build times, this one is still in the range of the last bunch of build times. So it seems this change seems to have a negligible effect on performance.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -477,12 +474,23 @@ public void tryAddImportToParentCompilationUnit(Class<?> clazz) {
...
-        return data.get(key) != null;
+        return data.containsKey(key);
+    }
+
+    /**
+     * Remove data by key.
+     * @see DataKey
+     */
+    public void removeData(DataKey<ResolvedType> key){
+        if(data!=null){
```
ftomassetti: nitpick: space after the if

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
```diff
@@ -266,27 +263,27 @@ private void solveArguments(Node node, NodeList<Expression> args, boolean solveL
...
+     * <p>
      * This method was originally intended to get the type of a value: any value has a type.
-     *
+     * <p>
      * For example:
      * <code>
      * int foo(int a) {
-     *         return a; // when getType is invoked on "a" it returns the type "int"
-     *     }
+     * return a; // when getType is invoked on "a" it returns the type "int"
```
ftomassetti: no idea how to put space into the tag, maybe `&nbsp;`?

ftomassetti (approved):
It seems much, much nicer!

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
matozoid: Check

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
matozoid: `Using <pre> instead of <code> helps.`

PR #1886, opened 2018-10-14
Introduce the toAST method

ftomassetti:
Frequently users want to go from resolved declarations back to source code. Currently they have to do a downcast. This PR introduce a method for that, which should make the thing much more obvious (so maybe more intuitive usage and less questions)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/AssociableToAST.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/AssociableToAST.java
@@ -1,0 +1,21 @@
+package com.github.javaparser.resolution.declarations;
+
+import com.github.javaparser.ast.Node;
+
+import java.util.Optional;
+
+/**
+ * A declaration that can be potentially associated with an AST node.
+ * @param <N> type of AST Node that can be associated
+ */
+public interface AssociableToAST<N extends Node> {
+
+    /**
+     * If the declaration is associated to an AST node return it, otherwise it return empty.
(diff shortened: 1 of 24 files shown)


matozoid:
- isn't this an improvement of getWrappedNode? Should that be deprecated?
- the Java coding standard would like you to spell `toAST` as `toAst`.

ftomassetti:
The difference is that `toAst` is present in all Resolved* declarations, while `getWrappedNode` is present only on the subclasses of the Resolved* declarations that derive from JP AST nodes. Therefore one which has a Resolved* declaration need to do a downcast before being able to use `getWrappedNode`.

Now, this means that `toAst` could potentially replace `getWrappedNode` (but not the other way around!). However `toAst` return an optional, which is less convenient than getting the direct node. Also,`toAst` sometimes have to return a more generic node. This is the case for subclasses of `ResolvedClassDeclaration`. In those cases `toAst` return a `Node`. Why? Because both anonymous class declarations and standard class declarations are subclasses of that. In one case the underlying AST node is an `ObjectCreationExpr`, while in the other case it is `ClassOrInterfaceDeclaration`. In these cases `getWrappedNode` is particularly nice because it returns the right type of AST node, not just a `Node`

ftomassetti:
Regarding the name, I renamed that

matozoid:
+1 then. Maybe you can put part of that explanation in the Javadoc somehow :-)

ftomassetti:
Comment added.
NOT merging now, to avoid breaking again the release :)

Issue #1882, opened 2018-10-12
Node#getData should return Optional, not null.


matozoid:
Oooops!

matozoid:
We could keep this intact if we change the contract into "get the data. If there is no data, illegal state." That would be viable since containsData exists too.

PR #1887, opened 2018-10-14
Clean up user data handling for nodes.

matozoid:
This slightly breaks the API, but it cleanly separates "data==null" and "data exists".


matozoid:
fixes #1882

PR #1890, opened 2018-10-18
Made getAccessSpecifier more generic

jjbr:
The `com.github.javaparser.ast.Modifier` class provides a method `getAccessSpecifier` to retrieve the access specifier from an EnumSet of modifiers. The function itself needs no functionality of an `EnumSet` and it would be convenient to have this method available for generic `Set`s, not just `EnumSet`s.


matozoid:
Thanks! This looks just fine.

PR #1892, opened 2018-10-18
[JsonSerializer] adding non-meta model properties, comments, and SYMBOL_RESOLVER_KEY

dpnolte:
Dear JavaParser maintainers,

First of all, thank you for the awesome library 👍 
I was experimenting with the JavaParserJsonSerializer and JavaParserJsonDeserializer. Besides the meta model properties, I needed to have more properties serialized (e.g., Range, Comment for javadoc). Additionally, after deserialization, I needed to apply the data SYMBOL_RESOLVER_KEY on the compilation unit again to be able to resolve types.  As such, I've  attempted to make the serializer a bit more extendable by the use of the following delegate interfaces:
```
 // serializer
    interface Delegate {
        void toJson(Node node, JsonGenerator generator);
    }
// deserializer
    interface Delegate {
        void fromJson(String propertyName, JsonValue jsonValue, Node node);
    }
```
These delegates are applied in the serializer/deserializer when required. The test - that I've added to JavaParserJsonDeserializerTest - best illustrates what my intentions are: 
```
@Test
    void testDelegate() {
        CompilationUnit cu = JavaParser.parse("public class X{} class Z{}");
        List<JavaParserJsonSerializer.Delegate> serializerDelegates = new LinkedList<>();
        JavaParserJsonSerializer.Delegate rangeSerializerDelegate = (node, generator) -> {
            if (node.getRange().isPresent()) {
                Range range = node.getRange().get();
                generator.writeStartObject("range");
                generator.write("beginLine", range.begin.line);
                generator.write("beginColumn", range.begin.column);
                generator.write("endLine", range.end.line);
                generator.write("endColumn", range.end.column);
                generator.writeEnd();
            }
        };
        serializerDelegates.add(rangeSerializerDelegate);
        String serialized = serialize(cu, false, serializerDelegates);

        Map<String, JavaParserJsonDeserializer.Delegate> deserializerDelegates = new HashMap<>();
        JavaParserJsonDeserializer.Delegate rangeDeserializerDelegate = (propertyName, jsonValue, node) -> {
            JsonObject jsonNode = (JsonObject)jsonValue;
            Position begin = new Position(jsonNode.getInt("beginLine"), jsonNode.getInt("beginColumn"));
            Position end = new Position(jsonNode.getInt("endLine"), jsonNode.getInt("endColumn"));
            node.setRange(new Range(begin, end));
        };
        deserializerDelegates.put("range", rangeDeserializerDelegate);

        Node deserialized = deserializer.deserializeObject(
                Json.createReader(new StringReader(serialized)),
                deserializerDelegates
        );
        Range range = deserialized.getRange().get();
        assertEquals(range.begin.line, 1);
        assertEquals(range.begin.line, 1);
        assertEquals(range.end.line, 1);
        assertEquals(range.end.column, 26);
    }
```
In any case, thanks for the awesome library :)


matozoid:
Hi! Thanks for the PR :-)

What I'm thinking is that the part you're putting behind a delegate can just as well be part of the main functionality - especially the range bit. The symbol solver part would be great too, but I guess this would include computer specific paths?

dpnolte:
Hi @matozoid ,
Thanks for your feedback! 
I've removed the delegates and instead extended the serializer and deserializer class. I've made these properties protected so that any user can create their own implementation.

**Non-meta properties**
This includes for now Range and TokenRange. In the deserializer, any property that is not found in the meta model is stored in the deferredJsonValues map. After the node is created, the method 'readNonMetaProperties' is invoked with the deferred json values.
```
    protected boolean readNonMetaProperties(String name, JsonValue jsonValue, Node node) {
        return readRange(name, jsonValue, node)
                || readTokenRange(name, jsonValue, node);
    }
```
when it returns false an exception is thrown indicating that it is a unknown property.

**Comments**
By delving a bit deeper into it, I noticed that comment is actually a property in the meta model. But, for some reason, comment is not part of the constructor parameters list: https://github.com/javaparser/javaparser/blob/1ec0b91cb78987f0227d41abedc8ebcc1c172875/javaparser-core/src/main/java/com/github/javaparser/metamodel/BaseNodeMetaModel.java#L71-L75
As such, the comment node is actually serialized and deserialized into the parameters map, but the comment is not set when constructing the node. I've therefore added the following:
```
            // Note: comment is a property meta model, but it is not listed as constructor parameter and not attached to node
            // @see BaseNodeMetaModel.getConstructorParameters
            if (parameters.containsKey("comment")) {
                node.setComment((Comment)parameters.get("comment"));
            }
```

**Symbol Resolver Key**
When deserializing a CompilationUnit node from json, the SYMBOL_RESOLVER_KEY will be set if the symbolResolver is set in the static configuration:
```
    private void setSymbolResolverIfCompilationUnit(Node node) {
        if (node instanceof CompilationUnit && JavaParser.getStaticConfiguration().getSymbolResolver().isPresent()) {
            CompilationUnit cu = (CompilationUnit)node;
            cu.setData(Node.SYMBOL_RESOLVER_KEY, JavaParser.getStaticConfiguration().getSymbolResolver().get());
        }
    }
```
As such, the deserialized CU and its children can be resolved with the current configured symbol resolver.

I've added a test for each of these 3 cases.

matozoid:
Okay, it's release time and I'd like to merge this, so the remarks I have will be in issues: #1895 #1896 #1897.

Thank you for this work :-) If you have the opportunity to look at the issues, that would be great

Issue #1397, opened 2018-02-03
Document why SymbolSolver.solveTypeInType is deprecated


matozoid:
... and what to use now?

Issue #1396, opened 2018-02-03
Document why JavaParserClassDeclaration.getContext is deprecated


matozoid:
... and what to use now?

matozoid:
Also: why are some deprecated, and some are not?

Issue #1910, opened 2018-10-30
Generify MethodResolutionLogic#solveMethodInType(...) to Allow for External Implementations


jjbr:
The method [`solveMethodInType(...)` in `com.github.javaparser.symbolsolver.resolution.MethodResolutionLogic` ](https://github.com/javaparser/javaparser/blob/8209c03065e69860be9b2bd65fa0753a27bbf38e/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java#L578-L619) tries to find a method in a type. It checks which kind of `ResolvedTypeDeclaration` was provided as argument and proceeds depending on the Type identified. 

I've recently started implementing a custom type solver with custom `ResolvedTypeDeclaration`, implementations which works fine so far. I call `resolve()` on an AST node and my type solver will also be used to resolve the node. Resolving method calls fails in the `solveMethodInType(...)` function, however, as the function does not accept new/external implementations of `ResolvedTypeDeclaration` and throws an exception when encountering one.

So far, I've come up with two possible solutions. I can modify the method to consider my custom `ResolvedTypeDeclaration` implementations. Since my implementations are very specific to my project and what I am doing and bring no additional value to anyone else, I would have to put this into a private fork and maintain it alongside the official fork of JavaParser. This seems suboptimal and anyone implementing a new `ResolvedTypeDeclaration` would have to do this in order to get method resolution to work.

The second and in my eyes preferred solution is to leave the method solving logic to the implementations of 'ResolvedTypeDeclaration'. All `solveMethodInType(...)` really does, is call `solveMethod` on the different implementations of `ResolvedTypeDeclaration` or their context, right after a cast. You could put a generic abstract `solveMethod` into a new interface (or would `ResolvedReferenceTypeDeclaration` work?), which all implementations of `ResolvedTypeDeclaration` can implement if they support it. Then `solveMethodInType` could be simplified to `typeDeclaration.solveMethod(...)`, probably even removing the need for that method entirely. At the same time, support for other implementations of `ResolvedTypeDeclaration` would not be blocked.

What do you think about this proposal?

ftomassetti:
[...]

PR #1915, opened 2018-11-01
Commenting

### LINKED ISSUE #1395: Document why JavaParserClassDeclaration.solveType is deprecated
... and what to use now?

ftomassetti:
Adding comments to fix #1395 #1396 #1397


matozoid:
There are lots of methods with these names, but only some are marked deprecated. Is it supposed to be that way?

ftomassetti:
Most of those methods are in contexts classes, where I think it is fine to have them. I have deprecated one more method, which was used only for other deprecated methods. There is also a `solveType` in the `SymbolSolver` class but that one does not take a `TypeSolver` so it should be fine

matozoid:
Thanks, +1 then.

matozoid:
If this also helps #1910, do continue the chat there.

MysterAitch:
Note that not all linked issues were automatically closed. Seems that GitHub is nitpicky about having "fixes" or similar immediately before every issue reference?

matozoid:
Here's the doc: https://help.github.com/articles/closing-issues-using-keywords/ . I'll go close them.

PR #1917, opened 2018-11-01
Adjust serializer to Java code style

matozoid:


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -575,7 +576,7 @@ public void register(AstObserver observer, ObserverRegistrationMode mode) {
         if (mode == null) {
             throw new IllegalArgumentException("Mode should be not null");
         }
-        switch (mode) {
+        switch(mode) {
```
ftomassetti, line 579: Do we have a style guideline? I tend to put spaces before parenthesis, I am happy to change my habits but I think we should write a common guideline somewhere (perhaps it exists and I am just missing it)

ftomassetti (approved):
Pretty straightforward and good.

Should we consider this an API breaking change? Should this go in a branch for the next "semi-major" release (3.8)?

matozoid:
If I'm not mistaken the API isn't touched here, or am I missing something?

ftomassetti:
I was referring to the name of the enum literals, which are public

matozoid:
Ah, oh. Well, I doubt they have more than one user yet, so no problem.

Issue #1689, opened 2018-07-13
Redesign LexicalPreservingPrinter


matozoid:
In my opinion, there are two main things mixed up in the logic right now, making everything complex:
- what tokens make up a part of the AST, and
- how do we want to format a source file.

The redesign should split the CSM into two part: the CSM, and the formatting rules. Then the LPP will get two phases: insert/delete/modify the tokens for the changed part, then apply the formatting rules to the changed part.

As a side effect we can modify the pretty printer into: print everything on one line, then apply all formatting rules.

This way there is one source of information about formatting for the whole library, and we can simplify two major parts of the library (PP and LPP.)

Now, I have no idea how to go about defining formatting rules, but Intellij has quite some documentation on how they do it.

matozoid:
@ftomassetti ?

ftomassetti:
It is a complex topic, I admit I have no clear ideas on this one :)

It is ture what you say: the CSM both specify what the syntax has strictly to be and some aspects of formatting (for example, where typically we want to get indentation). It could make sense to separate that. 

There is one thing I am thinking about: we could need a way to understand where spaces are strictly needed. For example, between a type and an identifier a need to put a space because otherwise I would get a different token:

```
// ok
int foo;
// not ok
intfoo;
```

While in other cases spaces are not strictly needed:
```
// ok
1 + 2;
// still ok
1+2;
```

ftomassetti:
I would love to see a prototype of the idea :D 
Any volunteer?

matozoid:
@jamierocks is volunteering! Let's go for a prototype first, indeed. Maybe just something that takes the AST, juggles the tokens around for a few rules and outputs it.

The most challenging parts of the AST to format are variable/field declarations and method declarations, because:
- array brackets can be all over the place, like `int[] a[], b[][];` and `int[] a()[]{}`
- multiple variables can be declared in one go, and the type is on the variable itself, not on the declaration (because the amount of ArrayTypes can be different for each variable.)

So if that is formattable, the rest is easy.

signed:
For formatting Java code there is [spotless](https://github.com/diffplug/spotless).
[prettier](https://github.com/prettier/prettier) puts its focus on web languages.
Maybe you can borrow ideas from there.

Going through the prettier documentation I found a link to the [A pretty printer](http://homepages.inf.ed.ac.uk/wadler/papers/prettier/prettier.pdf) paper.

matozoid:
[...]

PR #1936, opened 2018-11-24
Allow PrettyPrinterVistor's methods to be overriden

jamierocks:
## Preface

This will allow downstream projects to change the behaviour of pretty printer easily.

Currently, you have to override the `#visit` methods that use the utility methods - this avoid that necessity.

## Example

```java
    @Override
    protected void printMembers(final NodeList<BodyDeclaration<?>> members, final Void arg) {
        for (int i = 0; i < members.size(); i++) {
            final BodyDeclaration<?> member = members.get(i);

            if (i == 0 || !member.isFieldDeclaration()) {
                printer.println();
            }
            member.accept(this, arg);
            printer.println();
        }
    }
```

*In the future, I will go over my changes and make pull-requests :)*

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
@@ -65,13 +65,13 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
         return printer.getSource();
     }
 
-    private void printModifiers(final EnumSet<Modifier> modifiers) {
+    protected void printModifiers(final EnumSet<Modifier> modifiers) {
         if (modifiers.size() > 0) {
             printer.print(modifiers.stream().map(Modifier::asString).collect(Collectors.joining(" ")) + " ");
         }
     }
 
-    private void printMembers(final NodeList<BodyDeclaration<?>> members, final Void arg) {
+    protected void printMembers(final NodeList<BodyDeclaration<?>> members, final Void arg) {
         for (final BodyDeclaration<?> member : members) {
             printer.println();
             member.accept(this, arg);
@@ -79,7 +79,7 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
         }
     }
 
-    private void printMemberAnnotations(final NodeList<AnnotationExpr> annotations, final Void arg) {
+    protected void printMemberAnnotations(final NodeList<AnnotationExpr> annotations, final Void arg) {
         if (annotations.isEmpty()) {
             return;
         }
@@ -89,7 +89,7 @@ public class PrettyPrintVisitor implements VoidVisitor<Void> {
         }
     }
 
-    private void printAnnotations(final NodeList<AnnotationExpr> annotations, boolean prefixWithASpace,
+    protected void printAnnotations(final NodeList<AnnotationExpr> annotations, boolean prefixWithASpace,
                                   final Void arg) {
         if (annotations.isEmpty()) {
             return;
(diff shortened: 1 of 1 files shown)


matozoid:
Hi, thanks for the feedback, but those methods are private for a reason :-) When someone asks the pretty printer to do something different, there are two options:
- suggest a feature so we can see if we can make an option for it
- make a copy of the visitor and make it just how you like it.

The thing is complex enough without having to consider whether any change breaks user's code :-O

Do tell about the changes though!

jamierocks:
Well my biggest gripe with the default `PrettyPrinterVisitor` is that there is a newline between each field (the example given above is what I use to resolve that - see [1]). Looking through some other issues it seems as is source code formatting is something that you're looking into anyhow?

*edit: I'd be quite happy to open a new pr for that change, I think the rest are more personal preference - perhaps I'll take a shot at working on the configuration at some point :)*

With regards to those methods, while I appreciate they may not be 'public' API - I think there is a benefit to having them be protected, namely people can avoid having a huge file (20 lines vs 400 or so).

I also appreciate the point on breaking user's code, though I think if you are making modifications you ought to be prepared to have them break - perhaps I should add javadocs to make it clear that they are subject to frequent breakages?

[1]: https://gist.github.com/jamierocks/e2d4acd082f5c629db4f7f1157cf96f2

matozoid:
Generally the pretty print visitor is painful to maintain because it has no connection to the grammar. It's hard to generalize anything, any change has to be done manually. So the Concrete Syntax Tree was a nice change: it is a more abstract way of describing what a node should looks like. But it mixes structure and representation: there is formatting info and even a printer in there, but I don't think it should. So in the end I think the most maintainable approach would be to have a bunch of formatting rules that can be applied to the AST and the underlying tokens instead of having a pretty printer. That would mean a major change and a lot of work, so it's stuck at good ideas for now :-(

The current issue is clear, we do need to think about configuration though. I'll steal some ideas from IntelliJ :-) ... that one has a "blank lines" category with a lot of options. Do you have any more changes for blank line handling?

Let's ask @ftomassetti for an opinion, he tends to have good ideas.

ftomassetti:
I thought about this and I understand the fear that @matozoid is having: we do not want ending up breaking code that we do not control. Also, if people have to ask to us for some changes we can end up improving the solution for everyone.

I see two options:

1. We build in JP an option to allow not writing fields on separate lines
2. We work on a way to post-process the source code after we have serialized. After all we can produce a list of tokens and let you play with them before printing them. That would offer the greates flexibility

jamierocks:
Ideally, it'd be great to have javaparser's output configurable to be able to match my own code style (example below) - though the fields are the most important to me as it makes the output much longer.

```java
import java.io.IOException;
import java.io.InputStream;
import java.nio.file.Files;
import java.nio.file.Paths;

public class Demo {
    
    public void example() {
        try (final InputStream is = Files.newInputStream(Paths.get("beep.boop"))) {
        }
        catch (final IOException ex) {
            ex.printStackTrace();
        }
    }

}
```

As for including formatting information within the AST, IMO the AST should remain separated from formatting but I guess it would work (I'd need some demo code to see how that would work). I think the best move would be to work on a `FormattedPrinter`, having decided on what can be configured - perhaps even implement serialisers for the IntelliJ's file (that way the printer is designed to format correctly from the get-go).

matozoid:
Hey @jamierocks - what are your current thoughts on this?

jamierocks:
My ideal solution would be a printer that would output to a given format specification. I'm thinking that there would be an immutable `FormatSpecification` class with an associated builder class, in addition to serialisers for popular formats (IntelliJ and Eclipse perhaps). This would then be consumed by a `FormattedPrinter`.

I've been pretty busy with college work, but I'll try get to doing some ground work on this in the week.

matozoid:
Okay, let's move this discussion to the relevant issue then.

Do you mind if I close this?

matozoid:
@jamierocks relevant issue is #1689

PR #1922, opened 2018-11-06
[Improvement] Generified MethodResolutionLogic#solveMethodInType(...)

jjbr:
Addressing #1910

PR #1939, opened 2018-11-27
[Improvement] Generalize further parts of the symbol solver's resolution API

jjbr:
This is related to #1910 and #1922. `ContextHelper#solveMethodAsUsage(...)` now works on a new interface rather than a set of specific types that can solve method usages.


matozoid:
Thanks!

ftomassetti:
It looks good to me

Issue #1942, opened 2018-11-30
Unable to solve MethodCallExpr with enum paramter using ReflectionTypeSolver


HoldYourWaffle:
I'm trying to create a JUnit test that checks if all the translation keys I use in my project actually exist in the language definitions.
My (generalised) code is as follows:

```java
TypeSolver typeSolver = new ReflectionTypeSolver(false);
JavaSymbolSolver symbolSolver = new JavaSymbolSolver(typeSolver);
JavaParser.getStaticConfiguration().setSymbolResolver(symbolSolver);

for (CompilationUnit source : obtainSourcesOfProject()) {
    for (MethodCallExpr call : source.findAll(MethodCallExpr.class)) {
        // ... snip ...
        ResolvedMethodDeclaration dec = call.resolve();
        // ... snip ...
    }
}
```

This works for most method calls, but it chokes on the expression `something.equals(Enum.CONSTANT)`, throwing the following exception:
```
java.lang.RuntimeException: Unable to calculate the type of a parameter of a method call. Method call: something.equals(Enum.CONSTANT), Parameter: Enum.CONSTANT
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:225)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:240)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:125)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:142)
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:313)
	... snip ...
Caused by: java.lang.UnsupportedOperationException
	at com.github.javaparser.symbolsolver.reflectionmodel.ReflectionEnumConstantDeclaration.getType(ReflectionEnumConstantDeclaration.java:29)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.solveDotExpressionType(TypeExtractor.java:162)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:181)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:44)
[...]

Issue #1946, opened 2018-12-02
JavaParser not handling a failed solve correctly when using CombinedTypeSolver


HoldYourWaffle:
Please forgive me for the jumbled title. I have no idea how to describe this in a single sentence.
Related to #1942 and (I think) #1945.

I have run into another problem while parsing my source code. Starting to wonder if it's my code being bad, JSS not being stable or me using this completely wrong 😅 

It all started with a method call like this:
```java
UtilityClass.method(SomeClass.InnerEnum.CONSTANT);
```
Which throws the following exception:
```
java.lang.RuntimeException: Unable to calculate the type of a parameter of a method call. Method call: UtilityClass.method(UtilityClass.InnerEnum.CONSTANT), Parameter: UtilityClass.InnerEnum.CONSTANT
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:225)
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:240)
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:125)
    at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:142)
    at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:313)
    at TestJavaParser.testReflectionFirst(TestJavaParser.java:36)
    ... snip ...
Caused by: java.lang.UnsupportedOperationException
    at com.github.javaparser.symbolsolver.reflectionmodel.ReflectionEnumConstantDeclaration.getType(ReflectionEnumConstantDeclaration.java:29)
    at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.solveDotExpressionType(TypeExtractor.java:162)
    at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:181)
    at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:44)
    at com.github.javaparser.ast.expr.FieldAccessExpr.accept(FieldAccessExpr.java:90)
    at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:447)
[...]

PR #1947, opened 2018-12-02
Fix resolving of the type of enum constants obtained by reflection

ftomassetti:
Resolve the type of enum constants obtained by reflection.
See #1946


matozoid:
+1, that makes sense!

HoldYourWaffle:
Also fixes #1942.

ftomassetti:
Let's merge it then

Issue #1949, opened 2018-12-03
JSS can't resolve super-default method reference


HoldYourWaffle:
Found another issue. ([Sample project](https://github.com/HoldYourWaffle/JavaParserDefaultMethodChoke))
This one's quite simple, JSS is unable to solve a method call like this one:
```java
MyInterface.super.defaultMethod();
```

ftomassetti:
This should not be too hard to solve, just a missing functionality

matozoid:
@maartenc @HoldYourWaffle - anybody wants to give this a go?

HoldYourWaffle:
I would, but I have absolutely no idea how to.

matozoid:
Probably copy/paste some similar code into the place where the exception is thrown (if there is an exception?)

MysterAitch:
This threw me a little as I'd not seen this before (allowing the default methods to be called explicitly).

More info for me/others who look at this again in the future: 
https://stackoverflow.com/a/19976615/429733

```java
public class ChildClass implements A, C {
    @Override    
    public void foo() {
       //you could completely override the default implementations
       doSomethingElse();
       //or manage conflicts between the same method foo() in both A and C
       A.super.foo();
    }
    public void bah() {
       A.super.foo(); //original foo() from A accessed
       C.super.foo(); //original foo() from C accessed
    }
}
```

HoldYourWaffle:
I hadn't seen it before either, I was quite surprised when it popped up after throwing JSS at a big codebase.

Issue #1948, opened 2018-12-02
InvalidPathException when parsing interface with generic interface ancestors


HoldYourWaffle:
Closely related to #1876 (might even be identical), but I think I have some new (and probably better explained) information.
As with my previous issues, there's a [sample project](https://github.com/HoldYourWaffle/JavaParserGenericSuperInterfaceChoke). It mostly speaks for itself this time, there's a diagram in the readme that explains the class hierarchy.

The source of the error is method calls like [these](https://github.com/HoldYourWaffle/JavaParserGenericSuperInterfaceChoke/blob/7a2975344ca0450c6a91dca52121db208a6cc16b/src/main/java/main/MainClass.java#L12) which reference generic super interfaces.
It throws the following exception:
```
com.google.common.util.concurrent.UncheckedExecutionException: java.nio.file.InvalidPathException: Illegal char <<> at index 107: D:\VOLATILE\workspace-2\JavaParserGenericSuperInterfaceChoke\src\main\java/interfaces/GenericSuperInterface<T>.java
    at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2218)
    at com.google.common.cache.LocalCache.get(LocalCache.java:4147)
    at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:5053)
    at com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver.tryToSolveType(JavaParserTypeSolver.java:172)
    at com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver.tryToSolveType(CombinedTypeSolver.java:62)
    at com.github.javaparser.symbolsolver.javaparsermodel.contexts.CompilationUnitContext.solveType(CompilationUnitContext.java:163)
    at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.solveType(JavaParserInterfaceDeclaration.java:241)
    at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.toReferenceType(JavaParserInterfaceDeclaration.java:336)
    at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.getAncestors(JavaParserInterfaceDeclaration.java:250)
[...]

Issue #1945, opened 2018-12-01
JavaParser choking on multiple generic method calls on the same line


HoldYourWaffle:
(Part 2 of #1942)

I'm trying to parse some code using JavaParser, but it fails to resolve a method call involving multiple generic methods on a single line.
I could write a very long description of what this code looks like and how it fails, but since that isn't fun for anybody, I have created a small sample project that replicates my code structure in a *very* simplified way while still reproducing the error. You can find it [here](https://github.com/HoldYourWaffle/JavaParserGenericsChoke).

I've abstracted my code with the use of animals.
Some animals have hair, for which there is the `HairyAnimal` interface. Not all animals have the same type of hair, sheep have wool, dogs have fur, well you get the point. This is where the `HairType` interface comes in.
Finally we need to render the hair to the users screen, which is (unsurprisingly) done by the `HairTypeRenderer` ~interface~ abstract class*.
There's also a simple sheep implementation, using `HairTypeWool` and `WoolRenderer`.
<br>
Now for the actual choking code:
```java
sheep.getHairType().getRenderer().renderHair(sheep.getHairType(), sheep);
```
That line of code fails to resolve, triggering the following exception:
```
UnsolvedSymbolException{context='null', name='We are unable to find the method declaration corresponding to sheep.getHairType().getRenderer().renderHair(sheep.getHairType(), sheep)', cause='null'}
    at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:148)
    at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:313)
    at TestJavaParser.testChoke(TestJavaParser.java:31)
```
After some investigation I found out that if I split out the call like this everything resolves just fine:
```java
HairType<?> hairType = sheep.getHairType();
HairTypeRenderer<?> hairTypeRenderer = hairType.getRenderer();
hairTypeRenderer.renderHair(hairType, sheep);
```
I did some more variations with varying degree of 'split-out'-ness, which can all be found inside [`MainClass.java`](https://github.com/HoldYourWaffle/JavaParserGenericsChoke/blob/master/src/main/java/main/MainClass.java). The actual parsing is done inside [`TestJavaParser.java`](https://github.com/HoldYourWaffle/JavaParserGenericsChoke/blob/master/src/test/java/TestJavaParser.java).

[...]

PR #1952, opened 2018-12-03
Add an exception filter to the CombinedTypeSolver

HoldYourWaffle:
Actually fixes the "overlying" problem of #1946.
In that issue we concluded that ignoring exceptions and forwarding the solve to the next solver was undesirable, because we don't want any exceptions to go unnoticed. This makes a lot of sense from a development perspective, but while using JSS to parse real-world code it can lead to nasty roadblocks (which were #1942, #1945, #1946, #1948 and #1949 for me, and I still haven't been able to successfully parse the full codebase). I'm not trying to insult the stability of JSS, I'm the idiot for trying to use an incomplete piece of software 😉 

To solve this problem I have added an opt-in filter to the `CombinedTypeSolver`, allowing users to circumvent some shortcomings of JSS at their own peril. It simply consists of a `Predicate<Exception>` to which all thrown exceptions are tested. If the predicate returns `true`, the exception is ignored and the solve forwarded to the next solver. I also added some convenience filter implementations (black & white lists and some implementations thereof) for use-cases which I deemed "likely to happen to other people". I of course also added a test for this.

I'm not used to contributing to big projects like this one, so I'm pretty sure there's plenty of things to improve in this PR. Some things I think could be better with some help/suggestions:
- Variable names could be better I think.
- Some of the code wrapping seems a bit unfortunate to me, is it correct? (I just formatted the code using the provided eclipse template.)
- I often add a lot of whitespace to break up large blocks of code for readability. Not sure if that's wanted here.
- Given my very limited knowledge of the code this might be in the completely wrong place.
- Not sure if the `ExceptionFilters` class should be an inner-class

I would love to hear your feedback!

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
@@ -16,13 +16,14 @@
 
 package com.github.javaparser.symbolsolver.resolution.typesolvers;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Predicate;
+
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
 import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
-
-import java.util.ArrayList;
-import java.util.List;
 
 /**
  * A container for type solvers. All solving is done by the contained type solvers.
@@ -34,11 +35,27 @@ public class CombinedTypeSolver implements TypeSolver {
 
     private TypeSolver parent;
     private List<TypeSolver> elements = new ArrayList<>();
+    private Predicate<Exception> errorFilter;
 
     public CombinedTypeSolver(TypeSolver... elements) {
+        this(ExceptionFilters.IGNORE_ALL, elements);
+    }
+
+    /** @see #setFilter(Predicate) */
+    public CombinedTypeSolver(Predicate<Exception> errorFilter, TypeSolver... elements) {
+        setFilter(errorFilter);
+
         for (TypeSolver el : elements) {
             add(el);
         }
+    }
+
+    /**
+     * @param errorFilter A filter which determines if an exception raised while solving should be ignored.
+     *            Should return <code>true</code> when the exception must be <b>ignored</b>.
+     */
+    public void setFilter(Predicate<Exception> errorFilter) {
+        this.errorFilter = errorFilter;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
```diff
@@ -34,13 +35,29 @@
 
     private TypeSolver parent;
     private List<TypeSolver> elements = new ArrayList<>();
+    private Predicate<Exception> errorFilter;
 
     public CombinedTypeSolver(TypeSolver... elements) {
+        this(ExceptionFilters.IGNORE_ALL, elements);
```
ftomassetti: IGNORE_ALL should not be the default

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
```diff
@@ -34,13 +35,29 @@
...
         for (TypeSolver el : elements) {
             add(el);
         }
     }
 
+    /**
+     * @param errorFilter A filter which determines if an exception raised while solving should be ignored.
+     *            Should return <code>true</code> when the exception must be <b>ignored</b>.
+     */
+    public void setFilter(Predicate<Exception> errorFilter) {
```
ftomassetti: I would call it "ExceptionHandler" instead of Filter

ftomassetti: We could explain better the use case because it is far from being obvious

ftomassetti (commented):
I would approve it, with a few refinements, but it should not be active by default imho

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
HoldYourWaffle: Whooops that's due to a refactor, will fix.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/CombinedTypeSolver.java
HoldYourWaffle: Will do.

HoldYourWaffle:
@ftomassetti I have addressed your reviews. It wasn't meant to be activated by default (hence my metion of 'opt-in' in the initial comment), that was a mistake I made while refactoring.

ftomassetti:
Updating, let's see if tests pass

ftomassetti:
To me it looks good and this should be merged.

Thank you for your contribution!

Issue #1953, opened 2018-12-04
`ObjectCreationExpr` creation should default to the same state as the parser does


l0rinc:
It would seem logical for the following test to pass:
```java
String className = String.class.getCanonicalName();
ClassOrInterfaceType type = JavaParser.parseClassOrInterfaceType(className);
ObjectCreationExpr expected = JavaParser.parseExpression("new " + className + "()");
ObjectCreationExpr actual = new ObjectCreationExpr(null, type, new NodeList());
assertEquals(expected, actual);
```

since however
```java
public ObjectCreationExpr(final Expression scope, final ClassOrInterfaceType type, final NodeList<Expression> arguments) {
    this(null, scope, type, new NodeList<>(), arguments, null);
}
```
defaults to empty type arguments while `JavaParser.parseExpression` defaults to null, they won't be equal.

The following would fix the issue:
```java
public ObjectCreationExpr(final Expression scope, final ClassOrInterfaceType type, final NodeList<Expression> arguments) {
    this(null, scope, type, null, arguments, null);
}
```

(wanted to provide a PR but wasn't sure where to put tests for `ObjectCreationExpr`)

matozoid:
That makes sense. A PR is welcome :-)

l0rinc:
Pushed https://github.com/javaparser/javaparser/pull/1954/commits
I've added the test separately so that you can make sure it fails before the fix.
Let's wait for the build :)

PR #1954, opened 2018-12-04
#1953: `ObjectCreationExpr` creation should default to the same state as the parser does

l0rinc:
Fixes: https://github.com/javaparser/javaparser/issues/1953


matozoid:
Thanks! That's absolutely fine.

PR #1956, opened 2018-12-07
[Improvement] Generified type variable resolution

jjbr:
Relates to #1910, #1922, #1939.


ftomassetti:
The code seems good but I would suggest adding a test or more to prevent future regressions.

Thank you for your contributions!

jjbr:
> I would suggest adding a test or more to prevent future regressions.

Given this is a relatively straight-forward refactoring of existing functionality without the change of any logic, I am not sure which kind of test I would write. Do you think a test would be necessary and if yes could you point me in the direction of the kind of test you think is appropriate?

Looking through the code again, another question did occur to me: Do you think it is sensible to throw an `UnsupportedOperationException` here, or should we rather return `Optional.empty()`?
https://github.com/javaparser/javaparser/pull/1956/commits/5000b8ce6d741aa391ad54b5f886a8c622dcdca8#diff-e9c21a29406463dbd109d92e74bcf60aR195

matozoid:
@ftomassetti ?

ftomassetti:
I do not understand in which position but there is just one throw UnsupportedOperationException that seems fine

ftomassetti:
+1 (when tests finish after having merged master into this branch)

matozoid:
Thanks!

PR #1960, opened 2018-12-09
Merge testing into core

matozoid:
Well, why not?


HoldYourWaffle:
Suggestion: move the generated code to it's own seperate module. When I first started with this project this really confused me.
No idea how modules really work in maven, but I think it'd make sortof sense.

matozoid:
@HoldYourWaffle - what was the confusion exactly? I don't know why the generated code would be confusing since users don't need to generate anything.

ftomassetti:
Conflicts arising here. Nothing against doing this

matozoid:
Doing this would destroy all the [testcase] PR's, so I'm closing it.

PR #1965, opened 2018-12-13
JUnit test AnalyseJavaParserTest#parseTheRest() hangs on Windows

maartenc:
When srcFile is a file with a reserved name on Windows (like 'aux.java' for instance), Files.exists(srcFile) will return true, but opening an InputStream to it will hang forever. To avoid hangs, we check that srcFile is a regular file before trying to parse it.


ftomassetti:
Interesting! Should we point out the issue somehow to users? Being throwing some kind of exception or adding a warning somewhere

maartenc:
No I don't think this is the task of the parser.
In addition, the hang of this JUnit test for instance is because the file `javaparser-symbol-solver-testing/src/test/test_sourcecode/javaparser_src/proper_source/com/github/javaparser/ast/body/FieldDeclaration.java` contains a variable named `aux`. There isn't even an Aux.java file to warn the user about.

Maarten

maartenc:
An alternative place for this additional check could be in the `JavaParserTypeSolver#tryToSolveTypeUncached()` method where an unknown type is converted to a java sourcefile.

ftomassetti:
@matozoid for me it is +1

matozoid:
Okay, thanks for the contribution @maartenc :-)

PR #1966, opened 2018-12-14
[Bugfix] Fixed bug where the symbol solver can't resolve types in classes that…

jjbr:
… have an annotation as ancestor. The symbol solver tries to get the type parameters of parents which previously failed with an `UnsupportedOperationException`. The pull request includes both test and fix. Relates to #1833.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
@@ -192,7 +192,7 @@ public interface Context {
                 methodUsage = ((TypeVariableResolutionCapability) methodDeclaration)
                                       .resolveTypeVariables(this, argumentsTypes);
             } else {
-                return Optional.empty();
+                throw new UnsupportedOperationException();
             }
 
             return Optional.of(methodUsage);
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
@@ -192,7 +192,7 @@ public interface Context {
                 methodUsage = ((TypeVariableResolutionCapability) methodDeclaration)
                                       .resolveTypeVariables(this, argumentsTypes);
             } else {
-                throw new UnsupportedOperationException();
+                return Optional.empty();
             }
 
             return Optional.of(methodUsage);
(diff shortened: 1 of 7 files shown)


matozoid:
Thanks! Looks fine to me.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
```diff
@@ -192,7 +192,7 @@
                 methodUsage = ((TypeVariableResolutionCapability) methodDeclaration)
                                       .resolveTypeVariables(this, argumentsTypes);
             } else {
-                throw new UnsupportedOperationException();
+                return Optional.empty();
```
ftomassetti: mmm, why a method declaration should not have the TypeVariableResolutionCapability? Is this ok?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionAnnotationDeclaration.java
```diff
@@ -154,8 +151,8 @@ public String getName() {
 
     @Override
     public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
-        // TODO #1840
-        throw new UnsupportedOperationException();
+        // Annotation declarations cannot have type parameters - i.e. we can always return an empty list.
+        return Collections.emptyList();
```
ftomassetti, line 160: Good! Maybe we can move this to the javadoc?

ftomassetti (commented):
I have just a doubt about the Optional.empty

jjbr:
I've now added a couple of Javadoc comments and also fixed the corresponding methods in the other AnnotationDeclarations. The `UnsupportedOperationException` is back as well, that behaviour seems more desirable than to silently return an unresolved symbol, you're completely right!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/Context.java
```diff
@@ -192,7 +192,8 @@
                 methodUsage = ((TypeVariableResolutionCapability) methodDeclaration)
                                       .resolveTypeVariables(this, argumentsTypes);
             } else {
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException("Resolved method declarations should have the " +
+                                                        TypeVariableResolutionCapability.class.getName() + ".");
```
matozoid, line 196: Thanks so much for adding a message!

ftomassetti:
Updating branch. When tests are passing I would consider this good to go

Issue #1568, opened 2018-05-13
Wrong placement of spaces when adding modifiers to a method with annotations


ThLeu:
The other issue.

Result of addingModifiersWithExistingAnnotations
```
class X {
  @Test
  public final   void testCase() {
  }
```
one space too many gets added

Result of addingModifiersWithExistingAnnotationsShort
```
@Overridepublic final  void A(){}
```
the space seems to get placed at the wrong position

```
    @Test
    public void addingModifiersWithExistingAnnotationsShort() {
        MethodDeclaration it = consider("@Override void A(){}");
        it.setModifiers(EnumSet.of(Modifier.PUBLIC, Modifier.FINAL));
        assertTransformedToString("@Override public final void A(){}", it);
    }

    @Test
    public void addingModifiersWithExistingAnnotations() {
        considerCode(
                "class X {" + EOL +
                        "  @Test" + EOL +
                        "  void testCase() {" + EOL +
                        "  }" + EOL +
                        "}" + EOL
        );

        cu.getType(0).getMethods().get(0).addModifier(Modifier.FINAL, Modifier.PUBLIC);

        String result = LexicalPreservingPrinter.print(cu.findCompilationUnit().get());
        assertEqualsNoEol("class X {\n" +
                "  @Test\n" +
                "  public final void testCase() {\n" +
                "  }\n" +
                "}\n", result);
    }
```

ThLeu:
Might be connected with #1567

PR #1969, opened 2018-12-16
[TESTCASE] for lexical preservation and whitespace after modifiers.

matozoid:
See #1568


ftomassetti:
The first test fails because a newline was expected after the annotation typically. Given it is not there the lexical preserving printer does not add it because it thinks it is a choice of the user do not have one. Of course the resulting code does not compile...

ftomassetti:
The result from `extractReshuffledDiffElements` is not good I think...

matozoid:
@ftomassetti - do you mean this is a pointless test case? What is not good about the result?

ftomassetti:
Wasn't this integrated in #1994 ?
I think we should close it

matozoid:
Okay!

Issue #1963, opened 2018-12-12
UnsolvedSymbolException when using static final field from inherited interface


maartenc:
Hi,

consider the following example:

```
JDialog dialog = ...;
dialog.setDefaultCloseOperation(JDialog.DISPOSE_ON_CLOSE);
```

when trying to resolve the `JDialog.DISPOSE_ON_CLOSE` expression, an UnresolvedSymbolException is thrown. This happens because the DISPOSE_ON_CLOSE is not a direct field of the JDialog class, but is a field of the WindowConstants interface which is implemented by JDialog.

I have it fixed on a local checkout of the project, I'll try to create a PR for this.

Maarten

matozoid:
Great!

maartenc:
Thanks for merging the PR :-)

PR #1971, opened 2018-12-16
Fix for issue #1963

maartenc:
The ReflectionClassAdapter#hasField() method should also check fields from implemented interfaces. The implementation is now consistent with the getField() method. This fixes issue #1963.


ftomassetti:
Nice stuff @maartenc !

PR #1972, opened 2018-12-17
Fixes a RuntimeException when resolving a methodcall with a generic array as parameter.

maartenc:
Resolving the following construct will result in a RuntimeException:

```
List<String> foo = new ArrayList<>();
foo.toArray(new String[1]);

```
This PR fixes this.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
```diff
@@ -255,6 +255,12 @@ private void inferTypes(ResolvedType source, ResolvedType target, Map<ResolvedTy
             mappings.put(target.asTypeParameter(), source);
             return;
         }
+        if (source.isArray() && target.isArray()) {
+            ResolvedType sourceComponentType = source.asArrayType().getComponentType();
+            ResolvedType targetComponentType = target.asArrayType().getComponentType();
```
matozoid, line 260: The component type for `int[][][]` is `int[][]`, and its element type is `int`. Does this code work for multidimensional arrays too?

maartenc: Good question.
If you change the method parameter in the test to String[][][] it seems to work...

matozoid: Good enough for me :-)

matozoid:
Thanks! If @ftomassetti is also happy we'll merge it.

ftomassetti:
Infer types is such a cool feature! And it looks good to me. I just updated the branch, let's see what happens to test but for me it is good to go

ftomassetti:
Thank you @maartenc for your awesome contributions!

ftomassetti (approved):
Approved, pending tests

Issue #851, opened 2017-03-09
Rename AccessSpecifier.DEFAULT to PACKAGE_PRIVATE


matozoid:
Since `default` is now a keyword and a modifier, this is becoming confusing. Looking at https://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html `package-private` seems to be a good term.

PR #1975, opened 2018-12-17
Create a Modifier node (attempt 2)

### LINKED ISSUE #912: FieldDeclaration or MethodDeclaration should not auto-reorder the nodes (or provide a way to get the original raw text if necessary)

matozoid:
`Modifier` is no longer an enum, it's a real `Node` now, so it's officially part of the AST. The enum is still there, it's called `Modifier.Keyword` now. `Modifier` has a field of that type.
Modifiers are no longer stored in an `EnumSet`, but in a `NodeList`.

Fixes #912 

This reverts commit 9a03bc62


matozoid:
@ftomassetti can you help me with the lexical preservation part?

matozoid:
I'll have to piggy-back this one on this PR: #851

matozoid:
Back to the AccessSpecifier thing: I would agree that the access *level* or *scope* is a separate thing from the modifiers. We could have a seperate class for that.

matozoid:
All the remaining stuff is "dunno" for me.

ftomassetti:
> @ftomassetti can you help me with the lexical preservation part?

Ok, where I should look? There are 131 files changed and I got lost :D

matozoid:
Check why the tests fail - there seem to be only one or two distinct types of failure :relaxed:

ftomassetti:
PR on your PR sent

matozoid:
PR merged!

ftomassetti:
This is a big one!

matozoid:
Yep, part of the API breaking in minor releases promise/threat :-)

Issue #1667, opened 2018-06-25
'UnsupportedOperationException: removed child' with LexicalPreservingPrinter


rnorth:
Thanks for creating this project - it looks excellent!
I'm afraid I've hit a problem that occurs when using `LexicalPreservingPrinter`. I've included a code sample below that hits the problem on v3.6.10.

It could well be that I'm doing something wrong but it's unclear, especially as the problem does not occur when not using `LexicalPreservingPrinter`. (My intention in this code sample was simply to change the type of a variable; I reduced it to this from a larger trial refactoring).

```java
public class Main {

    public static void main(String[] args) throws Exception {
        String source = "public class Foo {\n" +
                "    public void foo() {\n" +
                "        Bar bar = new Bar();\n" +
                "    }\n" +
                "}";

        doWithoutPrettyPrint(source); // prints out as expected

        doWithPrettyPrint(source); // throws exception during transform:
        // Exception in thread "main" java.lang.UnsupportedOperationException: removed child(ClassOrInterfaceType) vs "Bar"   <89>   (line 3,col 9)-(line 3,col 11)
        //	at com.github.javaparser.printer.lexicalpreservation.Difference.applyRemovedDiffElement(Difference.java:389)
    }

    private static void doWithoutPrettyPrint(String source) {
        final CompilationUnit cu = JavaParser.parse(source);

        transform(cu);

        System.out.println(cu.toString());
    }

    private static void doWithPrettyPrint(String source) {
        CompilationUnit cu = JavaParser.parse(source);
        cu = LexicalPreservingPrinter.setup(cu);

        transform(cu);

        System.out.println(LexicalPreservingPrinter.print(cu));
    }

    private static void transform(CompilationUnit cu) {
        cu.findAll(VariableDeclarator.class).stream()
                .forEach(v -> {
                    v.setType("Baz");
                });
    }
}
```

matozoid:
Maybe @ThLeu can help? Or else @ftomassetti ?

ThLeu:
Hey @matozoid 
Sure, I can have a look at it, it may take some time though.
Will have a look as soon as I have a little more free time.

You can assign it to me if you like.

matozoid:
@ThLeu that would be great! Do you have anything that could help @rnorth right now?

ThLeu:
I fear not, after looking into it for a few minutes, it seems that this case is simply not supported yet or something with token and child element gets mixed up. Would have to look deeper into it first.

[...]

PR #1973, opened 2018-12-17
Re-add dependency for older junit tests

matozoid:
The build missed a lot of tests for a few months. Let's see how big the damage is.

PR #1978, opened 2018-12-18
[TESTCASE] failing testcase for #1667

MarvinWyrich:
I have reproduced the problem in the smallest possible way and hope that the test is in the right place. As noted by @rnorth, the problem does not occur when not using `LexicalPreservingPrinter`.

### DIFF
--- a/javaparser-core-testing/pom.xml
+++ b/javaparser-core-testing/pom.xml
@@ -92,6 +92,10 @@
             <artifactId>junit-vintage-engine</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.jbehave</groupId>
             <artifactId>jbehave-core</artifactId>
             <version>4.3.5</version>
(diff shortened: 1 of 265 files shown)


matozoid:
@MarvinWyrich - it seems to be passing - is your project on the latest version of JP?

MarvinWyrich:
Yes, I forked the current master branch and worked on it. If you are referring to the Travis checks, it seems that not all tests from javaparser-core-testing are executed, probably due to [this issue](https://github.com/javaparser/javaparser/pull/1973). But I need you to help me out here, because I'm not an expert on this.

And this is how it looks like in Eclipse (at least for me):

failing unit test

matozoid:
Ah right, I'll have to finish #1973 quickly then.

ftomassetti:
Updating the branch to benefit from #1973

ftomassetti:
Ok, this seems to be due to the fact that the type is not recognized as a child of the FieldDeclaration, because of the way fields are dealt with in FieldDeclaration. @matozoid this is biting us from time to time :)

So basically there is an issue because the NodeText sees "String" as a token, it does not understand it represent a child of the node. So when he expects to remove a child in the NodeText it cannot be found and the whole thing explode.

ftomassetti:
PR based on this proposed in #1993 

Thank you for submitting the failing test case. When that PR is accepted this one should be closed

matozoid:
@ftomassetti - I know, if you know of a real way to improve this, please open an issue.

MarvinWyrich:
> PR based on this proposed in #1993
>
> Thank you for submitting the failing test case. When that PR is accepted this one should be closed

Thank you very much for your time and contributions to this project. Since the linked PR was accepted I close this one.

PR #1979, opened 2018-12-18
UnsupportedOperationException when resolving method invocations on unbounded/extends generic types.

maartenc:
This PR fixed exceptions when resolving methods calls like:

```
Map<String, ?> map = ...;
map.get("").toString();

Map<String, ? extends Foo> map = ...;
map.get("").toString();
```

thanks,
Maarten


matozoid:
Good stuff!

ftomassetti (approved):
Big plus! Thank you!

PR #1981, opened 2018-12-19
Fixed #1980: Underscores in doubles are accounted for now.

### LINKED ISSUE #1980: Parsing of doubles fails in certain cases
An interesting edge case this one. According to the JLS you can define numeric values with underscores at any position (for code readability). In other words, `100000.0 == 100_000.0 == 1_0_0_0_0_0.0`. https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.2

However, according to the documentation of e.g. `java.lang.Double#valueOf(java.lang.String)` (which is referenced from `Double#parseDouble(String)` to use the same format) these underscores cause a `NumberFormatException` when parsing. https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#valueOf-java.lang.String-

In other words, it is not possible to blindly trust methods such as `Double#parseDouble` to parse any double as specified in Java source code.

**Update: I just found that we already consider this problem in ~some~ most cases.** It only appears to be missing in the `DoubleLiteralExpr#asDouble` case. Still leaving the text above for future reference, in case someone wonders why underscores need to be replaced in the string. Will submit a fix shortly.

jjbr:


ftomassetti:
+1

matozoid:
@jjbr do you happen to know if the same problem occurs with integers and longs?

jjbr:
> @jjbr do you happen to know if the same problem occurs with integers and longs?

The fundamental issue is the same for integers and longs but JavaParser already filters the underscores in those cases. I think it was simply forgotten in double (or perhaps someone thought it can't happen with doubles).

matozoid:
Thanks!

PR #1982, opened 2018-12-20
Returning from type search earlier if matching non-asterisk import was found unresolvable 

jjbr:
If an import for e.g. `a.b.Foo` is found while searching for the type `Foo` then we can stop looking any further, even if we cannot resolve `a.b.Foo`. Even if we were to find a `Foo` in e.g. `java.lang` later on, that result would be incorrect. This change hence improves performance and resolution accuracy in some edge cases.


ftomassetti:
Makes sense to me. Merging

Issue #1983, opened 2018-12-21
Comparing the type parameters when there are LazyType objects


zelinzhao:
I use Javaparser to analyze the following code.
```
public abstract class AbstractClass<E> {
	abstract public void method(Father<E> father);
}

public class Father<E> {
	E father;
	public Father(E father){
		this.father=father;
	}
}
public class Example extends AbstractClass<Integer>{
	public static class Son extends Father<Integer>{
		public Son(Integer father) {
			super(father);
		}
		int son;
	}

	@Override
	public void method(Father<Integer> father) {
        //MethodCallExpr here
		method((Son)father);
	}
	public boolean method(Son son) {
		return true;
	}
}
```
The code is actually from Tomcat8 and I change and remove some code for simplification.

The problem is that, I use Javaparser to parse MethodCallExpr `method((Son)father);` and call `resolve()` method on this MethodCallExpr. Javaparser then tries to figure out which `method` is called in this MethodCallExpr, `void method` or `boolean method`. Javaparser compares the type parameters in `compareConsideringTypeParameters` method in `ResolvedReferenceType` class.

The `compareConsideringTypeParameters` method calls `typeParametersValues()` on `boolean method(Son son)` to get the type parameter of `Son` class, and returns a `LazyType` object, whose `concrete` is `Integer`.

In the following comparison, the type parameters of `Father<Integer>` and `Son` are different. However, the two type parameters are both `Integer` and they should be equivalent.

I think we can compare the two type parameters here via class name.
```
    if(thisParam.describe().equals(otherParam.describe()))
    	return true;
```
I tried this locally and the Javaparser works as expected.

I don't know whether this is a real bug or Javaparser is designed in this way.
I will make a PR so that you can review the code. And there may be a better way to fix this if this is a real bug.

jlerbsc:
I think that [ABANDONNED] means not to be released. If this this is not the case you are free to reopen this PR.

PR #1984, opened 2018-12-21
Comparing type parameters via class names if there are LazyType objects

zelinzhao:
#1983


javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
```diff
@@ -471,6 +470,10 @@ protected boolean compareConsideringTypeParameters(ResolvedReferenceType other)
             for (int i = 0; i < typeParametersValues().size(); i++) {
                 ResolvedType thisParam = typeParametersValues().get(i);
                 ResolvedType otherParam = other.typeParametersValues().get(i);
+                //Sometimes, thisParam and otherParam may be LazyType.
```
ftomassetti: nitpick: I would use a space after the slash and a space between the if and the parenthesis

ftomassetti (approved):
To me it seems ok, however tests fail

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
ftomassetti: Perhaps that means that two types can be different but have the same description?

zelinzhao: Hi, @ftomassetti, thanks for the advice. I will be careful about the code format. 

I actually want to compare the qualified name of two types and I know the `describe()` method doesn't return the class name. I can call `getQualifiedName()` method on `ResolvedReferenceType` object, however in `LazyType`, there is only a `describe()` method.

How about the following patch? This is not a perfect way, but I think this is more robust than `describe()` method.
```
if (thisParam.isReferenceType() && otherParam.isReferenceType() 
    && thisParam.asReferenceType().getQualifiedName().equals(otherParam.asReferenceType().getQualifiedName()))
    return true;
```


I tried to import `LazyType` in this `ResolvedReferenceType` class, however, this seems failed because of the dependency between Javaparser-core and Javaparser-symbolsolver-core.
If we can access the `LazyType` in this method, we can get the concrete type from the `LazyType`.

ftomassetti: This seems reasonable, let's try that

matozoid:
@zelinzhao @ftomassetti - how is this PR going?

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
```diff
@@ -471,6 +470,10 @@ protected boolean compareConsideringTypeParameters(ResolvedReferenceType other)
...
+                if(thisParam.describe().equals(otherParam.describe()))
```
ftomassetti:
```suggestion
if (thisParam.isReferenceType() && otherParam.isReferenceType() 
        && thisParam.asReferenceType().getQualifiedName().equals(
                otherParam.asReferenceType().getQualifiedName())) {
    return true;
}
```

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
ftomassetti:
```suggestion
                // Sometimes, thisParam and otherParam may be LazyType.
```

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
ftomassetti:
```suggestion
if (thisParam.isReferenceType() && otherParam.isReferenceType() 
        && thisParam.asReferenceType().getQualifiedName().equals(
                otherParam.asReferenceType().getQualifiedName()))
```

ftomassetti:
Added two suggestions, as derived from the discussion. Updated the branch

matozoid:
Looks like @zelinzhao has abandoned this PR. @ftomassetti - can we rescue it?

matozoid:
Continued in https://github.com/javaparser/javaparser/pull/2048

PR #1987, opened 2018-12-21
Issue 1907 (Javadoc comments spacing)

### LINKED ISSUE #1907: Error in PrettyPrintVisitor.visit
I just updated from version 3.5.2 to 3.6.27. Now I have following issue:

Take this class:

```java
package de.baumato.loc;


public class SomeClass
{

    private int x;
    private int y;
		public SomeClass(int x, int y)
		{
			super();
			this.x = x;
			this.y = y;
		}
    
		
		/**
		 * mutli line comment	
		 */
		public void add(int x, int y) 
		{
			// single line comment
			this.x+= x;
			this.y+= y;
		}


		
		public int getX() 
		{
			return x;
		}


		
		public int getY() 
		{
			return y;
		}
		
		
}
```
and process it with:

`String code = JavaParser.parse(someClass).toString()`

then the resulting `code` always contains one space more in front of 'multi line comment' than the original source code.

I assume it has something to do with the logic in 
`PrettingPrintVisitor.visit(final JavadocComment n, final Void arg)`
The variable `prependSpace` is always evaluated to true for multi line comments.

matozoid:
Fixes #1907


ftomassetti (commented):
Two things:
1) There is a typo (mutli -> multi, two instances)
2) The tests pass for me when running it on the IDE or from the command line on mac, but I saw you commented about this in #1907 :)

PR #1993, opened 2018-12-24
Fix issue 1667 - handle field declaration type resolution

### LINKED ISSUE #1667: 'UnsupportedOperationException: removed child' with LexicalPreservingPrinter

ftomassetti:
Fix #1667 by handling the special case of types of field declarations.

Thank you @MarvinWyrich for your test case


ftomassetti:
Updating.
If tests pass @matozoid is ok for merging this one?

matozoid:
Fine with me, or do you want me to look at something specific @ftomassetti?

ftomassetti:
Nope. I will just update and wait the tests to pass before merging.

PR #1998, opened 2018-12-26
perform duplicate checking on all addImport methods

signed:
before this there was no duplicate checking for addImport(ImportDeclaration). This came as a surprise to me.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/builders/CompilationUnitBuildersTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/builders/CompilationUnitBuildersTest.java
@@ -31,6 +31,7 @@ import java.lang.annotation.ElementType;
 import java.util.List;
 import java.util.Map;
 
+import static com.github.javaparser.JavaParser.parseImport;
 import static com.github.javaparser.ast.Modifier.Keyword.PRIVATE;
 import static com.github.javaparser.utils.Utils.EOL;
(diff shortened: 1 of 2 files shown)


ftomassetti:
Well, I think that the API to build the AST should let you build an AST that is illegal for a variety of reasons. However we could have checks to later catch those errors.

signed:
I agree building an invalid AST should be possible.
And there are still ways to do it.

```
cu.setImports(NodeList<ImportDeclaration>)
cu.setImport(index, ImportDeclaration)
cu.getImports().add(ImportDeclaration)
```

My point is about api consistency and the Principle of least astonishment.
Initially I used `cu.addImport(Class<?>)` and for duplication purposes later moved to `cu.addImport(ImportDeclaration)`. Given it is an overloaded method I expected the same contract. That not being the case surprised me.

matozoid:
It's caused by the major Modifier overhaul from the last version, and I'm happy you're here to clean up the problems :-) It used to be a `Set`, so duplicates were impossible. I agree that the helper methods should help as much as possible, and the "basic" methods should be dumb.

However, conflicts :-(

signed:
@matozoid resolved the conflicts

matozoid:
@signed resolved more conflicts :-D

signed:
@matozoid travis download failed again. But otherwise this should be ready to merge.

matozoid:
Thanks again, @signed!

Issue #1999, opened 2018-12-26
anonymous class broken when statement is set as new method body


signed:
The following test fails when I run it as part of [MethodDeclarationTransformationsTest](https://github.com/javaparser/javaparser/blob/master/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java) on current head.
```
    @Test
    public void replaceBodyShouldNotBreakAnonymousClasses() {
        MethodDeclaration it = consider("public void method() { }");
        it.getBody().ifPresent(body -> {
            Statement statement = JavaParser.parseStatement("Object anonymous = new Object() {\n" +
                    "};");
            NodeList<Statement> statements = new NodeList<>();
            statements.add(statement);
            body.setStatements(statements);
        });

        String expected = "public void method() {\n" +
                "    Object anonymous = new Object() {\n" +
                "    };\n" +
                "}";
        assertTransformedToString(expected, it);
    }
```

ftomassetti:
Ok, this represents a new test that should be added, right? I will prepare a PR with this test. Thank you for submitting it.

ftomassetti:
To me the issue seems that we do not have a way to represent an empty body for the anonymous class

ftomassetti:
Proposing a PR in #2000

signed:
Thank you for looking into it.

ftomassetti:
My pleasure

PR #2000, opened 2018-12-27
[TESTCASE+Solution] adding test replaceBodyShouldNotBreakAnonymousClasses and fixing it

ftomassetti:
See #1999 . Thank you @signed for providing this

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser.printer.lexicalpreservation.transformations.ast.body;
 
+import com.github.javaparser.JavaParser;
 import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.MethodDeclaration;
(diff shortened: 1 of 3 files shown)


ftomassetti:
Updated to include a few derived tests and a solution

matozoid:
Well, if it's a testcase+solution, then it's just a normal PR, right? :-)

ftomassetti:
Yes, probably it is but I wanted to underline the fact I took the testcase prepared by someone else

matozoid:
Oh okay. I assume a [label] means "don't pull this PR", so no need for that.

matozoid:
Yay, @ftomassetti, you've got an EOL mistake :-/

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java
```diff
@@ -431,4 +434,40 @@ void removingAnnotationsWithSpaces() {
                         "}\n", result);
     }
 
+    @Test
+    public void parseAndPrintAnonymousClassExpression() {
+        Expression expression = JavaParser.parseExpression("new Object() {" + EOL +
+                "}");
+        String expected = "new Object() {\n" +
```
ftomassetti:
```suggestion
         String expected = "new Object() {" + EOL +
```

ftomassetti:
Let's see if this fix it

matozoid:
It did :-)

Issue #1976, opened 2018-12-17
Move all tests to JUnit 5


matozoid:

matozoid:
@signed is doing this one

PR #2002, opened 2018-12-30
Migrate tests to junit 5

signed:
This solves most of #1976.
As far as I know there is no junit5 version of jBehave.
All the other tests have been moved to junit5.

### DIFF
--- a/javaparser-core-testing/pom.xml
+++ b/javaparser-core-testing/pom.xml
@@ -92,6 +92,10 @@
             <artifactId>junit-vintage-engine</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.hamcrest</groupId>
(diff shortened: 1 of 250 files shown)


ftomassetti:
We should really be careful that all the tests keep running in all modules which have tests

ftomassetti:
In this PR the tests run are:
* 1463
* 11
* 928 / 825 (the tests are run twice and the number varies)

ftomassetti:
In master we had the same tests running so from my point of view this seems ok

matozoid:
Very cool, thank you very much @signed!

signed:
The least I could do. I need the migration for other projects and I could not have done it without javaparser.

matozoid:
For most users, the least they can do is nothing at all, so thank you anyway ;-)

PR #2003, opened 2019-01-02
Give LambdaExpr some useful constructors.

### LINKED ISSUE #1997: Why is it valid to create a `LambdaExpr` with `isEnclosingParameters=false` with non-single parameter list?
Hey,

```java
new LambdaExpr(new NodeList<>(), ..., false)
```
would result in illegal Java code, i.e. (`-> ...`, without params).
Is there maybe a way to ignore the `isEnclosingParameters` param in case it doesn't make sense?
Also, for multiple parameters it should also be ignored, i.e. the `isEnclosingParameters` parameter only makes sense for parameter lists with single values.

Would it make sense to only include this option for a constructor that accepts a single parameter and make it `true` for `NodeList`s?

matozoid:
Fixes #1997

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/LambdaExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/LambdaExprTest.java
@@ -2,20 +2,24 @@ package com.github.javaparser.ast.expr;
 
 import com.github.javaparser.TokenRange;
 import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.body.Parameter;
+import com.github.javaparser.ast.type.UnknownType;
 import org.junit.jupiter.api.Test;
 
 import static com.github.javaparser.JavaParser.parseExpression;
-import static org.junit.jupiter.api.Assertions.*;
+import static com.github.javaparser.JavaParser.parseStatement;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 class LambdaExprTest {
     @Test
-    void lambdaRange1(){
+    void lambdaRange1() {
         Expression expression = parseExpression("x -> y");
         assertRange("x", "y", expression);
     }
 
     @Test
-    void lambdaRange2(){
+    void lambdaRange2() {
         Expression expression = parseExpression("(x) -> y");
         assertRange("(", "y", expression);
     }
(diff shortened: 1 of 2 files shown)


matozoid:
I guess there could be a check to see if you're passing a single parameter to the multi-parameter constructor and not add () then... but most utility constructors are very straightforward.

@paplorinc - what do you think?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/LambdaExprTest.java
```diff
@@ -27,15 +31,45 @@ private void assertRange(String startToken, String endToken, Node node) {
...
+    @Test
+    void oneParameterAndExpressionUtilityConstructor() {
+        LambdaExpr expr = new LambdaExpr(new Parameter(new UnknownType(), "a"), parseExpression("5"));
+        assertEquals("a -> 5", expr.toString());
+    }
+
+    @Test
+    void oneParameterAndStatementUtilityConstructor() {
+        LambdaExpr expr = new LambdaExpr(new Parameter(new UnknownType(), "a"), parseStatement("return 5;"));
+        assertEquals("a -> return 5;", expr.toString());
```
l0rinc: I don't think this is valid Java code...

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/LambdaExprTest.java
```diff
@@ -27,15 +31,45 @@ private void assertRange(String startToken, String endToken, Node node) {
...
+    @Test
+    void multipleParametersAndExpressionUtilityConstructor() {
+        LambdaExpr expr = new LambdaExpr(new NodeList<>(new Parameter(new UnknownType(), "a"), new Parameter(new UnknownType(), "b")), parseExpression("5"));
+        assertEquals("(a, b) -> 5", expr.toString());
+    }
+
+    @Test
+    void multipleParametersAndStatementUtilityConstructor() {
+        LambdaExpr expr = new LambdaExpr(new NodeList<>(new Parameter(new UnknownType(), "a"), new Parameter(new UnknownType(), "b")), parseStatement("return 5;"));
+        assertEquals("(a, b) -> return 5;", expr.toString());
```
l0rinc: same

l0rinc (changes requested):
Thanks

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/LambdaExprTest.java
matozoid: Ooooooops!

PR #2004, opened 2019-01-02
resolve dependency in the install phase and cache them between builds

signed:
As far as I understand the travis documentation install should only resolve dependencies.
With the current `mvn install` command you do way more, `dependency:resolve` should be enough.


signed:
hmm, I was not able to get it working with a multi module build.
postpone for now.

Issue #1845, opened 2018-09-10
JEP 325: switch expressions


matozoid:
http://openjdk.java.net/jeps/325

- [x] Create SwitchExpr
- [x] add SwitchExpr() to grammar
- [x] add "value break"
- [x] add funky -> syntax
- [x] support multiple case values separated by commas.
- [x] add language level validation (don't allow SwitchExpr below JDK12, allow only certain types of labels and amounts of labels)
- [x] add language level validation (value break only allowed inside switch)
- [x] write Javadoc

matozoid:
https://bugs.openjdk.java.net/browse/JDK-8192963

matozoid:
https://bugs.openjdk.java.net/browse/JDK-8207241

matozoid:
@ftomassetti there's work for the symbol solver here. In the new JLS, 5.6.3 describes numeric promotion with switch expressions.

matozoid:
If I read the JLS change correctly...
- `case` can now take multiple arguments separated by `,`. `SwitchEntryStmt.label` has to become a `NodeList<Expression>`.
- `switch` is no longer a statement but an expression (so it should be wrapped in an `ExpressionStmt` when it is used "normally".)
- besides statements, we now als have `case` expressions that use `->` instead of `:`. I guess the most straightforward thing to do is to rename `SwitchEntryStmt` to `SwitchEntry` and give it a type enum and possible some extra fields. (The spec says the statements and expressions can't mix, but that's a nice semantic validation.)

I think that's all.

PR #2008, opened 2019-01-04
Add SwitchExpr node

matozoid:
Here's the first step towards supporting switch expressions. See #1845.

- a SwitchExpr node now exists.
- a SwitchExpr is created when a `switch` is found as part of an expression, according to the JLS grammar.

This does not enable you to do anything useful. It is only a first step, as can be seen from the test cases (all useful cases are disabled)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/SwitchExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/SwitchExpr.java
@@ -18,48 +18,49 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  */
-package com.github.javaparser.ast.stmt;
+package com.github.javaparser.ast.expr;
 
+import com.github.javaparser.TokenRange;
 import com.github.javaparser.ast.AllFieldsConstructor;
+import com.github.javaparser.ast.Generated;
+import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.NodeList;
-import com.github.javaparser.ast.expr.Expression;
-import com.github.javaparser.ast.expr.NameExpr;
 import com.github.javaparser.ast.nodeTypes.SwitchNode;
 import com.github.javaparser.ast.observer.ObservableProperty;
+import com.github.javaparser.ast.stmt.Statement;
+import com.github.javaparser.ast.stmt.SwitchEntryStmt;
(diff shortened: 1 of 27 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/SwitchExpr.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.Optional;
+import java.util.function.Consumer;
+import static com.github.javaparser.utils.Utils.assertNotNull;
+import com.github.javaparser.metamodel.SwitchExprMetaModel;
+
+/**
+ * A switch expression.
+ * <br/>In <code>switch(a) { ... }</code> the selector is "a",
+ * and the contents of the { ... } are the entries.
+ *
```
matozoid, line 46: The Javadoc is bogus, I'll fix it later.

javaparser-core/src/main/javacc/java.jj
```diff
@@ -1693,17 +1693,19 @@ Expression UnaryExpressionNotPlusMinus():
...
+		(
+			"~" { op = UnaryExpr.Operator.BITWISE_COMPLEMENT; begin=token(); } |
+			"!" { op = UnaryExpr.Operator.LOGICAL_COMPLEMENT;     begin=token(); }
+		) ret = UnaryExpression() { ret = new UnaryExpr(range(begin, token()), ret, op); }
+		| LOOKAHEAD( CastExpression() )
+			ret = CastExpression()
+		|
+			ret = PostfixExpression()
+		|
+			ret = SwitchExpression()
```
matozoid, line 1706: It's not very clear since I reformatted everything, but this choice is everything I added here.

javaparser-core/src/main/javacc/java.jj
```diff
@@ -2227,6 +2229,21 @@ SwitchStmt SwitchStatement():
...
+	SwitchEntryStmt entry;
+	NodeList<SwitchEntryStmt> entries = emptyList();
+	JavaToken begin;
+}
+{
+  "switch" {begin=token();} "(" selector = Expression() ")" "{"
+    ( entry = SwitchEntry() { entries = add(entries, entry); } )*
+  "}"
+
+  { return new SwitchExpr(range(begin, token()), selector, entries); }
```
matozoid, line 2244: This is and will probably stay the same as SwitchStatement(), not sure if javacc has a nice way of combining the two.

PR #2011, opened 2019-01-08
Run Maven in non-interactive mode to reduce log size

mernst:
Passing `-B` to Maven reduces log size by omitting progress messages such as "Progress: 125/150kB".  It also removes control characters from the output.


matozoid:
That looks a lot better, thanks :-)

PR #2012, opened 2019-01-08
Prefer ...ForEachStmt to ...ForeachStmt

mernst:
In version 3.7.0, the class ForeachStmt was renamed to ForEachStmt.
Several methods have multiple spellings:
```
  asForEachStmt
  asForeachStmt
  ifForEachStmt
  ifForeachStmt
  isForEachStmt
  isForeachStmt
  toForEachStmt
  toForeachStmt
```
This pull request deprecates the version of the method whose capitalization is inconsistent with the argument type and/or return type.


matozoid:
If you manage to get to the deprecated versions of these you have changed the capitalization already, so you are aware of the change. We can just as well throw them away (as was my intention)

matozoid:
Thanks by the way, I would never have noticed!

mernst:
OK, I'll let you make that change.  Thanks.

Issue #1876, opened 2018-10-08
Throw exception, when get ancestors of class that extended generic interface


ddepddep:
Hi, 

I want to get ancestors of some class or interface. (this could be in source file or class file.)

But when I try to get ancestors of the interface that extended generic interface, it throws exception.

Like below.
```
public interface A extends B<Long> {} 
```

This is my test code.

```
CombinedTypeSolver combinedTypeSolver = new CombinedTypeSolver(new ReflectionTypeSolver(false), new JavaParserTypeSolver(new File(srcDir), And more JarTypeSolver);
ResolvedReferenceTypeDeclaration rrtd = combinedTypeSolver.solveType("guru.springframework.repositories.ProductRepository.java");
rrtd.getAllAncestors(); // this code throws
```

And, this is exception stack trace 
```
Exception in thread "main" com.google.common.util.concurrent.UncheckedExecutionException: java.nio.file.InvalidPathException: Illegal char <<> at index 109: C:\Users\ddepddep\git\spring-boot-postgress-example\src\main\java/guru/springframework/repositories/CrudRepository<Product,Long>.java
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2218)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4147)
	at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:5053)
	at com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver.tryToSolveType(JavaParserTypeSolver.java:172)
	at com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver.tryToSolveType(CombinedTypeSolver.java:62)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.CompilationUnitContext.solveType(CompilationUnitContext.java:163)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.solveType(JavaParserInterfaceDeclaration.java:241)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.toReferenceType(JavaParserInterfaceDeclaration.java:331)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration.getAncestors(JavaParserInterfaceDeclaration.java:250)
[...]

PR #2013, opened 2019-01-09
Resolution fails on Windows for interfaces extending another typed interface.

maartenc:
Constructs like 

`public interface A extends B<C> {}`

results in 

`java.nio.file.InvalidPathException: Illegal char <<>`

exceptions on windows.


maartenc:
It looks like this issue has already been reported here: #1948 and #1876

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserInterfaceDeclaration.java
```diff
@@ -353,14 +353,18 @@ public ClassOrInterfaceDeclaration getWrappedNode() {
 
     private ResolvedReferenceType toReferenceType(ClassOrInterfaceType classOrInterfaceType) {
         SymbolReference<? extends ResolvedTypeDeclaration> ref = null;
-        String typeName = classOrInterfaceType.asString();
+        String typeName = classOrInterfaceType.getName().getId();
+        if (classOrInterfaceType.getScope().isPresent()) {
+            typeName = classOrInterfaceType.getScope().get().toString() + "." + typeName;
```
matozoid: Please use asString here, see http://javaparser.org/making-strings-in-javaparser/

javaparser-symbol-solver-testing/src/test/test_sourcecode/javaparser_new_src/javaparser-core/com/github/javaparser/ast/validator/TypedValidator.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.function.BiConsumer;
+
+/**
+ * A validator that validates a known node type.
+ */
+public interface TypedValidator<N extends Node> extends BiConsumer<N, ProblemReporter> {
+    /**
+     * @param node the node that wants to be validated
+     * @param problemReporter when found, validation errors can be reported here
```
matozoid: @ftomassetti is it okay to add new files in this directory?

matozoid:
Thanks!

maartenc:
ok, I've changed the `toString()` to `asString()`.

I did copy/past this code from the `JavaParserClassDeclaration#toReferenceType()` method, perhaps you should change it there as well?

Maarten

matozoid:
Okay, made a note to check for toString()s. This is now waiting for @ftomassetti (who seems incredibly busy)

javaparser-symbol-solver-testing/src/test/test_sourcecode/javaparser_new_src/javaparser-core/com/github/javaparser/ast/validator/TypedValidator.java
ftomassetti: No, it should basically be inserted anywhere else but here :)

ftomassetti: Of course also the Validator.java file should be moved. I would create a directory under the root of the resources folder and put both files there

ftomassetti (commented):
+1 if we move the files

maartenc:
Ok, I'll move the files and simplify them. 
I wanted to make this test similar to the other tests in that file, but I didn't knew that it was not ok to add new files to that directory.

I'll update this PR when I'm finished.

thanks for the feedback,
Maarten

maartenc:
I've removed the added files from the javaparser_new_src directory and reused some others I found under the resources directory. I hope that is ok?

matozoid:
+1 from me then!

maartenc:
Thanks Danny, maybe you could close the mentioned issues as well?
(#1948 and #1876)

Maarten

PR #2016, opened 2019-01-11
Added getLineCount() to the Range class. ref #1505

### LINKED ISSUE #1505: Range to have method like getLinesCount
Hi,
 
  do u mind if i will add the method getLinesCount into the Range class? I had found doing the following snippet quite some times... 

```
Range range = node.getRange().get();
if (range.end.line < range.begin.line) {
            return -1;
}
return range.end.line - range.begin.line + 1;
```

cdtpinto:
Fixes #1505


matozoid:
Thanks! Throw in a simple test if you can.

cdtpinto:
@matozoid will do!

matozoid:
Thanks!

PR #2017, opened 2019-01-11
Ignore bridged and synthetic methods when fetching the methods from a class using javassist

maartenc:
These methods are also filtered in the ReflectionTypeAdapter class and are causing issues later on in the resolution process.


matozoid:
Hopefully @ftomassetti can respond here.

ftomassetti:
Makes sense to me +1

ftomassetti:
Merging, thanks

PR #2022, opened 2019-01-14
Fix: retrieving the bounds on a JavassistTypeParameter could throw an UnsupportedOperationException if it extended another type.

maartenc:


ftomassetti (approved):
Looks good to me. Thank you for your contribution!

PR #2025, opened 2019-01-15
Switch arrow syntax

matozoid:
See #1845.

The `case 123 -> ...` syntax.

To do:
- implement symbol solving


matozoid:
@ftomassetti this is now failing on symbol solving. Heeeelp!

matozoid:
I've added multi-label case support so that this branch now represents full Java 12 support, EXCEPT the symbol solver part. That is broken. @ftomassetti ?

matozoid:
Everything solved now :-)

PR #2027, opened 2019-01-17
[TESTCASE] remove method modifiers should preserve indentation

signed:
I'm currently running version 3.8.3 and my [testcase](https://github.com/signed/convert-junit4-to-junit5/blob/extend/convert-junit4-to-junit5/src/test/java/jb/convert/ast/TestAnnotationConversionTest.java) has the proper indentation.This test breaks when I upgrade to 3.10.0 because of the indentation.
Is this a regression in the printer? I adjusted one of the existing tests to have the indentation I'm expecting.


matozoid:
There was a big change in how modifiers work. Let's see if @ftomassetti knows.

ftomassetti:
Yes, this makes sense in principle, however the whole preserving the indentation is tricky and brittle

signed:
For another [testcase](https://github.com/signed/javaparser/blob/new_line_part_of_codemodel/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/MethodDeclarationTransformationsTest.java#L229) I had a look at the ConcreteSyntaxModel and the diff calculation.
The current ConcreteSyntaxModel looks to me like it is optimized for pretty printing.
It does not track newline and whitespace characters in all the locations they can occure.
Therefore they are not taken into account when the diff is calculated.

For my particular test case I added a newline to the [MethodCallExpression](https://github.com/signed/javaparser/blob/new_line_part_of_codemodel/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java#L394).
This results in the newline character being available in the diff algorithm and the printed code to be formatted in the way I expect. But now the pretty print tests for method declarations are failing because of an additional newline that is printed.

Maybe there could be a 2nd Concrete Syntax Model optimized for Lexical Preserving printing?
It should contain/preserve all newline and whitespace tokens it encounters.
This way the information is available in the diff algorithm.

What do you thing?

matozoid:
#1689 might be of interest.

PR #2037, opened 2019-01-24
fix #2033, do not throw exception when incompleteList is set

### LINKED ISSUE #2033: UnsupportedOperationException thrown after setting acceptIncompleteList=true
https://github.com/javaparser/javaparser/blob/cb194db28ea321862242bd2472cd5d25011633a5/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java#L230

Here `UnsupportedOperationException` is thrown out even when `acceptIncompleteList=true`.
as `getCorrespondingDeclaration` called on unsolved `implementedDeclRef` later.

I think here should be another if branch to ignore unsolved `implementedDeclRef`, if acceptIncompleteList is _true_.

villadora:
Fixes #2033.


malteskoruppa:
Makes sense to me. :)

matozoid:
@malteskoruppa - any idea how this could be unit tested?

malteskoruppa:
If I understood @villadora correctly, he stumbled upon an `UnsupportedOperationException` when attempting to perform some resolution in some piece of code. With this fix, this exception should not be thrown anymore.

So all that would be needed to unit test this would be to take that piece of code and perform that resolution -- and check that an `UnsupportedOperationException` does *not* occur.

I would be hard-pressed to come up with code that would have triggered that exception on the spot, but I suppose the OP has it, since he did see that exception. :-) @villadora Could you perhaps provide us with that code?

ftomassetti:
I think this `acceptIncompleteList` is dangerous and could lead to wrong results, so we should state that. That said the suggested behavior makes sense

matozoid:
This one misses the release because there are suggestions waiting for an answer.

matozoid:
@villadora - are you still supporting this PR?

matozoid:
@ftomassetti - shall I rescue this PR? (Copy it to a branch of my own)

matozoid:
@ftomassetti - all this "incompleteList" stuff... It sounds like it could be a configurable flag instead of a parameter. Where could that be set?

ftomassetti:
Yes, I guess it could be a flag. I am still a bit worried about this feature

MysterAitch:
Closing this PR as it seems to no longer apply to the current code.

_( #2033 seems to have been fixed in dea27fa )_

PR #2038, opened 2019-01-24
Improved method resolution

jjbr:
It turns out that method resolution in Java is quite involved when it comes to overloaded methods. The `MethodResolutionLogic#isMoreSpecific` method was a bit too simplistic and has been improved with this pull request (see the tests added in this pull request for some examples). I don't want to guarantee a 100% compliance with the JLS (we'd have to look at larger parts of the symbol solver code to verify this) but it's better than before.


matozoid:
Thanks! But the build fails - did you run the slow tests too?

jjbr:
No, will have a look and fix the remaining failing test!

matozoid:
If you haven't figured it out yet: `mvn test -P AlsoSlowTests`

jjbr:
There was one logic flaw that resulted in occasional misidentifications whether a method is more specific when the argument type is an array type (to be passed to method with variadic parameter). Most of the test failures were due to the expected resolution not being 100% correct and with the improved resolution logic the symbol solver now occasionally resolved a method more accurately than before. I fixed the logic flaw and the tests - everything should be fine now. :slightly_smiling_face:

matozoid:
Thanks a lot!

malteskoruppa:
Good work! :-)

PR #2040, opened 2019-01-25
Issue 1888 - path equality -> file equality (ignore this)

MysterAitch:
Also happy to add unit tests (see #1888), though I'm not sure how best to do so given that the drive lettering is Windows-specific and `isSameFile` requires the files to actually exist.

Issue #1888, opened 2018-10-14
CollectionStrategy - directory comparison


MysterAitch:
It appears that if the collection strategy fails when there is a mismatch between using the windows-style paths with an initial drive letter versus other style paths (e.g. `C:/var/test`) with a leading slash (e.g. `/var/test`). 

```
        final ProjectRoot projectRoot =
                new ParserCollectionStrategy()
                        .collect(Paths.get(x));
```

```
        final ProjectRoot projectRoot =
                new SymbolSolverCollectionStrategy()
                        .collect(Paths.get(x).toAbsolutePath());
```

The comparison occurs on the following lines of the following files:

https://github.com/javaparser/javaparser/blob/4f682b22ae276ed7e18a2928cfdabd4813775aa3/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/utils/SymbolSolverCollectionStrategy.java#L68 

https://github.com/javaparser/javaparser/blob/4f682b22ae276ed7e18a2928cfdabd4813775aa3/javaparser-core/src/main/java/com/github/javaparser/utils/ParserCollectionStrategy.java#L60

In my use case that has made this visible, I am storing paths normalised to unix-style path string (not a perfect solution, but forward slashes do get explicitly converted to back-slashes on Windows so hopefully safe enough!).

A test case demonstrating the different results of path comparison using this technique is shown below. Note that exactly how Windows deals with paths with a leading slash can be variable, according to a few online resources (though I don't fully understand when it differs).

```
    @Test
    public void filePathComparisonTest() {
        Path x1 = Paths.get("/var/test/");
        Path x2 = Paths.get("C:/var/test/");
        Path x3 = Paths.get("C:\\var\\test\\");

        assertTrue(x1.equals(x2)); // Fails  (Windows 7)
        assertTrue(x2.equals(x3)); // Passes (Windows 7)
        assertTrue(x1.equals(x3)); // Fails  (Windows 7)

        assertTrue(x1.toAbsolutePath().equals(x2.toAbsolutePath())); // Passes (Windows 7)
        assertTrue(x2.toAbsolutePath().equals(x3.toAbsolutePath())); // Passes (Windows 7)
        assertTrue(x1.toAbsolutePath().equals(x3.toAbsolutePath())); // Passes (Windows 7)
    }
```

MysterAitch:
To resolve this, I suggest that the paths be converted to the absolute path (or some other normalised version). Example:

[...]

PR #2041, opened 2019-01-25
Issue 1888 - path equality -> file equality (attempt 2)

MysterAitch:
Changed from using `Path.equals` to `File.isSameFile`. See #1888. Also added null check (`Path.equals` would previously have simply returned false, while `Files.isSameFile` throws NPE).

(Attempt 2 as #2040 was getting extremely messy/unwieldy).


MysterAitch:
Phew, both checks passed. That was much harder work than it likely should have been - sorry for the noise!! :grimacing: :D 

I still think that `SourceRoot` should also receive the same fix, and the null check seems redundant given that if a null is passed to `path` then it'll be flagged higher up, and `dir` will always have a value as it is the current 'focus' of the visitor. 

I'll leave that alone and move on for now though!

matozoid:
I'm kind of warming up to the idea of fixing SourceRoot now :-)

But let's merge this one.

pkoziol:
Is there a reason why `SymbolSolverCollectionStrategy` was skipped in this PR (it was modified in #2040)?

I'm asking, because I replaced `ParserCollectionStrategy` with `SymbolSolverCollectionStrategy` and my program broke. Apparently `ParserCollectionStrategy` works for relative paths and `SymbolSolverCollectionStrategy` doesn't. If `SymbolSolverCollectionStrategy` would use `Files.isSameFile` everything would be OK.

matozoid:
I don't think there is a reason. Would you mind making a PR for this change?

Issue #2034, opened 2019-01-23
Should ForEachStmt#getVariable() return a VariableDeclarator instead of a VariableDeclarationExpr?


malteskoruppa:
Hey guys, sorry, haven't been around for some time, due to lots of work and deadlines... :-)

Here's a simple question. I noticed that the method `ForEachStmt#getVariable()` returns a `VariableDeclarationExpr`. Wouldn't it be better if it returned a `VariableDeclarator` instead?

The reason is this: In any foreach statement, *exactly one* variable is declared (not zero, not two or more). Therefore if I want to get information about that variable, I have to basically write `forEachStmt.getVariable().getVariable(0)`, which is a bit convoluted. In addition, it is not "statically safe" to just access the 0'th variable: To write "clean" code, I would have to check that `foreachStmt.getVariable().getVariables().size() == 1` before accessing it. This however, feels a bit useless, as it should *always* be true and I should be able to trust JavaParser that this is indeed the case.

So I was thinking it would be nicer to have `ForEachStmt#getVariable()` return the single `VariableDeclarator` instead, without wrapping it in a `VariableDeclarationExpr`.

Are there good reasons not to do so, and if yes, what are they? :-)

matozoid:
I had to look it up, but yes, there are good reasons :-)
Check the grammar here: https://docs.oracle.com/javase/specs/jls/se7/html/jls-14.html#jls-14.14.2
`VariableModifiers` is the hard part here: it supports "final" which `VariableDeclarator` does not.
The first feeling might be to tighten this up to a `VariableDeclarator` and maybe a final flag or so, but the rule is to have the grammar accept as much as possible and catch the errors in post-validations. So I suggest we add a few helper methods here instead. One that gets variabledeclarator 0, and one that says if it is final. You can suggest the names, or why not make a PR :-) Mention them in the Javadoc, and maybe check if there is a validation for this case (`expression.length==1`, and only modifier may be `final`.)

malteskoruppa:
Ah, that makes sense. :-) I'll have a look!

PR #2047, opened 2019-01-30
Convenience methods for ForEachStmt

malteskoruppa:
See #2034.

I also added validation for the invariants on the `VariableDeclarationExpr` held by a `ForEachStmt` to `ForEachStmt`'s `setVariable(...)` method. However, it appears that this method is auto-generated, so I am not sure if those checks won't be overwritten when the methods are generated next time? If so, what is the "proper" way to add custom code to auto-generated methods?

Of course I could have added these checks to the new convenience getter methods directly, but I feel that the proper place to perform such checks is in the setter, not in the getter. If this was done in the getter instead of the setter,

1. a `ForEachStmt` could potentially have an invalid state without raising an exception, as long as the getters are not called; and
2. having the getter methods check the invariants would mean that we uselessly check them over and over again each time the getter is called; and
3. having a getter method throw an exception is rather unusual behavior and likely very surprising to users.

Anyhow, I'm happy to take suggestions for improvements of course. :)


matozoid:
Thanks!
Yes, it will definitely be overwritten. No way around that right now - only some basic stuff like "not null" and "not empty" are supported. Yes, the state of the AST can be invalid. No more explanation here, this was simply never built :-) There are no issues for it either, so I'm not worried.

The validation phase happens right after parsing- the code can be found in `com.github.javaparser.ast.validator`. This can also be run after constructing your own AST so you'll know if there are any errors, but it is far from fail safe. A validation can be added here, or open an issue to add a validation and I'll do it somewhere in the future :-)

The rest seems fine! 

Do you think we want setters too? Maybe add a few utility constructors?

matozoid:
Oh, to regenerate the code: run `run_core_generators.sh`

malteskoruppa:
Thanks for the info! I tried to make the validation happen as explained. Is this ok?

matozoid:
The only thing remaining is unit tests on the validations. Check these: https://github.com/javaparser/javaparser/tree/master/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator
I'm undecided if these validations should be for Java 1.5+ or for all Java versions, since the ForEach can always exist in an AST, even if you set the language level to 1.0. Anyway, if you keep it at 1.5, another test that the validation don't pop up on 1.4 would be good :-)

malteskoruppa:
> I'm undecided if these validations should be for Java 1.5+ or for all Java versions, since the ForEach can always exist in an AST, even if you set the language level to 1.0.

Well there's already a post-validation for Java 1.0 to 1.4 called `noForEach` that ensures that foreach statements do not pop up in ASTs of those early-days Java versions in the first place, so I think it makes sense to have the validation for the proper form of foreach statements for Java 1.5+.

See here:
https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java1_0Validator.java#L74

> Anyway, if you keep it at 1.5, another test that the validation don't pop up on 1.4 would be good :-)

Well, since foreach statements can't occur in Java 1.0 to Java 1.4 at all, the only thing we can do here is to check that a problem is reported when we do parse a for-each loop when the language level is set to something between 1.0 and 1.4. But then I would really only be checking that the `noForEach` post-validation (which already existed before this PR!) is working properly, so it doesn't really have anything to do with the new validator I added. Anyhow, I added such a test. :)

In addition, there already exists a test that checks that a valid foreach-loop is parsed without problems starting with Java 1.5:
https://github.com/javaparser/javaparser/blob/master/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java5ValidatorTest.java#L131

I now also added another test that checks that modifiers other than `final` will lead to post-validation problems.

I would have liked to add another test that checks that multiple variables are not allowed in a foreach loop. So I tried to parse stuff like `forach (int i, int j : nums) {}`, but that just generates a parse exception before the post-validation phase is even triggered. :) Nevertheless, a post-validation is clearly useful, just to be on the safe side.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java5ValidatorTest.java
```diff
@@ -133,6 +133,15 @@ void foreach() {
         assertNoProblems(result);
     }
 
+    @Test
+    void noModifiersInForEachBesideFinal() {
+        ParseResult<Statement> result = javaParser.parse(STATEMENT, provider("for(static transient int i : nums){}"));
+        assertProblems(result,
+                "(line 1,col 5) 'static' is not allowed here.",
+                "(line 1,col 5) 'transient' is not allowed here.",
+                "(line 1,col 1) A foreach statement's variable declaration may have at most one 'final' modifier, and no other modifiers. Given: [static , transient ].");
```
matozoid: Hmmm, three messages are a little overkill?

matozoid:
I think the multiple variables that should parse but is wrong is like `for(int a,b,c:...)...`

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java5ValidatorTest.java
malteskoruppa: Sure, what should we do instead? The two first messages are already generated by existing validators; those are not my work. :-)

malteskoruppa:
Ah, yes, why didn't I think of that? :grinning: 

Ok, so now there's also a test for checking that the validator that ensures that a foreach loop can have at most one variable works as expected. :)

The only thing that remains to do is find out what we want to do about the illegal modifiers in foreach loops, which generate multiple messages: one for each illegal modifier (that was already there before) plus one (to say that the foreach loop can only have a final modifier; this is the new validator I wrote).

matozoid:
I just checked: the final check is done in ModifierValidator, and its generic for all `VariableDeclarationExpr`s, so it would be best to leave it as it is - even though the new message is better.

malteskoruppa:
Makes sense, I removed that particular validation.

I nevertheless kept the `noModifiersInForEachBesideFinal()` test, even though it is not actually specific to a foreach statement, because a test is a test is a test. :-)

matozoid:
More tests is better than less tests :-D

matozoid:
Thanks!

PR #2048, opened 2019-01-30
[ABANDONED] Comparing type parameters via class names if there are LazyType objects

### LINKED ISSUE #1983: Comparing the type parameters when there are LazyType objects

matozoid:
Fixes #1983.

Copy of #1984, since that seems abandoned.


jlerbsc:
I think [ABANDONNED] means that this PR is deprecated and it must not be released. If this is not the case you are free to reopen it.

PR #2049, opened 2019-01-30
Delegate parent handling to underlying JarTypeSolver in AarTypeSolver

xiphirx:
Previously if you were to call `getAllAncestors()` you may run into a `UnsolvedSymbolException` if the type originated from an AAR since `getAncestors()` will only search for types within the same AAR rather than other solvers defined in a `CombinedTypeSolver`, this was because we never set the parent for the wrapped `JarTypeSolver`.


matozoid:
Makes sense, thanks!

PR #2080, opened 2019-02-08
Fix toString issue

### LINKED ISSUE #2079: Log.trace(...) causing UnsolvedSymbolException leading to inconsistent method resolution
Here is a fun bug :). When trying to resolve a method call in the IntelliJ Community project, resolution failed... but only the first couple of times. The resolution worked from the 3rd resolution attempt onward. (See pull request #2080 for code).

The underlying issue is that when the method arguments are resolved, an unexpected `UnsolvedSymbolException` is thrown when `JavaParserFacade#solveArguments(...)` calls `JavaParserFacade#getType(...)`. The exception is propagated and the entire method call expression cannot be resolved. But the argument's type was successfully cached. Now the next time the same method call expression is resolved, the `getType` method simply returns the cached type, not throwing an `UnsolvedSymbolException` anymore. The method call here has two problematic arguments (the first one is fine), hence this happens twice, before the method call can be resolved the third time.

There are a couple of underlying issue, in my opinion:
- The call to `Log.trace(...)` in `getType(...)` immediately evaluates the arguments, even if tracing is disabled. This is not very performant and should probably be changed (unless you use tracing differently to what I am used to?). 
- The `toString()` method of node should probably not throw an UnsolvedSymbolException
- Is there something else amiss that the `toString()` method throws such an exception?

I hope I've explained the issue in enough detail - have a look at the test case for more insight, that should help :).

jjbr:
Fix an issue with a Log.trace(...) call in JavaParserFacade where exceptions are thrown inconsistently and unexpectedly.

Fixes #2079 .


matozoid:
@jjbr - fixed it :-)

PR #2085, opened 2019-02-14
Fixing Problem when JarTypeSolver doesn't resolve raw method parameters correctly

kgeilmann:
This PR fixes a problem of the JarTypeSolver with raw types for method parameters.

When trying to resolve the method call of getLogger in `Logger.getLogger(getClass())` (declared as `static Logger getLogger(Class c)`, Logger loaded by the JarTypeSolver), JavaParser throws an exception: 

    java.lang.UnsupportedOperationException: com.github.javaparser.resolution.types.ResolvedWildcard
	at 
    com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.matchTypeParameters(MethodCallExprContext.java:356)
    ...

The reason is, that for non-generic methods the JavassistMethodDeclaration ignores the type of the clazz parameter in the signature and returns the type as it was declared in the corresponding class, i.e. as java.lang.Class with a TypeParametersMap containing T as wildcard.
The JavaParserTypeSolver can resolve these calls, because it correctly returns just java.lang.Class with an empty TypeParameterMap.

With this PR the JarTypeSolver respects the method signature also for non generic methods, such that it behaves the same as the JavaParserTypeSolver.


matozoid:
Thanks!

PR #2094, opened 2019-02-20
Fix: matching a lambda expression with a functional-interface that wasn't annotated with @FunctionalInteface didn't work.

maartenc:


matozoid:
Thanks! Let's see what @ftomassetti says.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
```diff
@@ -106,7 +107,7 @@ public boolean isAssignableBy(ResolvedType other) {
             }
         }
         if (other instanceof LambdaArgumentTypePlaceholder) {
-            return this.getTypeDeclaration().hasAnnotation(FunctionalInterface.class.getCanonicalName());
+            return FunctionalInterfaceLogic.isFunctionalInterfaceType(this);
```
ftomassetti, line 110: Seems better

ftomassetti (approved):
Seems good to me

ftomassetti:
Update code, let's see if tests still pass

PR #2096, opened 2019-02-20
Fix: JavaParserEnumDeclaration getAncestors() broken when enum implements an interface

maartenc:


maartenc:
I did a copy/past of the code from JavaParserClassDeclaration and modified it a little bit.
Perhaps this should be refactored later by someone who knows the code better than me, because the code in JavaParserClassDeclaration, JavaParserInterfaceDeclaration and JavaParserEnumDeclaration is very similar in this area.

matozoid:
Do we have an okay from @ftomassetti ?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java
```diff
@@ -226,16 +226,65 @@ public int hashCode() {
         ancestors.add(enumClass);
         if (wrappedNode.getImplementedTypes() != null) {
             for (ClassOrInterfaceType implementedType : wrappedNode.getImplementedTypes()) {
-                SymbolReference<ResolvedTypeDeclaration> implementedDeclRef = new SymbolSolver(typeSolver).solveTypeInType(this, implementedType.getName().getId());
-                if (!implementedDeclRef.isSolved() && !acceptIncompleteList) {
-                    throw new UnsolvedSymbolException(implementedType.getName().getId());
+                ResolvedReferenceType ancestor;
+                try {
+                    ancestor = toReferenceType(implementedType);
```
ftomassetti: perhaps we can do the add to ancestors directly here, avoiding the temp var and the null assignment

ftomassetti (approved):
I am not sure why before it was not working and now it is. However the code seems correct and the tests passing so it is an ok from me

maartenc:
Thanks for the feedback, I've updated the code like you suggested.

matozoid:
Thanks again @maartenc!

PR #2097, opened 2019-02-20
Skip space before javadoc inline tag content

### LINKED ISSUE #2092: Javadoc inline tags have a stray space before the content

matozoid:
Fixes #2092.

@ftomassetti - am I fixing a bug or a feature here?


javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocTest.java
```diff
@@ -89,7 +90,8 @@ void descriptionAndBlockTagsAreRetrievable() {
     @Test
     void inlineTagsAreParsable() {
         String docText =
-                "Returns the {@link TOFilename}s of all files that existed during the requested" + EOL +
+                "{@inheritDoc}" + EOL +
```
ftomassetti, line 93: Why changing the test instead od adding a new one? Isn't this one different?

javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocTest.java
```diff
@@ -166,7 +180,7 @@ void issue1533() {
         List<JavadocDescriptionElement> elements = compilationUnit.getType(0).getJavadoc().get().getDescription().getElements();
         assertEquals(3, elements.size());
         assertEquals(new JavadocSnippet("hallo "), elements.get(0));
-        assertEquals(new JavadocInlineTag("link", LINK, " Foo"), elements.get(1));
+        assertEquals(new JavadocInlineTag("link", LINK, "Foo"), elements.get(1));
```
ftomassetti, line 183: I am not sure here. What if we had a tab o three spaces before Foo? Would they be relevant or not?

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
```diff
@@ -41,7 +41,11 @@ public static JavadocDescriptionElement fromText(String text) {
         text = text.substring(2, text.length() - 1);
         String tagName = nextWord(text);
         Type type = Type.fromName(tagName);
-        String content = text.substring(tagName.length());
+        int contentStart = tagName.length();
+        while (contentStart < text.length() && text.charAt(contentStart) == ' ') {
```
ftomassetti, line 45: Should we use trim? Should we skip any kind of whitespace? What about new lines?

javaparser-core/src/main/java/com/github/javaparser/javadoc/description/JavadocInlineTag.java
```diff
@@ -101,7 +105,7 @@ public String getName() {
 
     @Override
     public String toText() {
-        return "{@" + tagName + this.content +"}";
+        return "{@" + tagName + " " + content + "}";
```
ftomassetti, line 108: Is there always just a space between the tagName and content? Is it acceptable to lose the precise original form or not?

ftomassetti (approved):
I think it is good, just sharing some doubts

javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocTest.java
matozoid: This test is supposed to check inline tags. I can split it up?

javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocTest.java
matozoid: As far as I can see it's only relevant for printing. I should to check though.

PR #2103, opened 2019-02-22
[TESTCASE] issue 263 missing comments

### LINKED ISSUE #263: DumpVisitor loses comments II
```
package a;

public class Test {

            // one line of comment
            // another line of comment

            static {
                System.out.println("Hello");   
            }   

}
```

the comments get lost wenn dumping the parsed file

### LINKED ISSUE #307: if condition contains leading comments
Use DumpVisitor against  DumpVisitor.java itself

Comment will appear inside If condition in public void visit(final IfStmt n, final Object arg)

Method in error is DumpVisitor public void visit(final IfStmt n, final Object arg) or IfStmt.java itself

### LINKED ISSUE #498: Bug in parsing multiple line comments
If I parse Java code containing mulitple line comments, only the last line comment is parsed.
If you run the attached snippet, the parsed source contains only "line 3", but "line 1" and "line 2" are missing.
This bug seems to be present in all versions up to 3.0.0-alpha.4.

```
import com.github.javaparser.ast.visitor.DumpVisitor;

public class ShowBug {

        String src =
                "public class TestDumpVisitor {\r\n" +
                "\r\n" +
                "   public boolean test() {\r\n" +
                "       // line 1\r\n" +
                "       // line 2\r\n" +
                "       // line 3\r\n" +
                "       return false;\r\n" +
                "   }\r\n" +
                "}\r\n" +
                "";

        CompilationUnit cu = JavaParserTools.getCompilationUnit(src);
        DumpVisitor dv = new DumpVisitor();
        cu.accept(dv, null);
        String dumpSrc = dv.getSource();

        System.out.println("Original:");
        System.out.println(src);
        System.out.println("Parsed:");
        System.out.println(dumpSrc);
    }
}
```

`

matozoid:
Fixes #263.
Fixes #307.
Fixes #498.


MysterAitch:
This test case now fails only due to whitespace differences, as opposed to missing comments.

With some tweaks to the test case, perhaps this and the linked issues can be closed?

PR #2104, opened 2019-02-23
Make toString's printer configuration accessible

matozoid:
As suggested by @lucamolteni.


matozoid:
@lucamolteni will this work for you?

lucamolteni:
@matozoid perfect thank you!

matozoid:
Here it goes then!

PR #2112, opened 2019-03-01
Fixes issue 2099

### LINKED ISSUE #2099: Comment shift to new statement when adding statement
Hi, I'm trying to add a statement using addStatement on a BlockStmt (the method 'test' below) and the line comment seems to shift from one statement to the new (from the query.getLong(0); to the if statement), is this a bug in JavaParser? If it is, how can I work around that?

```
public class Teste {
    public void test(Uri uri, ContentProvider provider) {
        Cursor query = provider.query(uri, null, null, null, null);
        query.getLong(0); // Teste
    }
}
```

```
public class Teste {
    public void test(Uri uri, ContentProvider provider) {
            Cursor query = provider.query(uri, null, null, null, null);
            query.getLong(0); if(query != null) {
            query.close();
        }
        // Teste
    }
}
```

Thanks in advance

moraispgsi:
This pull request fixes #2099. Basically it adds a condition in the applyAddedDiffElement method to account for the commentary in the same line.


matozoid:
Cool! Let's see what @ftomassetti says.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/stmt/BlockStmtTest.java
```diff
@@ -1,10 +1,18 @@
 package com.github.javaparser.ast.stmt;
 
+import com.github.javaparser.JavaParser;
```
moraispgsi: The changes in this file are not necessary.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/TransformationsTest.java
```diff
@@ -140,4 +145,54 @@ void exampleParam5() throws IOException {
         md.getBody().get().getStatements().add(new ReturnStmt(new NameExpr("p1")));
         assertTransformed("Example_param5", cu);
     }
+
+    @Test
+    void issue2099AddingAddingStatementAfterTraillingComment1() {
```
moraispgsi: issue2099AddingAddingStatementAfterTraillingComment1 -> issue2099AddingStatementAfterTraillingComment1

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/TransformationsTest.java
```diff
@@ -140,4 +145,54 @@ void exampleParam5() throws IOException {
...
+                "    value2(); // Test\n" +
+                "    if(value != null) {\n" +
+                "        value.value();\n" +
+                "    }\n" +
+                "}";
+        assertEqualsNoEol(expected, s);
+    }
+
+    @Test
+    void issue2099AddingAddingStatementAfterTraillingComment2() {
```
moraispgsi: issue2099AddingAddingStatementAfterTraillingComment2 -> issue2099AddingStatementAfterTraillingComment2

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -572,6 +572,17 @@ private void applyAddedDiffElement(Added added) {
         }
 
         if (!used) {
+            if(nodeText.numberOfElements() > originalIndex + 1 &&
+                nodeText.getTextElement(originalIndex).isComment()) {
+                String expanded = nodeText.getTextElement(originalIndex).expand();
+                if(expanded.startsWith("/*")) {
```
ftomassetti, line 593: mmm. I would use an helper function, maybe putting it directly on the nodetext, to check if the node text is a comment. We should also consider line comments

ftomassetti (approved):
To me it works but I would refactor a little bit the change in Difference for clarity

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
moraispgsi: It is considering line comments. You can see that both tests pass.

matozoid:
@moraispgsi @ftomassetti - shall we merge and release this, or do I need to wait for refactoring?

moraispgsi:
I have detected other problems concerning indentation and adding statements to a blockstmt that I will be submitting corrections for in the immediate future. As for the refactoring, I'm not sure what you would like me to do.

ftomassetti:
I would wait for the refactoring based on the comments provided

matozoid:
@moraispgsi normally PR's get merged when all comments on it are resolved. I hear this PR fixes quite a few problems, so it might be more important to merge it right away and resolve things later. But I see it turned red, so the decision is made - it will wait until next week :-)

moraispgsi:
Yes, It is failing because I added two more tests where it fails in the indentation... trying to fix them before we merge.

matozoid:
No hurry, I've started releasing so it will come next Sunday.

moraispgsi:
I am having trouble with addASecondStatementToExistingMethod test, I think my solution is right and this test is wrong. It's funny because this test is ancient leading back to issue #926. The problem I am facing is that the assertation only considers a portion of the original code. [You can see it in this picture](https://pasteboard.co/I3Ljs5y.png). Am I missing something?

moraispgsi:
PR is back to green but I had to modify the test that I mentioned in the previous comment.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
```diff
@@ -803,7 +803,7 @@ void addASecondStatementToExistingMethod() throws IOException {
         assertEquals("public void someMethod() {" + EOL
                 + "        String test = \"\";" + EOL
                 + "        String test2 = \"\";" + EOL
-                + "}", LexicalPreservingPrinter.print(methodDeclaration));
+                + "    }", LexicalPreservingPrinter.print(methodDeclaration));
```
ftomassetti, line 808: Mmm, I think that ideally the right brace should not be intended because the opening method was not indented. However that would be really tricky to implement as we should remember where the brace was opened and what indentation we had there

moraispgsi: What should we do about it?

matozoid: @ftomassetti - what do you think?

ftomassetti: We could leave it as it is for now, perhaps adding a comment in the test because if later someone implement the complex but more correct behavior he will break this test

matozoid: Okay, @moraispgsi, let's get that comment in and merge this thing.

moraispgsi: @matozoid I added the comment.

matozoid:
Thanks for supporting your PR, @moraispgsi !

PR #2115, opened 2019-03-01
Make StaticJavaParser thread safe

### LINKED ISSUE #2113: StaticJavaParser might not be thread safe
I think `StaticJavaParser` is not thread safe:

The "old" JavaParser class called `com.github.javaparser.JavaParser#simplifiedParse` internally which in turn created a new `JavaParser` for each parse call.

`StaticJavaParser` just re-uses the same `JavaParser` instance (which is fine I guess) but that is quite a different behavior.

It would be nice if at least this was documented somewhere so people like me don't wonder why concurrency-related tests are suddenly failing.

matozoid:
Fixes #2113.


matozoid:
Is this better, @dhardtke?

dhardtke:
Yes, much better! muchas gracias 😁

PR #2116, opened 2019-03-04
Fix: issue resolving generic types with nested class types

maartenc:
When trying to replace a generic parameter with a concrete type, an exception is thrown if that concrete type is a double nested class.

Maarten


ftomassetti (approved):
Looks good to me.
@maartenc you are doing great contributions. Thanks!

maartenc:
With pleasure :-)
Thanks!
Maarten

PR #2117, opened 2019-03-07
Fix: issue with resolving overloaded methods for javassist types

maartenc:
The wrong method could be returned when trying to resolve overloaded methods for javassist types. The first method with the same name and number of arguments was returned without considering the type of the parameters. 

I've changed that logic and made it more or less similar to the reflection-based implementation.

Maarten

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistMethodDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistMethodDeclaration.java
@@ -34,10 +34,7 @@ import javassist.bytecode.BadBytecode;
 import javassist.bytecode.SignatureAttribute;
 
 import java.lang.reflect.Modifier;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 
 /**
@@ -160,7 +157,7 @@ public class JavassistMethodDeclaration implements ResolvedMethodDeclaration, Ty
     public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
         try {
             if (ctMethod.getGenericSignature() == null) {
-                return Collections.emptyList();
+                return new ArrayList<>();
             }
             SignatureAttribute.MethodSignature methodSignature = SignatureAttribute.toMethodSignature(ctMethod.getGenericSignature());
             return Arrays.stream(methodSignature.getTypeParameters()).map((jasTp) -> new JavassistTypeParameter(jasTp, this, typeSolver)).collect(Collectors.toList());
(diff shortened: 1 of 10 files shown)


matozoid:
Thanks again :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistMethodDeclaration.java
```diff
@@ -160,7 +157,7 @@ public boolean isAbstract() {
     public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
         try {
             if (ctMethod.getGenericSignature() == null) {
-                return Collections.emptyList();
+                return new ArrayList<>();
```
ftomassetti, line 160: Why is that?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
```diff
@@ -70,7 +70,7 @@ private void collectDeclaredFields(CtClass ctClass, List<ResolvedFieldDeclaratio
 
   public List<ResolvedTypeParameterDeclaration> getTypeParameters() {
     if (null == ctClass.getGenericSignature()) {
-      return Collections.emptyList();
+      return new ArrayList<>();
```
ftomassetti: Again, why? We are not trying to modify this list, right?

ftomassetti (approved):
Great stuff, thank you!

ftomassetti:
Just a minor comment but this is great stuff, really needed

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistMethodDeclaration.java
maartenc: Yes, the list is modified in the class MethodResolutionLogic, line 271.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
maartenc: ok, this is probably not needed here, but only in JavassistMethodDeclarationAdapter. I'll check if junit tests are still working when I revert this change.

maartenc: I've reverted the new ArrayList creation here in this class, it was not necessary.

Issue #2091, opened 2019-02-18
Symbol resolution of switch expressions


matozoid:
Continued from #1845

The new Java 12 constructs need symbol resolution logic. 

* [ ] do symbol resolution on the switch expressions
* [ ] do symbol resolution on value breaks

matozoid:
Neither of these is needed, since neither is a type.

PR #2119, opened 2019-03-09
Resolve switch expressions

matozoid:
See #2091. I don't know what to do at the places with TODO. I'm not even sure this is the right approach at all.


matozoid:
@maartenc @ftomassetti can you give me a hand here?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -504,4 +505,44 @@ public ResolvedType visit(FieldDeclaration node, Boolean solveLambdas) {
         }
         throw new IllegalArgumentException("Cannot resolve the type of a field with multiple variable declarations. Pick one");
     }
+
+    @Override
+    public ResolvedType visit(SwitchExpr switchExpr, Boolean solveLambdas) {
+        List<ResolvedType> allSwitchEntiesResolved = switchExpr.getEntries().stream()
+                .map(entry -> entry.accept(this, solveLambdas))
+                .collect(toList());
+        // TODO find common type of all switch entries
```
ftomassetti: We should have a utility for that somewhere

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -504,4 +505,44 @@ public ResolvedType visit(FieldDeclaration node, Boolean solveLambdas) {
...
+            default:
+                throw new IllegalStateException("This kind of switch entry is not supported yet.");
+        }
+    }
+
+    private ResolvedType findCommonTypeOfAllBreakStatements(Node node, Boolean solveLambdas) {
+        List<ResolvedType> allBreakStmtsResolved = node.findAll(BreakStmt.class).stream()
+                .map(bs -> bs.accept(this, solveLambdas))
+                .collect(toList());
+        // TODO find common type of all break statements
```
ftomassetti: Again, we should have an utility for that somewhere. And if not we should create it

ftomassetti (approved):
I do not see anything clearly wrong with this code

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
matozoid: I'll go hunt for it then.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/SwitchExprTest.java
```diff
@@ -0,0 +1,68 @@
...
+    void switchEntryBreaksInsideStatementBlock() {
+        CompilationUnit ast = javaParser.parse("class X{void x(){Object a=switch(a){case 1-> {a(); b(); break 3+4;}}; }}").getResult().get();
+        ResolvedType resolvedType = ast.findFirst(SwitchExpr.class).get().calculateResolvedType();
+        assertEquals(ResolvedPrimitiveType.INT, resolvedType);
+    }
+
+    @Test
+    void multipleSwitchEntriesReturnDifferentTypes() {
+        CompilationUnit ast = javaParser.parse("class X{void x(){Object a=switch(a){case 1-> 1; case 2->\"\"; default->1.0; };}}").getResult().get();
+        ResolvedType resolvedType = ast.findFirst(SwitchExpr.class).get().calculateResolvedType();
```
matozoid: @ftomassetti - is the type of the switch expression a union type (like I made it here) because of the values of its cases, or is it `Object` because it is assigned to `Object a`?

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedUnionType.java
```diff
@@ -22,23 +22,27 @@
...
  * A union type is defined in java as list of types separates by pipes.
  *
  * @author Federico Tomassetti
  */
 public class ResolvedUnionType implements ResolvedType {
-    private List<ResolvedType> elements;
+    private final List<ResolvedType> elements;
 
-    public ResolvedUnionType(List<ResolvedType> elements) {
+    public ResolvedUnionType(Collection<ResolvedType> elements) {
```
ftomassetti, line 36: Is order really irrelevant here?

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedUnionType.java
```diff
@@ -22,23 +22,27 @@
...
         if (elements.size() < 2) {
             throw new IllegalArgumentException("An union type should have at least two elements. This has " + elements.size());
         }
         this.elements = new LinkedList<>(elements);
     }
 
+    /**
+     * Note: this algorithm doesn't seem to work as it should, nor is it in the right module.
```
ftomassetti, line 44: Should we deprecate it then?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/SwitchExprTest.java
ftomassetti: To me the switch expression should have its own type, then we could check it is compatible with the destination

ftomassetti (approved):

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/SwitchExprTest.java
matozoid: Its own type? I was thinking that instead of (ab)using a union, it would be better to simply return a list of possible result types. Is that the idea?

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedUnionType.java
matozoid: I would say yes, it is irrelevant. Still, if you pass in a list, order will be retained (for printing purposes mostly)

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedUnionType.java
matozoid: I have no idea why it was added in the first place :-(

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/SwitchExprTest.java
matozoid: And *can* we check compatability, or do we *need to* check compatability with the destination?

ftomassetti: We can check if for validation but we do not need to check it for symbol resolution, imho

matozoid:
@ftomassetti I was trying to add a new kind of resolved type of SwitchEntry or SwitchExpr, and I finally understood the problem: I'm trying to resolving something that is not a type to a type. That's why it doesn't fit in what's there. So the "union type" part is something new here.

Issue #1391, opened 2018-02-02
Ways to free memory?


AwesomeLemon:
Hello.
I'm using JavaParser and JavaSymbolSolver when creating a dataset by processing thousands of git repos.
My problem is that the RAM consumption increases the longer my dataset-creating-program runs until there's no more RAM left.
I found that calling `JavaParserFacade.clearInstances();` helps by clearing the cache, about which I don't really care since I have already moved on to processing of the next project.
So I wonder if there're other similar methods for some other caches, or generally any ways to clear memory?

matozoid:
@ftomassetti can comment on the caches.

There's another option: you can look at `ParserConfiguration` and check to see which features you don't need, then turn those off.

AwesomeLemon:
Thanks for the pointer, I've set only lexical preservation to false, since all other options seem to impact comments, which I need.

ftomassetti:
I could find these caches:
* in PhantomNodeLogic
* in JavaParserTypeSolver (3 caches)
* in SourceRoot
* TypeInferenceCache
* in JavaParserFacade (2 caches)

ftomassetti:
I would say: let's see which ones are causing an issue and add methods to reset them

dhardtke:
What I find problematic is that there is no way not to use `JavaParserFacade.instances` since `JSS`s internal classes use `JavaParserFacade.get()` internally :(

matozoid:
For `SourceRoot`: read the Javadoc carefully and you'll be able to avoid the cache. Also: the cache is scoped to the lifecycle of `SourceRoot`, so once it gets garbage collected the cache is gone. Meaning: regularly instantiate a new one if the cache is bothering you.

daanschipper:
I'm currently investigating the problem while parsing many files with JavaParser. I'm running it against a project with ~290 source roots and ~20k .java files. 

My findings are:
- The cache in `SourceRoot` is not used since I'm calling `SourceRoot.parseParallelized`. By invoking `tryToParse` or `save` (manually adding a newly created Java file) will add the path to the file together with the compilation unit to the cache. 
[...]

PR #2122, opened 2019-03-13
Performance improvement getting all methods

maartenc:
Avoid unnecessary conversion ResolvedMethodDeclaration -> MethodUsage -> ResolvedMethodDeclaration. The MethodUsage constructor does type solving of paramters and return type, which can be time consuming, especially if there are a lot of generified methods. 

This change made type resolving of some big classes 2x or 3x faster.

Maarten

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -382,9 +382,7 @@ public abstract class ResolvedReferenceType implements ResolvedType,
      * that have been overwritten.
      */
     public List<ResolvedMethodDeclaration> getAllMethods() {
-        List<ResolvedMethodDeclaration> allMethods = new LinkedList<>();
-        allMethods.addAll(this.getDeclaredMethods().stream().map(MethodUsage::getDeclaration)
-                .collect(Collectors.toList()));
+        List<ResolvedMethodDeclaration> allMethods = new LinkedList<>(this.getTypeDeclaration().getDeclaredMethods());
         getDirectAncestors().forEach(a ->
                 allMethods.addAll(a.getAllMethods()));
         return allMethods;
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
```diff
@@ -406,18 +404,9 @@ public boolean isRawType() {
...
 
     public List<ResolvedMethodDeclaration> getAllMethodsVisibleToInheritors() {
-        List<ResolvedMethodDeclaration> res = new LinkedList<>(this.getDeclaredMethods().stream()
-                .map(m -> m.getDeclaration())
+        return this.getAllMethods().stream()
                 .filter(m -> m.accessSpecifier() != PRIVATE)
-                .collect(Collectors.toList()));
-
-        // We want to avoid infinite recursion in case of Object having Object as ancestor
-        if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
```
ftomassetti, line 415: Are we sure this is not necessarily anymore?

ftomassetti:
Now I am a bit curious about what @maartenc is building with Javaparser :)

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
maartenc: I think so because:
- the getAllMethods() didn't had that safety-check
- the implementation of getDirectAncestors() seems to handle the Object.class ancestor already

ftomassetti: Then it looks good to me

ftomassetti (approved):
+1. Great work, as usual

maartenc:
thanks :-)
We are using javaparser to analyse our source base for unused dependencies.
Some of our biggest source files took about 10 minutes to analyse (for a single file!), this change took it down to 5 minutes (but which is still too slow in my opinion).

Maarten

matozoid:
@maartenc have you seen this issue? https://github.com/javaparser/javaparser/issues/1391

maartenc:
Thanks for the tip @matozoid . I'll have a look it that issue.

PR #2082, opened 2019-02-10
Fixed issue #2024, can now remove line and block comments with LexicalPreservingPrinter

JustinKissling:
Fixes #2024 by modifying LexicalPreservingPrinter to handle removed block and line comments.
Also includes test cases.

PR #2127, opened 2019-03-14
Fixed problem with LexicalPreservingPrinter removing line comments

JustinKissling:
This is an addition to my previous change here: https://github.com/javaparser/javaparser/pull/2082
I found that when parsing a java file with Unix style line endings, removing a comment while using the LexicalPreservingPrinter failed (the exact same code with Windows style line endings worked fine). It looks like some comments are in a ChildTextElement instead of a TokenTextElement, which is why they couldn't be found by the current method. This change attempts to fix that, by searching through ChildTextElements as well after no suitable TokenTextElement could be found. 
Unfortunately, I was unable to reproduce this issue with a test case, as it only happened to me when parsing an actual file with Unix style line endings.
I'm not sure if this is the best solution but I'd like to hear your thoughts on it.


ftomassetti (approved):
This seems reasonable to me. I think the issue is when parsing code with a line separator different from the one of the current machine

matozoid:
It shouldn't be hard to make a string containing \r\n or \n, whatever breaks the code. Or am I missing something?

Anyway, the PR runs on Appveyor (Windows) and on Travis (Linux) so you can try pushing some tests and see if they break on one or the other :-)

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -189,7 +230,7 @@ public void concretePropertyChange(Node observedNode, ObservableProperty propert
                 matchingTokens = nodeText.getElements().stream()
                     .filter(e -> e.isToken(SINGLE_LINE_COMMENT))
                     .map(e -> (TokenTextElement) e)
-                    .filter(t -> t.getText().equals("//" + oldValue.getContent() + System.lineSeparator()))
```
matozoid, line 192: Ah yes, this could never have gone right because you never know what EOL's are in the input.

JustinKissling: Yeah, that was one part of the problem. The other was that the comment is in a TokenTextElement with a Windows file, but in a ChildTextElement with a Unix file

JustinKissling:
Just added a test case for this, with Unix line endings (\n) in the input and comparison string. This breaks the old code, but should work with the new one.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
matozoid: It sounds like something else is wrong there. There are several EOL tokens, and it seems there is a hard comparison with one of them somewhere in the code :-/

matozoid:
I've made a follow-up bug report: https://github.com/javaparser/javaparser/issues/2128

Issue #2138, opened 2019-03-25
Problem adding a statement to block previously added ifstmt


moraispgsi:
Added new test to my fork where this problem can be shown https://github.com/TQRG/javaparser/commit/1548b57218350b41cce462ec73ca0b12ee36088d

```
java.lang.UnsupportedOperationException: kept token(}) vs ChildTextElement{value.value();}

	at com.github.javaparser.printer.lexicalpreservation.Difference.applyKeptDiffElement(Difference.java:484)
	at com.github.javaparser.printer.lexicalpreservation.Difference.apply(Difference.java:141)
	at com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter$Observer.concreteListChange(LexicalPreservingPrinter.java:269)
	at com.github.javaparser.ast.observer.PropagatingAstObserver.listChange(PropagatingAstObserver.java:72)
	at com.github.javaparser.ast.NodeList.lambda$notifyElementAdded$1(NodeList.java:466)
	at java.util.ArrayList.forEach(ArrayList.java:1249)
	at com.github.javaparser.ast.NodeList.notifyElementAdded(NodeList.java:466)
	at com.github.javaparser.ast.NodeList.add(NodeList.java:70)
	at com.github.javaparser.ast.nodeTypes.NodeWithStatements.addStatement(NodeWithStatements.java:54)
	at com.github.javaparser.printer.lexicalpreservation.TransformationsTest.addingStatement4(TransformationsTest.java:311)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:515)
	at org.junit.jupiter.engine.execution.ExecutableInvoker.invoke(ExecutableInvoker.java:115)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$invokeTestMethod$6(TestMethodTestDescriptor.java:171)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:72)
[...]

PR #2139, opened 2019-03-25
[TESTCASE]  New add statement test

moraispgsi:
Test for issue #2138


moraispgsi:
@matozoid The problem seems to occur because there are two observers and because of that the difference is applied twice.

matozoid:
Two observers? @ftomassetti - how can that happen?

moraispgsi:
I need some help correcting this issue. Help @matozoid @ftomassetti

ftomassetti:
I have no idea how two observers should end up being attached to the same node

moraispgsi:
@ftomassetti I am not sure why this happens, I add an ifstmt to a blockstmt, I think this adds the first observer, after that I add a new statement to the ifstmt, and this I think adds a second observer. And when it comes to the second change the two observers are called in sequence, so two differences are computed and applied instead of one.

matozoid:
Well, it should be easy to debug.

matozoid:
It is caused by calling `LexicalPreservingPrinter.setup` three times on three different pieces of code. It creates three different observers that happily start propagating around when the pieces of code get put together.

@ftomassetti - what is the idea in this case? Should this be seen as bad user behaviour? Should the observers not propagate when one already exists? I guess this is more or less a useful use case...

moraispgsi:
@ftomassetti

PR #1820, opened 2018-09-01
fix #1814: Avoid infinite recursion when resolving java.lang.Object

dhardtke:
This was a surprisingly easy fix.

PR #2141, opened 2019-03-27
Fix StackOverflowError in getAllMethods() and getAllFieldsVisibleToInheritors()

### LINKED ISSUE #2140: StackOverflowError in getAllMethods() and getAllFieldsVisibleToInheritors()
... this is due to `getDirectAncestors()` always including `Object` and the aforementioned methods calling themselves on each ancestor returned by `getDirectAncestors()`.

I really don't know why `getDirectAncestors()` includes `Object` because it requires us to always make exceptions when processing the returned List.

This issue is probably related to https://github.com/javaparser/javaparser/pull/1820

PR incoming...

dhardtke:
Fixes #2140.


maartenc:
Wouldn't it be better to let getDirectAncestors() returns an empty List for java.lang.Object ?

matozoid:
@dhardtke - what do you think about what @maartenc says?

dhardtke:
I do agree that returning an empty `List` would be a better idea :P you wan't me to change the implementation?

No point in changing something if it's not gonna get merged, hence my asking.

matozoid:
It feels like we could break some thing here - could you change the implementation, then write a line in `changelog.md` warning people? (Like: *this* is going to break because of *this*. Doing *this* will fix it.)

dhardtke:
Sure.

dhardtke:
Now this is weird: the test from the other PR that I linked (https://github.com/javaparser/javaparser/pull/1820) called `Issue1814` is failing locally but either it has not been executed in Travis' CI, or it executed successfully (https://travis-ci.org/javaparser/javaparser/builds/511782047) - how's that?

No need to write a separate test class if `Issue1814` already covers this.

matozoid:
@dhardtke - looking into it.

matozoid:
Right, I've got it. Tests classes have to end with "Test" since JUnit 5...

matozoid:
You can rename Issue1814.java to Issue1814Test.java, and I'll do the other tests that aren't running.

dhardtke:
Done.

matozoid:
Thanks!

PR #2151, opened 2019-04-03
Fix for endless recursion bug #2150

### LINKED ISSUE #2150: Endless recursion when typesolving within type extending similarily-named type
Say we have an interface (or class, don't think it matters) that extends a type of the same unqualified name but in a different package:
```
package foo;
public interface MyInterface extends bar.MyInterface {
```

We may not know bar.MyInterface (could be in a library we don't have available, for instance). So resolving `bar.MyInterface` as ancestor of foo.MyInterface fails (see `JavaParserInterfaceDeclaration#toReferenceType`). The symbol solver then falls back on resolving the unqualified name `MyInterface`. That obviously succeeds, as our current context is called the same. Now we've resolved to the wrong type and worse, we've looped back to the same type we started out from. The `JavaParserTypeDeclarationAdapter#checkAncestorsForType` method then loopes until the stack overflows, always checking the same type again. 

I haven't tested it yet but I can imagine this could also produce loops in more complex cases, say we have extends-relations such as `A -> B -> C -> D -> bar.A` and `bar.A` is not known.

jjbr:
…lasses?) extend similarily-named interfaces (/classes).

Fixes #2150


malteskoruppa:
This PR makes a lot of sense to me, since:
1. it eliminates a currently existing bug that traps the symbol solver in an endless recursion, as the test in this PR clearly demonstrates (just comment out the fix in the main code and run the test and you'll get a `StackOverflowError`). A `StackOverflowError` is obviously an extremely undesirable outcome for users; and
2. on top of that, the bug may clearly lead to incorrect resolutions. Why would one try to do the resolution on an unqualified name as a fallback when the resolution already failed on the qualified name? I cannot picture any scenario where this fallback would lead to a successful and correct resolution; I can however easily picture scenarios where this will lead to a successful, but incorrect resolution (e.g., if the extended type is `a.b.c.Foo`, but the compilation unit imports a type `x.y.z.Foo`).

Digging a bit deeper, I also made the observation that the discussed fallback was not implemented by @ftomassetti himself. Rather, it was done by @FredLL in commit c990d23490f5c4ac4322119b03d5951b63eef055, a commit with the message *"loads of changes to support our code"*. I am almost positive that the change made by Fred in this particular case is incorrect, as explained above and also explained by @jjbr in #2150.

I hope this PR will make it into a release in the near future, perhaps even the next one? :smile:  @matozoid could you perhaps add @ftomassetti as a reviewer?

matozoid:
Well, @malteskoruppa, I think you just reviewed it :-) So here it goes. Next release is Sunday of course.

I'm not sure why I missed this one?

PR #2153, opened 2019-04-04
Disallow adding fields to annotation declarations

### LINKED ISSUE #676: It should not be possible to add a Field to an AnnotationDeclaration
The methods `addField` should be overridden in `AnnotationDeclaration` to return `UnsupportedOperationException`

matozoid:
Fixes #676.


javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/AnnotationDeclarationTest.java
```diff
@@ -0,0 +1,12 @@
...
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class AnnotationDeclarationTest {
+    @Test
+    void cantAddField() {
+        assertThrows(IllegalStateException.class, () -> new AnnotationDeclaration().addField(Object.class, "oo"));
+    }
+}
```
ftomassetti: nitpick: new line at the end of the file
```suggestion
}

```

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationDeclaration.java
```diff
@@ -135,4 +136,9 @@ public ResolvedAnnotationDeclaration resolve() {
     public Optional<AnnotationDeclaration> toAnnotationDeclaration() {
         return Optional.of(this);
     }
+
+    @Override
+    public FieldDeclaration addField(Type type, String name, Modifier.Keyword... modifiers) {
+        throw new IllegalStateException("Cannot add a field to an annotation declaration.");
```
ftomassetti, line 142: This seems a good idea

ftomassetti (approved):
+1

javaparser-core/src/main/java/com/github/javaparser/ast/body/AnnotationDeclaration.java
matozoid: @ftomassetti you suggested it years ago, see issue :-)

ftomassetti: I was so wise when I was younger :)

PR #2156, opened 2019-04-05
Copy user data while cloning.

### LINKED ISSUE #1943: SymbolSolver is unable to .resolve()
Hello,

i am trying to find the fully qualified name of interface `main.MyInterface<E>` that is being implemented in class `main.MyClass<E>`.

```
package main;
public class MyClass<E> implements MyInterface<E> { }
```

```
package main;

public class MyInterface<E> { }
```

```
for (ClassOrInterfaceType type : classDeclaration.getImplementedTypes()) {
	ResolvedReferenceType rrt = type.resolve(); // exception gets thrown here
	if (rrt != null) {
		String interfaceName = rrt.getQualifiedName();
		// do other stuff with it
	}
}
```

Sadly, this yields me an exception:

```
Symbol resolution not configured: to configure consider setting a SymbolResolver in the ParserConfiguration
	at com.github.javaparser.ast.Node.lambda$getSymbolResolver$5(Node.java:682)
	at java.base/java.util.Optional.map(Optional.java:254)
	at com.github.javaparser.ast.Node.getSymbolResolver(Node.java:679)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:292)
```

I am setting a SymbolSolver before I start running `.resolve()`:

```
CombinedTypeSolver typeSolver = new CombinedTypeSolver(new ReflectionTypeSolver());
// optionally some more external libraries added (required for what I am doing)
CombinedTypeSolver finalTypeSolver = new CombinedTypeSolver(typeSolver);
if (path.endsWith(".jar")) {
	finalTypeSolver.add(new JarTypeSolver(path));
} else if (path.endsWith("src")){
	finalTypeSolver.add(new JavaParserTypeSolver(path));
}

JavaSymbolSolver symbolSolver = new JavaSymbolSolver(finalTypeSolver);
JavaParser.getStaticConfiguration().setSymbolResolver(symbolSolver);
```

I also double checked without the generic types and it always yields the same exception. Same issue with extending from classes.
Am I doing something wrong or why does this not work as intended?

I just updated to the 3.7.1  (from 3.6.23) and now the exception is now this:

```
java.lang.IllegalStateException: No data of this type found. Use containsData to check for this first.
	at com.github.javaparser.ast.Node.getData(Node.java:457)
	at com.github.javaparser.ast.Node.lambda$getSymbolResolver$5(Node.java:699)
	at java.base/java.util.Optional.map(Optional.java:254)
	at com.github.javaparser.ast.Node.getSymbolResolver(Node.java:698)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:292)
```

Since this did work before and only started to fail recently, I double checked what I did differently. And it turns out that I previously was using a path to a source folder (`JavaParserTypeSolver`). Now I am trying to read a JAR file and therefore use the `JarTypeSolver`, but it does not seem to work with that one. Whats going on here?

matozoid:
Fixes #1943.

User data doesn't really get cloned, it gets copied. I can't expect people to implement correct cloning for all the data they want to store. Or can I?

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
@@ -54,6 +54,7 @@ public class CloneVisitorGenerator extends VisitorGenerator {
 
         body.addStatement(builder.toString());
         body.addStatement("r.setComment(comment);");
+        body.addStatement("copyData(n, r);");
         body.addStatement("return r;");
     }
 }
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/DataKeyTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/DataKeyTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser.ast;
 
+import com.github.javaparser.ast.expr.Name;
 import com.github.javaparser.ast.expr.SimpleName;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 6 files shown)


matozoid:
I think copying is fine. I'm creating this because the symbol solver doesn't get set in cloned nodes, and that thing should not get a clone() call.

PR #2168, opened 2019-04-10
Don't put EOL inside the line comment tokens

### LINKED ISSUE #2166: Line comment includes the EOL character
This was never seen as a problem, but when dealing with source code at the token level it's quite nasty.

matozoid:
Fixes #2166.


ftomassetti (approved):
My only comment is: do we need to preserve the information on the type of newline used? Do we want to support using a platform specific newline?

ftomassetti:
Ops... I pressed "merge" by mistake. Sleep deprivation.
Sorry about that!

matozoid:
@ftomassetti with this PR the EOL will be better preserved because it will be an EOL-category token after the comment. It seems this PR changes everything for the better, but I've only guessed what the change in the lex pres printer ought to be.

PR #2174, opened 2019-04-15
Make a shortcut to set the type of all variables declared in a variable declaration

### LINKED ISSUE #1362: Create utility methods to change variable declaration type in one go
Since we can have `int a, b, c;` and the type `int` is stored three times: in a, b, and c, the only way to change the type to something else is to set the new type on a, b, and c separately. That takes three method calls, and the AST state is inconsistent until they are all done.

Create a utility method that takes a type and sets it on all the variables, keeping the array brackets consistent, so `int[] a[], b;` with type changed to `float` will become `float[] a[], b;`, not `float a, b;`.

Document well that the `Type` that is being set will be cloned before inserting it into the AST, so the one that the user supplies will never end up in the AST.

matozoid:
Fixes #1362.


matozoid:
@rpau - okay, here's a PR. What's going to happen now? :-)

PR #2175, opened 2019-04-15
Reintroduce accesspecifier

matozoid:
Reintroduce accesspecifier after some mild complaints.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocExtractorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocExtractorTest.java
@@ -42,7 +42,7 @@ class JavadocExtractorTest {
 
     private void processFile(File file) throws FileNotFoundException {
         try {
-            CompilationUnit cu = parse(file, StandardCharsets.UTF_8);
+            CompilationUnit cu = parse(file);
             new VoidVisitorAdapter<Object>() {
                 @Override
                 public void visit(JavadocComment n, Object arg) {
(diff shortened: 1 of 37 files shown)


javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocExtractorTest.java
```diff
@@ -42,7 +42,7 @@ void canParseAllJavadocsInJavaParser() throws FileNotFoundException {
 
     private void processFile(File file) throws FileNotFoundException {
         try {
-            CompilationUnit cu = parse(file, StandardCharsets.UTF_8);
+            CompilationUnit cu = parse(file);
```
ftomassetti, line 45: Why this?

javaparser-core/src/main/java/com/github/javaparser/ast/AccessSpecifier.java
```diff
@@ -0,0 +1,47 @@
...
+package com.github.javaparser.ast;
+
+/**
+ * Access specifier. Represents one of the possible levels of
+ * access permitted by the language.
+ *
+ * @author Federico Tomassetti
+ * @since July 2014
+ */
+public enum AccessSpecifier {
```
ftomassetti, line 31: The good old AccessSpecifier. It is almost five years old!

ftomassetti (approved):
Nothing surprsing, +1

javaparser-core/src/main/java/com/github/javaparser/ast/AccessSpecifier.java
matozoid: I will never delete it again!

javaparser-core-testing/src/test/java/com/github/javaparser/javadoc/JavadocExtractorTest.java
matozoid: A recent PR moved the character encoding to ParserConfiguration, making parse methods with an explicit encoding redundant.

(And the default is UTF8 anyway)

PR #2186, opened 2019-04-21
[TESTCASE] lex pres: ClassOrInterfaceDeclaration addMember using index 

### LINKED ISSUE #2137: ClassOrInterfaceDeclaration addMember using index
Hi, currently I am not able to place a class  inside another class in a particular index. When I try getMembers() and add the class to the collection the indentation breaks. I think the library needs a method overload e.g. addMember(int index, ...).

moraispgsi:
Fixes #2137


moraispgsi:
@matozoid @quanturium

moraispgsi:
@matozoid In this test you can see what happens and what we mean by indentation breaking.

moraispgsi:
@matozoid I'm trying to debug this issue, I ran into a situation where the token range does not match the parsed node representation. In this picture https://imgur.com/a/OXap3ld we can see the mismatch of **\r\n**, in the token range we only see one **\r\n** however in the node representation we get **\r\n\r\n**. I think this is the issue because when the difference algorithm is applied there are extra **\r\n**, and that creates the conflict. I need your help validating this hypothesis.

moraispgsi:
@matozoid Also I'm not sure how to resolve this if this is in fact the issue.

matozoid:
I think you're getting misled by `toString()` - that's the pretty printer output you're seeing. That thing puts whitespace wherever it likes.

moraispgsi:
@matozoid In my test when I debug it, the second \r\n is considered when calculating the difference. It happens right here https://imgur.com/a/3lHPGHU.

moraispgsi:
@matozoid https://imgur.com/a/vizDpEA Here you can see that there are \r\n that should not be there. Maybe this is because the new code that is added does not have a LexicalPreservingPrinter setup.

moraispgsi:
@matozoid Could you take a look at it please?

matozoid:
@moraispgsi - take a look at this please: https://javaparser.org/contributing/

moraispgsi:
@matozoid Anything in specific?

matozoid:
There is no maintainer for this part of the library. You either solve the issue, or wait for someone to come along who wants to do it.

moraispgsi:
@matozoid Ok, thanks.

matozoid:
No need to close this, it's still an issue.

4everTheOne:
Hi all!

I think this PR can now be closed, since issue #2137 has been fixed in PR #3763.

I would like to also use this message, to wish all the members and contributors a wonderful festive season!

Issue #2195, opened 2019-04-25
Resolution of implmented/extended types falsely resolves to nested types


jjbr:
Say we have three types: 
- interface `pckg.A`
- class `pckg.B implements A`
- class `pckg.B.A extends B` 

and we now try to resolve the type that is implemented by `B` then we would expect to get `pckg.A`. Instead we get `pckg.B.A` as the scope chosen for the resolution is the class `pckg.B` and the resolution first searches through the members of `B`, where we find the nested `pckg.B.A` with the same simple name as `A`.

matozoid:
Thanks!

MysterAitch:
Just submitted a PR that fixes this issue as described, with an additional case where `class B extends A` (rather than implements).

I've not dug deeper into this but I'm also seeing oddities with resolving the declared types of fields and I'm unsure if this is a separate issue or an extension of this one. 

For example, in the code below the field `field` resolves to `implements_duplicate.A`. This seems to be true both with and without the changes in the above pull request but I'll need to get my head around it another time.

Further investigations would be warranted for local variables as well as fields, but as above these might be worth a separate issue?

`DuplicateTypeName.java`


```java
package implements_duplicate;

public interface DuplicateTypeName {

}
```

`A.java`:

```java
package implements_duplicate;

public class A implements DuplicateTypeName {

    DuplicateTypeName field;

    class DuplicateTypeName extends A {

    }
}
```

PR #2196, opened 2019-04-25
[TESTCASE] for #2195 - resolution of extended/implemented types falsely…

jjbr:
… considers nested types of the extending/implementing type.

Testcase for #2195.


jjbr:
Given that #2211 includes a test that is basically the same one as in this pull request (apart from naming) this can be closed :).

PR #2197, opened 2019-04-25
fix Removing 2 nodes with the same parent fail #2187

### LINKED ISSUE #2187: Removing 2 nodes with the same parent fail
In this block, if we try to remove 2 methodCallExpr (log), the second fail (nothing is happening) because it seems that the parent node of the ExpressionStmt is setted to null after the first remove. NodeList.remove invokes notifyElementRemoved() and after this method is called all statements of the BlockStmt have the parent set to null.

    catch (final Exception e) {
        log.error("context", e);
        log.warn("another context", e);
        throw new ApplicationException(e);
    }
Below the test case
I am not sure that this code is correct but I entrust this analysis to your expert eye

    @Test
	public void test1() throws ParseException {
		String sample = "{ try {/* empty */} catch (final Exception e) {\r\n" + 
				"    log.error(\"context\", e);\r\n" + 
				"    log.warn(\"another context\", e);\r\n" + 
				"    throw new ApplicationException(e);\r\n" + 
				"}}";
		JavaParser jp = new JavaParser();
		Statement stmt = jp.parseStatement(sample).getResult().get();
		stmt = LexicalPreservingPrinter.setup(stmt);
		BlockStmt bs = stmt.getChildNodesByType(CatchClause.class).get(0).getChildNodesByType(BlockStmt.class).get(0);
		remove(bs.getStatements().get(0));
    //		System.out.println(""+stmt.toString());
		remove(bs.getStatements().get(0));
    //		System.out.println(""+stmt.toString());
	}
	// remove the node and parent's node until response is true
	boolean remove(Node node) {
		boolean result = node.remove();
		if (!result && node.getParentNode().isPresent())
			result = remove(node.getParentNode().get());
		return result;
	}

jlerbsc:
Fixes #2187.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/remove/NodeRemovalTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/remove/NodeRemovalTest.java
@@ -22,49 +22,78 @@
 package com.github.javaparser.remove;
 
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.Statement;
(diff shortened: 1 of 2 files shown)


matozoid:
Thanks! This seems to break no existing functionality, so here it goes :-)

PR #2201, opened 2019-04-28
Fixes #2139

moraispgsi:
Fixes #2139.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/NodeList.java
@@ -481,7 +481,9 @@ public class NodeList<N extends Node> implements List<N>, Iterable<N>, HasParent
 
     @Override
     public void register(AstObserver observer) {
-        this.observers.add(observer);
+        if(!this.observers.contains(observer)) {
+            this.observers.add(observer);
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


moraispgsi:
@matozoid Not sure why it failed in travis.

matozoid:
Looks like Travis dropped support for Java 11 :-/

Anyway, @ftomassetti wasn't replying, but I would say that it is a user error to mix multiple lexical preserving printers in the same AST. It should be `setup` only once. So this PR would ignore a user error, probably leading to more trouble down the line.

moraispgsi:
@matozoid If this is a user error how would I go about joining two or more pieces of code? What the PR does is keep the same observer for the LexicalPreservingPrinter, when the code is joined together as if it was a single piece of code.

matozoid:
Ah, of course, this PR does not check if there is *any* observer already there, but only if the current one is there. But then how does it solve your problem? It was mixing several observer instances if I'm not mistaken?

moraispgsi:
Well the NodeList only adds observer instances with different references, and the LexicalPreserverPrinter does not create an observer for every setup, instead keeps it as a static variable and instantiates it lazily in the setup method.

moraispgsi:
@matozoid So now, with this PR, only one instance of the observer exists for the LexicalPreservingPrinter.

matozoid:
Travis is having a certificate problem and can be ignored.

I can't remember that clearly what I saw so I'll trust your judgement. Looks like fine code anyway :-)

PR #2208, opened 2019-05-02
Sonar issues : squid : S1166 - Exception handlers should preserve the…

jlerbsc:
Fix sonar issues
squid : S1166 - Exception handlers should preserve the original exception
pmd : EmptyCatchBlock - Catch blocks should not be left empty
This fix was realized with our own product based on javaparser.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -45,7 +45,8 @@ import java.nio.file.Path;
 import static com.github.javaparser.ParseStart.*;
 import static com.github.javaparser.Problem.PROBLEM_BY_BEGIN_POSITION;
 import static com.github.javaparser.Providers.*;
-import static com.github.javaparser.utils.Utils.assertNotNull;
+import static com.github.javaparser.utils.Utils.assertNotNull;import org.apache.log4j.Logger;
+
 
 /**
  * Parse Java source code and creates Abstract Syntax Trees.
@@ -54,6 +55,9 @@ import static com.github.javaparser.utils.Utils.assertNotNull;
  * @see StaticJavaParser
  */
 public final class JavaParser {
+    protected static Logger LOG = Logger.getLogger(JavaParser.class.getName());
+    
+
     private final ParserConfiguration configuration;
 
     private GeneratedJavaParser astParser = null;
@@ -128,7 +132,7 @@ public final class JavaParser {
         } finally {
             try {
                 provider.close();
-            } catch (IOException e) {
+            } catch (@SuppressWarnings("squid:S1166") IOException e) {
                 // Since we're done parsing and have our result, we don't care about any errors.
             }
         }
--- a/javaparser-core/src/main/java/com/github/javaparser/UnicodeEscapeProcessingProvider.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/UnicodeEscapeProcessingProvider.java
@@ -1,12 +1,16 @@
 package com.github.javaparser;
 
-import java.io.IOException;
+import java.io.IOException;import org.apache.log4j.Logger;
+
 
 /**
  * An implementation of interface CharStream, where the stream is assumed to
  * contain only ASCII characters (with java-like unicode escape processing).
  */
 class UnicodeEscapeProcessingProvider implements Provider {
+    protected static Logger LOG = Logger.getLogger(UnicodeEscapeProcessingProvider.class.getName());
+    
+
     private static int hexval(char c) throws java.io.IOException {
         switch (c) {
             case '0':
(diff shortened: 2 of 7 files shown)


matozoid:
Looks like it needs a little more work ;-) Most of these changes are not really applicable here though. But feel free to test a bit :-)

matozoid:
I'll add the reasons why.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -45,15 +45,19 @@
 import static com.github.javaparser.ParseStart.*;
 import static com.github.javaparser.Problem.PROBLEM_BY_BEGIN_POSITION;
 import static com.github.javaparser.Providers.*;
-import static com.github.javaparser.utils.Utils.assertNotNull;
+import static com.github.javaparser.utils.Utils.assertNotNull;import org.apache.log4j.Logger;
+
```
matozoid, line 49: javaparser-core has no dependencies and it should stay that way. So no log4j. And even then there would have been a preference for slf4j.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -128,7 +132,7 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
         } finally {
             try {
                 provider.close();
-            } catch (IOException e) {
+            } catch (@SuppressWarnings("squid:S1166") IOException e) {
                 // Since we're done parsing and have our result, we don't care about any errors.
```
matozoid, line 136: This project does not use Sonar, so no need to suppress its warnings. Besides that: there is a clear explanation on the next line.

javaparser-core/src/main/java/com/github/javaparser/UnicodeEscapeProcessingProvider.java
```diff
@@ -110,7 +114,7 @@ private void expandBuffer(boolean wrapAround) {
                 bufpos -= tokenBegin;
             }
         } catch (Exception t) {
-            throw new RuntimeException(t.getMessage());
+            throw new RuntimeException(t.getMessage(), t);
```
matozoid, line 117: This whole class has been copied from JavaCC. There is an issue to completely refactor it because it is a horror.

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
```diff
@@ -9,13 +9,17 @@
 import java.nio.file.PathMatcher;
 import java.util.Optional;
 
-import static com.github.javaparser.StaticJavaParser.parse;
+import static com.github.javaparser.StaticJavaParser.parse;import org.apache.log4j.Logger;
+
```
matozoid, line 13: Is this one of those `LexicalPreservingPrinter` issues?

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
```diff
@@ -24,9 +28,9 @@
             return parse(file.toFile()).getStorage()
                     .map(CompilationUnit.Storage::getSourceRoot);
         } catch (ParseProblemException e) {
-            Log.info("Problem parsing file %s", () -> file);
+            Log.info("Problem parsing file %s", () -> file, e);
```
matozoid, line 31: This is not log4j :-)

javaparser-core/src/main/java/com/github/javaparser/utils/Log.java
```diff
@@ -47,7 +47,7 @@ private void printStackTrace(Throwable throwable) {
                 throwable.printStackTrace(pw);
                 trace(sw::toString);
             } catch (IOException e) {
-                throw new AssertionError("Error in logging library");
+                throw new AssertionError("Error in logging library", e);
```
matozoid, line 50: This is supposed to be unreachable code, but if it *is* reached it might be good to know exactly what happened - so this one is useful :-)

matozoid:
Maybe it's a good idea to try your tool on a more "normal" project :-)

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
jlerbsc: I don't know if there is a LexicalPreservingPrinter issue. The tool try to add an import with a default log adapter but there is an issue in the tool. I'm working on this.

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
jlerbsc: Yes i see it too late :-( com.github.javaparser.utils.Log

matozoid: It is of course a weird thing. Here is the background if you're interested: https://javaparser.org/javaparsers-logging-framework-in-one-file/

PR #2209, opened 2019-05-03
Line ending remains when removing a statement

MarvinWyrich:
In the following example a statement (`int result = a * b;`) is removed. Not removing the EOL character causes the formatting to break. Unfortunately, this is the case for files with Unix line endings.

```
\t void multiply(int a, int b) { \n
\t \t int result = a * b; \n
\t \t return a * b; \n
\t }
```
becomes
```
\t void multiply(int a, int b) { \n
\t \t \n
return a * b; \n
\t }
```
where the `\n` in the second line should have been removed (at least this is what happens for the same file with Windows line endings). I would be happy to help with the fix, but would appreciate the input from more experienced project developers first. Maybe related: #2127.


matozoid:
Thanks!

matozoid:
We have @ftomassetti who created this printer, but I'm guessing he will say "yes, that \n should have been removed" :-)

A fix is very welcome.

ftomassetti:
yes, that \n should have been removed

:D

matozoid:
I can add: there are three (or more?) tokens representing EOL's. All of these tokens will return "true" when asked if they are an EOL token. If some part of the code is explicitly checking \n or \r\n, or is checking for a specific EOL token, it is wrong.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
```diff
@@ -390,7 +418,51 @@ void printASimpleMethodAddingAStatement() {
...
+    			+ "}";
+    			
+    	
+    	CompilationUnit cu = parse(code);
+    	LexicalPreservingPrinter.setup(cu);    	
+    	ExpressionStmt stmt = cu.findAll(ExpressionStmt.class).get(0);
+    	stmt.remove();
+    	
+        assertEqualsNoEol("class A {" 					+ EOL
+    			+ "\t"		+  "foo(int a, int b) {"	+ EOL
```
matozoid: I think (can't check right now) that assertEqualsNoEol matches any EOL in "actual" with "\n" in "expected" - so the left argument should have \n, not EOL.

MarvinWyrich: It replaces all EOL characters in "expected" (left argument) with `System.getProperty("line.separator");`. I think the confusion comes from my unnecessary use of the function when the EOL constant is already used (we could also just use `assertEquals` here). I have also just noticed that this test is not ideal because it either passes or fails depending on the system. I could change the EOLs in this test to "\r\n" (both in expected and in actual) to demonstrate the different behavior to the unit test that follows. Would that also make sense from your point of view?

matozoid: Yes, that seems fine, since that's the combination that's giving the trouble here.

MarvinWyrich:
Update on the debugging status:

The `differenceElements` are calculated correctly:
```
{\r\n
CsmIndent Removed(int result = a * b;)Removed(\r\n)
return a * b;\r\n
CsmUnindent}
```

when applying the difference (`apply()` in `Difference.java`), it first removes the expression. Then it fails to remove the second to be removed token due to the following condition evaluating to false:

```
else if (removed.isToken() && originalElementIsToken
                && (removed.getTokenType() == ((TokenTextElement) originalElement).getTokenKind())) {
            nodeText.removeElement(originalIndex);
            diffIndex++;
        }
```

This is because `removed.getTokenType()` always returns **2** (= token type for `\r\n`).  
The kind of token for `originalElement` is either **2** in the test case for `\r\n` line endings or **3** in the test case for `\n` line endings or **4** in the test case for `\r` line endings. This mismatch then causes the algorithm not to remove the element and just to incrase the `diffIndex`.

I have seen that the `JavaToken.java` also has an enum `Category`. Maybe this can be useful at this point. But it could also be the case that `removed.getTokenType()` should not always return **2**. I will take a closer look at this later. If you spontaneously have some input for me, you are welcome to tell me.

matozoid:
Random thought: I'm guessing that eol's need some special casing here. It's also worth investigating if more of these checks exist in the code and if they contain the same bug.

MarvinWyrich:
The root cause seems to be that when creating the diffelements the EOL of the respective operating system is used instead of the EOL of the original code (which can result in a mismatch during later comparison). CRLF and CR fail for the [travis build](https://travis-ci.org/javaparser/javaparser/builds/528768711), LF and CR fail on my local machine (Windows).

Not sure where exactly this happens and if it would be easily fixable, but that seems to be the reason for the behavior. I could probably fix the symptoms with a special case handling, but I would like to hear the opinion of @ftomassetti before :)

matozoid:
What would happen with a source file where someone messed up and mixed the EOL's? (Not that this happens too often, but it's a good problem for designing a solution)

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -1,18 +1,37 @@
...
+import static com.github.javaparser.GeneratedJavaParserConstants.CHAR;
+import static com.github.javaparser.GeneratedJavaParserConstants.DOUBLE;
+import static com.github.javaparser.GeneratedJavaParserConstants.FLOAT;
+import static com.github.javaparser.GeneratedJavaParserConstants.INT;
+import static com.github.javaparser.GeneratedJavaParserConstants.LBRACE;
+import static com.github.javaparser.GeneratedJavaParserConstants.LONG;
+import static com.github.javaparser.GeneratedJavaParserConstants.RBRACE;
+import static com.github.javaparser.GeneratedJavaParserConstants.SHORT;
+import static com.github.javaparser.GeneratedJavaParserConstants.SPACE;
+
```
matozoid: Careful with the formatter there! :-)

matozoid:
Cool :-)

PR #2211, opened 2019-05-04
Issue 2195

### LINKED ISSUE #2195: Resolution of implmented/extended types falsely resolves to nested types

MysterAitch:
Potentially further work required re: resolving the types of fields, but fixes #2195 as described in the linked issue.


matozoid:
Nice :-) Let's see if @ftomassetti has time to check it.

PR #2226, opened 2019-05-13
[TESTCASE] Lexpres fails to keep EOL tokens behind line end comments

MarvinWyrich:
Given:

```
class A {
  foo() {
    int result = 53; // comment
    return 53;
  }
}
```

Removing the first statement `int result = 53;` gives me the following output:

```
class A {
  foo() {
    // comment return 53;
  }
}
```

but I would expect:

```
class A {
  foo() {
    // comment
    return 53;
  }
}
```

Tests run successfully for version 3.13.6, but not for newer versions.  
Looking at the title and the discussion, my bet would be that #2168 introduced this bug.


matozoid:
That would be me then, but I honestly don't know how that part of the lex printer is supposed to work :-/

MarvinWyrich:
It looks like all previous tests for handling line comments assume that comments are in a separate line. End of line comments cause the problem described above. I also find it very difficult to understand where the actual problem lies in the code, and I'm surprised it doesn't affect more people.

Is there a chance that @ftomassetti will come back one day? 😅

matozoid:
Actually, chucking a breakpoint on the changed part of the lex pres printer will show the context for what it's doing. I don't know what it is supposed to be doing, so I just changed things until it worked :-)

jlerbsc:

PR #2232, opened 2019-05-17
Implement fully qualified name logic for type declarations

### LINKED ISSUE #2230: get the fully qualified type/class name for a TypeDeclaration
on the JP gitter we discussed the possibility of having a convenience method to resolve the fully qualified type name from a `TypeDeclaration`.  this will involve either manually walking the parent nodes until there are none or we reach a TopLevelClass, and then extracting the package name from the TopLevelClass's CompilationUnit.

I wrote a "quick and dirty" variant that works (for the most simple cases... Top level and nested types (classes, enums, interfaces and annotations) but probably needs more testing to be considered a more "general tool" and included in JavaParser...(The other question is "in what class should this method reside")

Some cases which need to be tested (or specifically noted in the Javadoc as "not supported") are:
Handling Anonymous Classes (since these are going to be created inside methods or constructors or Static initializer blocks)
(i.e. new Serializable(){})
... although
Handling Local Classes 
(i.e. 
```java
package aaaa.bbbb;
class Top{
    void m(){
        class local{
        }
    }
}
```
(these local classes are ALSO constructed in methods or constructors or static blocks) 

some side issue that probably needs to be determined is... do we maintain a reference to the method/constructor/staticBlock that contains the LocalClass/Anonymous class
(i.e. in the local class `local` above, how do we signify the path to the class?
"aaaa.bbbb.Top.local" (omitting the method name "m")
"aaaa.bbbb.Top#m.local (using the # notation to refer the method name "m")
"aaaa.bbbb.Top.m.local (using the . notation to refer the method name "m") 

...also a consideration needs to be made to overloaded methods:
```java
package aaaa.bbbb;
class Top{
    void m(){
        class local{
        }
    }
    void m(int i){
      class local{
      }
    }
}
```

because in the above, any of the (3) notations above are ambiguous (because they could mean  any version of local)

...also if we have a local class in a static block, how do we represent that

```java
package aaa.bbb;
class Top{
    static {
      class Local {
      }
    }
}
```
it's totally up in the air, but we could do something like #init:
"aaa.bbb.Top#init"

Finally just to exhast other things that should be considered... you CAN define a class inside a lambda body:

```java
Consumer<String> c = c-> {
                class FFF{
                          
                }
            };
            
            Consumer<Integer> u = y-> { 
                class FFF{
                    
                }
            };
```

and the above (2) implementations of FFF could be on any unnamed Lambda, so again we have a problem not being able to follow a unique reference.

(TBH sorry for being this thorough, but I could very well see that many of these situations (i.e. Local Classes) are simply unsupported by the method (it makes the logic easier, and the corner cases are not that prevalent in real code)  (I'm sorta just rationalizing why I didn't want to go too far down the rabbit hole" and or "share the original code" if it gets misused for those cases it did not support.

Below is the original naive code for the purpose to be used as reference or just throw away entirely
```java
    static String getFullName(TypeDeclaration astType){
        if( astType.isTopLevelType() ){
            if( astType.findCompilationUnit().get().getPackageDeclaration().isPresent()) {
                String pkgName = astType.findCompilationUnit().get().getPackageDeclaration().get().getNameAsString();
                return pkgName +"."+ astType.getNameAsString();
            }
            return astType.getNameAsString(); //top level TYPE but no package declaration
        }
        if(!astType.getParentNode().isPresent()){
            return astType.getNameAsString();
        }
        //prefix back to parents
        String name = astType.getNameAsString();
        Node n = astType.getParentNode().get();
        if( n instanceof TypeDeclaration ){
            return getFullName( (TypeDeclaration)n)+"."+name;
        }
        if( n instanceof LocalClassDeclarationStmt ){
          LocalClassDeclarationStmt lc = (LocalClassDeclarationStmt) n;
          return lc.getClassDeclaration().getNameAsString();
        } else {
            NodeWithName nwn = (NodeWithName) n;
            name = nwn.getNameAsString() + "." + name;
            if (n.getParentNode().isPresent()) {
                return getFullName((TypeDeclaration) n) + "." + name;
            }
        }
        return name;
    }
```

matozoid:
Fixes #2230.

@edefazio - is this useful to you?

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/TypeDeclarationTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/TypeDeclarationTest.java
@@ -1,0 +1,64 @@
+package com.github.javaparser.ast.body;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import org.junit.jupiter.api.Test;
+
+import static com.github.javaparser.utils.TestParser.parseBodyDeclaration;
+import static com.github.javaparser.utils.TestParser.parseCompilationUnit;
+import static java.util.stream.Collectors.joining;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class TypeDeclarationTest {
+    @Test
+    void qualifiedNameOfClassInDefaultPackage() {
+        assertFQN("X", parseCompilationUnit("class X{ }"));
+    }
+
+    @Test
+    void qualifiedNameOfClassInAPackage() {
+        assertFQN("a.b.c.X", parseCompilationUnit("package a.b.c; class X{}"));
+    }
+
+    @Test
+    void qualifiedNameOfInterfaceInAPackage() {
+        assertFQN("a.b.c.X", parseCompilationUnit("package a.b.c; interface X{}"));
+    }
+
+    @Test
+    void qualifiedNameOfEnumInAPackage() {
+        assertFQN("a.b.c.X", parseCompilationUnit("package a.b.c; enum X{}"));
+    }
+
+    @Test
+    void qualifiedNameOfAnnotationInAPackage() {
+        assertFQN("a.b.c.X", parseCompilationUnit("package a.b.c; @interface X{}"));
+    }
+
+    @Test
+    void qualifiedNameOfNestedClassInAPackage() {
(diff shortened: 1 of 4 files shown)


edefazio:
Thanks, that's much more concise than my code... (I'm going to happily retire the old stuff in draft when the new build gets done) 

BTW, your functional lambda-fu is strong,...I am a mere functional novice

matozoid:
@edefazio don't fight the lambdas and they will guide you :-)

PR #2233, opened 2019-05-18
The Javadoc for Range.strictlyContains now fits its implementation

### LINKED ISSUE #2231: Javadoc of Range.strictlyContains doesn't match its implementation
The Javadoc for `Range.strictlyContains` is as follows:
```
/**
     * Do this strictly contains other? It means that this has to be larger than other and it has to start as other
     * or before and end as other or after.
     */
```
However, the implementation does not account for the `start as other` part. the implementation is currently as follows:
```
    public boolean strictlyContains(Range other) {
        return begin.isBefore(other.begin) && end.isAfter(other.end);
    }
```

However, according to the Javadoc I'd expect it to be like this:
```
public boolean strictlyContains(Range other) {
        return (begin.isBefore(other.begin) || begin.equals(other.begin))&& end.isAfter(other.end);
    }
```

Which one is correct: the Javadoc (which has my preference) or the implementation? I can change it via a PR, but first wanted to verify this problem via an issue.  The `strictlyContains` method does not seem to be used anywhere else in Javaparser, so I assume changing the implementation to the one staged above wouldn't be an issue.

SimonBaars:
Fixes #2231.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/Range.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/Range.java
@@ -82,8 +82,8 @@ public class Range {
     }
 
     /**
-     * Does this strictly contain other? It means that this has to be larger than other and it has to start before other
-     * and end after other.
+     * Does this strictly contain other? It means that this has to be larger than other and it has to start before
+     * other and end after other.
      */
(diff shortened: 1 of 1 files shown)


matozoid:
Perfect!

PR #2238, opened 2019-05-21
Added encoding parameters to all save functions

### LINKED ISSUE #2237: The save() Method of CompilationUnit should not hardcode its encoding
Hey guys, I noticed something while working on a codebase which has to be encoded using ISO-8859-1. The save Function of CompilationUnit currently has no way of specifying the Charset.
```java
public void save(Function<CompilationUnit, String> makeOutput) {
    try {
        Files.createDirectories(path.getParent());
        final String code = makeOutput.apply(getCompilationUnit());
        Files.write(path, code.getBytes(UTF8));
    } catch (IOException e) {
        throw new RuntimeException(e);
    }
 }
```
Shouldn't it be possible to pass an optional Charset to this Method? Like this:

```java
public void save(Function<CompilationUnit, String> makeOutput, charset: Charset) { 
    ...
    Files.write(path, code.getBytes(charset));
    ...
}
```
Or is there any reason for this behaviour? If you want me to, I could create a PR for this

Funkschy:
Closes #2237.
Added a bunch of optional Parameters for the various save methods, to allow users to specify a file encoding.
When the user does not specify an encoding we should use the encoding of ParserConfiguration.
Also added the encoding to CompilationUnit.Storage

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -173,7 +173,7 @@ public final class JavaParser {
     @Deprecated
     public ParseResult<CompilationUnit> parse(final File file, final Charset encoding) throws FileNotFoundException {
         ParseResult<CompilationUnit> result = parse(COMPILATION_UNIT, provider(file, encoding));
-        result.getResult().ifPresent(cu -> cu.setStorage(file.toPath()));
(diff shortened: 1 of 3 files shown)


matozoid:
This is absolutely wonderful!

Funkschy:
@matozoid this project is absolutely wonderful, I'm happy to help, even though it's just a small thing.
Thank you for your hard work :)

PR #2240, opened 2019-05-22
Use the encoding from the config in tryParse()

Funkschy:
I only changed the methods related to save() in my PR yesterday, but today i realised,
that the Parser should also use the configured encoding in the parse Method.
With this change, it is now possible to parse statements like 'int äöü = 0' and save them
with the same encoding used in the original file.


matozoid:
Thanks again!

PR #2244, opened 2019-05-24
Return early when searching for a method without parameters

maartenc:
When searching for methods without parameters, you can stop searching when you find one in the given class. 

Although this PR shouldn't change the behaviour of the method, it fixes an issue I encounter when trying to resolve a method without having the full class hierarchy in my classpath. As an extra benefit, it should make method-resolution a little bit faster for methods without parameters :)


matozoid:
It's Tomassetti Time again! @ftomassetti

maartenc:
Sorry @matozoid , the next issue I'll create will be for you ;-)

ftomassetti:
I guess that also this makes sense. Still, you should put all your
dependencies on the silver, not relying on tricks :)
Yet it seems a very common case so the optimization seems very useful
Thank you!
BTW, what are you building with JavaParser?

PR #2267, opened 2019-06-12
Re-implemented unicode preprocessor

### LINKED ISSUE #1930: Clean up unicode preprocessor
See  #1928

Specifically:
- the code has a lot of redundant stuff in it.
- it throws a runtime exception when a bad escape is encountered (like `\uz`). It should just pass those characters through. (Which is expected behaviour starting from Java 12, but is nice to have earlier too, because the bad escapes get passed into the parser which can produce a decent error instead of a runtime exception.)

haumacher:
Fixes #1930.


matozoid:
Thanks!

PR #2268, opened 2019-06-13
Issue2258: NullType must not fail matchTypeParameters

### LINKED ISSUE #2258: Type resolution issue of static methods call expressions
Hi,

I ran into a problem when calling calculateResolvedType() on a MethodCallExpr.
It seems that the call would throw an UnsupportedOperationException when:
1. The called method is static
2. The expression calls the method in the format of "{class name}.{method name}" (the {class name} part is not omitted)
3. A parameter of the method has an array type
4. The actual parameter passed to that array-type parameter is null

Here's an example piece of code to reproduce the issue:
```
public class TestClass {

	public static void foo(Object[] nullable) {
		System.out.println("Verify program execution."); // This line is not necessary.
	}

		TestClass.foo(null);
	}
}
```
JavaParser throws when it is resolving the type of the call in the main method.
The stack trace looks like:
```
java.lang.UnsupportedOperationException: com.github.javaparser.symbolsolver.model.typesystem.NullType
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.matchTypeParameters(MethodCallExprContext.java:361)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.resolveMethodTypeParameters(MethodCallExprContext.java:345)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:86)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:584)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:267)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:44)
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:115)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:448)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:310)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:292)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.calculateType(JavaSymbolSolver.java:250)
	at com.github.javaparser.ast.expr.Expression.calculateResolvedType(Expression.java:564)
```

For now, I'm using the workaround of calling resolve() on the method call expression and then getting the return type of the resolved method. This seems to work fine.
I thought it would be nicer if calculateResolvedType() works as well.

jlerbsc:
Fixes #2258. Accept NullType when trying to match type parameters on MethodCallExpr

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
@@ -357,12 +357,13 @@ public class MethodCallExprContext extends AbstractJavaParserContext<MethodCallE
             }
             matchedTypeParameters.put(expectedType.asTypeParameter(), actualType);
         } else if (expectedType.isArray()) {
-            if (!actualType.isArray()) {
+        	// Issue 2258 : NullType must not fail this search
+            if (!(actualType.isArray() || actualType.isNull())) {
                 throw new UnsupportedOperationException(actualType.getClass().getCanonicalName());
             }
             matchTypeParameters(
                     expectedType.asArrayType().getComponentType(),
-                    actualType.asArrayType().getComponentType(),
+                    actualType.isNull() ? actualType : actualType.asArrayType().getComponentType(),
                     matchedTypeParameters);
         } else if (expectedType.isReferenceType()) {
             // avoid cases where the actual type has no type parameters but the expected one has. Such as: "classX extends classY<Integer>"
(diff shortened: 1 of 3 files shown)


matozoid:
Thanks, let's see what @ftomassetti says.

matozoid:
Ouch, this PR breaks the build - can you check it out @jlerbsc ?

jlerbsc:
Yes. It's because i use junit4 in the eclipse project, I will correct that.

ftomassetti (approved):
Yes, this makes sense. Null can be used as an array value

PR #2271, opened 2019-06-15
Correct positions with unicode escape processing

### LINKED ISSUE #2270: Incorrect source code positions in standards-complient unicode escaping mode
With standards-complient unicode escaping (`com.github.javaparser.ParserConfiguration.setPreprocessUnicodeEscapes(true)`), the source code positions reported in the AST are incorrect: Unicode escape sequences are processed before parsing so that the scanner consumes the decoded stream counting only positions in the decoded version of the source file. Thes positions are reported in the AST.

In the example where braces are replaced by corresponding unicode escape sequences
```
@interface AD \u007B String foo(); \u007D
```
the AST reports the range (1,17)-(1,29) for the method declaration `String foo();`, while it spans the range (1, 22)-(1, 34) in the original source.

haumacher:
Fixes #2270.

A known problem in the position mapping so far is that when the unicode decoder encounters a broken unicode escape sequence it does not adjust the internal position after pushing back the already read chars. This would result in incorrect positions if the parse operation continues after the broken unicode escape. Currently, the scanner immediately aborts after seeing a broken unicode escape, therefore, the problem cannot be observed in a complete parse test.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -21,34 +21,41 @@
 
 package com.github.javaparser;
 
-import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.NodeList;
-import com.github.javaparser.ast.body.AnnotationMemberDeclaration;
-import com.github.javaparser.ast.body.MethodDeclaration;
-import com.github.javaparser.ast.expr.*;
-import com.github.javaparser.ast.stmt.*;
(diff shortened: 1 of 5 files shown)


matozoid:
What happens when a Unicode escape is broken can be moved to another issue. This PR doesn't seem to make it worse. I'm thinking that somehow it should continue, marking every broken escape as a problem. Since error recovery like this has been implemented in only few places, it's low priority. (Until someone starts trying to build an IDE on top of JavaParser again ;-) )

matozoid:
@haumacher - are you happy with this PR? If so, I'll merge it.

PR #2278, opened 2019-06-23
Copy orphan comments.

### LINKED ISSUE #2277: Missing orphan comments on clone.
Cloning a node, the orphan comments are missing.

anelam:
Copy all orphan comments on cloning a node.

Fixes #2277.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java
@@ -54,6 +54,7 @@ public class CloneVisitorGenerator extends VisitorGenerator {
 
         body.addStatement(builder.toString());
         body.addStatement("r.setComment(comment);");
+        body.addStatement("n.getOrphanComments().stream().map(Comment::clone).forEach(r::addOrphanComment);");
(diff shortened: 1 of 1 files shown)


matozoid:
Hi! Thanks for the PR :-)

It's the right thing to do, with one problem: the `CloneVisitor` is generated by the https://github.com/javaparser/javaparser/blob/master/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/visitor/CloneVisitorGenerator.java

The generators are run every few months when the AST changes, so that would overwrite the changes in this PR.

If you want to change the generator, the easiest way to run them is by running this: https://github.com/javaparser/javaparser/blob/master/run_core_generators.sh (if you're on Windows it should be easy to make it a batch file.) If you want to run it from an IDE, you need to set some parameter - if you can't figure it out, I can help.

anelam:
Oh thanks, I will fix the PR.

anelam:
I updated the PR, hoping that it will fix #2277 now.

matozoid:
It works fine, thanks!

malteskoruppa:
@matozoid If I see correctly, the generators have not been rerun since this PR, so that the fix has not made it into the `CloneVisitor` yet. If possible, I would be much obliged if you could run the generators for the next release, so that this fix becomes active. :)

matozoid:
Oops! PR opened!

PR #2294, opened 2019-07-14
Use File.isSameFile instead of Path.equals in SymbolSolverCollectionStrategy

pkoziol:
Similar change as done for ParserCollectionStrategy in #2041.
Fixes problems with different separators and relative paths on Windows.


matozoid:
Everything is green, so here it goes :-) Thanks! Next release is next Sunday.

PR #2299, opened 2019-07-16
Added a overlapsWith method for Ranges

SimonBaars:
This way we can check whether two ranges overlap with each other. I also created testcases. I also created alternatives for the `contains` and `strictlyContains` methods, to allow for positions.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/RangeTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/RangeTest.java
@@ -128,10 +128,11 @@ class RangeTest {
     }
     
     @Test
-    void rangesOverlapBegin() {
+    void rangesOverlap() {
         Range r1 = Range.range(1, 1, 3, 10);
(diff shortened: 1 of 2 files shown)


javaparser-core-testing/src/test/java/com/github/javaparser/RangeTest.java
```diff
@@ -102,5 +102,53 @@ void lineCountIsReturned() {
...
+
+    @Test
+    void aRangeContainsItsEnd() {
+        Range r = Range.range(1, 1, 3, 10);
+        assertTrue(r.contains(r.end));
+    }
+    
+    @Test
+    void aRangeStrictlyContainsItsBegin() {
+        Range r = Range.range(1, 1, 3, 10);
```
matozoid, line 120: does *not* contain, right?

matozoid:
Thanks!

matozoid:
Oh wait, I still had comments open :-D

PR #2302, opened 2019-07-19
Should containsWithin() compare files as well?

malteskoruppa:
_Please note that some tests are failing in this PR. Therefore it's not an actual pull request, but more of an issue accompanied by some code. :)_

I recently had a weird bug where some node sayed it contained some other node, even though the two nodes were in completely different files. I had used `Node#containsWithin()` for checking whether one node contained the other.

I then realized that `containsWithin()` compares _only the ranges_, but _not the files_ that these nodes are contained in. At first I thought it was a bug, which is why I starting writing a fix to prepare a pull request. I then realized this may not be a bug at all, because:

1. In order to compare the files as well, we need to find the compilation unit, and thus we need to add an abstract method `findCompilationUnit()` to the interface `NodeWithRange`, which is a bit ugly maybe (see changes in this PR)
2. Also, once I had changed `NodeWithRange` as described, various tests started failing (well, 12 tests, actually), especially the PrettyPrintVisitorTests and the LexicalPreservingPrinterTests.

So I'm not sure now... is it intentional that `containsWithin()` compares based on ranges only, but not based on files? If so, I would suggest:

1. at least making that clear in a Javadoc, because it can be confusing. The name `containsWithin()` suggests that one node is contained within the other, so perhaps the method name is not ideal, though I'll admit I have no better idea on the spot. ;-)
2. providing a method (perhaps in `Node` rather than in `NodeWithRange`) that _actually_ checks whether some node is a descendant node of another node. This would be a pretty generic and useful method I think, and it can be implemented in a straightforward way by using `findFirst()` behind the scenes.

I could take care of that, if you agree.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/nodeTypes/NodeWithRangeTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/nodeTypes/NodeWithRangeTest.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2016 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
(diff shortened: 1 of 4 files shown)


matozoid:
I agree that it is confusing as it is right now. It would be interesting to analyze the errors, those might be bugs in the code. @danilomo - are you interested?

Naming... maybe deprecate this one and call it `containsWithinRange` and another called `containsNode`?

malteskoruppa:
As far as I am concerned, this can be closed. This has been (so to say) fixed in #2306. Though it may still be interesting to determine why some tests failed when I changed `containsWithin()` to compare files as well, as described above. If @danilomo is interested in investigating, I'd be happy to provide any help necessary. :)

matozoid:
@danilomo said he's going to be busy for a while, so let's close this then :-)

PR #2306, opened 2019-07-22
Introduced isDescendantOf(Node) and isAncestorOf(Node) methods

malteskoruppa:
See #2302.

This PR introduces new `isDescendantOf(Node)` and `isAncestorOf(Node)` convenience methods. Without them, users (myself included) may accidentally use `containsWithin(Node)`, which does something a bit different. :)

Also, `containsWithin(Node)` has been deprecated in favor of a new method `containsWithinRange(Node)`.

There is one thing that slightly bugs me still. There's an inconsistency that currently already exists in JavaParser and that became particularly obvious when I implemented the new methods. Namely, the method `findFirst()` starts walking the tree downwards from the _current_ node, whereas the method `findAncestor()` does not start walking the tree upwards from the _current_ node, but starts walking the tree upwards from the _parent_ node.

The new `isDescendantOf()` and `isAncestorOf()` methods are just straightforward implementations based on `findAncestor()` and `findFirst()`, respectively. The problem is that I wanted to be able to consistently say "a node is a descendant of itself" and "a node is an ancestor of itself" (either both, or neither; but we can't say one, but not the other, that would just be inconsistent). So the implementation of `isDescendantOf(Node)`, which uses `findAncestor()`, has to use an additional check whether the given node is equal to the current object, whereas the method `isAncestorOf()`, which uses `findFirst()`, doesn't have to do that check.

I think it would be more consistent to rewrite `findAncestor()` in such a way that it also starts from the _current_ node instead of the _parent_ node, what do you think? I didn't want to do that now, though, because that would be a change in behavior of an existing method which may surprise users... so at best, it'd be a story for another PR. What do you think?

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/AncestorDescendantTests.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/AncestorDescendantTests.java
@@ -12,7 +12,7 @@ import static org.junit.jupiter.api.Assertions.*;
 public class AncestorDescendantTests {
 
     @Test
-    void nodeIsAncestorOfItself() throws IOException {
+    void nodeIsNotAncestorOfItself() throws IOException {
         JavaParser parser = new JavaParser();
 
         Provider provider = Providers.resourceProvider("com/github/javaparser/range/A.java");
@@ -26,11 +26,11 @@ public class AncestorDescendantTests {
                 .getBody().get().getStatement(0).asExpressionStmt() // int a = 42;
                 .getExpression().asVariableDeclarationExpr(); // a = 42
 
-        assertTrue(node.isAncestorOf(node));
+        assertFalse(node.isAncestorOf(node));
     }
 
     @Test
-    void nodeIsDescendantOfItself() throws IOException {
+    void nodeIsNotDescendantOfItself() throws IOException {
         JavaParser parser = new JavaParser();
 
         Provider provider = Providers.resourceProvider("com/github/javaparser/range/A.java");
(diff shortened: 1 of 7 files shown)


matozoid:
Thanks!

And oh yeah, that issue again :-)
In my opinion: walk and find are simple tree walking algorithms. If you walk a tree, you include the root. JP has a bunch of methods that use the terms "parent", "ancestor", "child", "descendant", and those do *not* include the root. Being your own parent or own child would just be weird.

My feeling is that we could add `findDescendant` and such, being like `find` but excluding the root. Possibly, to avoid confusion, we could rename `find` to `findInTree` or so.

malteskoruppa:
Yes, having a `findDescendant()` that is kind of the opposite of `findAncestor()` would maybe make things clearer. None of these would include the root; and if you want to walk the subtree from the current position, you use `find{First,All}()` (or `find{First,All}InTree()`, but I personally don't feel that it would be necessary to rename the current methods) instead. I like this idea. :)

Hmm, if JP generally doesn't include the root for the methods that use "ancestor" or "descendant" in their names, should I change the methods `isDescendantOf()` and `isAncestorOf()` in this PR to not include the root either? That way, we would consistently define a node to _not_ be an ancestor or descendant of itself in the context of a JavaParser AST. I think there is no globally agreed-upon definition as far as that matter is concerned anyway, although my feeling is that in most theory textbooks on graphs and trees, nodes are defined to be ancestors or descendants of themselves (but I may be wrong). And anyway, noone forces us to adhere to some theory textbooks in the first place. :)

matozoid:
Here, I started a terminology block: https://javaparser.org/code-style-architecture/

malteskoruppa:
Ok then, I fixed the methods `isAncestorOf()` and `isDescendantOf()` accordingly. :-)

matozoid:
Awesome :-)

Issue #2152, opened 2019-04-04
Parent node is not set for comment nodes


malteskoruppa:
It appears that `Comment` nodes have no parent node set, at least not always.

For instance, parse this file using `JavaParser.parse(File)`:

```java
public class Foo {

  public void test() {
    // blah
    int i = 42;
  }
}
```

Find the comment node representing `// blah` (it's under the `ExpressionStatement` representing `int i = 42;`). This comment node does not have the `ExpressionStatement` as its parent; instead, its parent is set to `null`.

This is a problem, because: I have an application that scans lots of files and a visitor which inspects various nodes within these files. Under certain criteria, nodes are reported with their file name and range. Amongst other nodes, sometimes comments may be reported as well. In this situation, however, the file that the comment belongs to cannot be retrieved, because to retrieve the file that a node belongs to, I need to find its `CompilationUnit` first. But I can't do that if the parent is `null`.

Of course I can work around that in some hacky way, but it's not very nice. :)

Is it intentional that `Comment` nodes have no parent node?
If yes: Given only a `Comment` node (and nothing else), what would be the preferred way to retrieve the file that this `Comment` node belongs to (assuming it was parsed from a file rather than just a code string)?

matozoid:
it seems the code generator doesn't want to generate `Node.setComment` for us... That's odd.

matozoid:
Oh look: https://github.com/javaparser/javaparser/blob/master/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/node/PropertyGenerator.java#L52

So I guess we kind of forgot to keep that up to date manually.

malteskoruppa:
So if I understand correctly, a `this.comment.setParentNode(this);` after `this.comment.setCommentedNode(this);` in the method `Node#setComment(Comment)` should fix this? :-)

matozoid:
Just compare it to another normal setter.

A question to ask before this would be: why are we ignoring the setter, and can we stop ignoring it? I'm curious what breaks when we remove the exception and run the generators.

matozoid:
Okay, I got it again! A comment is made to point to the node it is associated with through its "commentedNode" field. So everything is as it should be right now, and you can get back into the AST proper by calling `getCommentedNode()` :-)

Odd design, but remember that comment support is a bit of an oddity anyway.

matozoid:
@malteskoruppa - is this answer okay?

malteskoruppa:
Hi!

[...]

PR #2307, opened 2019-07-22
Fix findRootNode() and associated methods for comment nodes

malteskoruppa:
See #2152. This PR does not yet change the whole architecture of comments to take _all_ comments (including orphan comments) out of the AST, as discussed in that issue. But it already fixes `findRootNode()` (and methods which use `findRootNode()`, such as `findCompilationUnit()`) for comments nodes for the time being, as discussed at the very beginning of that issue. :-)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -239,12 +239,9 @@ public abstract class Node implements Cloneable, HasParentNode<Node>, Visitable,
      *
      * @param comment to be set
      */
-    public final Node setComment(final Comment comment) {
+    public Node setComment(final Comment comment) {
         if (this.comment == comment) {
             return this;
-        }
-        if (comment != null && (this instanceof Comment)) {
-            throw new RuntimeException("A comment can not be commented");
         }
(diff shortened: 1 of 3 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -694,6 +694,10 @@ public boolean replace(Node node, Node replacementNode) {
      */
     public Node findRootNode() {
         Node n = this;
+        // (Non-orphan) comments are not integrated into the normal AST; we need to get the commented node first.
+        if (this instanceof Comment && ((Comment) this).getCommentedNode().isPresent()) {
+            n = ((Comment) this).getCommentedNode().get();
+        }
```
matozoid: This could be done with a method override in comment, or not?

malteskoruppa: Uh oh, I completely agree!

I fixed that. Also, while doing so, I moved the special logic required when `setComment(Comment)` is called on a Comment node (-> comments on comments are not allowed) to the Comment class. There, we had exactly the same glitch: a `this instanceof Comment` was used rather than a proper method override. :-)

matozoid:
Thanks again!

PR #2316, opened 2019-08-05
JEP 354

### LINKED ISSUE #2266: Revert `break` changes and implement JEP 354
https://openjdk.java.net/jeps/354

matozoid:
Fixes #2266.


edefazio (commented):
ONly thing I saw was on the PreetyPrintVisitor:
````java
public void visit(YieldStmt n, Void arg) {
        printer.print("yield");
        n.getExpression().accept(this, arg);
}
````
Since there is only a conditional space after the "break" keyword (if there is a label): 

````
public void visit(final BreakStmt n, final Void arg) {
        printer.print("break");
        n.getValue().ifPresent(value -> {
            printer.print(" ");
            value.accept(this, arg);
        });
        n.getLabel().ifPresent(l -> printer.print(" ").print(l.getIdentifier()));
        printer.print(";");
    }
````

arent we going to need a space after "yield"... i.e. "yield " since we assume we will always have a expr after the "yield" keyword and only conditionally on the "break" keyword

matozoid:
- [x] note that the serialization will break, delivering a lot of bad tokens in the tokenlist. (But will probably work okay for the rest)
- [x] write a test for the pretty printer and fix the space bug
- [x] make YieldStmt implement `NodeWithExpression`
- [x] note that Java 12 support no longer includes break-with-value

edefazio (commented):
:)

PR #2317, opened 2019-08-06
Allow var in try-with-resources statements

malteskoruppa:
I believe there is currently a small bug in the post-validators for Java 10.

Namely, the post-validator ensures that the keyword `var` _cannot_ occur within a try-with-resources statement. However, I believe this is incorrect: It is perfectly fine to use `var` within a try-with-resources statement.

- I tried to compile a simple `.java` file that uses `var` within a try-with-resources statement using Java 10, and it worked fine.
- My IDE (IntelliJ) does not complain when I try to use `var` within a try-with-resources statement.
- The JLS does not appear to forbid `var` within a try-with-resources statement.
   - In [JLS 14.20.3](https://docs.oracle.com/javase/specs/jls/se10/html/jls-14.html#jls-14.20.3), the given syntax rules allow to construct a try-with-resources with the `var` keyword. That possibility is even mentioned later on:
      > The type of a variable declared in a resource specification is determined as follows:
      > - If LocalVariableType is an UnannType, then (...)
      > - If LocalVariableType is var, then (...)
   - In [JLS 14.4](https://docs.oracle.com/javase/specs/jls/se10/html/jls-14.html#jls-14.4), a bunch of restrictions is given where a local variable declaration may not use `var`. A try-with-resources statement is _not_ among those restrictions.
- Try-with-resources statements with `var` exist in practice. I stumbled over this bug while parsing a file that used this. As proof, I managed to find some article where `var` is explicitly used within a try-with-resources statement: https://dzone.com/articles/finally-java-10-has-var-to-declare-local-variables

This PR fixes the post-validator for Java 10 (and up) to allow `var` within try-with-resource statements. It also fixes a test accordingly and introduces two new tests.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/stmt/TryStmtTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/stmt/TryStmtTest.java
@@ -25,21 +25,18 @@ class TryStmtTest {
     void multipleTest() {
         TryStmt tryStmt = parse9("try(Reader x = new FileReader(); Reader x = new FileReader()){}");
         assertInstanceOf(VariableDeclarationExpr.class, tryStmt.getResources().get(0));
-
     }
 
     @Test
     void modifiersTest() {
         TryStmt tryStmt = parse9("try(final @A Reader x = new FileReader()){}");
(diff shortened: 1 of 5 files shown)


matozoid:
Looks like I completely missed this. Quality PR!

PR #2336, opened 2019-08-16
Add getConstructors() methods to NodeWithMembers

malteskoruppa:
`NodeWithMembers` has several helper methods to retrieve a `NodeWithMembers`' members that are methods (`getMethods(...)`) or fields (`getFields(...)`), but none to retrieve members that are constructors, so let's add them. :-)


matozoid:
Check out `NodeWithConstructs` :-)

malteskoruppa:
Whoops. :sweat_smile: 

I had a `TypeDeclaration` and wanted its methods and its constructors, and noticed that `TypeDeclaration` was missing a `getConstructors()` method, and so I created the PR... ;-)

But, obviously, a `TypeDeclaration` could be an `AnnotationDeclaration`, which does not have constructors. So I needed to check that it's a `NodeWithConstructors` (-> `ClassOrInterfaceDeclaration` or `EnumDeclaration`) first. Understood. :-)

Thanks!

But I noticed another thing:
While the `getMethodsByParameterTypes(...)` methods in `NodeWithMembers` return _all_ methods with a given set of parameter types, the `getConstructorByParameterTypes(...)` methods in `NodeWithConstructors` return only the _lexicographically first_ constructor with a given set of parameter types.

So if we have this:
```java
class Foo {
  void foo(int i, Object o) {}
  void foo(int i) {}
  Foo(int i, Object o) {}
  Foo(int i) {}
}
```

...then `getMethodsByParameterTypes(int.class)` will return both `foo` methods, while `getConstructorByParameterTypes(int.class)` will return only the first constructor, is that right? I could add the `getConstructorsByParameterTypes(...)` that I wrote to `NodeWithConstructors` instead of `NodeWithMembers` to add that functionality, what do you think? :-)

malteskoruppa:
(btw, the tests should have passed, but it looks like there is some kind of Travis hiccup)

matozoid:
Merging NodeWithMembers and NodeWithConstructors would not be a bad thing actually - if someone really wants the constructors of an annotation declaration, then that's their own fault :-)

About getConstructorByParameterTypes: it seems correct, since many methods (with different names) can have the same parameters, but only one constructor can. Am I missing something?

malteskoruppa:
Hmm, I guess NodeWithMembers and NodeWithConstructors could be merged as far as I am concerned. :-) It's a rather subtle distinction, I think having only a NodeWithMembers would be just fine. I could take care of it.

For the `getConstructorByParameterTypes(Class...)` problem: Yes, either you or I are missing something here. Here's the thing: It appears that the way the `getMethodsByParameterTypes(Class...)` / `getConstructorByParameterTypes(Class...)` is written, the `Class...` parameter does _not_ intend to reflect the _exact_ parameter list taken by some constructor or method. Instead, what happens is that _any_ methods / constructors are matched that have, for each given type, _at least_ one parameter with a corresponding type. 

So for instance:
```java
class Foo {
  void foo(int i, Object o) {}
  void bar(int i) {}
  void baz(Object o) {}
  void qux(Object o, int i) {}
  Foo(int i, Object o) {}
  Foo(int i) {}
}
```

...when parsing this as a compilation unit, if I call `getMethodsByParameterTypes(int.class)`, this will _not_ only return method `bar`, but also methods `foo` and `qux` (but not method `baz`).

The same thing logic is used for constructors, except that only the first one among all the constructors found is returned: Therefore, calling `getConstructorByParameterTypes(int.class)` above will _only_ return the `Foo(int, Object)` constructor, but not the `Foo(int)` constructor. I find this behavior somewhat surprising.

So I see two possibilities:

1. Modify the behavior of the `get...ByParameterTypes(Class...)` method to match the given parameter types _exactly_, rather than simply checking that the given parameter types are among the parameter types taken by the methods / constructors. If I read you correctly, this is the way you understood these methods (I also originally thought that this is the way they worked until I had a closer look). If that was the behavior, I agree it would indeed make perfect sense to return only one constructor in `getConstructorByParameterTypes(Class...)`, since each constructor is uniquely identified by the parameters it takes.

2. Preserve the current (perhaps somewhat unintuitive) behavior. But in this case, there should be a ` getConstructorsByParameterTypes(Class...)` method (plural constructors!), since more than one constructor can have, for example, an int parameter among its parameters, as in the above example.

matozoid:
Okay, I didn't realize this! Check the method `getMethodsBySignature` which checks the method name, then calls `getMethodsByParameterTypes`. It is definitely supposed to check the types of the parameters in order of appearance, not just the current "more or less this signature" behaviour. So that can be opened as a bug report (or this PR can be used to fix it immediately :-) )

I'm kind of confused how this ever got this way? :-D

malteskoruppa:
Okay, I will fix it!

Btw, I was not entirely correct. It's true that the `getMethodsByParameters()` is buggy in the sense that it has a "more or less this signature" behavior rather than a "exactly this signature" behavior. But it's slightly different from what I said. The problem is that it throws all parameters of a method declaration into a _set_ rather than a list, then throws the given parameter types into a _set_ as well, and then compares the sets. :-)

So basically,

1. it disregards the order of the parameters; and
2. since it's a set, multiple identical types are in the set only once, so, e.g., `getMethodsByParameters(int.class)` would return all methods whose parameters are all ints, even those methods that have more than one int parameter, i.e., it would return methods `foo(int i)`, `foo(int i, int j)`, etc.; and vice-versa: the call  `getMethodsByParameters(int.class, int.class)` is equivalent to the call `getMethodsByParameters(int.class)`, and so on.

Anyway... I'll do this:
1. Fix the `getMethodsByParameters()` methods to match the given parameter list exactly.
2. Move the `getConstructorByParameters()` methods from `NodeWithConstructors` to `NodeWithMembers`.
3. Remove `NodeWithConstructors` (and make appropriate modifications where `NodeWithConstructors` was used, which is not a lot of locations anyway).

What do you think? :-)

matozoid:
I think that would be great :-)

malteskoruppa:
Okay, here goes.

I met a few additional problems:

1. The `hasParametersOfType(Class<?>...)` method was unable to handle generics. For instance, if you have a node with a parameter `List<String>`, then that method would always return `false` for any argument, even when you call it as `hasParametersOfType(List.class)`, because `List` is not the same thing as `List<String>`. And you can't write something like `hasParametersOfType(List<String>.class)`, so basically with this method you had no chance at all to match parameters that used generics. I fixed that by considering only the erasure of a type, as the JVM does to some degree as well ([JLS 8.4.2](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.4.2)).

2. Both `NodeWithMembers` and `NodeWithConstructors` declared a method `getConstructors()` (though only `NodeWithConstructors` additionally had the `getConstructorByParameterTypes(...)` methods). By removing `NodeWithConstructors` and putting its methods into `NodeWithMembers`, this redundancy has now been removed.

3. Moving the method `addConstructor(Modifier.Keyword...)` from `NodeWithConstructors` to `NodeWithMembers` proved to be slightly problematic, since that method creates a new `ConstructorDeclaration`, which needs to be given a name. The name is the same as the name of the enclosing node with members, but in order to get access to the `getName()` method, `NodeWithMembers` needs to extends `NodeWithSimpleName`. This was not previously the case. Now it is. I think that's fine, since any node with members is also always a node with a simple name. Btw, in the case of `NodeWithConstructors` it worked the same way: `NodeWithConstructors` used to simply extend `NodeWithSimpleName`.

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithParameters.java
```diff
@@ -133,30 +134,47 @@ default Parameter addAndGetParameter(Parameter parameter) {
...
+     * @return {@code true} if all parameters match one by one, in the given order.
      */
     default boolean hasParametersOfType(String... paramTypes) {
         return getParameters().stream()
-                .map(p -> p.getType().toString())
-                .collect(toSet())
-                .equals(Stream.of(paramTypes).collect(toSet()));
+                .map(p -> p.getType().asString())
+                .collect(toList())
+                .equals(Stream.of(paramTypes).collect(toList()));
```
matozoid: Arrays,asList prevents a stream/collect

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithParameters.java
```diff
@@ -133,30 +134,47 @@ default Parameter addAndGetParameter(Parameter parameter) {
...
+     * @return {@code true} if all parameters match one by one, in the given order.
      */
     default boolean hasParametersOfType(Class<?>... paramTypes) {
-        return getParameters().stream().map(p -> p.getType().toString())
-                .collect(toSet())
-                .equals(Stream.of(paramTypes).map(Class::getSimpleName).collect(toSet()));
+        return getParameters().stream()
+                // if p.getType() is a class or interface type, we want to consider its erasure, i.e., if the parameter
+                // is "List<String>", we want to consider it as "List", so we need to call getName()
+                .map(p -> p.getType() instanceof ClassOrInterfaceType
```
matozoid: There's a nice shorthand for that: https://javaparser.org/less-casting-in-javaparser/

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithParameters.java
```diff
@@ -133,30 +134,47 @@ default Parameter addAndGetParameter(Parameter parameter) {
...
+                        ? ((ClassOrInterfaceType) p.getType()).getName().asString()
```
matozoid: There should be getNameAsString :-)

matozoid:
Okay, a few nitpicks but overall this looks good :-)

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithParameters.java
malteskoruppa: Oh nice, I didn't know about those `to...()` methods, it's a nice trick. :-)

malteskoruppa:
Okay, fixed, thanks for the comments. :-)

matozoid:
Thanks for the PR :-)

PR #2339, opened 2019-08-20
Add TRANSIENT case to toToken() method

hecaidi:
The TRANSIENT case was missing at LexicalDifferenceCalculator toToken() method


matozoid:
Ugh. Thanks for this :-)

matozoid:
I think there are more missing, I'll open an issue.

PR #2341, opened 2019-08-20
ListAdditionChange.java, getValue() method: the returned NodeList mus…

hecaidi:
ListAdditionChange.java, getValue() method: the returned NodeList must be a child of 'node' parameter


matozoid:
Sounds sensible.

PR #2342, opened 2019-08-20
Add visitor with default actions before and after each visit() method

hecaidi:
I propose you to add a new visitor with two defaults: one before invoking the parent method and another one after.
For info, I have used it to log node contents indenting its children:

```
	public static void logNode(Node node) {
		logger.debug("** logNode({}) **", node.getClass().getSimpleName());
		AtomicReference<String> indent = new AtomicReference<String>("");

		node.accept(new VoidVisitorWithDefaultsAfter<Void>() {
			@Override
			public void defaultAction(Node node, Void arg) {
				indent.set(indent.get() + "  ");
				logger.debug("{}{} - typeInfo=[{}]: [{}]", indent, node.getClass().getSimpleName(), getTypeInfo(node), node);
			}

			@Override
			public void defaultActionAfter(Node node, Void arg) {
				indent.set(indent.get().substring(2));
			}

		}, null);
		logger.debug("** end logNode({}) **", node.getClass().getSimpleName());
	}
```


matozoid:
Hi! Have you tried using the `walk` method for this purpose?

hecaidi:
No, I hadn't tried. Thanks, much simpler.

PR #2343, opened 2019-08-21
Printer to print orphan comments before most nodes

hecaidi:
I have been having problems using the lexical preserving printer.
Instead I have finally decided to use pretty printer with PrettyPrintVisitor.java.
As I was losing a lot of comments I have made a copy of it that prints orphan comments in every visit method: **PrettyPrintVisitorOrphanComments.java** 

I am not sure why you haven't written like that PrettyPrintVisitor.java at first place and if there could be some point I am missing here. This is the reason I have not overwritten PrettyPrintVisitor.java directly.

Do you think it could be useful adding PrettyPrintVisitorOrphanComments.java to JavaParser?


matozoid:
The current way the pretty printer is set up when dealing with comments is very fragile: it may be printing comments multiple times now. I hope it doesn't, then this would be a good improvement :-) Can you replace the PPV itself in this PR so we can see what the test suite says?

(Comment problems have received their own label in the issue list: https://github.com/javaparser/javaparser/labels/Comment%20support)

matozoid:
Surprisingly little is failing!

hecaidi:
It does not longer fail, but I want to review a pair of things.
Please, do not merge by the moment.

hecaidi:
In this version of the PrettyPrintVisitor we print orphan comments for all the nodes for which we are already printing its comment.
If you agree with the merge, it's OK for me.

matozoid:
Let's do it then. Thanks!

RickDerrick:
it may be printing comments multiple times now. Can you replace the PPV itself in this PR .

<a href="https://123hp-com-setup.net/">123.hp.com/setup</a>

Regards,
Rick Derrick

PR #2345, opened 2019-08-21
Fix issue #2344

### LINKED ISSUE #2344: AbstractSymbolResolutionTest.isJavaVersion9OrAbove() fails for jdkVersion=11
```
   protected boolean isJavaVersion9OrAbove() {
        String jdkVersion = System.getProperty("java.version");
        return Integer.parseInt(jdkVersion.substring(0, jdkVersion.indexOf('.'))) >= 9;
    }
```
As there is no point when jdkVersion is 11, a StringIndexOutOfBoundsException is thrown

malteskoruppa:
(Hopefully) fixes #2344. :-)

The IndexOutOfBoundsException is a trivial fix of course, but I personally wonder why this `isJavaVersion9OrAbove()` method exists at all... it looks like the tests that rely on this method "expect" a different test outcome when the JVM version used to execute the tests is >= 9. That seems surprising. The outcome of the tests should ideally not depend on the used JVM version.

More precisely, it turns out that this method is used by some tests in order to change the path to the file which contains the "expected output" when the used JVM is >= 9: for instance, from `./javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/expected_output/java-symbol-solver-model/com_github_javaparser_symbolsolver_model_typesystem_PrimitiveType.txt` to `./javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/expected_output/java-symbol-solver-model/com_github_javaparser_symbolsolver_model_typesystem_PrimitiveType_J9.txt`.

Fortunately, this happens only in exactly four cases. However, for each of the four `*_J9.txt` files, the corresponding counterpart file without the `_J9` is identical. In other words, it looks to me like this is a relic from older times, and nowadays the tests do produce the same output, regardless of the used JVM version, as it should be. :) Thus, we can remove those `*_J9.txt` files and the `isJavaVersion9OrAbove()` method, I think, though of course @ftomassetti will know best.

At any rate, after removing this method and the `*_J9.txt` files, all tests still pass.


matozoid:
This would patch the direct problem, but does not solve the underlying issue which I've just commented on. Still, better than nothing :-)

PR #2348, opened 2019-08-24
add MethodDeclaration ParseStart & convenience methods

MDZhB:
Found this useful. No apparent way to add it without modifying the grammar.


matozoid:
You can use parseBodyDeclaration for that, or am I missing something?

But this one has a clearer name, so it is good.

PR #2355, opened 2019-09-05
fix issue #2351 Can add an enum constant with lexical preserving printer

### LINKED ISSUE #2351: Can add an enum constant with lexical preserving printer
When I run the JavaParser it removes all empty lines within a method even if I do nothing to the CompilationUnit (some of them are there for a better overview and when I just add an EnumConstant I don't want half of the lines of the file to be changed...)
Is there any possibility to just modify what was added/removed to the CompilationUnit?

btw. Thanks for this tool its really useful!

jlerbsc:
Fixes #2351 Can add an enum constant with lexical preserving printer.


matozoid:
Cool! Let's have it.

PR #2358, opened 2019-09-11
Fix the getterName generation for `Boolean`s and `boolean`s

### LINKED ISSUE #2357: CodeGenerationUtils should generate getter with `is` prefix for `boolean`s, not for `java.lang.Boolean`
According to [this](https://stackoverflow.com/a/28450498/2891426) StackOverflow answer, and the JavaBeans specification referenced from that answer, the `
CodeGenerationUtils.getterName(Class<?> type, String name)` should work as the following:

 - return `name` if `name.startsWith("is") && class.equals(Boolean.TYPE)` 
 - return `"is" + capitalize(name)` if `class.equals(Boolean.TYPE)` 
 - return `"get + capitalize(name)`

Currently it generates `is` prefixed getters for any `Boolean` type, and `get` prefixed for the `boolean` ones.

vilmosnagy:
Fixes #2357.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
@@ -15,8 +15,10 @@ class CodeGenerationUtilsTest {
     @Test
     void getters() {
         assertEquals("getValue", getterName(Object.class, "value"));
-        assertEquals("isBlue", getterName(Boolean.class, "isBlue"));
-        assertEquals("isBlue", getterName(Boolean.class, "blue"));
+        assertEquals("isBlue", getterName(Boolean.TYPE, "isBlue"));
+        assertEquals("isBlue", getterName(Boolean.TYPE, "blue"));
+        assertEquals("getBlue", getterName(Boolean.class, "blue"));
+        assertEquals("getIsBlue", getterName(Boolean.class, "isBlue"));
     }
 
 }
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
@@ -15,8 +15,8 @@ class CodeGenerationUtilsTest {
     @Test
     void getters() {
         assertEquals("getValue", getterName(Object.class, "value"));
-        assertEquals("isBlue", getterName(Boolean.TYPE, "isBlue"));
-        assertEquals("isBlue", getterName(Boolean.TYPE, "blue"));
+        assertEquals("isBlue", getterName(boolean.class, "isBlue"));
+        assertEquals("isBlue", getterName(boolean.class, "blue"));
         assertEquals("getBlue", getterName(Boolean.class, "blue"));
         assertEquals("getIsBlue", getterName(Boolean.class, "isBlue"));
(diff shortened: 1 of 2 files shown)


javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
```diff
@@ -15,8 +15,10 @@ void setters() {
     @Test
     void getters() {
         assertEquals("getValue", getterName(Object.class, "value"));
-        assertEquals("isBlue", getterName(Boolean.class, "isBlue"));
-        assertEquals("isBlue", getterName(Boolean.class, "blue"));
+        assertEquals("isBlue", getterName(Boolean.TYPE, "isBlue"));
+        assertEquals("isBlue", getterName(Boolean.TYPE, "blue"));
```
matozoid: Oh, I didn't know about this! Isn't it the same as `boolean.class` though?

vilmosnagy: Edit: oh, sorry, I haven't drunk my coffee before writing this post. Yeah, `Boolean.TYPE` is the same as `boolean.class`, but I didn't know that. I can refactor the `Boolean.TYPE`s to `boolean.class`es if you think that's more expressive.

--------
original post

No, `Boolean.class` is not the same as `Boolean.TYPE`, see this example:

```java
import java.util.Objects;

class Scratch {

    private boolean primitive;
    private Boolean boxed;

    public static void main(String[] args) throws NoSuchFieldException {
        final var primitiveField = Scratch.class.getDeclaredField("primitive");
        final var boxedField = Scratch.class.getDeclaredField("boxed");

        System.out.println("primitive field's type is Boolean.TYPE: \t" + Objects.equals(primitiveField.getType(), Boolean.TYPE));
        System.out.println("primitive field's type is Boolean.class:\t" + Objects.equals(primitiveField.getType(), Boolean.class));
        System.out.println("    boxed field's type is Boolean.TYPE: \t" + Objects.equals(boxedField.getType(), Boolean.TYPE));
        System.out.println("    boxed field's type is Boolean.class:\t" + Objects.equals(boxedField.getType(), Boolean.class));
        System.out.println("\n\nboolean.class is the same as Boolean.TYPE:\t" + Objects.equals(boolean.class, Boolean.TYPE));
    }
}
```

`Boolean.TYPE` is the `Class` representation of the primitive type `boolean`, while `Boolean.class` is the `Class` representation of the boxed one. `Boolean.TYPE` is the same as `boolean.class`, but I didn't know that `boolean.class` is a valid expression in java :slightly_smiling_face: 

I can refactor the `Boolean.TYPE`s to `boolean.class`es if you think that's more expressive.

matozoid: Yeah, I thought the subtle capitalization change might cause misunderstanding ;-) I think it's more expressive because "TYPE" vs "class" is not as direct as `boolean` vs `Boolean`... although you could argue against it because of the subtle caps. Make it as you like.

vilmosnagy: I've replaced it with `boolean.class`. I agree with you that it's more expressive, and IDEs usually highlights the primitive type `boolean` as shown below, so it's harder to confuse it with the boxed type `Boolean`.

image

PR #2359, opened 2019-09-11
Fix: Handle all modifiers in lexical preserving printer #2340

### LINKED ISSUE #2340: Handle all modifiers in lexical preserving printer
#2339 - check that all modifiers are handled.

jlerbsc:
Fixes #2340 Handle all modifiers in lexical preserving printer

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/PrettyPrinterIssue2340Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/PrettyPrinterIssue2340Test.java
@@ -2,6 +2,7 @@ package com.github.javaparser.printer.lexicalpreservation;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.util.List;
 import java.util.Optional;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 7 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -387,7 +392,7 @@ private void applyRemovedDiffElement(RemovedGroup removedGroup, Removed removed,
         } else if (originalElementIsToken && originalElement.isWhiteSpaceOrComment()) {
             originalIndex++;
         } else if (removed.isPrimitiveType()) {
-            if (isPrimitiveType(originalElement)) {
+            if (originalElement.isPrimitive()) {
```
matozoid, line 395: Nice!

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/PrettyPrinterIssue2340Test.java
```diff
@@ -0,0 +1,45 @@
...
+import com.github.javaparser.ast.body.EnumDeclaration;
+
+class PrettyPrinterIssue2340Test extends AbstractLexicalPreservingTest {
+
+    @Test
+    void printingVariableDeclarationWithAddedModifier() {
+        String def2 = "List i";
+        considerVariableDeclaration(def2);
+        expression.asVariableDeclarationExpr().getModifiers().addFirst(Modifier.privateModifier());
+//        System.out.println(LexicalPreservingPrinter.print(expression));
```
matozoid: Do you want to keep the println's in here?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -495,6 +515,39 @@ private void applyKeptDiffElement(Kept kept, TextElement originalElement, boolea
...
+
+    /*
+     * Returns the number of tokens to skip in originalElements list to synchronize with the DiffElements list
+     * This is due to the fact that types are considered as token in the originalElements list.
+     * For example,
+     * List<String> is represented by 4 tokens ([List][<][String][>]) while it's a CsmChild element in the DiffElements list
+     * So in this case, getIndexToNextTokenElement(..) returns 3 because we have to skip 3 tokens ([<][String][>]) to synchronize
+     * DiffElements list and originalElements list 
+     * The end of recursivity is reached when there is no next token or if the next token is a whitespace,EOL or tab, to take into account this type of declaration
+     * List<List<String>> l
```
matozoid: This doesn't take this one into account, I think: `List      <String> a;`. It is an overall improvement though (although a solid algorithm is preferred)

jlerbsc: You are right!

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/PrettyPrinterIssue2340Test.java
jlerbsc: No it's useless.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
matozoid: Does that mean you're going to improve it, or that you consider this a good enough improvement? ;-)

jlerbsc: I have pushed an improvment which take into account your remark.

matozoid:
Thanks!

PR #2372, opened 2019-09-19
Fix indent of added comments

Implex1v:
After adding a comment to an existing node of an AST, the node after the comment loses it's indent when printing it with the `LexicalPreservingPrinter`.
Given this example
```java
public class Foo {
    void mymethod() {
    }
}
```
the resulting string would look like this:
```java
public class Foo {
    //line
void mymethod() {
}
}
```
I fixed the LPP so the result looks like this
```java
public class Foo {
    //line
    void mymethod() {
    }
}
```

I've added the indents to the LPPTest. Should I add another test or is it fine?

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
@@ -1124,8 +1124,8 @@ class LexicalPreservingPrinterTest extends AbstractLexicalPreservingTest {
                 .setBlockComment("block");
         assertEqualsNoEol("public class Foo {" + EOL +
                           "    /*block*/" + EOL +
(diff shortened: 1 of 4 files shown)


matozoid:
Hey, that's awesome!

I'll bounce the test question back: do you think it's well tested this way?

Implex1v:
It's never well tested :p
BBT: I'll add a test for a block comment and maybe another test.

Implex1v:
It's never well tested ^^
BTT: I might add a test for the test file used for IntelliJ IDEA/Eclipse and check, if the indentions are as expected.

matozoid:
Alright, Sunday I'll merge this in and we'll see if something has happened by then :-)

Implex1v:
Yeah I found one problem with my new test and I'm going to fix it. I'll tell you, when I am done :)

Implex1v:
I've fixed the metioned bug and added a new test. For me it looks fine and mergeable.

matozoid:
Thanks!

PR #2378, opened 2019-09-23
Fix and test for issue #2367, resolving NameExpr when it represents a private field name

### LINKED ISSUE #2367: Problems resolving a NameExpr when it represents a private field name
Hello,

I am having problems with 
`((NameExpr) node).resolve().getType().describe()`
when the node represents the name of a private field.

An exception is thrown at the `resolve()` call. I think the problem comes from class **com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext**, at the method:
```
@Override
    public SymbolReference<? extends ResolvedValueDeclaration> solveSymbol(String name) {
        if (typeSolver == null) throw new IllegalArgumentException();

        if (this.getDeclaration().hasVisibleField(name)) {
            return SymbolReference.solved(this.getDeclaration().getVisibleField(name));
        }

        // then to parent
        return getParent().solveSymbol(name);
    }
```
I don't understand why it is only searching for visible fields, being the **ClassOrInterfaceDeclaration** a parent of my **NameExpr** node.

Should I do this in other way?

SimonBaars:
Fixes #2367. When solving a symbol as value, private members should also be considered.


matozoid:
Cool!

PR #2380, opened 2019-09-24
Fix indent of inserted code blocks

Implex1v:
Concidering the following example with LPP setup
```java
public class Example1 {
    public void foo() {
        System.out.println("Hello World!");
    }
}
```
adding code blocks and if statments results in
```java
public class Example1 {
    public void foo() {
        System.out.println("Hello World!");
        if (name.equals("foo")) {
            int i = 0;
        System.out.println(i);
        if (i < 0) {
            i = 0;
    }
        new Object() {
        };
    } else {
    }
    }
}
```
but I would expect
```java
public class Example1 {
    public void foo() {
        System.out.println("Hello World!");
        if (name.equals("foo")) {
            int i = 0;
            System.out.println(i);
            if (i < 0) {
                i = 0;
            }
            new Object() {
            };
        } else {
        }
    }
}
```
I've fixed this problem by applying the indention modifications of `CsmIndent` and `CsmUnindent` to the current indention modifier. There is also a test including the example above.


matozoid:
It does not seem to break any tests, so great!

PR #2381, opened 2019-09-24
Fix issue 2374 Comments of added Nodes are ignored in LexicalPreservi…

### LINKED ISSUE #2374: Comments of added Nodes are ignored in LexicalPreservingPrinter
Given this example code
```java
public class Bar {
    public void foo() {
        System.out.print("Hello");
    }
}
```
```java
```
```java
public class Bar {
    public void foo() {
System.out.print("Hello!");
System.out.println("World!");
    }
}
```

jlerbsc:
Fixes #2374. I'm not sure that's the way to go

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
@@ -319,17 +319,4 @@ public class MethodCallExpr extends Expression implements NodeWithTypeArguments<
     public Optional<MethodCallExpr> toMethodCallExpr() {
         return Optional.of(this);
     }
-    
-    @Override
-    /**
-     * Returns the comment or an Optional.empty if there is no comment on this method call.
-     * In case of MethodCallExpr, Comment are setted in the parent ExpressionStmt comment attribute. 
-     */
-    public Optional<Comment> getComment() {
-        Optional<Comment> comment = super.getComment();
-        if (!comment.isPresent() && getParentNode().isPresent()) {
-            comment = getParentNode().get().getComment();
-        }
-        return comment;
-    }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
@@ -323,7 +323,7 @@ public class MethodCallExpr extends Expression implements NodeWithTypeArguments<
     @Override
     /**
      * Returns the comment or an Optional.empty if there is no comment on this method call.
-     * In case of MethodCallExpr, Comment are setted in an the parent ExpressionStmt comment attribute. 
+     * In case of MethodCallExpr, Comment are setted in the parent ExpressionStmt comment attribute. 
      */
     public Optional<Comment> getComment() {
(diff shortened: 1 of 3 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -521,7 +522,10 @@ private static NodeText interpret(Node node, CsmElement csm, NodeText nodeText)
 
             pendingIndentation = false;
             if (element instanceof LexicalDifferenceCalculator.CsmChild) {
-                nodeText.addChild(((LexicalDifferenceCalculator.CsmChild) element).getChild());
+                Node child = ((LexicalDifferenceCalculator.CsmChild) element).getChild();
+                if (child.getComment().isPresent())
+                    nodeText.addToken(getTokenKind(child.getComment().get()), child.getComment().get().toString());
```
matozoid: Optionals are intended to be used with functional style. `ifPresent` would work better here.

matozoid:
Can you add a little test to demonstrate that it works?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
```diff
@@ -318,4 +319,17 @@ public ResolvedMethodDeclaration resolve() {
     public Optional<MethodCallExpr> toMethodCallExpr() {
         return Optional.of(this);
     }
+    
+    @Override
+    /**
+     * Returns the comment or an Optional.empty if there is no comment on this method call.
+     * In case of MethodCallExpr, Comment are setted in an the parent ExpressionStmt comment attribute. 
+     */
+    public Optional<Comment> getComment() {
```
matozoid: This is not logical: a comment is always set on the node where you want it. This problem should not be solved by making a single node behave inconsistently.

jlerbsc: Ok it's probably caused by 

    stmt.setLineComment(lineComment);

matozoid: Ah, you mean that for something like a `MethodCallExpr`, which can end up in a `ExpressionStmt`, we have to consider two nodes when deciding whether it has a comment? So we might need to be able to query whether a node is a "pseudonode" which doesn't have any source code text of its own? I think the lex.pres.printer has some "ghost node" logic (or so) for that. But @ftomassetti can explain if he has some time :-)

matozoid: @jlerbsc - I still don't really get what you want to happen, but I guess this PR can be merged.

jlerbsc: @Matozoid, i'm completely wrong and this PR should not be merged. In the issue declared by @Implex1v, he add a comment to a statement (not a MethodCallExpr) so i don't really fix the issue.

matozoid: Well, it's merged already and it did fix something, I think? Otherwise I (or you) will make a revert PR.

PR #2389, opened 2019-10-03
Fix issue #2374 Comments of added Nodes are ignored in LexicalPreserv…

### LINKED ISSUE #2374: Comments of added Nodes are ignored in LexicalPreservingPrinter

jlerbsc:
…ingPrinter

Fixes #2374 .


matozoid:
@jlerbsc this PR is not doing so well yet :-/

PR #2390, opened 2019-10-04
Fix issue #2374 Comments of added Nodes are ignored in LexicalPreserv…

### LINKED ISSUE #2374: Comments of added Nodes are ignored in LexicalPreservingPrinter

jlerbsc:
…ingPrinter

Fixes #2374 .


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalDifferenceCalculator.java
```diff
@@ -146,6 +149,19 @@ private void calculatedSyntaxModelForNode(CsmElement csm, Node node, List<CsmEle
...
+                // fix issue #2374
+                // Add node comment if needed (it's not very elegant but it works)
+                // We need to be sure that the node is an ExpressionStmt because we can meet
+                // this class definition
+                // a line comment <This is my class, with my comment> followed by
+                // class A {}
+                // In this case keyworld [class] is considered as a token and [A] is a child element
+                // So if we don't care that the node is an ExpressionStmt we could try to generate a wrong definition
+                // like this [class // This is my class, with my comment A {}]
+                if (node.getComment().isPresent() && node instanceof ExpressionStmt) {
```
matozoid, line 161: What is the defining characteristic of nodes like ExpressionStmt that make them need behaviour like this? Maybe we can make that a property of the node, so we can avoid node specific exceptional logic here.

matozoid: @jlerbsc ?

jlerbsc: I think it's more specific to lexical printer than expression statement.

matozoid: Okay, but do you think there are other nodes likes this? If not, it's okay.

jlerbsc: Sincerely, I don't know! I don't know enough JavaParser to answer this question. Maybe the author could answer it.

matozoid: Let's just merge it then :-)

Issue #2227, opened 2019-05-14
Make a generator for the packages in bnd.bnd


matozoid:
These are often out of date for a long time, making JP only partially usable from OSGi.

alessiostalla:
@matozoid I would like to lend a hand on this. I've already looked at generators. Were you thinking about a generator that writes the actual bnd.bnd file? Or would printing to stdout suffice, like the GrammarLetterGenerator does?

matozoid:
Hi! Thanks for the offer. I think it would be best to do it automatically because it gets forgotten all the time. It shouldn't be too hard to put another one in here: https://github.com/javaparser/javaparser/blob/master/javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/CoreGenerator.java

GrammarLetterGenerator would have been nice to automate too, but it has to output a very specific format right in the middle of the grammar - so printing to stdout is the lazy implementation (and it is unlikely that anybody will notice if it doesn't get run for a year or two)

alessiostalla:
@matozoid thanks for the info. I will sketch an implementation and open a PR.

MysterAitch:
#2394 seems to fix this -- can close this issue?

PR #2394, opened 2019-10-08
#2227 Make a generator for the packages in bnd.bnd

alessiostalla:
Generated file inside the archive (GitHub wouldn't let me attach it directly): [bnd.bnd.zip](https://github.com/javaparser/javaparser/files/3704342/bnd.bnd.zip)


matozoid:
Thanks! The output seems to look just fine :-) This can be merged. Would you like some comments on the code?

alessiostalla:
If you have comments, please share :)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+
+    public BndGenerator(SourceRoot sourceRoot) {
+        super(sourceRoot);
+    }
+
+    @Override
+    public void generate() throws IOException {
+        Log.info("Running %s", () -> getClass().getSimpleName());
+        File root = sourceRoot.getRoot().toFile();
+        visit(root, "");
```
matozoid: SourceRoot was built for common file traversal tasks - since you're avoiding it, can you say what it is lacking?

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+        List<String> sortedPackages = new ArrayList<>(packages);
+        sortedPackages.sort(String::compareTo);
```
matozoid: I think the list of packages and its transformations can be changed into a big streaming operation. It would be fun to see how that looks.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+        File projectRoot = root.getParentFile().getParentFile().getParentFile();
```
matozoid: `Path` more or less replaces `File`, so there is an attempt to use it as much as possible in JP. It's a decent API once you get used to it.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+        File template = new File(projectRoot, "bnd.bnd.template");
+        File output = new File(projectRoot, "bnd.bnd");
+        String lineSeparator = System.getProperty("line.separator");
+        StringBuilder packagesList = new StringBuilder("\\" + lineSeparator);
+        boolean first = true;
+        for(String pkg : sortedPackages) {
+            if(!first) {
+                packagesList.append(", \\").append(lineSeparator);
```
matozoid: This kind of stuff is nowadays solved with String.join and Collectors.joining.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+            packagesList.append("    ").append(pkg);
+            first = false;
+        }
+        try(BufferedReader reader = new BufferedReader(new FileReader(template))) {
+            try(PrintWriter writer = new PrintWriter(new FileWriter(output))) {
+                reader.lines().forEach(l -> writer.println(l.replace("{exportedPackages}", packagesList)));
+            }
```
matozoid: You could read the whole template into a single string (https://howtodoinjava.com/java/io/java-read-file-to-string-examples/) and do a replace on it :-) There's probably a write variant of it too, I don't know.

javaparser-core/.gitignore
```diff
@@ -0,0 +1,2 @@
+#The bnd.bnd file is generated by the run_core_generators.sh script (specifically, by the BndGenerator).
+bnd.bnd
```
matozoid: I was a little scared of this because the generators are run by hand. It doesn't happen for every release. So bnd.bnd might be missing for a few releases until I remember to run the generators (but I guess the file will stay around then.)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
alessiostalla: I didn't use it because it doesn't contain all the packages and files, probably because it only contains those files that are explicitly parsed by other generators. It doesn't offer a method to enumerate all the source files, including unparsed ones.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
alessiostalla: Check my new commit :)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
alessiostalla: I didn't know about the `Files` utility class. With that, `Path` is much more usable, so I've rewritten everything with it.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
alessiostalla: This has been superseded by the use of the stream API.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
alessiostalla: I think efficiency-wise it's the same but it uses less memory, not that it matters much with such small amounts of data.

javaparser-core/.gitignore
alessiostalla: You're right, I've reverted it and added a comment.

alessiostalla (commented):
I've added a commit that addresses your comments.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
matozoid: Hmmm, maybe the generators should start by running SourceRoot over the whole project then... Thanks, that's for another time :-)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
matozoid: It looks really fun :-)

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/BndGenerator.java
matozoid: It's more that it's nice to get rid of some of those streams.

alessiostalla: Makes sense. I have updated the PR. Thanks for your comments, very valuable!

matozoid:
Wonderful! Thank you very much.

PR #2395, opened 2019-10-09
Allow $ in class-names, use getCanonicalName() instead

### LINKED ISSUE #2391: CompilationUnit#addImport produces duplicate imports
<artifactId>javaparser-core</artifactId>
<version>3.14.159265359</version>

```
public CompilationUnit addImport(ImportDeclaration importDeclaration) {
        if (getImports().stream().noneMatch(im -> im.toString().equals(importDeclaration.toString()))) {
            getImports().add(importDeclaration);
        }
        return this;
    }
```
should be
```
public CompilationUnit addImport(ImportDeclaration importDeclaration) {
        if (getImports().stream().noneMatch(im -> im.getNameAsString().equals(importDeclaration.getNameAsString()))) {
            getImports().add(importDeclaration);
        }
        return this;
    }
```
otherwise comments and other nodes affect the result. 
i.e.
```
package a.b.c;

/*
a package comment
*/

import x.y.z.Type;
import x.y.z.Type;
```

Maybe we should improve this method to ignore implicit imported packages like "java.lang" ?
```
public CompilationUnit addImport(ImportDeclaration importDeclaration) {
        if (!importDeclaration.getNameAsString().startsWith("java.lang") && getImports().stream().noneMatch(im -> im.getNameAsString().equals(importDeclaration.getNameAsString()))) {
            getImports().add(importDeclaration);
        }
        return this;
    }
```

gitmotte:
Fixes #2391. 
* allow $ in class-names, use getCanonicalName() instead of replacing $ of getName() for inner classes
* ignore null-names (getCanonicalName could return null)

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/builders/CompilationUnitBuildersTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/builders/CompilationUnitBuildersTest.java
@@ -34,217 +34,243 @@ import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 
+import org.junit.jupiter.api.Test;
+
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.PackageDeclaration;
 import com.github.javaparser.ast.body.AnnotationDeclaration;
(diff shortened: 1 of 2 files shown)


matozoid:
Thanks again :-)

PR #2398, opened 2019-10-24
bugfix: when parsing chaining method invoked,NotFoundException occured

leoChaoGlut:
At the beginning, i have load all the necessary Jars into JavaParserFacade.

When parsing "ClassA.getClassB().methodB()", NotFoundException occured.

"javassist.CtMethod#getReturnType" using the wrong "javassist.ClassPool" to get "ClassB", so NotFoundException occured.

The solution is: when "ctMethod.getReturnType()" throws NotFoundException, using typeSolver to solve the return type.


edefazio:
Thanks for submitting the PR.

I am guessing English is not your first language, and I am looking over what you are doing (in the code) and I believe I understand it, but I wanted to see if I could add more detail in the comments to the code you submitted.

for starters; do you have a unit test that might help people understand this problem?  

That would go a long way in helping the team understand how to classify what this fixes... 
From your description it seems like you have a class
```java
//a class to be resolved
package aaaa.bbbb;

class TopLevel{
    class Nested{
         returnType[][] methodToBeResolved(){} 
    }
}
```
then you are using the `JavaParserFacade` to try and resolve the type of `returnType` on `methodToBeResolved` which in the above case SHOULD BE  the type:
`returnType[][]`

... and there are some issues ... specifically with the `ctMethod.getReturnType()' method call which could (in some cases) throw a `NotFoundException` which propagates back to the caller

to avoid this, you manually try/catch the `NotFoundException`

first question... does the `NotFoundException` happen
1) for any returnType? 
2) only for returnTypes that ARE NOT generic?

...in general what causes this Exception to be triggered?

other than that, I can see you grabbed some code from javassist to manually parse the String returned from `ctMethod.getMethodInfo().getDescriptor()` to "manually resolve" the Type using the TypeSolver.

Does that do a decent job describing it? 

I'm just trying to ensure we communicate precisely the problem, when it occurs, and how your code goes about fixing it (the code looks fine, I'd just like the record, and a few comments describe the scenario that this code solves)

Thanks,
Eric

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistMethodDeclaration.java
```diff
@@ -198,4 +210,32 @@ public ResolvedType getSpecifiedException(int index) {
...
+        char c = desc.charAt(i);
+        int arrayDim = 0;
+        while (c == '[') {
+            ++arrayDim;
+            c = desc.charAt(++i);
+        }
+        if (c == 'L') {
+            i2 = desc.indexOf(';', ++i);
+            return desc.substring(i, i2++).replace('/', '.');
+        }
```
edefazio: @matozoid pointed out that, while we are collecting the arrayDim, it isn't being used.

Is this the correct behavior?
... or should the []'s be appended to the end of line 236
`return desc.substring(i, i2++).replace('/', '.') //(add the array dimensions here) ;`

edefazio: Here is the part from the original (javassist code) that is missing
```java
 if (arrayDim > 0) {
    StringBuffer sbuf = new StringBuffer(name);
    while (arrayDim-- > 0)
    sbuf.append("[]");
    name = sbuf.toString();
}
```
source: [http://www.docjar.com/html/api/javassist/bytecode/Descriptor.java.html](http://www.docjar.com/html/api/javassist/bytecode/Descriptor.java.html)

leoChaoGlut: This is a good comment. I miss this scenario.

leoChaoGlut:
> Thanks for submitting the PR.
> 
> I am guessing English is not your first language, and I am looking over what you are doing (in the code) and I believe I understand it, but I wanted to see if I could add more detail in the comments to the code you submitted.
> 
> for starters; do you have a unit test that might help people understand this problem?
> 
> That would go a long way in helping the team understand how to classify what this fixes...
> From your description it seems like you have a class
> 
> ```java
> //a class to be resolved
> package aaaa.bbbb;
> 
> class TopLevel{
>     class Nested{
>          returnType[][] methodToBeResolved(){} 
>     }
> }
> ```
> 
> then you are using the `JavaParserFacade` to try and resolve the type of `returnType` on `methodToBeResolved` which in the above case SHOULD BE the type:
> `returnType[][]`
> 
> ... and there are some issues ... specifically with the `ctMethod.getReturnType()' method call which could (in some cases) throw a `NotFoundException` which propagates back to the caller
> 
> to avoid this, you manually try/catch the `NotFoundException`
> 
> first question... does the `NotFoundException` happen
> 
> 1. for any returnType?
> 2. only for returnTypes that ARE NOT generic?
> 
> ...in general what causes this Exception to be triggered?
> 
> other than that, I can see you grabbed some code from javassist to manually parse the String returned from `ctMethod.getMethodInfo().getDescriptor()` to "manually resolve" the Type using the TypeSolver.
> 
> Does that do a decent job describing it?
> 
> I'm just trying to ensure we communicate precisely the problem, when it occurs, and how your code goes about fixing it (the code looks fine, I'd just like the record, and a few comments describe the scenario that this code solves)
> 
> Thanks,
> Eric

Unit Test is already done.

edefazio (approved):
Thanks for the PR

edefazio (approved):
Thanks for the PR

matozoid:
@leoChaoGlut  @edefazio - what is the state of this PR?

edefazio:
I reviewed the changes, and they looked fine at the time, I wanted someone who uses the JSS to review them in case I missed something and I dont think anyone did.

(Im not really set up locally to test the JSS so I *think* its ok to merge but hey, I'm not giving it my normal high confidence rating since its been a bit)

matozoid:
Okay, let's go do some live user testing then :-)

matozoid:
Thanks @leoChaoGlut!

PR #2403, opened 2019-11-02
removing unused dependencies

cesarsotovalero:
Hello, I noticed that javassist and guava are declared in the pom of module javaparser-symbol-solver-logic. However, they are not used in this module.


matozoid:
Thanks! It's all green so here it goes.

PR #2407, opened 2019-11-06
Fix: Remove "new ParserConfiguration()" in parallelized parse methods and add encoding parameter to provide()

pac0rro:
On "parseParallelized" methods I've removed instantiation of ParserConfiguration in order to make use of internal custom ParserConfiguration instance.

Also, encoding was provided default to UTF8 because the JavaParser was invoked with the provider() method versions without the "encoding" parameter

(No related bugs open)


pac0rro:
Um, the CI's build has finished successfully but it seems the status is not verified. Probably a bug in travis-ci system?

matozoid:
Thanks! This makes sense. I see Travis is really green, so let's merge this.

PR #2419, opened 2019-11-16
Issue 2416 - if the range begin/end are backwards, swap them

### LINKED ISSUE #2416: Implementation of `Range` presumes an order of start/end
The current implementation of `Range` blindly accepts the given parameters as being in order. 

While this is not necessarily problematic, this assumption is inconsistent with the implementation in the rest of the class. Examples include the implementations `isBefore` which presume the end of the current range is the "later" position in the document.

Additionally, the line count assumes a specific ordering and will return a negative number if the begin/end positions are reversed.

I'll submit a PR for this shortly :)

MysterAitch:
Fixes #2416.


MysterAitch:
Just noticed that I accidentally included javadoc polishing/additions into this commit when I'd intended to submit that separately. 

Happy to split it out if needed/requested, but for now I'll leave it as-is.

MysterAitch:
Please note comments at https://github.com/javaparser/javaparser/issues/2416#issuecomment-554685477 before accepting.

matozoid:
I've read the whole thing and, seeing that there is a problem (people writing wrappers hints that JP's API is not good enough) this work is very welcome.

I've considered complaining that allowing the construction of a range with positions reversed might hide user bugs, but there were some convincing arguments to just swap them instead.

PR #2420, opened 2019-11-17
Issue 2418 - have range overlap use contains, not strictContains

### LINKED ISSUE #2418: Range overlap should use contains, not strictContains
Range overlap should use `contains`, not `strictContains`. A PR shall be coming shortly.

Rationale: Given that `Position` represents a specific character as opposed to the boundary between characters, both range 1 and range 2 in the example below contain the letter `C`, thus should be flagged as overlapping each other.

```
Text: ABCDE

Range 1: ABC
Range 2:   CDE
```

https://github.com/javaparser/javaparser/blob/9374f39045f55e1bdb706bb5a923b4e172c340dd/javaparser-core/src/main/java/com/github/javaparser/Range.java#L109-L111

MysterAitch:
Fixes #2418.


matozoid:
Have you looked through the code to see where it is used, and if it could change behaviour? (Hopefully for the better ;-) )

This also seems one to mention in the changelog since it is sneaky API breakage.

matozoid:
Oops, merging things has introduced conflicts :-( Can you solve them?

MysterAitch:
> Have you looked through the code to see where it is used, and if it could change behaviour? (Hopefully for the better ;-) )
> 
> This also seems one to mention in the changelog since it is sneaky API breakage.

Perhaps naively, I've assumed that tests passing = no issues :grinning: 

I'll resolve the conflicts and scan the repo later this evening.

matozoid:
I've branched this so I could fix the conflict. PR is here: #2451

PR #2421, opened 2019-11-17
Refactor range tests

MysterAitch:
Refactoring of the unit tests for class `Range`.

Note that some failing tests depend on the bugfixes contained within #2419 / #2420 .

- use named/labelled range pairs, as opposed to line/column permutations being defined/recreated on a per-test basis; 

- all test cases are renamed to a consistent and descriptive format; 

- added a few test cases e.g. for touching but not overlapping ranges;


matozoid:
Copied this to #2452 so I could fix the conflicts.

Issue #2044, opened 2019-01-27
T#<method declared in java.lang.Object>, where T is a Type Parameter, cannot be resolved.


MarioIshac:
Let's look at this sample method I have:

```
private int getPartition(final K key) {
    return key.hashCode() / getHashes().length;
}
```

Based on what I've seen so far, the symbol solver cannot find a `ResolvedMethodDeclaration` from the `MethodCallExpr` `key.hashCode()` because `key` is of type `K` a type parameter. However, `key.hashCode()` is a valid `MethodCallExpr`, because K implicitly extends `Object`. 

I may be way off here, and I may have not set up my resolver correctly (though I am using `ReflectionTypeSolver`). But is this what is happening, and if so, does this sound like a bug or intended behavior?  I can get around it either way.

matozoid:
@ftomassetti and friends?

MysterAitch:
How are you doing getting the resolved type?

Something similar came up as a problem for me recently where `.resolve().getResolvedType()` (or similar - I don't have the code in front of me) would seemingly return the declared generic type, while `.calculateResolvedType()` would return the instantiated type. This came up when trying to resolve entrysets and entries in a set or list for example.

NB: This comes with the disclaimer that I've not fully investigated/confirmed the difference, but it might be related to the above.

MarioIshac:
> How are you doing getting the resolved type?
> 
> Something similar came up as a problem for me recently where `.resolve().getResolvedType()` (or similar - I don't have the code in front of me) would seemingly return the declared generic type, while `.calculateResolvedType()` would return the instantiated type. This came up when trying to resolve entrysets and entries in a set or list for example.
> 
> NB: This comes with the disclaimer that I've not fully investigated/confirmed the difference, but it might be related to the above.

I'm still using the old API, so I get the `ResolvedMethodDeclaration` through `JavaParserFascade#solveMethodAsUsage(methodCallExpr).getDeclaration();` I haven't discovered a problem with resolving the type yet, `getJavaParserFacade().getType(methodCallExpr.getScope().get());` works completely fine (assuming that the `Optional` contains a value.

Additional Info: This is the stacktrace:
```
Caused by: java.lang.RuntimeException: Method 'hashCode' cannot be resolved in context key.hashCode() (line: 40) MethodCallExprContext{wrapped=key.hashCode()}. Parameter types: []
[...]

PR #2422, opened 2019-11-17
[TESTCASE] Issue #2044 - type parameters inheriting methods from `Object`

MysterAitch:
Testcase for #2044 as described at https://github.com/javaparser/javaparser/issues/2044#issuecomment-457981251


MysterAitch:
Fixed in #2428

Issue #2148, opened 2019-04-02
Zenodo / DOIs


MysterAitch:
I'm trying to figure out how to reference/cite JavaParser in a research paper and I've just come across [Zenodo](https://github.com/zenodo/zenodo). It appears to enable releases of software to be assigned a "Digital Object Identifier" (DOI). Zenodo is run by CERN and is referenced from the [GitHub docs](https://guides.github.com/activities/citable-code/) in its article about citable code.

Linking Zenodo to GitHub repos [appears straightforward](https://guides.github.com/activities/citable-code/) and seems to require minimal effort/setup - set it up once and the rest happens automatically behind the scenes. I can't vouch for this first-hand though!

From what I can see [in the FAQs](http://help.zenodo.org/#versioning), one DOI is assigned to the project (resolves to the latest version) and each release is then assigned its own DOI (minting being triggered via GitHub release webhook).

Is this something that would be considered?

matozoid:
Hi! We've had this question before (how to reference JP) so it's nice to see that it can be resolved easily. I've made an account and switched on preservation for the main project: https://github.com/javaparser/javaparser . Let's see what happens next Sunday when I release :-)

matozoid:
Nothing happened :-( I've googled for issues but haven't found anything I could act on directly. Would you like admin access to the project?

MysterAitch:
It would be rude to say not to a generous offer like that! ;)

I'm not sure how much farther I'll get with it but I'm happy to have a poke around and see if I can narrow things down at least?

matozoid:
Okay, I've added you as an admin. Let me know if that worked :-)

Sure, go ahead. Just don't break what's working :-)

MysterAitch:
Apologies for the delay. I've had a look around but I don't seem to have any additional access to the project (e.g. to look at the commit hooks) and the project doesn't appear when I login to Zenodo. 

Assuming that you gave organisation access etc, I'm wondering whether the first upload needs to be done manually and then subsequent ones will be automatic, but that's just a hunch. Alternatively it might be the commit webhooks.

matozoid:
Ah, github says "Awaiting MysterAitch’s response" - so I assume there's an email or so for you.

I did google this, and there are some reports that automatic releases just don't get counted. I don't give a damn about JavaParser's github release list (this one: https://github.com/javaparser/javaparser/releases ) so feel free to mess with it.

MysterAitch:
[...]

PR #2423, opened 2019-11-17
Add a "github action" to mark version tags as releases

MysterAitch:
Adds a "github action" to trigger a github release (as opposed to tag) every time a tag is pushed with the specified pattern (currently `javaparser-parent-[0-9].*` as this matches the tags added by the maven release plugin).

The benefit here is that it should then trigger the release webhooks, thus triggering actions such as those in #2148 to recognise the pushed tags as a new release.


MysterAitch:
I've been experimenting with this at https://github.com/MysterAitch/test-release-actions (feel free to have a play also!)

It looks like all tags that get pushed (e.g. if pushing three or four tags in one push) are processed, and it seems to treat annotated tags (to be generally avoided) and unannotated tags (preferred) equally.

Note `git push --follow-tags` is preferred (it'll only push annotated tags) over `git push --tags` (pushes all tags)

matozoid:
This looks alright, but I can't see from what you're saying if this is ready to merge? When do tags get annotated?

MysterAitch:
Apologies for the lack of clarity!!

The annotated tag comments were more a general "think aloud" about manually-added tags.

In JavaParser's case, the tags added and pushed (I believe) by the maven release plugin are sufficient to trigger this action.

matozoid:
Here goes then! Thanks :-)

Issue #2032, opened 2019-01-21
Add `asString` to class `Node`


MysterAitch:
I'm currently investigating the best way to print the content of an arbitrary node, and there doesn't seem to be a reliable means of (programmatically) determining whether `toString` or `asString` should be called.

A single interface that can be checked against using `instanceof` or similar would make this significantly more straightforward, as would implementing `asString` within the class `Node` (or similar?). 

How might this be done best? 

1. A simple default implementation is to delegate to `toString` and have subclasses override it as required.
    - This would provide exactly what is currently available, but it doesn't meet the definition of `asString` (i.e. [the node minus comments and annotations](http://javaparser.org/making-strings-in-javaparser/)).   
1. A more robust implementation would be to clone the node, strip it of comments and annotations, and then call `toString` on the clone.
    - This doesn't "feel" right though and I can't quite put my finger on why. 
    - Perhaps it's due to not having a single location to strip out both annotations and comments? (`Node` has comments, but a range of other classes have annotations).

matozoid:
Good thing you've made this an issue first. `asString` is the opposite of a print function: it gives you something you can use programmatically, as a key in a hashmap, or to compare to something in a configuration file. Now that I'm writing this I'm thinking the name might be all wrong.

The best way to print the content of an arbitrary node is to use the pretty printer, or the lexical preserving printer, or a custom pretty printer. This sounds like you want a copy of the pretty print visitor and delete the annotations and comment printing in there. Or yeah, your point two, it's fine what you're doing there :-)

MysterAitch:
I'm not sure I quite follow what you're saying and I'm getting a little muddled. It's likely I'm not familiar enough to understand the use case for it (and my motivation for using `asString` is a rather vague feeling based on "`asString` is probably better" rather than anything more concrete) so shall ponder on a little longer :) 

Fair point re: use of printers - pretty sure you're spot on there and I'll dig a little further along those lines for my immediate needs. Somehow the connection with 'toString` using printers was completely missed!

matozoid:
Do keep asking, it educational to see what happens when people use JP!

- Lexical Preserving Printer: tries to print with indentation and whitespace as much as the parsed original.
- Pretty Printer: happily forces its formatting rules on everything.
[...]

PR #2424, opened 2019-11-17
[DISCUSSION] Issue 2032 - add string output interface

MysterAitch:
Ref: #2032 

My understanding is that the `asString` method is to be used to give a representation of a declaration/definition that remains stable following changes to meta-data such as comments and annotations.

Regardless, for code that handles arbitrary nodes, it is not currently possible to determine which nodes has an `asString` representation other than searching the codebase and maintaining a local list (which may then go out of date thus must be reviewed each release).

For that reason I am strongly in favour of introducing an interface to describe this behaviour/contract. 

A nice and simple approach _(the one taken in this PR)_ is to attach such an interface to `Node` and delegate to `toString` by default. This doesn't break anything or change existing behaviour, and avoids having to make hard decisions.

An acceptable but slightly harder-work alternative would be to have the interface applied to specific nodes/node typess (this would allow `instanceof` checks to verify if it has an `asString` method), but in this case I am a little less clear _where_ to place it, and what the criteria would be for its use/inclusion. 

---- 

Long-term, my preference would be to deprecate all usage of `toString` in favour of:

- `toPrettyString` (or `toPrettyPrintString`) - this applies the pretty printer to the node (overloaded with an optional config as `toString` currently does), and then have `toString` delegate to `toPrettyString` to maintain backwards compatability.

- `toNormalisedString` - this is what `asString` is currently trying to be, but in need of a better name to describe a node which is "stable" in that it is unaffected by changes to meta-data (such as comments and annotations and position in file). Also worth explicitly noting in the interface contract that this is not long-term stable thus is subject to change and human change over time. 

Perhaps something that could be added to the definition of `asString` is that it refers to the "definition" of something -- e.g in the case of a class/method, it's the class/method header. Similarly, in the case of a loop, it's the loop "header" (i.e. omitting the body), with two string representations of "loop from 1 to 10" being roughly equivalent even if the body of the loop differs. In the case of variables we would have a declaration/assignment/combined representation. Not sure about the rest!

I did consider suggesting that a config could be used re: the `asString`/`toNormalisedString` behaviour (optionally include/exclude: comments, annotations, children), but then having it programmatically/algorithmically defined brings it closer to just a variation of a pretty printer, and limits any scope for manual intervention to override it (as is the case with current examples of the `asString` method.

----

Anyhow, this has turned into somewhat an essay! Going forward I suggest we introduce an interface of some kind now per this pr (perhaps flag it as alpha/pre-release/unstable/subject to change/whatever?) until things have firmed up a bit more?

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
@@ -26,6 +26,7 @@ import com.github.javaparser.TokenRange;
 import com.github.javaparser.ast.comments.BlockComment;
 import com.github.javaparser.ast.comments.Comment;
 import com.github.javaparser.ast.comments.LineComment;
+import com.github.javaparser.ast.nodeTypes.NodeAsString;
 import com.github.javaparser.ast.nodeTypes.NodeWithRange;
 import com.github.javaparser.ast.nodeTypes.NodeWithTokenRange;
 import com.github.javaparser.ast.observer.AstObserver;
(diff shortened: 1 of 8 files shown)


matozoid:
What the heck :-D Coming out of nowhere with a new philosophy and a PR about something that I didn't think needed work is a little hard on the maintainer. Can we please go back to #2032 and discuss there?

MysterAitch:
> What the heck :-D Coming out of nowhere with a new philosophy and a PR about something that I didn't think needed work is a little hard on the maintainer. Can we please go back to #2032 and discuss there?

Apologies for the sudden side-swipe!

I've renamed this to indicate the pr was more as an experiment about that would be involved in delivering it and as a concrete basis for discussion. 

Since creating this I've hopefully gotten better/less shy about it, but I'm still feeling my way around how pull requests and similar work! 

PS: I'm on mobile at the moment and can't seem to find the option to mark this PR as a draft - if you get to this before I manage to get to a computer, please do so :)

matozoid:
Okay, let's start with the point that seems to be central: what does `asString()` do? I didn't get any further than "it's an arbitrary representation of the node" which is certainly quite a lousy design. Useful though. Now that you talk about normalisation, definition, etc, I'm thinking maybe this is closer to `getDeclarationAsString` than I thought? There is also `CallableDeclaration.Signature` that is somewhat related. Do you think we can unify this into one thing?

(Postponing the rest of the points to later ;-) )

Issue #2411, opened 2019-11-14
addStatement(int index, Statement statement) puts statements at the beginning of the BlockStmt


vin-car:
I'm trying to refactor automatically some code in order to create local variables for boolean parts of if that have binary conditions. I would like that my class:

```java
public class CheckIf {
    public void myMethod() {
        if( check1() && check2() ){ System.out.println("Some code"); }
        if( check3() || check4() ){ System.out.println("Something else happen here"); }
        if( check5() && check6() ){ System.out.println("Some other code"); }
    }
}
```
Becomes like this
```java
public class CheckIf {
    public void myMethod() {
        final boolean condLeft0 = check1();
        final boolean condRight0 = check2();
        if( condLeft0 && condRight0 ){ System.out.println("Some code"); }
        final boolean condLeft1 = check3();
        final boolean condRight1 = check4();
        if( condLeft1 || condRight1 ){ System.out.println("Something else happen here"); }
        final boolean condLeft2 = check5();
        final boolean condRight2 = check6();
        if( condLeft2 && condRight2 ){ System.out.println("Some other code"); }
    }
}
```

I'm using the following code:
```java
public class IfConditionVistor extends ModifierVisitor<Void> {
    private int varIndex = 0;
    
    @Override
    public Visitable visit(IfStmt ifStmt, Void arg) {
        Expression condExpr = ifStmt.getCondition();
        if (condExpr instanceof BinaryExpr) {
            BinaryExpr binaryExpr = (BinaryExpr) condExpr;
            Expression leftExpr = binaryExpr.getLeft();
            Expression rightExpr = binaryExpr.getRight();
            
            PrimitiveType boolType = new PrimitiveType(Primitive.BOOLEAN);
            
            String leftVarName = "condLeft" + varIndex;
            VariableDeclarationExpr leftVarDeclExpr = getVarDeclExpr(boolType, leftVarName, leftExpr);
            
            String rightVarName = "condRight" + varIndex;
            VariableDeclarationExpr rightVarDeclExpr = getVarDeclExpr(boolType, rightVarName, rightExpr);
            
            ifStmt.findAncestor(BlockStmt.class).ifPresent(parentBlock -> {
                List<Node> childNodes = parentBlock.getChildNodes();
                int idx = childNodes.indexOf(ifStmt); // in the sample class idx is 0, then 1 end then 2
                if( idx == -1 ) { throw new RuntimeException(); /* tmp */ }

[...]

PR #2426, opened 2019-11-17
added notes re: sort order of `getChildNodes` vs `getStatements`

MysterAitch:
dded some notes to the javadoc to clarify that the order of the nodes return from `getChildNodes` does not necessarily match the order that they appear within the source code, and that `getStatements()` may be a better alternative if it available -- see #2411


javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -31,6 +31,7 @@
 import com.github.javaparser.ast.observer.AstObserver;
 import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.ast.observer.PropagatingAstObserver;
+import com.github.javaparser.ast.stmt.BlockStmt;
```
MysterAitch, line 34: This bit confused me just now as I was reading through -- turns out this is due to the `{@link}` on line 316. 

Reading up on this, the alternative seems to be to state the fully qualified name instead.

Happy to edit if requested as I don't have a strong view either way (perhaps a small leaning towards keeping the import).

matozoid:
Oh, it's not about using getstatement, it's about using a specific property, not the general "child nodes"

MysterAitch:
@matozoid -- would you mind expanding on that please?

I'm not sure if it's a general statement being made, or if there's something I misunderstood that you're correcting me on? If it's the latter then I'm not seeing it yet!

matozoid:
Sorry, I was in a hurry. I think we can word it a bit stronger by suggesting not to use the method, and to see it as internal. Also: suggesting getstatements is a bit arbitrary, although I'm not sure how to say "a property of a node, you know, the ones that form the tree, not the others. The fields, you want the fields. But not the childnodes field" :-)

matozoid:
Waiting for a response from @MysterAitch :-)

MysterAitch:
> 
> 
> Waiting for a response from @MysterAitch :-)

I'm still looking at this every so often not really understanding what you mean, sorry! :(

matozoid:
Please ask on Gitter, because I do not know what you don't get :-)

jlerbsc:

PR #2428, opened 2019-11-19
Provide an implicit bound of "extends Object" when resolving TypeVariables with an empty bounds list

tophyr:
Fixes issue #2044 raised by @MysterAitch


matozoid:
@MysterAitch - are you okay with this fix?

MysterAitch:
I'll do some additional digging in the next day or two, but nothing jumps out as being questionable and it seems to pass the testcases so prima-facie it seems okay to me :)

matozoid:
@MysterAitch okay, give a shout when you're done digging, preferably before the Sunday release :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
```diff
@@ -150,12 +152,18 @@ private MethodUsage resolveMethodTypeParametersFromExplicitList(TypeSolver typeS
     @Override
     public SymbolReference<ResolvedMethodDeclaration> solveMethod(String name, List<ResolvedType> argumentsTypes, boolean staticOnly) {
         Collection<ResolvedReferenceTypeDeclaration> rrtds = findTypeDeclarations(wrappedNode.getScope());
+
+        if (rrtds.isEmpty()) {
```
MysterAitch, line 161: My preference would be to add a line comment here with a short note about why we're doing this and to include a link to the GitHub issue. Doing so will help any future readers of the code :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
```diff
@@ -384,12 +392,21 @@ private void matchTypeParameters(ResolvedType expectedType, ResolvedType actualT
     }
 
     private Optional<MethodUsage> solveMethodAsUsage(ResolvedTypeVariable tp, String name, List<ResolvedType> argumentsTypes, Context invokationContext) {
-        for (ResolvedTypeParameterDeclaration.Bound bound : tp.asTypeParameter().getBounds()) {
+        List<ResolvedTypeParameterDeclaration.Bound> bounds = tp.asTypeParameter().getBounds();
+
+        if (bounds.isEmpty()) {
```
MysterAitch, line 406: Ditto here

My preference would be to add a line comment here with a short note about why we're doing this and to include a link to the GitHub issue. This will help any future readers of the code :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
```diff
@@ -150,12 +152,18 @@ private MethodUsage resolveMethodTypeParametersFromExplicitList(TypeSolver typeS
     @Override
     public SymbolReference<ResolvedMethodDeclaration> solveMethod(String name, List<ResolvedType> argumentsTypes, boolean staticOnly) {
         Collection<ResolvedReferenceTypeDeclaration> rrtds = findTypeDeclarations(wrappedNode.getScope());
+
+        if (rrtds.isEmpty()) {
+            rrtds = Collections.singleton(typeSolver.solveType(Object.class.getCanonicalName()));
```
MysterAitch, line 166: Question: Should we manually specify `java.lang.Object` here as opposed to trying to resolve the object class?

Are there any cases where it won't resolve to `java.lang.Object`? 

I do seem to recall a discussion about supplying a custom `Object` class (perhaps in the context of Android app development and supplying a jar file which contains an Object.java???), but I'll need to have a search to figure out if it is relevant/applicable here, so my current thought is "maybe"..?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
MysterAitch, line 161: Question: Is there a reason that we're only adding it if it's empty? Could/should we do it in all cases?

There might be an obvious answer here (or in linked discussions) that I'm not seeing while on mobile!

MysterAitch (changes requested):
I'm not quite sure how this GitHub review process works, but let's experiment!

I've added some comments/notes about specific things I'll investigate later unless comments come in beforehand answering some of the things I'm not sure about right now off the top of my head!

In summary, I'm leaning heavily toward approval because it seems to solve the issue - hesitating **_only_** because I don't know the code well enough to know if this is the "best" or "most complete" way!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
tophyr: no, you're right - `<T>` is always equivalent to `<T extends java.lang.Object>`

I'm not sure about manually specifying vs resolving - the way I wrote it seemed like the most straightforward way to obtain a `ResolvedReferenceTypeDeclaration` for `java.lang.Object`, but I'm open to better ideas.

Android still uses `java.lang.Object`, so I can't think of any "custom implementation" issues - the public API's the public API

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
tophyr: if we add it when it's not empty then we're effectively adding a bound that's almost certainly more permissive than the actual type. consider `<T extends InputStream>` - if we add `Object` unconditionally here then we'd effectively be transforming that type parameter into `<T extends InputStream | Object>`.

MysterAitch (approved):

MysterAitch:
I am sorry for the substantial delays here -- turns out I had somewhat less time last weekend than I originally anticipated!

Looks good to me - I've not yet found a way to break the fix or found a reason to hold this back :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
MysterAitch: It still niggles at me that perhaps there is a way for the reference to Object not be `java.lang.Object` but can't yet figure out if/how this might actually be the case (e.g. an inner class somewhere, or perhaps an `Object` located somewhere other than `java.lang`). 

Personally I would be more comfortable explicitly referring to `java.lang.Object`, but this shouldn't block merging of the PR as-is though :+1:

tophyr:
Ohhh I see the concern I think. Since I'm not looking the class up by
string "Object", but rather using the Object.class type, wouldn't that
resolve to the same absolute package/class that Object.class is (aka
java.lang?)

matozoid:
I'm not sure about the state of this PR, but it's been in review long enough so I'm merging it :-)

PR #2429, opened 2019-11-21
Constructor resolution fails for generic javassist types

maartenc:
The list returned by constructor.getTypeParameters() is not always writeable (for instance the empty list returned by JavassistConstructorDeclaration), this patch avoids changing this list.


matozoid:
Looks good! Thanks again

PR #2431, opened 2019-11-25
fix for solving annotations with qualified names

### LINKED ISSUE #2412: AstResolutionUtils hasDirectlyAnnotation throws UnsolvedSymbolException when an annotation used on a class with qualified name
If I have a class that looks like this:
```java
@com.foo.bar.SomeAnnotation
class MyClass { }
```

And I resolve this class and call the `ResolvedReferenceTypeDeclaration.hasAnnotation(String)` on the resolved type with any annotation, it will throw an `UnsolvedSymbolException` even if the `com.foo.bar.SomeAnnotation` can be solved by the configured type solver. 

The exception is thrown by the `AstResolutionUtils`, and if I'm not mistaken this is the part that causes it:
https://github.com/javaparser/javaparser/blob/374848ca4377ccf48f4009d0d55da9c7324163d0/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/AstResolutionUtils.java#L90

I'm not sure if it is a good solution, but I guess the `annotationExpr.getName().getId()` could be replaced with:
`annotationExpr.getName().getQualifier().map(Name::asString).orElse("") + annotationExpr.getName().getIdentifier()` 

In which case, the `CompilationUnitContext` would resolve the type here:
https://github.com/javaparser/javaparser/blob/6b9344dcd41302a1e81d8ff55baa8f8ff43dc5e6/javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-core/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java#L190

If this is acceptable as a solution I can create a PR with this fix.

PeterCsubak:
Fixes #2412.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -85,15 +85,15 @@ public class CompilationUnitContext extends AbstractJavaParserContext<Compilatio
         // Look among statically imported values
         if (wrappedNode.getImports() != null) {
             for (ImportDeclaration importDecl : wrappedNode.getImports()) {
-                if(importDecl.isStatic()){
-                    if(importDecl.isAsterisk()) {
+                if (importDecl.isStatic()) {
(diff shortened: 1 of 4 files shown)


matozoid:
Looks good to me, thanks!

Issue #1757, opened 2018-08-16
ObjectCreationExpr with anonymous class body cannot be resolved on interface types


kgeilmann:
```JavaParserFacade.get(typeSolver).solve( expr )``` fails with a ClassCastException for ObjectCreationExpr nodes which have an anonymous class body and whose type is an interface, like in 
```
new Comparator<String>() {
   public int compare(String o1, String o2) {
       return 0;
    }
}
```
I've seen it failing with the following exceptions:

java.lang.ClassCastException: com.github.javaparser.symbolsolver.reflectionmodel.ReflectionInterfaceDeclaration cannot be cast to com.github.javaparser.resolution.declarations.ResolvedClassDeclaration

java.lang.ClassCastException: com.github.javaparser.symbolsolver.javassistmodel.JavassistInterfaceDeclaration cannot be cast to com.github.javaparser.resolution.declarations.ResolvedClassDeclaration

java.lang.ClassCastException: com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration cannot be cast to com.github.javaparser.resolution.declarations.ResolvedClassDeclaration

malteskoruppa:
Quick comment: This is very similar to #1599.

There, the problem was that one cannot resolve method calls to methods defined in an anonymous class.

Here, the problem is that one cannot resolve object creation expressions to constructors defined in an anonymous class. (In this case, it is an implicit declaration of a default constructor).

This particular class cast expression problem stems from the fact that JSS first resolves the type of the instantiated object and simply assumes that it must always be a `ResolvedClassDeclaration`, since the type is being instantiated. But, in the case of an anonymous class, it could also be a `ResolvedInterfaceDeclaration`, of course. (As far as your class cast exceptions are concerned: Note that all of `ReflectionInterfaceDeclaration`, `JavassistInterfaceDeclaration`, and `JavaParserInterfaceDeclaration` are instances of `ResolvedInterfaceDeclaration`, which cannot be cast to `ResolvedClassDeclaration`.)

In the big picture, though, all of this is one big issue "implement support to resolve call expressions to callable declarations in anonymous classes".

malteskoruppa:
Ah, re-reading that other issue, I see that the very problem mentioned here was already reported there too: https://github.com/javaparser/javaparser/issues/1599#issuecomment-392761331.

matozoid:
[...]

Issue #1599, opened 2018-05-24
Resolving anonymous classes and methods within them doesn't work


jjbr:
Currently I am struggling to resolve the type of variables that have anonymous class types and any methods called of those anonymous types. Take following code:

```
public abstract class A {
  public abstract void bar();
}

public void foo() {
  A myVar = new A() {
    public void bar() {System.out.println("bar");}
    public void bar2() {System.out.println("bar2");}
  };

  myVar.bar();
  myVar.bar2();
}
```

If I try to calculate the resolved type of `myVar.bar()` I get the `public abstract void bar();` from the original abstract class declaration. If I try to resolve `myVar.bar2()`, I get an `UnsolvedSymbolException`. Similarly, if I try to resolve the type of myVar, I only get the abstract class A.

Is this the desired behaviour? It seems as though it is not currently possible to work with anonymous classes properly. Or is there something I am missing?

matozoid:
Does this have anything to do with #1436?

jjbr:
I believe they are related, yes, although the approach described in #1436 seems to be slightly different. From what I can see they may, in essence, come down to the same issue, though.

matozoid:
Then I'll declare this a bug report too.

jjbr:
I just stumbled on a case where the core of this issue produces an arguably even less desirable result. If I implement and instantiate an interface with an anonymous implementation and try to resolve the constructor call, JavaParser throws a ClassCastException. Take the following code as example:

```
	CompilationUnit cu = JavaParser.parse(
			"public class X {" +
			"   interface Foo {" +
			"       public void foo();" +
			"   }" +
			"   public void bar() {" +
			"       Foo f = new Foo() {" +
			"           @Override" +
			"           public void foo() {System.out.println(\"Out\");}" +
			"       };" +
			"   }" +
			"}");

	ClassOrInterfaceDeclaration xClass = cu.getClassByName("X").get();
	ObjectCreationExpr creationExpr = (ObjectCreationExpr)
				((VariableDeclarationExpr)xClass.getMethods().get(0).getBody().get().getStatement(0).getChildNodes().get(0)).getVariable(0).getInitializer().get();
	creationExpr.resolve(); // throws ClassCastException
```

[...]

Issue #1436, opened 2018-02-22
Problem with solving anonymous class


Eransho:
Hi,
First of all - great project!
I've encountered the following issue: When trying to solve a code containing an anonymous class usage, you get the following exception:
```
java.lang.ClassCastException: com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserInterfaceDeclaration cannot be cast to com.github.javaparser.resolution.declarations.ResolvedClassDeclaration
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:207)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:132)
```
The relevant part of the code I tried to solved:
```
TypeA obj = ...
obj.doSomething(new TypeIfc() {
    @Override
    public void noReturn() { ... }
	
    @Override
    public TypeIfc me() { ... }
});
```
Where `TypeIfc` is a local interface in the local project I'm trying to solve. 

I wasn't sure if this is a known issue or not, didn't find it on the list.
Cheers

matozoid:
Thanks for the report!

matozoid:
I tried to prepare a test case, but it doesn't fail. Can you help out? https://github.com/matozoid/javaparser/blob/issue_1436_solve_anon_class/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1436.java

Eransho:
Instead of calling `oce.calculateResolvedType()` I used `JavaParserFacade.get(typeSolver).solve(oce)`. 
My motivation is to get the resolved type of both return type and params of methods/ctors calls

matozoid:
Alright, which version are you using?

Eransho:
3.5.15

PR #2433, opened 2019-11-26
Fix: resolving constructors of anonymous inner classes implementing an interface fails

maartenc:
This might also fix issue #1757, although I didn't get the exception reported in that issue.


matozoid:
We should check #1599 and #1436 too then, because they get referred too. But that doesn't block this PR.

maartenc:
Hi @matozoid, I've added a testcase for issue #1436, and this PR also seems to fix that issue.

In the newly added testcase I also check the name of the resolved type, and there might be an issue about the name of the anonymous inner class, but I'm not sure what the spec says. I think the correct type should be B$1 (the first anonymous inner class of B), but JavaParser returns 'B.Anonymous-UUID' as type name. Anyway, if the name is incorrect, that is out of the scope of this PR imho.

Maarten

maartenc:
About issue #1599 : I think that's about resolving methods of anoymous inner classes. This PR only addresses constructor resulotion, so I don't think it will solve that issue.

In issue #1599, the problem about resolving the constructor is mentioned in a comment, that should be fixed by this PR.

matozoid:
> In the newly added testcase I also check the name of the resolved type, and there might be an issue about the name of the anonymous inner class, but I'm not sure what the spec says. I think the correct type should be B$1 (the first anonymous inner class of B), but JavaParser returns 'B.Anonymous-UUID' as type name. Anyway, if the name is incorrect, that is out of the scope of this PR imho.

Can you create a new issue for that?

If this is all done, then we can merge :-)

Issue #2162, opened 2019-04-09
JavaSymbolSolver is unable to solve generic method return types


pgrepds:
I'm trying to solve the generic return type of a `MethodCallExpr`. 

e.g.: `getView().getTest()`, where `getView()` returns `V` with the superclass `AWT.Component`.

 I'm using `JavaParserFacade.solve(methodCallExpr).getCorrespondingDeclaration().getReturnType()`, where `methodCallExpr` is the expression above (`getView().getTest()`).

`getView()` is resolved in the type `AWT.Component`, which does not obviously has a method called `getTest`.

I get `java.lang.RuntimeException: Unable to calculate the type of a parameter of a method call`.

matozoid:
So this is your base class?

```java
abstract class Screen {
    abstract <V extends java.awt.Component> V getView();
}
```

pgrepds:
Yes, exactly. And V has a method, which is not available in Component.

matozoid:
This?
```java
import java.awt.*;

abstract class Screen {
    abstract <V extends Component> V getView();
}

class D extends Component {
    void test() {
    }
}

class B extends Screen {
    @Override
    D getView() {
        return new D();
    }
}
```

matozoid:
Or maybe this?

```java
import java.awt.*;

abstract class Screen <V extends Component> {
    abstract V getView();
}

class D extends Component {
    void test() {
    }
}

class B extends Screen<D> {
    @Override
    D getView() {
        return new D();
    }
}
```

pgrepds:
It's the second one.

MysterAitch:
I've spent some time on this and my conclusion is that the issue described has presumably been solved since being posted (or I have misunderstood the problem) as I have not been able to replicate it.

Extending @matozoid 's code above, I have been using the following example code sample to run tests against (evaluating the output of the method calls in the class `Run`).

I'll submit a testcase PR shortly to show tests using this code sample passing.

```java

import java.awt.*;

abstract class Screen <V extends Component> {
    abstract V getView();
}

class D extends Component {
    void getTest() {
    }
}

class B extends Screen<D> {
    @Override
    D getView() {
        return new D();
    }
}

class Run {
    public static void main(String[] args) {
        B b1 = new B();
        b1.getView(); // b1.getView() -> B#getView(), overriding Screen#getView() -> returns object of type D.
        
[...]

PR #2434, opened 2019-11-27
[TESTCASE] Issue 2162

MysterAitch:
Testcase hopefully showing that #2162 can be closed as the problem (as I understand it) is no longer present.


matozoid:
That seems to be the case! Shall I merge it?

MysterAitch:
Assuming you're happy, sure! :)

matozoid:
I'm very happy every time the tower of symbol solver bugs gets reduced :-)

Issue #2210, opened 2019-05-04
Resolution of Method References


MassimilianoPoggi:
Hi,
  I can't find a way to resolve method references (such as System.out::println) anywhere. Is this not supported by the current version of JSS?
In the process of trying to implement something of the sort myself I've also noticed that JavaParser parses the left side of the method reference as a TypeExpr: isn't this incorrect behaviour, given that instances can be the scope of a method reference (such as the example provided above)?

matozoid:
Hi! It might be related to #2160.

MassimilianoPoggi:
While eventual problems in the resolution might be related to the mentioned issue, the problems I'm outlining in this issue are unrelated:

1. the MethodReferenceExpr class does not implement Resolvable at all
2. JavaParser is incorrectly parsing method references by automatically assuming the left side of the expression to be a TypeExpr, while it should be possible to also parse it as a NameExpr/FieldAccessExpr for cases where the left side is an instance and not a class.

matozoid:
Sorry, I didn't mean to close it!

scootafew:
I'm also facing the same problem and trying to investigate, @MassimilianoPoggi did you ever get anywhere with what you were trying to implement for this?

MassimilianoPoggi:
@scootafew Being on a deadline at the time and noticing a few other problems within the tool I ended up using what seemed like a more mature alternative in [Spoon](https://github.com/INRIA/spoon). I'm sorry I can't be of further help with your problem.

scootafew:
@MassimilianoPoggi no worries at all, thanks for the quick reply! :)

matozoid:
Spoon uses the compiler libraries from Eclipse behind the scenes, so their symbol resolution is definitely more mature. JavaParser has other good things (like not secretly being Eclipse ;-) )

jlerbsc:
@MassimilianoPoggi could you give us a test case to explore this issue?

MassimilianoPoggi:
While I don't have time to write a fully-fledged test case, here's a snippet quickly reproducing the "parsing" issue (I'm not sure whether to label it as a parsing issue exclusively as I'm fairly certain you can't differentitate between field accesses and fully qualified type references without a resolution step in-between, e.g. `java.lang.Object::toString`, so maybe just not calling it `TypeExpr` could suffice as it would be less ambiguous)

```   
public static void main(String... args) {
        ParseResult<Statement> methodReference = new JavaParser()
[...]

PR #2450, opened 2019-12-15
Add method reference resolution to the AST

matozoid:
Related to #2210.

Disclaimer: I don't know what I'm doing. I just copied stuff around until the test worked.


matozoid:
@MassimilianoPoggi @scootafew @jlerbsc any comments on this?

scootafew:
@matozoid this looks good to me! I'd be interested to test on different types of method reference (e.g. constructor calls) but can be raised as a separate issue later

matozoid:
Yeah, at least there is a resolve method that does something now :-)

jlerbsc:
Very good ... No need to look at Spoon now ;-)

PR #2456, opened 2019-12-18
Fix resolving methods with generic return types

maartenc:
Fix an issue with resolving generic return types.

Consider the example below, when resolving the `foo()` methodcall in the Bar class, the return type was not resolved to `String`, but was still `T`.

```
public class Foo <T> {
    public T foo() {
        return null;
    }
}

public class Bar extends Foo<String> {
    public void bar() {
        foo();
    }
}
```


maartenc:
Hmz, I have broken some tests... I'll look into it...

matozoid:
Hey @maartenc - how is this one going?

maartenc:
Sorry, I didn't had much time yet

matozoid:
Hey cool! Time to merge it.

Issue #2455, opened 2019-12-16
Method References on object instances or field accesses parsed to TypeExpr instead of expected NameExpr or FieldAccessExpr


scootafew:
Since the addition of a `resolve` method to `MethodReferenceExpr` I've been trying to resolve some different types of method references. If we take the example of a call to `System.out::println`, when it is expressed as a method reference it cannot be resolved (instead throws `UnsupportedOperationException`), whilst the equivalent regular method call is resolved with no problems.

So this test passes:
```
@Test
    void resolveMethodCallExpr() {
        JavaParser parser = createParserWithResolver(new ReflectionTypeSolver());
        MethodCallExpr methodCall = parser.parse("class X{void x(){System.out.println(\"Hello world!\");}}")
                .getResult().get()
                .findFirst(MethodCallExpr.class).get();
        ResolvedMethodDeclaration resolvedMethodRef = methodCall.resolve();
        assertEquals("println", resolvedMethodRef.getName());
        assertEquals("void", resolvedMethodRef.getReturnType().describe());
    }
```

But this one does not:
```
    @Test
    void resolveMethodReferenceExpr() {
        JavaParser parser = createParserWithResolver(new ReflectionTypeSolver());
        MethodReferenceExpr methodRef = parser.parse("class X{void x(){Consumer<? super String>r=System.out::println;}}")
                .getResult().get()
                .findFirst(MethodReferenceExpr.class).get();
        ResolvedMethodDeclaration resolvedMethodRef = methodRef.resolve();
        assertEquals("println", resolvedMethodRef.getName());
        assertEquals("void", resolvedMethodRef.getReturnType().describe());
    }
```

From the debugging I've done and the comments made in #2210 this does seem to be down to the way method references are being parsed. In the examples above, the scope of the call is `System.out`, which for a method reference is parsed to a `ClassOrInterfaceType` wrapped in a `TypeExpr`. The equivalent for a method call is parsed to a `FieldAccessExpr` which has the correct symbol-solver logic.

I'm hoping someone could please shed some light on where this behavior would be defined (I'm hoping not in JavaCC :P)

matozoid:
The AST seems reasonable to me (correct me if I'm wrong - I often am :-) )
System is a type
out is a name

In JavaParserFacade line 385, the solver is rather dumb. It grabs `id`, which is `out`, and looks for the its declaration, completely ignoring `System`. What it should have done is solve the whole name expression!

jlerbsc:
[...]

PR #2473, opened 2019-12-27
Improving Symbol-Solving of Method References

scootafew:
Improved how `MethodReferenceExpr`'s are resolved to `MethodUsage`'s to more closely align with how it is done for regular `MethodCallExpr`'s. Also added plenty of test cases to verify new behavior.

Please note: I've left two test cases disabled as they currently fail due to #2455. I believe the solution lies in modifying the grammar so that field accesses and name expressions are parsed as such in the scope of `MethodReferenceExpr`'s, rather than copying symbol solving logic from these to the method solving a `TypeExpr`.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -132,8 +132,7 @@ public class JavaParserFacade {
     }
 
     public SymbolReference<ResolvedMethodDeclaration> solve(MethodReferenceExpr methodReferenceExpr) {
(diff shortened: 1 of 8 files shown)


matozoid:
Awesome!

matozoid:
Looks like there is a JDK-dependent test there. It's best to avoid solver tests that assume something about the contents of the JDK :-)

scootafew:
Test failure:
```
MethodReferenceResolutionTest.superclassMethodOverridden:94 expected: <java.util.AbstractList.hashCode()> but was: <java.util.ArrayList.hashCode()>
```

I always love it when a test passes locally and then not on the build :D Makes it slightly harder to debug. Do you have any idea why behavior may be different to locally @matozoid ? Also considering that `ArrayList` itself doesn't override hashcode I'm scratching my head...

scootafew:
Ahh we commented at the same time, I understand. Will fix!

matozoid:
Thanks! I'll do a release on Sunday.

PR #2485, opened 2020-01-18
Added asNumber function to literal and long literal expressions

### LINKED ISSUE #2483: IntegerLiteralExpr#asInt produces NumberFormatException for 2147483648 literal
We have been running into an issue while parsing the expression `-2147483648`.  According to the Java Spec (https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.1), the largest number a decimal literal can be is `2147483648`, but only in combination with the unary expression, since it then represents the lowest number in Java `-2147483648` (-2^31). Larger numbers or the use of it without the unary expression is considered a compile error.

JavaParser will (correctly) deduce that `-2147483648` is a unary expression containing a literal. However, by calling `asInt` on the `IntegerLiteralExpr` it will (correctly) produce a `NumberFormatExpression` since it cannot represent the value of the literal `2147483648` in a Java `int`. However, the combination itself is a valid Java construct which we cannot correctly parse without an exception.

Quick fix would probably be to return the literal as a `long`, not sure if this would introduce other problems.

oxisto:
- [x] Fixes #2483
- [x] Documentation as to when to use `toNumber` and deprecate `toInt`
- [x] Ensure that the number is only returned in combination with a unary minus expression

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
@@ -28,6 +28,7 @@ import com.github.javaparser.ast.visitor.VoidVisitor;
 import com.github.javaparser.metamodel.IntegerLiteralExprMetaModel;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
 import com.github.javaparser.TokenRange;
+import java.util.Objects;
 import java.util.function.Consumer;
 import java.util.Optional;
 import com.github.javaparser.ast.Generated;
@@ -101,6 +102,22 @@ public class IntegerLiteralExpr extends LiteralStringValueExpr {
             return Integer.parseUnsignedInt(result.substring(1), 8);
         }
         return Integer.parseInt(result);
+    }
+
+    /**
+     * @return the literal value as an integer while respecting different number representations
+     */
+    public Number asNumber() {
+        /* we need to handle the special case for the literal 2147483648, which is used to
+         * represent Integer.MIN_VALUE (-2147483648) as a combination of a UnaryExpr and an
+         * IntegerLiteralExpr. However 2147483648 cannot be represented in an integer, so we
+         * need to return a long
+         */
+        if (Objects.equals(value, "2147483648")) {
+            return 2147483648L;
+        } else {
+            return asInt();
+        }
     }
 
     public IntegerLiteralExpr setInt(int value) {
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
@@ -52,6 +52,9 @@ import static com.github.javaparser.utils.Utils.hasUnaryMinusAsParent;
  */
 public class IntegerLiteralExpr extends LiteralStringValueExpr {
 
+    public static final String MAX_31_BIT_UNSIGNED_VALUE_AS_STRING = "2147483648";
+    public static final long MAX_31_BIT_UNSIGNED_VALUE_AS_LONG = 2147483648L;
+
     public IntegerLiteralExpr() {
         this(null, "0");
     }
@@ -96,7 +99,9 @@ public class IntegerLiteralExpr extends LiteralStringValueExpr {
 
     /**
      * @return the literal value as an integer while respecting different number representations
-     * @deprecated Will be made private or merged with {@link IntegerLiteralExpr#asNumber()} in future releases
+     * @deprecated This function has issues with corner cases, such as 2147483648, so please use {@link
+     * IntegerLiteralExpr#asNumber()}. It will be made private or merged with {@link IntegerLiteralExpr#asNumber()} in
+     * future releases
      */
     public int asInt() {
         String result = value.replaceAll("_", "");
@@ -113,9 +118,10 @@ public class IntegerLiteralExpr extends LiteralStringValueExpr {
     }
 
     /**
-     * This function returns a representation of the literal values as a number. This will return an integer, except for
-     * the case when the literal has the value 2147483648 (which is only allowed in the expression
-     * <code>-2147483648</code>) and returns a long.
+     * This function returns a representation of the literal value as a number. This will return an integer, except for
+     * the case when the literal has the value <code>2147483648</code>. This special literal is only allowed in the
+     * expression <code>-2147483648</code> which represents <code>Integer.MIN_VALUE</code>). However 2147483648 (2^31)
+     * is out of range of int, which is -(2^31) to (2^31)-1 and thus a long must be returned.
      * <p>
      * Note, that this function will NOT return a negative number if the literal was specified in decimal, since
      * according to the language specification an expression such as <code>-1</code> is represented by a unary
@@ -132,8 +138,8 @@ public class IntegerLiteralExpr extends LiteralStringValueExpr {
          * IntegerLiteralExpr. However 2147483648 cannot be represented in an integer, so we
          * need to return a long
          */
-        if (Objects.equals(value, "2147483648") && hasUnaryMinusAsParent(this)) {
-            return 2147483648L;
+        if (Objects.equals(value, MAX_31_BIT_UNSIGNED_VALUE_AS_STRING) && hasUnaryMinusAsParent(this)) {
+            return MAX_31_BIT_UNSIGNED_VALUE_AS_LONG;
         } else {
             return asInt();
         }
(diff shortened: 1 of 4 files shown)


matozoid:
Okay, that's good. Now we need to explain to the users why we have both toInt and toNumber (I say deprecate toInt), and it would be helpful to explain how negative numbers work.

oxisto:
Yeah this needs some more documentation, and probably also the equivalent of the `setInt` function. I also wanted to add code that validates whether 2147483648 is only used in combination with an unary expression (it is considered as a parsing error otherwise according to spec) but i did not find a propery place to check this. I tried overloading the `setValue` function and checking it there but here the parent is not set yet. I have also seen something about validators? Are they the correct place for such a check? I implemented one but it did not seem to do anything, although maybe i forgot to re-generate the generators

matozoid:
`setValue` is probably the generated setter? Don't change that one.

The validators are currently intended to run right after parsing to check if the AST conforms to the selected language level, and some other sanity checks. You can run them again after modifying the AST, but I don't expect many people to do this, nor can I think of a good point to run them automatically. Before pretty printing? Not everyone pretty prints. At some other points? Is it worth the performance "hit"? The *good* thing is that nobody every complained about negative literals before, so this is a niche problem. You may well be the only person ever to run into this issue, so I'm not too worried :-)

oxisto:
Ready for review. I did not touch the `setInt` functions though, they will be a pain to modify, not sure if it is worth it so considering this "done".

matozoid:
Maybe just @deprecate the `setInt`s. That way we annoy users into reading the documentation :-)

One final thing: when using @deprecate, I try to say clearly why it is deprecated, and what to use instead.

MysterAitch:
Minor documentation tweak request: include comments/notes to make it explicit how these numbers are derived please, and _why_ they are valid/invalid please -- helps to make it clear there's a specific reason for the values, rather than being arbitrary "magic numbers" :)

e.g.: 
```java
// 2^31 = 2,147,483,648
// 2^63 = 9,223,372,036,854,775,808
// 2147483648 = 2^31 = Integer.MAX_VALUE + 1 // thus out of range of int, which is -(2^31) to (2^31)-1
// Long.MIN_VALUE = -(2^63) = -(9,223,372,036,854,775,808)
// Long.MAX_VALUE = ((2^63)-1) = 9,223,372,036,854,775,807
String max31BitUnsignedValueAsString = "2147483648"; // 2^31 = 2,147,483,648
```

Not a biggie and not worth holding this up to add in, just thinking ahead :)

oxisto:
> Maybe just @deprecate the `setInt`s. That way we annoy users into reading the documentation :-)

Not sure if we can @deprecate `setInt`, since there is no new `setNumber`. I will update the documentation and bit and then hopefully this is ready to go.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
```diff
@@ -88,6 +99,9 @@ public boolean remove(Node node) {
 
     /**
      * @return the literal value as an integer while respecting different number representations
+     * @deprecated This function has issues with corner cases, such as 2147483648, so please use {@link
+     * IntegerLiteralExpr#asNumber()}. It will be made private or merged with {@link IntegerLiteralExpr#asNumber()} in
+     * future releases
      */
     public int asInt() {
```
MysterAitch, line 109: Please add the annotation `@deprecated`

javaparser-core/src/main/java/com/github/javaparser/ast/expr/LongLiteralExpr.java
```diff
@@ -88,6 +96,9 @@ public boolean remove(Node node) {
 
     /**
      * @return the literal value as an long while respecting different number representations
+     * @deprecated This function has issues with corner cases, such as 9223372036854775808L, so please use {@link
+     * LongLiteralExpr#asNumber()}. It will be made private or merged with {@link LongLiteralExpr#asNumber()} in future
+     * releases
      */
     public long asLong() {
```
MysterAitch, line 111: Please add the annotation `@deprecated`

MysterAitch:
> > Maybe just @deprecate the `setInt`s. That way we annoy users into reading the documentation :-)
> 
> Not sure if we can @deprecate `setInt`, since there is no new `setNumber`. I will update the documentation and bit and then hopefully this is ready to go.

There are two constructors - one which takes an int value and one which takes a string value.

Rather than deprecate, let's overload it and have a `setInt` method that takes a string?

Or, have a pair of more generally named `setValue(String)`/`setValue(Number)` methods?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/IntegerLiteralExpr.java
```diff
@@ -103,6 +117,34 @@ public int asInt() {
...
     }
 
+    /**
+     * This function returns a representation of the literal value as a number. This will return an integer, except for
+     * the case when the literal has the value <code>2147483648</code>. This special literal is only allowed in the
+     * expression <code>-2147483648</code> which represents <code>Integer.MIN_VALUE</code>). However 2147483648 (2^31)
+     * is out of range of int, which is -(2^31) to (2^31)-1 and thus a long must be returned.
+     * <p>
+     * Note, that this function will NOT return a negative number if the literal was specified in decimal, since
+     * according to the language specification an expression such as <code>-1</code> is represented by a unary
```
MysterAitch: Is this section 15.15.4 or a different section?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/LongLiteralExpr.java
```diff
@@ -107,6 +118,34 @@ public long asLong() {
...
 
+    /**
+     * This function returns a representation of the literal value as a number. This will return a long, except for the
+     * case when the literal has the value <code>9223372036854775808L</code>. This special literal is only allowed in
+     * the expression <code>-9223372036854775808L</code> which represents <code>Long.MIN_VALUE</code>). However
+     * 9223372036854775808 (2^63) is out of range of long, which is -(2^63) to (2^63)-1 and thus a BigInteger must be
+     * returned.
+     * <p>
+     * Note, that this function will NOT return a negative number if the literal was specified in decimal, since *
+     * according to the language specification an expression such as <code>-1L</code> is represented by a unary *
```
MysterAitch: Is this section 15.15.4 or a different section?

matozoid:
I was suggesting deprecation for `setInt` because it makes it easy to do the wrong thing. This whole thread proves that the conversion from and to `int` is tricky. Maybe even deprecate the constructor to herd people into thinking "why can I only set a string?" and coming up with "oh, it is explained in the Javadoc, AND there is a warning about negative numbers there."

As a maintainer I'm certainly not looking forward to explaining this thing over and over in the issue tracker ;-)

oxisto:
Ok I will deprecate setInt and the int constructor then. I actually also forgot to set the @Deprecated annotation and not just the javadoc. I will upload a new version later this evening.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/LongLiteralExpr.java
oxisto: It is defined in 3.10.1 https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.10.1. I can directly refer to the chapter.

matozoid:
I'm happy :-) @MysterAitch - any remarks left?

MysterAitch:
Nothing substantive from me - just one query and a couple of bits of "polish".

- Can we confirm that it is intentional to deprecate `setInt(int)` and `setLong(long)` _without_ adding in `setInt(String)` and `setLong(string)`? ... In favour of the user constructing a new object and replacing it (or using a deprecated method)? 
_I have no strong preference either way, this just feels like it's creeping into the mutable vs immutable territory_

- Yes please, re: adding in the JLS section numbers

- I note the `<li>` tags are left open!

matozoid:
All I can say is that the my thoughts on this are: 1: have a way to correctly determine the integer value of the literal, and 2: help people in keeping their AST valid. Maybe even nicer would be to remove all the utility stuff from the node, and stick the node->int algorithm in a helper class. That would reduce discoverability though.

Don't read this as a suggestion, it's just a comment :-)

oxisto:
> Nothing substantive from me - just one query and a couple of bits of "polish".
> 
> * Can we confirm that it is intentional to deprecate `setInt(int)` and `setLong(long)` _without_ adding in `setInt(String)` and `setLong(string)`? ... In favour of the user constructing a new object and replacing it (or using a deprecated method)?
>   _I have no strong preference either way, this just feels like it's creeping into the mutable vs immutable territory_
> * Yes please, re: adding in the JLS section numbers
> * I note the `<li>` tags are left open!

Compromise: Instead of referring to the constructor I will refer to the setValue(string) methods that exist because of inheritance to StringValueLiteral (which is better btw anyway since the setInt and setLong values did not include property change listeners, but setValues does).

The JLS section numbers should be there, I will fix the li-Tag

matozoid:
... and that's how it will look for this release :-)

OH! What should I write in the changelog.md file?

PR #2425, opened 2019-11-17
Issue 1801 - update copyright notices

MysterAitch:
This PR updates / normalises the copyright notices in all java files. 

Given the two histories for the two sections of the project, this PR uses two copyright notices (one part originally written by Julio, one part originally written by Federico).

This PR also updates the end year in the notice to 2019 (many are 2016, or 2018, or similar).

- ***JavaParser Core*** copyright notices have been updated to: 
```
/*
 * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
 * Copyright (C) 2011, 2013-2019 The JavaParser Team.
 *
 * This file is part of JavaParser.
 *
 * JavaParser can be used either under the terms of
 * a) the GNU Lesser General Public License as published by
 *     the Free Software Foundation, either version 3 of the License, or
 *     (at your option) any later version.
 * b) the terms of the Apache License
 *
 * You should have received a copy of both licenses in LICENCE.LGPL and
 * LICENCE.APACHE. Please refer to those files for details.
 *
 * JavaParser is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Lesser General Public License for more details.
 */
```

- ***JavaParser Symbol Solver*** copyright notices have been updated to: 

```
/*
 * Copyright (C) 2015-2016 Federico Tomassetti
 * Copyright (C) 2017-2019 The JavaParser Team.
 *
 * This file is part of JavaParser.
 *
 * JavaParser can be used either under the terms of
 * a) the GNU Lesser General Public License as published by
 *     the Free Software Foundation, either version 3 of the License, or
 *     (at your option) any later version.
 * b) the terms of the Apache License
 *
 * You should have received a copy of both licenses in LICENCE.LGPL and
 * LICENCE.APACHE. Please refer to those files for details.
 *
 * JavaParser is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Lesser General Public License for more details.
 */
```

See also: 

#655 
#1801 
#1833 
#1835

https://github.com/javaparser/javasymbolsolver/issues/316

Issue #1801, opened 2018-08-27
Change symbol solver's copyright notice


matozoid:
See https://github.com/javaparser/javasymbolsolver/issues/316

PR #2486, opened 2020-01-21
Update copyright notices (update to 2020, and include in generated metamodel files)

MysterAitch:
Follow-up to #1801 #2425 and #2453 

Generators completely overwrite files rather than "edit" them, thus need to have their copyright notices explicitly inserted during file generation - this is now done.

Happy to resubmit as separate pull requests if desired (one for the updated year and one for fixing the generators to include the copyright notice in generated files).


matozoid:
Yay!

PR #2487, opened 2020-01-21
Issue 1511 - IndexOutOfBoundsException when attempting to resolve super()

### LINKED ISSUE #1511: IndexOutOfBoundsException when attempting to resolve super()
Consider the following file `A.java` that we want to parse:

```java
class A {
    A() {
	super();
    }
}
```

This class implicitly extends `java.lang.Object`, and calling `super()` invokes the (default) empty-bodied constructor of `java.lang.Object` (see JLS 8.8.9: _"If the class being declared is the primordial class Object, then the default constructor has an empty body."_).

Now let's try to get the `ExplicitConstructorInvocationStmt` that represents the `super();` statement and then call `resolveInvokedConstructor()` on it. I would expect JSS to return a `ReflectionConstructorDeclaration` representing the parameterless constructor `java.lang.Object#Object()`.

What really happens: an `IndexOutOfBoundsException` is thrown. Stack trace:

```
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0

	at java.util.ArrayList.rangeCheck(ArrayList.java:657)
	at java.util.ArrayList.get(ArrayList.java:433)
	at com.github.javaparser.ast.NodeList.get(NodeList.java:128)
	at com.github.javaparser.ast.nodeTypes.NodeWithExtends.getExtendedTypes(NodeWithExtends.java:39)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:152)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:136)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:147)
	at com.github.javaparser.ast.stmt.ExplicitConstructorInvocationStmt.resolveInvokedConstructor(ExplicitConstructorInvocationStmt.java:291)
```

Explanation: What happens is that while trying to resolve `super();`, JSS attempts to get the class that `A` extends. This happens in `JavaParserFacade.java` in line 152:
`ResolvedType classDecl = JavaParserFacade.get(typeSolver).convert(classNode.getExtendedTypes(0), classNode);`

However, `classNode.getExtendedTypes(0)` throws an `IndexOutOfBoundsException` since class `A` does not explicitly extend any type. Instead, it _implicitly_ extends `java.lang.Object`.

@ftomassetti My question is essentially: **Should this be considered expected behavior (JSS has no understanding that classes that do not explicitly extend anything implicitly extend `java.lang.Object`), or is this a bug/missing feature?**

Just for completeness, I used the following code to try and resolve the `super();` statement:

```java
	@Test
	public void tempTest() throws FileNotFoundException {

		// configure symbol solver
		CombinedTypeSolver typeSolver = new CombinedTypeSolver();
		typeSolver.add(new ReflectionTypeSolver());
		typeSolver.add(new JavaParserTypeSolver(new File("tmp/")));
		JavaSymbolSolver symbolSolver = new JavaSymbolSolver(typeSolver);
		JavaParser.getStaticConfiguration().setSymbolResolver(symbolSolver);

		// get compilation unit & extract explicit constructor invocation statement
		CompilationUnit cu = JavaParser.parse(new File("tmp/A.java"));
		ExplicitConstructorInvocationStmt ecis = cu.getPrimaryType().get()
				                                         .asClassOrInterfaceDeclaration().getMember(0)
				                                         .asConstructorDeclaration().getBody().getStatement(0)
				                                         .asExplicitConstructorInvocationStmt();

		// attempt to resolve explicit constructor invocation statement
		ResolvedConstructorDeclaration rcd = ecis.resolveInvokedConstructor(); // <-- exception occurs
	}
```

MysterAitch:
While this fixes #1511 , is there a better way?

It feels like there should be a better/less-"bitty" solution to have JSS recognise that an absence of an explicit parent should default to `java.lang.Object`. I can't see it at the moment though!

_(if there's no explicit `extends`, use `java.lang.Object` as the ancestor extended type)_

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
@@ -28,9 +28,10 @@ import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 
 /**
- * A node that extends other types.
+ * A node that explicitly extends other types, using the <code>extends</code> keyword.
  */
 public interface NodeWithExtends<N extends Node> {
+
     NodeList<ClassOrInterfaceType> getExtendedTypes();
 
     void tryAddImportToParentCompilationUnit(Class<?> clazz);
@@ -56,6 +57,7 @@ public interface NodeWithExtends<N extends Node> {
     /**
      * @deprecated use addExtendedType
      */
+    @Deprecated
     default N addExtends(Class<?> clazz) {
         return addExtendedType(clazz);
     }
@@ -63,6 +65,7 @@ public interface NodeWithExtends<N extends Node> {
     /**
      * @deprecated use addExtendedType
      */
+    @Deprecated
     default N addExtends(String name) {
         return addExtendedType(name);
     }
@@ -70,7 +73,7 @@ public interface NodeWithExtends<N extends Node> {
     /**
      * Add an "extends" to this and automatically add the import
      *
-     * @param clazz the class to extand from
+     * @param clazz the class to extend from
      * @return this
      */
     default N addExtendedType(Class<?> clazz) {
@@ -89,4 +92,5 @@ public interface NodeWithExtends<N extends Node> {
         getExtendedTypes().add(parseClassOrInterfaceType(name));
         return (N) this;
     }
+
 }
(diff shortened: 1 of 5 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
```diff
@@ -28,9 +28,10 @@
 import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 
 /**
- * A node that extends other types.
+ * A node that explicitly extends other types, using the <code>extends</code> keyword.
```
matozoid, line 31: "A node that can explicitly " etc. would be even better :-)

matozoid:
I have no idea what's going on, so I'll probably pull this tomorrow.

matozoid:
Forgot to put this in the release, so it will be in the next :-)

PR #2491, opened 2020-01-21
Issue 1526 - resolve imports before same-package files

### LINKED ISSUE #1526: CompilationUnitContext.solveType(String name, TypeSolver typeSolver) checks package and imports in wrong order
Assume you have a class like this:

```java
package a.b.c;

import f.g.h.DataObjectFactory;

public class ExampleClass{

protected final DataObjectFactory doFactory;
...
 public DataObject getDataObject(String objectID){
	return (DataObject) doFactory.getDataObject(objectID);
 }
}
```
with the imported class
```java
package f.g.h;

public class DataObjectFactory{
...
 public DataObject getDataObject(String objectID){
	...
 }
}
```

And in package a.b.c there exists also a class called `a.b.c.DataObjectFactory` but which doesn't has a method called `getDataObject `. 
(Please don't argue why it's a bad idea having two classes with the same name - in this case I can't change it anyway).
Now when trying to call `ResolvedType type = JavaParserFacade.get(typeSolver).getType(expr);` where expr is the corresponding MethodCallExpr to `doFactory.getDataObject(objectID)`, one gets a RuntimeException `Exception in thread "main" java.lang.RuntimeException: Method 'getDataObjectMap' cannot be resolved in context doFactory.getDataObjectMap() (line: xxx) MethodCallExprContext{wrapped=doFactory.getDataObjectMap()}. Parameter types: [].`

The reason is: When resolving the Type you end up in `CompilationUnitContext.solveType(String name, TypeSolver typeSolver)` (see below). Here you first check if there is a class called "DataObjectFactory" in the current package (which is true in this case) and return that type. 

So in my case I'll get `a.b.c.DataObjectFactory` instead of `f.g.h.DataObjectFactory` which causes the RuntimeException because there is no method called `getDataObject` in `a.b.c.DataObjectFactory` .

A check for imports would have followed right behind the search in the current package. In my opinion you have to swap these two cases so that you first look at the imports and afterwards check the current package (which I think is the 'natural order' java is working with).

```java
 @Override
    public SymbolReference<ResolvedTypeDeclaration> solveType(String name, TypeSolver typeSolver) {
        if (wrappedNode.getTypes() != null) {
            for (TypeDeclaration<?> type : wrappedNode.getTypes()) {
                if (type.getName().getId().equals(name)) {
                    if (type instanceof ClassOrInterfaceDeclaration) {
                        return SymbolReference.solved(JavaParserFacade.get(typeSolver).getTypeDeclaration((ClassOrInterfaceDeclaration) type));
                    } else if (type instanceof AnnotationDeclaration) {
                        return SymbolReference.solved(new JavaParserAnnotationDeclaration((AnnotationDeclaration) type, typeSolver));
                    } else if (type instanceof EnumDeclaration) {
                        return SymbolReference.solved(new JavaParserEnumDeclaration((EnumDeclaration) type, typeSolver));
                    } else {
                        throw new UnsupportedOperationException(type.getClass().getCanonicalName());
                    }
                }
            }
        }

        // Look in current package
        if (this.wrappedNode.getPackageDeclaration().isPresent()) {
            String qName = this.wrappedNode.getPackageDeclaration().get().getName().toString() + "." + name;
            SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
            if (ref != null && ref.isSolved()) {
                return SymbolReference.adapt(ref, ResolvedTypeDeclaration.class);
            }
        } else {
            // look for classes in the default package
            String qName = name;
            SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
            if (ref != null && ref.isSolved()) {
                return SymbolReference.adapt(ref, ResolvedTypeDeclaration.class);
            }
        }

        if (wrappedNode.getImports() != null) {
            int dotPos = name.indexOf('.');
            String prefix = null;
            if (dotPos > -1) {
                prefix = name.substring(0, dotPos);
            }
            // look into type imports
            for (ImportDeclaration importDecl : wrappedNode.getImports()) {
                if (!importDecl.isAsterisk()) {
                    String qName = importDecl.getNameAsString();
                    boolean defaultPackage = !importDecl.getName().getQualifier().isPresent();
                    boolean found = !defaultPackage && importDecl.getName().getIdentifier().equals(name);
                    if (!found) {
                        if (prefix != null) {
                            found = qName.endsWith("." + prefix);
                            if (found) {
                                qName = qName + name.substring(dotPos);
                            }
                        }
                    }
                    if (found) {
                        SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
                        if (ref.isSolved()) {
                            return SymbolReference.adapt(ref, ResolvedTypeDeclaration.class);
                        }
                    }
                }
            }
            // look into type imports on demand
            for (ImportDeclaration importDecl : wrappedNode.getImports()) {
                if (importDecl.isAsterisk()) {
                    String qName = importDecl.getNameAsString() + "." + name;
                    SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType(qName);
                    if (ref.isSolved()) {
                        return SymbolReference.adapt(ref, ResolvedTypeDeclaration.class);
                    }
                }
            }
        }

        // Look in the java.lang package
        SymbolReference<ResolvedReferenceTypeDeclaration> ref = typeSolver.tryToSolveType("java.lang." + name);
        if (ref.isSolved()) {
            return SymbolReference.adapt(ref, ResolvedTypeDeclaration.class);
        }

        // DO NOT look for absolute name if this name is not qualified: you cannot import classes from the default package
        if (isQualifiedName(name)) {
            return SymbolReference.adapt(typeSolver.tryToSolveType(name), ResolvedTypeDeclaration.class);
        } else {
            return SymbolReference.unsolved(ResolvedReferenceTypeDeclaration.class);
        }
    }
```

Does anyone agree?
And if so, could anybody do this change? I'm sorry, but I don't know how to work with pull requests etc.

MysterAitch:
Fixes #1526.


MysterAitch:
Seems Travis failed due to a network issue, as opposed to a test failure!

Is there an easy way to re-trigger a build other than pushing again?

> [ERROR] Failed to execute goal biz.aQute.bnd:bnd-maven-plugin:4.1.0:bnd-process (default) on project javaparser-core: Execution default of goal biz.aQute.bnd:bnd-maven-plugin:4.1.0:bnd-process failed: Plugin biz.aQute.bnd:bnd-maven-plugin:4.1.0 or one of its dependencies could not be resolved: Could not transfer artifact biz.aQute.bnd:biz.aQute.bndlib:jar:4.1.0 from/to google-maven-central (https://maven-central.storage-download.googleapis.com/maven2/): GET request of: biz/aQute/bnd/biz.aQute.bndlib/4.1.0/biz.aQute.bndlib-4.1.0.jar from google-maven-central failed: Connection reset -> [Help 1]

matozoid:
I normally click in the green-checks-list to find the details of the travis build. On travis itself there is a rebuild button.

PR #2492, opened 2020-01-21
#1370 -- Solving generic types that are of type java.lang.Object

### LINKED ISSUE #1370: Solving generic types that are of type java.lang.Object
Sorry for the non-descriptive title, but I didn't come up with anything better...

Look at this code:

```java
final String source = String.join(System.lineSeparator(),
        "package graph;",
        "class Vertex<Data> {",
        "    private final Data data;",
        "    public Vertex(Data data) { this.data = data; }",
        "    public Data getData() { return this.data; }",
        "}",
        "",
        "public class Application {",
        "    public static void main(String[] args) {",
        "        System.out.println(new Vertex<>(42).getData().equals(42));",
        "    }",
        "}");

final JavaParserFacade facade = JavaParserFacade.get(new ReflectionTypeSolver(false));

JavaParser.parse(source).accept(new VoidVisitorAdapter<Void>() {
    @Override
    public void visit(final MethodCallExpr n, final Void arg) {
        super.visit(n, arg);

        try {
            System.out.printf("Node: %s, solved Type: %s%n", n, facade.solve(n));
        } catch (RuntimeException e) {
            e.printStackTrace();
        }
    }
}, null);
```

> Method 'equals' cannot be resolved in context new Vertex<>(42).getData().equals(42) (line: 10) MethodCallExprContext{wrapped=new Vertex<>(42).getData().equals(42)}. Parameter types: [PrimitiveTypeUsage{name='int'}]

How can I solve the chained `equals()` call? Doing this works:
```java
Integer d = new Vertex<>(42).getData();
System.out.println(d.equals(42));
```

But what about more complex cases, like this:

```java
public class Application<Data> {
    public static void main(String[] args, Data data) {
        System.out.println(data.equals(42));
    }
}
```

> Method 'equals' cannot be resolved in context data.equals(42) (line: 10) MethodCallExprContext{wrapped=data.equals(42)}. Parameter types: [PrimitiveTypeUsage{name='int'}]

Thanks!

MysterAitch:
Testcase that verifies the reported issue is no longer present / tests pass.

Closes #1370

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1370Test.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1370Test.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2015-2016 Federico Tomassetti
+ * Copyright (C) 2017-2020 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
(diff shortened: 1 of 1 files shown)


matozoid:
Cool!

PR #2494, opened 2020-01-22
fix for canBeAssignedTo bad result bug

### LINKED ISSUE #2387: JavaParserClassDeclaration canBeAssignedTo gives wrong result when reaches Object as super type
If I configure a `ReflectionTypeSolver` and try to determine wether a parsed class extends an other class with the `JavaParserClassDeclaration.canBeAssignedTo()` method, and the parsed java class doesn't extend any other class this method will still report as a match.

I've done some debugging and found this part of the code, which I think causes the wrong results:
```
        ResolvedClassDeclaration superclass = (ResolvedClassDeclaration) getSuperClass().getTypeDeclaration();
        if (superclass != null) {
            if (Object.class.getCanonicalName().equals(superclass.getQualifiedName())) {
                return true;
            }
            if (superclass.canBeAssignedTo(other)) {
                return true;
            }
        }
```

It checks if the superclass is the `Object` but it doesn't check if the other class is also the `Object` class, which means it will return true for every class if the `java.lang.Object` can be resolved by a `TypeSolver`.

Am I guessing right, or did I miss something?

PeterCsubak:
Fixes #2387.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -243,7 +243,8 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration impleme
         ResolvedClassDeclaration superclass = (ResolvedClassDeclaration) getSuperClass().getTypeDeclaration();
         if (superclass != null) {
             // We want to avoid infinite recursion in case of Object having Object as ancestor
-            if (Object.class.getCanonicalName().equals(superclass.getQualifiedName())) {
+            String objectClassName = Object.class.getCanonicalName();
+            if (objectClassName.equals(superclass.getQualifiedName()) && objectClassName.equals(other.getQualifiedName())) {
                 return true;
             }
             if (superclass.canBeAssignedTo(other)) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -242,11 +242,6 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration impleme
         }
         ResolvedClassDeclaration superclass = (ResolvedClassDeclaration) getSuperClass().getTypeDeclaration();
         if (superclass != null) {
-            // We want to avoid infinite recursion in case of Object having Object as ancestor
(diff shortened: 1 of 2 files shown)


MysterAitch:
It seems we've stumbled into a separate bug! :smiley: 

While this edit does fix the bug in #2387, removing the `if(){}` block also seems to fix it too because this condition is also tested within `canBeAssignedTo` too -- note specifically line 233 below. 

https://github.com/javaparser/javaparser/blob/7d77dc58d62e5b7f4a9f76e9b0e810f557391c04/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionClassDeclaration.java#L228-L247

Reading through the pre-existing comment, it seems that what the `if(){}` is supposed to be doing is detecting an infinite loop in the class ancestry (which would turn the ancestry tree from a "tree" into a graph). Unfortunately the implementation of this is broken, leading to the issue you describe in #2387 !

What pre-existing comment suggests is that the intention is to only check the case where "this class" is equal to the "super class" (where object is a superclass of object). 

Would you like to have a go at editing this pr to sort that other issue out? :smile:

PeterCsubak:
@MysterAitch I've added the checks for other cases you mentioned in the issue, and added one more where the parameter is the `Serializable` interface, just to check if it works in the case where the class doesn't implement the specified interface. 

Also I've removed the `if` statement completely, since as you pointed out, the check will be done in the `ReflectionClassDeclaration` implementation. I do not see the point of checking the superClass being the same class as the checked class, because that isn't really possible, is it? 
I mean the Object class will most likely be resolved by the `ReflectionTypeSolver`, which will check the super class here:
```java
     if (this.clazz.getSuperclass() != null 
             && new ReflectionClassDeclaration(clazz.getSuperclass(), typeSolver).canBeAssignedTo(other)) { 
```
And the `Object.class.getSuperclass()` returns null, so there isn't any infinite recursion in the `ReflectionClassDeclaration`. 

The only way I can imagine this could cause problems, if the Object isn't resolved by the `ReflectionTypeSolver`, but how could that be? 
`JarTypeSolver` could be used to load `rt.jar` but I guess javassist wouldn't like that. :) or maybe the Object class could be parsed with the parser and resolved that way, but I'm not sure if that even would work. 
If I would use the type solver without the reflection solver, and without the Object class resolvable some other way, the `object()` method would fail in the `JavaParserClassDeclaration`, with `UnsolvedSymbolException`.

So I'm a bit lost. Could you give me an example where the infinite recursion can cause problems?

MysterAitch:
> So I'm a bit lost. Could you give me an example where the infinite recursion can cause problems?

Honestly, no not really :p

I can only presume that it was added for a reason, but I'm on the move away from a computer now so will struggle to hunt back through the logs to find out more.

I do have a vague recollection of discussion about having duplicate object classes - possibly when including both the reflection and also a jar (maybe android?). 

PR looks good to me (thanks for the additions!) -- my thoughts are that this is good for closing the issue, and a new PR / issue can be created for the potential infinite loop.

PR #2279, opened 2019-06-24
[DRAFT] Initial changes made to get (most) of the package rename working 

edefazio:
I wanted to submit this separate from the package name changes because basically
the other changes change every single .java file 

since the directory structure would change

It's hard to do any diffing to determine where the problems are and what manual changes were made

For simplicity/ brevity you can 
do a text search for "** MED" to find any annotations about what my thoughts are
& to find any issues or comments or stacktraces that are occurring to the code 
(AFTER I do the package name change) 

This should clean install fine (as is)
For brevity, here are the (3) Test Classes with (14) individual tests that fail AFTER the package name change:
[ERROR]   JavassistClassDeclarationTest.testGetAllInterfaces:276 » UnsolvedSymbol Unsolv...
[ERROR]   JavassistClassDeclarationTest.testGetAllInterfacesWithoutParameters:369 » UnsolvedSymbol
[ERROR]   JavassistClassDeclarationTest.testGetInterfaces:263 » UnsolvedSymbol Unsolved ...
[ERROR]   JavassistClassDeclarationTest.testGetInterfacesWithoutParameters:325 » UnsolvedSymbol
[ERROR]   ContextTest.resolveCascadeOfReferencesToMethod:259 » UnsolvedSymbol Unsolved s...
[ERROR]   ContextTest.resolveGenericReturnTypeOfMethodInJar:355 » UnsolvedSymbol Unsolve...
[ERROR]   ContextTest.resolveLambdaType:544 » UnsolvedSymbol Unsolved symbol in cu.getTy...
[ERROR]   ContextTest.resolveReferenceToCallOnLambdaParam:586 » Runtime Error calculatin...
[ERROR]   ContextTest.resolveReferenceToLambdaParam:561 » Runtime Error calculating the ...
[ERROR]   ContextTest.resolveReferenceToMethod:230 » UnsolvedSymbol Unsolved symbol in c...
[ERROR]   ContextTest.resolveTypeUsageOfCascadeMethodInGenericClass:521 » Runtime Error ...
[ERROR]   ContextTest.resolveTypeUsageOfFirstMethodInGenericClass:477 » UnsolvedSymbol U...
[ERROR]   ContextTest.resolveTypeUsageOfMethodInGenericClass:499 » UnsolvedSymbol Unsolv...
[ERROR]   MethodsResolutionWithJavassistTest.testOverloadedMethods:61 » Runtime Error ca...

 I also tested including the slow tests, and the following (additional) failure:
[...]

Issue #2249, opened 2019-05-27
Change package names part 2


matozoid:
Move to org.javaparser.

matozoid:
@edefazio is picking this one up.

matozoid:
@ftomassetti - we'll need new release rights on Sontaype then! Can you organize that?

ftomassetti:
Ok, I will look into that

PR #2497, opened 2020-01-23
[discussion] Follow up of #2279 -- change packaging

MysterAitch:
Follow-up from #2279 

I'm not sure if there's ANY appetite for this at all or what the motivation for the original PR was, but I stumbled across the old commit #2279 and wanted to see what it would take to make such a change... can't be THAT hard, right??

In fairness it wasn't particularly painful -- the main "hiccup" is the test cases which depend on a copy of the source code using the current packages so I had to go back and "undo" those. fwiw that commit seems to be a decent resource to highlight where this is the case..? 

The other "issue" that came up is a few tests which have content in the resources file -- it took a small while to realise that the tests fail not because they're doing some odd reflection or something, but because the text files had jp imports and I had been somewhat overzealous in migrating those across. Details can be seen in the second commit.

Sidenote: Maybe there's a way to separate out tests which rely on a supplied copies of source code?


matozoid:
Calling @edefazio :-)

edefazio:
"can't be THAT hard, right??"
---Those were my famous last words before I wandered into the swamp

I would have wanted to "categorize" the failing tests (like 9 or so) that were just too hard for my silly little monkey brain to fix... 

I think most of the changes I made were REALLY easy and straight forward, I tried to document what I was doing, I think i got javaparser-core done in like an hour and a half.
 
...But (in the end of the day) the juice wasnt worth the squeeze,... meaning it's kinda pointless to adopt the changes unless you are going to change the package structure, so (for something like tests) it's kinda like "why bother"  ... and I hate saying this because I spent probably 30 hours or so trying to make sense of things... 
(Nothing I changed made the tests "better" or anything, but it was just grunt work to try to be the next step in allowing the package rename to occur)

(Keep in mind the entirety of the JSS vexes me, so, perhaps you'll have better luck)

MysterAitch:
Link: #2249

MysterAitch:
Closing this PR -- this is useful as a reference, but would pretty much need to be re-done from scratch to be mergable.

PR #2500, opened 2020-01-23
Introduce `@OpenIssueTest` to annotate tests that are not (yet) expected to pass / should normally be ignored

MysterAitch:
Following the chat on gitter, here's an implementation of using tags to add to test cases that are not expected to pass yet (thus should be excluded by default).

This way, the testcases currently added as pull requests to the project can be merged in and evaluated in-bulk.

Note that the pom has been updated to exclude these tests by default (using a profile) and this gets taken into account when running tests via maven, but I'm not sure about how well supported this is re: running tests directly within the ide (it can be done via manually editing run configs, but may be a bit startling to see!)


MysterAitch:
Auto tests pass fine, but I'll leave it as a draft for now for discussion re: if this is the desired approach to take -- I'm certainly open to alternatives!

I did see that it's possible to disable tests, but I've not seen how to disable them by default (perhaps include `@Disabled` in the annotation interface `OpenIssueTest`, then some kind of logic to override this and actually run the test given a condition...?). 

This would be beneficial to anyone running tests via an ide as by default they seem to appear as failing tests.

matozoid:
Haven't looked carefully, but I can say the slow tests are disabled by default using configuration in 
pom.xml

MysterAitch:
> Haven't looked carefully, but I can say the slow tests are disabled by default using configuration in
> pom.xml

.. Which is fine if running via maven goal, less fine if running tests via IDE :)

Because the maven tests take so long to complete, during development often I'll just run/rerun a directory of tests. If that directory happens to contain an "expected to be failing" test then at best this is "noise", and at worst it's derailing/confusing.

matozoid:
I don't understand - the point is not to run the slow and the "unsolved" tests by default, right? And if I'm not mistaken you can enable these from the maven command line, and from the IDE with a few clicks. What am I missing?

MysterAitch:
> What am I missing?

Probably nothing substantive! :D

My workflow normally involves triggering tests either via the context menu or keyboard shortcut. By default, these appear to trigger _ALL_ tests within the context selected (directory/file/single method). 

It is reasonably straightforward to setup a run config that includes/excludes tags. The catch seems to be that it means running tests via specifc run configs (selected from a run-config menu) as opposed to a context menu or keyboard shortcut -- which I'm usually too lazy to do ;) . 

---

In terms of impact, it's probably limited as it will only ever affect anybody running the project source code (and anybody doing this will either see the annotation/documentation, or can have it explained) -- meaning that it's likely not something that should hold this back!

Perhaps `@OpenIssueTest_FailExpected` would be a more explicit annotation name?

Assuming you're happy with this "proof of concept", this can probably be progressed past "draft" status and perhaps merged in.

---

image


image
 
image

.travis.yml
```diff
@@ -1,7 +1,7 @@
 sudo: false
 language: java
 install: mvn -B install -DskipTests=true
-script: mvn -B test -P AlsoSlowTests
+script: mvn -B test -P AlsoSlowTests,ExcludeOpenIssueTests
```
matozoid, line 4: Wait, does that mean the default run *includes* the open issue tests? So builds fail by default?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue2361Test.java
```diff
@@ -0,0 +1,93 @@
...
+ * </ul>
+ *
+ * @see <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2">https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2</a>
+ * @see <a href="https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.25">https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.25</a>
+ * @see <a href="https://github.com/javaparser/javaparser/issues/2361">https://github.com/javaparser/javaparser/issues/2361</a>
+ */
+public class Issue2361Test {
+
+
+    @OpenIssueTest(issueNumber = {2361, 2435})
```
matozoid: The issue numbers are really neat :-)

.travis.yml
MysterAitch: I'm not sure how best to answer as I'm not 100% sure of how maven/travis etc handle things, but I can say that the profile which skips these is enabled by default.

My best understanding at the moment is that we either need to explicitly pass all profiles, or none of them.

(meaning that `script: mvn -B test` and `script: mvn -B test -P AlsoSlowTests,ExcludeOpenIssueTests` are effectively equivalent)

I might be completely off-base with this though!

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue2361Test.java
MysterAitch: It would be nice to be able to use them in some way.

- Maybe a "fun project idea" to search for all these annotations and use the github api to find out if the issue is open/closed..? 

- Or maybe an open issue test is now green, post to gitter or the pr/issue itself..?

- Or maybe just add a javadoc comment with a url link to the github issue that can be copy/pasted..?


For now though, it can just be manually edited into the url or searched for via the web ui ;)

.travis.yml
matozoid: Alright, well, I think it would have been easier to name both positively: "AlsoSlowTests" and "AlsoOpenIssueTests". It's a bit confusing like this :-) If you agree and have time, do change, otherwise it gets merged like this in a few hours.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue2361Test.java
matozoid: It is a fun project idea, but the label "fun project idea" is for stuff that is a good idea, but is not going to be part of JP itself :-)

.travis.yml
MysterAitch: I'm not sure how to make the logic of this happen as I don't entirely understand how specifying `AlsoSlowTests` allows them to run.

I _think_ what happens is `activeByDefault` applies only when no profile is specified, and is ignored when one or more profiles are applied. 

This would mean that `AlsoSlowTests` should be renamed to `RunAllTests` or `RunAllTests_IncludingThoseExpectedToFail` or similar _IF_ this assumption about how maven works is true.

If renamed this way, it makes more sense why we need to specify <s>`AlsoSlowTests`</s> `RunAllTests` and `ExceptOpenIssueTests` (the full set, followed set subtraction to exclude some).


```xml
        <profile>
            <id>NonSlowTests</id>
            <activation>
                <activeByDefault>true</activeByDefault>
            </activation>
            <build>
                <plugins>
                    <plugin>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <configuration>
                            <excludedGroups>com.github.javaparser.SlowTest</excludedGroups>
                            <parallel>methods</parallel>
                            <threadCount>4</threadCount>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>

        <profile>
            <id>AlsoSlowTests</id>
        </profile>



        <profile>
            <id>ExcludeOpenIssueTests</id>
            <activation>
                <activeByDefault>true</activeByDefault>
            </activation>
            <build>
                <plugins>
                    <plugin>
                        <artifactId>maven-surefire-plugin</artifactId>
                        <configuration>
                            <excludedGroups>com.github.javaparser.OpenIssueTest</excludedGroups>
                            <parallel>methods</parallel>
                            <threadCount>4</threadCount>
                        </configuration>
                    </plugin>
                </plugins>
            </build>
        </profile>
```

Issue #1392, opened 2018-02-03
Turn "consider annotations" flag into a "logical attribution" flag


matozoid:
Instead of saying "never attribute comments to annotations," let's make it do what its real intention is: "put comments on the main thing that is following the comment, like a method, a class, ..."

PR #2502, opened 2020-01-24
Issue #398 - Inspect TODO < or <= 

### LINKED ISSUE #398: Inspect TODO < or <= ?

MysterAitch:
Closes #398 (related: #1390 )

I think it's ready to go, but I'm leaving it as a draft for now mostly because I'm not 100% confident in it. 

Please pick it apart / add new tests etc!


javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
```diff
@@ -63,7 +61,7 @@ void issue624() throws IOException {
 
     @Test
     void issue200EnumConstantsWithCommentsForceVerticalAlignment() {
-        CompilationUnit cu = parse("public enum X {" + EOL +
+        CompilationUnit cu = StaticJavaParser.parse("public enum X {" + EOL +
```
matozoid: There is actually a `TestParser` or so in the project, which is meant to have the best possible configuration for tests.

matozoid:
Looks good to me, especially because other tests aren't failing :-)

matozoid:
Is there anything in particular you're doubting?

javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
MysterAitch: Didn't realise this, cheers! 
Just pushed a change to use it :)

MysterAitch:
> 
> 
> Is there anything in particular you're doubting?

Mostly just that I'm still not 100% sure what the bits of code are "supposed" to be doing :D 

I think I noticed that sometimes the comments are being added to the `SimpleName` (as opposed to the `ClassOrInterfaceDeclaration`) and a few other oddities, but I didn't sit with it long enough to figure out exactly what is "supposed" to be happening (nevermind compare it against what's _actually_ happening!!)

---

image

---

image

---

Similarly, while I **THINK** I got the renaming of `firstNonAnnotationNode` correct (from `beginNodeWithoutConsideringAnnotations`) including the modifications to take into account annotations, it might be that I fudged things somehow? I don't have any reason to believe this is the case, but equally I would need to spend some additional time to be confident.

This is perhaps the topic of another issue/PR and shouldn't hold this one up though..? I'm not sure and won't have much time this weekend to go into detail on this.

MysterAitch:
I think a "simple" question is re: the intention here...

Is `nodeContains` supposed to return true when the node is _"contained"_ with respect to the source code or the (concrete?) syntax tree?

_(The code I submitted assumes the former, but then it feels strange to be talking about if a "Node" contains text, as opposed to a "NodeRange" or similar..)_

matozoid:
Since you're in unknown land (I don't know either, maybe @ftomassetti remembers but this is from long ago!) - just try to make things better in general. Add documentation, fix stuff that is probably wrong, make decisions about how it should work... It sound like comments shouldn't be attached to `SimpleName`s for example (well maybe in case of `int /*z*/ x;`) so you can change that and see what happens.

MysterAitch:
> 
> 
> Since you're in unknown land (I don't know either, maybe @ftomassetti remembers but this is from long ago!) - just try to make things better in general. Add documentation, fix stuff that is probably wrong, make decisions about how it should work... It sound like comments shouldn't be attached to `SimpleName`s for example (well maybe in case of `int /*z*/ x;`) so you can change that and see what happens.

Makes sense, cheers!

I've just pushed some additional tests / comments which clarify where some of the "oddities" occur (d823efc). 

It seems that for the purpose of solving #398 this is probably complete, but it seems to be a "can of worms" just underneath the surface that require additional issues to be opened :P

matozoid:
Please write a line in the changelog.md like:
BREAKING: what users can expect. why this happens. what they should do.

A nice follow-up for this would be to move the focus from "looking at a node to find its comments" to "looking at a comment in CommentsCollection to find the nodes it relates to." That requires mixing up CommentsCollection and CommentsInserter.

MysterAitch:
> Please write a line in the changelog.md like:
> BREAKING: what users can expect. why this happens. what they should do.

I don't _**think**_ any behaviours have changed so far (the method name change is private).

Edit: scratch that -- just remembered that it changes slightly where a comment within the method modifiers is now considered as being "in" the method.


> A nice follow-up for this would be to move the focus from "looking at a node to find its comments" to "looking at a comment in CommentsCollection to find the nodes it relates to." That requires mixing up CommentsCollection and CommentsInserter.

I'm not 100% following this but I think I agree. Personally I need a bit more time to figure out in my own head where I'm expecting comments to "apply", and which nodes are relevant.

For example, where a javadoc comment appears between two annotations of a method, I see that applying to the method declaration which includes the annotations/modifiers/body.

Currently, however, if I select the modifier `public` or the return value `void` and query for the javadoc that applies to it, it is empty because it is not the specific node that applies.

That gets a bit trickier if wanting to go "up" the tree though -- should this also extend to javadoc of the class that a method is within? I'm inclined to say yes that it is potentially relevant, but then we're moving towards a model of "which javadoc comments apply to this node or it's context?" and "which javadoc comments apply specifically to this node?".

Or the reverse -- which nodes does this javadoc apply? Arguably the class javadoc is relevant to all members of the class (and sub-classes/interfaces etc?). In this case we're saying it applies to the sub-tree of the ast perhaps? Though in some cases we're only interested in the head of that sub-tree..?


... it's an interesting topic and I suspect I've already gotten a little carried away with it here already!

matozoid:
> > Please write a line in the changelog.md like:
> > BREAKING: what users can expect. why this happens. what they should do.
> 
> I don't _**think**_ any behaviours have changed so far (the method name change is private).
> 
> Edit: scratch that -- just remembered that it changes slightly where a comment within the method modifiers is now considered as being "in" the method.

Well, if it is more of a bug fix, no explanation is necessary I think. I just imagine users being incredibly happy that something is working better than before :-)

> ... it's an interesting topic and I suspect I've already gotten a little carried away with it here already!

Well, it seems you are now the expert in comment attribution too ;-)

matozoid:
Oops, conflicts!

MysterAitch:
sorted out the conflict...

not sure if there's already an issue for this or not, but the bits that remain are more geared towards "given this input, which node should it apply to?" 

e.g. similar to #1392 which talks about comments being attached to the node being annotated, rather than the annotation itself

given this, I'm happy to merge this pr and have any follow ups done in a new issue/pr

PR #2504, opened 2020-01-26
#2501 - Updated maven version in appveyor script

### LINKED ISSUE #2501: Appveyor / maven 3.2.5?
The appveyor script installs maven v3.2.5 (released in 2014)

I'm not aware of any particular reason/need to upgrade, but 2014 seems to be extremely old!

_(I just noticed that this is the last version that has a minimum java 6, and the next [version 3.3.1 has a minimum of java version 7](https://maven.apache.org/docs/history.html) --- is this the rationale for not updating?)_

https://github.com/javaparser/javaparser/blob/d1ee0363f4f1e0e93c7d34510f48194a899405ce/appveyor.yml#L5-L15

MysterAitch:
Closes #2501

I noticed that in #100 the minimum build version is JDK8 and in appveyor.yml it specifies JDK10 as the version it actually runs during the test.

Given this, (presumably!) it should be fine to bump the maven version that gets run to the latest available (this is v3.6.3 released 2019-11-25 according to the [maven site](https://maven.apache.org/docs/history.html))

Let's see what happens when it triggers! :)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,15 +7,15 @@ install:
       Add-Type -AssemblyName System.IO.Compression.FileSystem
       if (!(Test-Path -Path "C:\maven" )) {
         (new-object System.Net.WebClient).DownloadFile(
-          'http://www.us.apache.org/dist/maven/maven-3/3.2.5/binaries/apache-maven-3.2.5-bin.zip',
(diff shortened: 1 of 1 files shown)


matozoid:
Thanks!

PR #2505, opened 2020-01-26
minor bugfix -- now checking presence of both `getBegin()` _and_ `getEnd()`

MysterAitch:
minor bugfix -- now checking presence of both `getBegin()` _and_ `getEnd()` (as opposed to checking the presence of `getBegin()` twice).

Sidenote: would it be better to throw an exception rather than fail silently?


matozoid:
Phew, asking questions with answers lost to time. I vaguely remember that the whole nullability of ranges and tokenranges is a bit awkward.

PR #2516, opened 2020-02-07
Fixes issue with setLexicalPreservationEnabled=true where the Phantom…

### LINKED ISSUE #2515: isPhantomNodeCache is never cleaned up. Let's add a way to clean it up manually
Context: This comes from lexical preservation: [https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java](https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java)

This cache seems to be used to preserve the lexical grammar. It is not used if setLexicalPreservationEnabled is set to false inside the ParserConfiguration object.

Previous conversation on Glitter chat:

@myself : Hello, isPhantomNodeCache seems to be growing indefinitely. What would be a good way to clean it up? I'm iterating over a list of files that I might or might not want to modify. They are all independent of each other. I'd like to cleanup isPhantomNodeCache instead of letting it grow linearly.
[...]
@matozoid : I guess someone will need to look into that. Can you open an issue?

I will follow up with a diff allowing manual cleanup.

quanturium:
…NodeCache is never cleared. This adds a way to clean it up manually if needed

Fixes #2515.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/PhantomNodeLogic.java
@@ -34,7 +34,7 @@ import static java.util.Collections.synchronizedMap;
 /**
  * We want to recognize and ignore "phantom" nodes, like the fake type of variable in FieldDeclaration
  */
-class PhantomNodeLogic {
+public class PhantomNodeLogic {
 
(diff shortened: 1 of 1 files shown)


matozoid:
That's pretty straightforward, so let's merge it. Thanks!

PR #2518, opened 2020-02-09
JavaParserEnumDeclaration can have access specifier

iTakeshi:
According to [JLS-8.9](https://docs.oracle.com/javase/specs/jls/se11/html/jls-8.html#jls-8.9), an enum declaration can have access modifiers.
This PR fixes `JavaParserEnumDeclaration.accessSpecifier()` to return a `AccessSpecifier` constant, which currently throws `UnsupportedOperation`.

I'm not sure about the place where I put the test case is good or not. Please tell me if there is elsewhere appropriate.


matozoid:
> I'm not sure about the place where I put the test case is good or not. Please tell me if there is elsewhere appropriate.

Nobody really knows. I can say that if you think it should be in a new test file, create one.

iTakeshi:
> if you think it should be in a new test file, create one.

Understood, thanks.
For this time I think it's okay as it currently is.

matozoid:
In it goes, thanks!

PR #2519, opened 2020-02-09
Annotations have only one ancestor

iTakeshi:
According to [JLS-9.6](https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-9.6), an annotation has only one direct super interface `java.lang.annotation.Annotation`.
This PR fixes `JavaParserAnnotationDeclaration.getAncestors()` to return a list containing one element, which currently throws an `UnsupportedOperation`.


matozoid:
Sounds sensible to me. Thanks!

PR #2520, opened 2020-02-10
solve annotation element types in AnnotationDeclarationContext

iTakeshi:
Currently `JavaParserAnnotationMemberDeclaration.getType()` throws an `UnsupportedOperationException`, but it should return the corresponding type for the member.
Also, its type should be resolved in the enclosing `@interface`, which could have nested classes/interfaces/enums, thus I created `javaparsermodel.contexts.AnnotationDeclarationContext`.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AnnotationDeclarationContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AnnotationDeclarationContext.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (C) 2015-2016 Federico Tomassetti
+ * Copyright (C) 2017-2020 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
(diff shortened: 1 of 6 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AnnotationDeclarationContext.java
```diff
@@ -0,0 +1,78 @@
...
+import com.github.javaparser.resolution.declarations.ResolvedValueDeclaration;
+import com.github.javaparser.resolution.types.ResolvedType;
+import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserAnnotationDeclaration;
+import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
+import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
+
+import java.util.List;
+
+/**
+ * @author Federico Tomassetti
```
MysterAitch: Please double check the `@author` annotation for this new file :)

iTakeshi:
@MysterAitch Thank you very much for your review, and I'm sorry for a couple of my careless misses.
Fixed the author name and rebased to the latest master, then force-pushed (because I don't want to pollute log).

matozoid:
Thank you both! I'll merge this.

PR #2521, opened 2020-02-11
Symbolsolver import order

iTakeshi:
This PR solves two problems.

### 1: 896e8da
A single-type-import ([JLS 7.5.1](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.5.1)) and a single-static-import([JLS 7.5.3](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.5.3)) can import same name into a compilation unit, when the static-imported name doesn't refer a type (e.g., enum member, etc.).
In such cases, those static imports can just be ignored and other imports should be evaluated.
(This commit also includes minor refactoring around nested `if`)

### 2: 2e71583
According to [JLS 7.5.2](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.5.2) and [JLS 6.4.1](https://docs.oracle.com/javase/specs/jls/se11/html/jls-6.html#jls-6.4.1), asterisk imports should be evaluated after package members.


iTakeshi:
The travis build failed seemingly because of a network connection error. Could you trigger to rebuild?  (i.e., rebuild request needs a write-access to the repo)
Thanks for your help! (Cc: maybe @matozoid )

iTakeshi:
Thanks @MysterAitch for your review.
(I'm sorry that your comment got invisible due to my force push... I didn't know that.)

As you commented, the assertion messages were not descriptive enough for the test cases.
I think it was because of the class name `Name` is highly confusing, thus I renamed those classes to just class `A`.
Also I revised the assertion messages based on your suggestion.

iTakeshi:
@matozoid @MysterAitch 
Also a management question: On revising PRs, should I just add new commits on top of original PR commits, or should I force-push them just like I did for this PR?
Personally I don't like the commit log gets dirty because of those revising commits, thus I prefer force-pushing the revised commits.

If you could tell me your preference for this project, I will follow the guideline.

matozoid:
@iTakeshi please no force pushing if it confuses github.

matozoid:
... and thanks for the PR! :-)

iTakeshi:
> please no force pushing if it confuses github.

Understood, I will add commits except for when there are no side effects. Thanks for the guideline.

PR #2522, opened 2020-02-11
solve implemented/extended types with scope

iTakeshi:
An interface or a class can implement/extend another interface/class using their "scoped" name, but in such case, the implemented/extended one can be referred by its "simple" name, e.g.
```java
class A { class B {} }
class Foo extends A.B {
    B field_B;
}
```
To resolve the type `B` for `field_B`, `ClassOrInterfaceDeclarationContext` should search the type with regard to its "scoped" name, which currently only uses the "simple" name.


matozoid:
Thanks for all this work - I'm trying to find someone to review these PR's :-)

iTakeshi:
No need to rush :D
Thanks for this great OSS!

MysterAitch:
In the situation where a type isn't being found, either the query is wrong or the search results are wrong (or some odd combination of both!).

This solution fixes the query (changing the input argument to `.solveType()`), while I'm wondering if we should instead be fixing the search results (fixing `solveType()` itself).

I've been digging into this a little during my lunch break so I've not had time to come to any meaningful conclusion on this yet, but I note this down now while I think about it and would definitely appreciate some additional inputs here to help sway one way or the other!

MysterAitch:
Just to expand on this as I wasn't able to put it down! :D ... 

Note in the screenshot below that for the `wrappedNode` of `CompilationUnit`, `wrappedNode.getTypes()` returns only the types that are immediately within the cu itself (note that none of the nested classes/interfaces are included).

For this reason, the name `C2_1` is never matched (thus never solved). 

So this raises the question: should `getTypes` also return the types (classes/interfaces/enums etc.) found within the children of the compilation unit? Should those be qualified etc? Should it be filtered to only include "visible" types (public/package-private)? 

_(as above, I'm not yet sold one way or another)_

image

image

iTakeshi:
@MysterAitch Thank you for reviewing.

I think the current behavior of `CompilationUnitContext` that
> returns only the types that are immediately within the cu itself

is correct, and I think resolving nested types is NOT a responsibility of a `cu`.
If the target name is qualified (i.e. contains a dot like `C2.C2_1`), `cu` delegates the type resolution to the outermost class declaration `C2` (see https://github.com/javaparser/javaparser/blob/3e4ca2c233fe50b3c14160fecb5a6ab9e49a633c/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java#L141-L162)

Thus, in my opinion, the target name (query) should be appropriately qualified.
Then, `ClassOrInterfaceDeclarationContext` (or `JavaParserTypeDeclarationAdapter` to whom it delegates the resolution) can handle the query.

(In my understanding, the `JavaParserTypeDeclarationAdapter` solves types with regard to the access modifiers, but I'm not completely sure if it can interpret the visibility under any circumstances. Probably there are some cases in which some types are resolved to types which should be invisible)

iTakeshi:
I just found another edge case...
When a class implements/extends a type with the same simple name, the class should be prioritized to the implemented/extended type.
```java
package main;
class A implements another.A {
  A field_a;  // <= this should refer main.A, but not another.A
}
```

I added a commit to solve this issue, but I'd like you to review if it's ok to have redundant statements among `ClassOrInterfaceDeclarationContext` and `JavaParserTypeDeclarationAdapter`...
https://github.com/javaparser/javaparser/pull/2522/files#diff-35fec71f6603467347504bdd005702cbR98-R101
https://github.com/javaparser/javaparser/blob/3e4ca2c233fe50b3c14160fecb5a6ab9e49a633c/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java#L64-L66

iTakeshi:
I found yet another edge case:
```java
package main;
import yet.another.A;
class B implements another.A {
  A field_a;
}
```
In the above case, the type of `field_A` should be `yet.another.A`, but not `another.A` (that is, imported types are prioritized to implemented/extended types)

Now I start thinking that fixing the search algorithm instead of the query might be the better option.
I'll do another couple of experiments this weekend to figure it out.
Until then, I mark this PR as [NOT READY].

iTakeshi:
Digging into the JLS, but so far I couldn't find any reference for this behavior...
This is apparently related to #2195 / #2211, but do you guys know any resources to help this situation?

matozoid:
My references are mostly:
- the JLS
- what does the compiler or intellij do?
- the JDK test suite

MysterAitch:
Hi @iTakeshi -- reading through, I'm not sure where we're at with this. 

Before I dig back into this (my memory is rather awful!), could you share some thoughts about what is working and what is left to do please?

Cheers!

iTakeshi:
I'm really sorry, but I couldn't do much work about this because I've been completely occupied by another project...
To be honest, I don't even know how the latest version of javaparser behaves in those cases.
I'll try to get back to you until next weekend.
Thanks for your patience!

MysterAitch:
No worries at all! 

Just having a sweep through issues/PRs at the moment and I'm just wondering if this PR is in a place where it can be merged (or tweaked and then merged) as an incremental improvement, and the areas that aren't all there yet can be dealt with separately.

My head's not fully in it at the moment though and would need to spend some time understanding what's going on before I know for sure! 

Hope you are staying safe and well :)

iTakeshi:
@MysterAitch I just confirmed the current status.

Facts:
1. The test cases added in this PR
    - do not pass without the modification in this PR, and
    - do pass with the modification in this PR.
2. The problem I mentioned in [this comment](https://github.com/javaparser/javaparser/pull/2522#issuecomment-586313579) remains unsolved.

Because of the fact#1, at least this PR can benefit as `an incremental improvement` like you said.
For the fact#2, I can submit a "testcase" PR once (if) this PR is merged.

One thing I'm now concerned about is [your comment](https://github.com/javaparser/javaparser/pull/2522#issuecomment-584717764) regarding the library design, that is
> This solution fixes the query (changing the input argument to .solveType()), while I'm wondering if we should instead be fixing the search results (fixing solveType() itself).

To be honest, I can't tell which is the "right" way to go, but now I'm certainly sure that the latter way (fixing `solveType()` itself) would be much easier than the former one.

Hope this helps you.

PR #2526, opened 2020-02-25
Issue #1574 - Change getComments

### LINKED ISSUE #1574: JavaParser doesn't remove comments before package declaration
I posted this on Stackoverflow, and a maintainer of Javaparser told me to post an issue here.
With his help, I managed to solve my problem with another solution, but I think that this problem should be solved.
Here is my post:

I am writing a Java application that removes comments from Java files.

I wrote this code:

```
static void removeComments(Node node) {
        for (Comment child : node.getAllContainedComments()) {
            child.remove();
        }
    }
```

CompilationUnit cu = JavaParser.parse(projectDir); // projectDir is received from parameter
removeComments(cu);
It kind of works. However, if there are block comments before the package declarations, the code won't remove those comments.

For example, this code:

```
/*
 * Block comment
 */
// Line comment
package test;

/*
 * Block comment
 */
// comment
/**
 *
 * @author Me
 */
public class Test {

        // TODO code application logic here
        System.out.println("print 1"); // comment in code line
        // comment 2
        System.out.println("print 2");
        /*
        block comment
         */
    }

}
```
Becomes this code:

```
/*
 * Block comment
 */
package test;

public class Test {

        System.out.println("print 1");
        System.out.println("print 2");
    }
}
```
Is it a bug from JavaParser, or am I missing something?

**Edit:**

If I put a line comment at the beginning of the file, the first block comment (the one before the package declaration) will be removed, but not this line comment. I think JavaParser doesn't consider that the first line of a file may be a comment.

Mathiponds:
Fixes #1574.
The `getComments` method in the `CompilationUnit` was not doing the right thing. It was not returning the comment associated to the compilation unit. I changed that and I added some tests for the functions that get comments.
As discussed with Matozoid, getComments is not a good name and getAllComments would be better. This will be done in another PR.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+Version 3.15.13
+------------------
+[issues resolved](https://github.com/javaparser/javaparser/milestone/166?closed=1)
+
 Version 3.15.12
 ------------------
 [issues resolved](https://github.com/javaparser/javaparser/milestone/165?closed=1)
--- a/javaparser-core-generators/pom.xml
+++ b/javaparser-core-generators/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
-        <version>3.15.13-SNAPSHOT</version>
+        <version>3.15.14-SNAPSHOT</version>
     </parent>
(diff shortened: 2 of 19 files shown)


matozoid:
Looks great, thanks!

PR #2527, opened 2020-02-25
Issue #2480 - Deprecate getComments

### LINKED ISSUE #2480: CompilationUnit.getComments does not return comment before package statement
**Steps to reproduce:**
```java
CompilationUnit u = StaticJavaParser.parse("/*Comment*/ package a; class A {}");
System.out.println(u.getComments());
```

**Expected result:**
A singleton list containing the comment that reads `/*Comment*/`

**Observed result:**
An empty list

I can get the comment with `u.getComment()`. I don't know if this is intended, but I think it's kind of ugly to have to call both `getComment` and `getComments` if I want to get ALL the comments in the compilation unit.

Mathiponds:
Fix #2480 
It deprecates `getComments` because the name is confusing with other method dealing with comments like `getComment` and `getAllContainedComments`
I implemented `getAllComments` that does the same thing as `getComments` but with a better name

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
@@ -135,16 +135,8 @@ public class CompilationUnit extends Node {
     }
 
     /**
-     * Return a list containing all comments declared in this compilation unit.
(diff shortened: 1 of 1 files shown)


matozoid:
Great! Thanks!

PR #2528, opened 2020-02-25
Solve issue 2482

### LINKED ISSUE #2482: Comment is not setted to the correct node
It seems that in the statement bellow the comment is not setted to the correct node

    // a comment before parent
    if (file != null) {}

is parsed to 

    if (// a comment before parent
      file != null) {
    }

It seems that CommentsInserter.insertComments(..) postprocessor run on child nodes but in this case it should take into account the node itself.

What do you think about this?

axelkennedal:
Fixes #2482 

One comment; it seems like maybe a change to the javaparser readme got accidentally accepted in a pull request by @Mathiponds (a friend of mine). This pull request removes that change (a link to a Google doc).

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/Issue2482Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/Issue2482Test.java
@@ -1,5 +1,6 @@
 package com.github.javaparser;
 
+import com.github.javaparser.ast.expr.AssignExpr;
 import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.stmt.Statement;
 import org.junit.jupiter.api.Test;
@@ -26,7 +27,7 @@ public class Issue2482Test {
                 "{ if (file != null) {} }");
         assertTrue(st.getComment().isPresent());
         assertTrue(st.getOrphanComments().isEmpty());
-        assertEquals(1, st.getAllContainedComments());
+        assertEquals(1, st.getAllContainedComments().size());
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


matozoid:
Ah, funny how I missed that link :-) Let's see if @edefazio can take a look at this.

edefazio:
Ohh, so basically we merged changes that updated the JavaParser README.md documentation and pointed to a particular project that was related to JavaParser (a team project)?

...And this change reverts the original changes to the README.md ? 

anything else or is this just a simple (whoopsie /fix)... 
(I hope we aren't mining Dogecoins or something now)

javaparser-core-testing/src/test/java/com/github/javaparser/Issue2482Test.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test
+    public void commentBeforeBlock() {
+        Statement st = StaticJavaParser.parseBlock(
+                "// a comment before parent" + System.lineSeparator() +
+                "{ if (file != null) {} }");
+        assertTrue(st.getComment().isPresent());
+        assertTrue(st.getOrphanComments().isEmpty());
+        assertEquals(1, st.getAllContainedComments());
```
edefazio: Im confused how this works...
st.getAllContainedComments() returns a list RIGHT?
so how can a list be equal to 1?
shouldnt it be :

`st.getAllContainedComments.size()`

javaparser-core-testing/src/test/java/com/github/javaparser/Issue2482Test.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test
+    public void commentBeforeIfStatement() {
+        Statement st = StaticJavaParser.parseStatement(
+                "// a comment before parent" + System.lineSeparator() +
+                        "if (file != null) {}");
+        assertTrue(st.getComment().isPresent());
+        assertTrue(st.getOrphanComments().isEmpty());
+        assertEquals(1, st.getAllContainedComments());
```
edefazio: Im confused how this works...
st.getAllContainedComments() returns a list RIGHT?
so how can a list be equal to 1?
shouldnt it be :

`st.getAllContainedComments.size()` 
?

edefazio:
Hmm I apologize if I am not doing this correctly

I want to be honest, but I'm having a hard time understanding the changes GitHub is showing me here. (I appear to have commented on an outdated version of a file)

I noticed the original  links from the README.md and it looks like they have been removed... but (from a code perspective) I kinda cant make heads or tails of what I'm supposed to be reviewing.

(Maybe I need remedial github training or something)

matozoid:
@edefazio I always look at the "files changed" tab. I'm also not sured what happened - maybe @axelkennedal confused github with creative git usage?

edefazio (commented):
looks good to me :)

matozoid:
There it goes then!

PR #2529, opened 2020-02-25
Fix 2503

### LINKED ISSUE #2503: `EnclosedExpr` -- inner may be null?
Within `EnclosedExpr`, `removeInner()` calls `setInner(null)`. Additionally, the javadoc for `setInner(Expression value)` indicates that the value may be null.

The first line of `setInner` is an assertion that `value` is not null. 

<s>I suspect this might be as simple as removing the assertion, but there are other wide-impacting side-effects of doing so such as allowing `getInner()` to return null (as opposed to being optional). </s>

_Edit: just spotted #1010 and #1018 which indicates that the intention is inner must not be null._

If inner **_is_** required (i.e. must not be null), then different changes are needed -- minimum edit the documentation and to do _something_ to `removeInner` (not sure what? maybe throw an error?) and perhaps the constructors?

Either way, it seems some additional tests would be helpful here too!

Thoughts? :) 

https://github.com/javaparser/javaparser/blob/d1ee0363f4f1e0e93c7d34510f48194a899405ce/javaparser-core/src/main/java/com/github/javaparser/ast/expr/EnclosedExpr.java#L83-L101

https://github.com/javaparser/javaparser/blob/d1ee0363f4f1e0e93c7d34510f48194a899405ce/javaparser-core/src/main/java/com/github/javaparser/ast/expr/EnclosedExpr.java#L111-L114

axelkennedal:
Fixes #2503 by removing old remove-methods.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,7 @@
+Version 3.15.13
+------------------
+[issues resolved](https://github.com/javaparser/javaparser/milestone/166?closed=1)
+
 Version 3.15.12
 ------------------
 [issues resolved](https://github.com/javaparser/javaparser/milestone/165?closed=1)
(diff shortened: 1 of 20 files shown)


matozoid:
Cool, thanks!

PR #2532, opened 2020-02-26
Issue #546 - Add possibility in switch node

### LINKED ISSUE #546: PrettyPrinter: specify the indentation style for switch/case

Mathiponds:
It is now possible to have a `switch` node indented as
```java
switch(x)
case 1:
    return y;
case 2:
    return z;
}
```
You have to set the parameter `indentInSwitch` of
`PrettyPrinterConfiguration` to do so. Its default value is `true`

Fixes #546.


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1164,11 +1164,15 @@ private void printSwitchNode(SwitchNode n, Void arg) {
...
+                printer.indent();
+            }
             for (final SwitchEntry e : n.getEntries()) {
                 e.accept(this, arg);
             }
-            printer.unindent();
+            if(configuration.isIndentInSwitch()) {
+                printer.unindent();
+            }
         }
```
Mathiponds, line 1183: Or maybe it is better to write it as 
```java 
            if(configuration.isIndentInSwitch()) {
                printer.indent();
                for (final SwitchEntry e : n.getEntries())
                    e.accept(this, arg);
                print.unindent();
            }else{
                for (final SwitchEntry e : n.getEntries()) 
                    e.accept(this, arg);
            }
```

MysterAitch: Personally, I think a cleaner/tidier alternative would be:

```java
printer.indent(configuration.isIndentInSwitch());
```

Or a helper method `indentIf(boolean shouldIndent)` (or similar).

MysterAitch:
Thinking out loud re:naming -- there are multiple places within a switch where indentation may be desirable.

Indenting the case line, and indenting the body of the case.

Is it worth renaming the config option to explicitly refer only to indenting the case body?

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
Mathiponds: Good idea I will do that

matozoid: I don't see why the indenting printer should be modified for a single case?

matozoid:
Thanks! It's good enough like this :-)

PR #2536, opened 2020-02-27
Implement issue 545: spaces between operators

### LINKED ISSUE #545: PrettyPrinter: optional spaces around operators
1 + 1
vs
1+1

hmellin97:
Fixes #545 

A boolean in the PrettyPrinterConfiguration was added that lets the user choose if it wants spaces between operators or not.


matozoid:
That's a good start, but it looks like some tests fell over :-(

hmellin97:
image
For some reason this assertEquals fails, even though IntelliJ thinks they are identical

javaparser-core-testing-bdd/src/test/resources/com/github/javaparser/pretty_printing_scenarios.story
```diff
@@ -193,14 +193,14 @@ package a.b.c;
...
+    a=2+1;
+    b=3+1;
 }
 When the block is parsed by the Java parser
 Then it is printed as:
 {
-    a = 2;
-    b = 3;
+    a = 2 + 1;
+    b = 3 + 1;
```
MysterAitch, line 203: Could you add new test cases, rather than editing existing ones?

MysterAitch:
@hmellin97 It appears that there are differences in line separators (`\n` versus `\r\n` -- also displayed in some places within IntelliJ as `LF` versus `CRLF`).

The helper method `TestUtils#assertEqualsNoEol` can assert equality with normalised line endings and may be helpful here.

MysterAitch:
Looks good, thanks!

- The configuration option relates to spacing **around** operators (as opposed to _between_ them) - please rename the configuration option and methods to reflect this. 


Only the one above is important to me, and below are a couple more suggestions to "polish" things up.

- Having two configurations (one for `BinaryExpr` and one for `AssignExpr`) would be preferable imo (this would allow more granular `x += 1+1`/`x += 1 + 1`/`x+=1+1`/`x+=1 + 1`), but this is not necessarily required as it could be added in later!
  Sidenote: The `TernaryExpr` ([JLS§15](https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html)) seems to be named (more descriptively?) as a `ConditionalExpr` within JavaParser. 

- Some of the test cases seem to test multiple things (e.g. true/false, input, output). Parameterised tests would be useful where there are multiple assertions in a row with only changed input/output data, as would renaming the test method to describe the case being tested (e.g. expecting spaces or not, binary vs unary expressions).

matozoid:
From the maintainers side: @MysterAitch two configurations seem to be a bit much right now - remember that this a feature nobody asked for, in a component that will be removed in the distant future (instead of a pretty printer, we should have a code formatter that works on the tokens underlying the AST)

matozoid:
Thanks! It's fine like this :-)

matozoid:
Oops, I merged the wrong PR :-D

It would be great if the comments could still be addressed!

Issue #1295, opened 2017-12-17
Refactor Javadoc parser API


matozoid:
... to match more with the rest of JP

PR #2540, opened 2020-02-28
Proposed Structure for #1295

simonsirak:
Provides base work towards #1295. This PR could be ignored or extended/modified by future contributors.

#### Added
* A node structure that mirrors the Javadoc custom parser structure (JavadocDescription, JavadocBlockTag, JavadocDescriptionElement, JavadocSnippet and JavadocInlineTag);
* An integration to the overall node structure including the Meta Model;
* Relevant unit tests that are translated from the custom javadoc parser. Tests related to the
  parsing will be added with the grammar;
* Add Grammar for Javadoc;
* Integrate into main parser;
* Unit tests.

The new proposed JavadocContent node is injected into the existing JavadocComment AST node. This allows users to choose between sticking with the old way of handling JavadocComments, or choose to use our proposed JavadocContent nodes. However, currently the grammar only manages to successfully parse Javadoc; only empty JavadocContent nodes are injected into JavadocComment nodes upon parse success. 

#### Not Added 
The following was not added due to time constraints, and could be extended by future contributors:

* Pretty Printing;
* Concrete Syntax Model Support;
* Node Construction in parser (currently, an empty JavadocContent is returned from our Javadoc parser). This might potentially require modification of the Javadoc grammar.

A unit test was commented for presentation purposes (as part of a school project), as it would otherwise fail the build.


javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/JavadocTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    void toTextForEmptyJavadoc() {
+        // REQ 2
+        JavadocContent doc = new JavadocContent();
+        assertEquals("", doc.toText());
+    }
+
+    @Test
+    void toTextForJavadocWithTwoLinesOfJustDescription() {
+        // REQ 2
```
matozoid, line 57: What do all these REQ's refer to?

javaparser-core/src/main/java/com/github/javaparser/ast/comments/JavadocBlockTag.java
```diff
@@ -0,0 +1,203 @@
...
+import com.github.javaparser.ast.visitor.CloneVisitor;
+import com.github.javaparser.metamodel.JavadocBlockTagMetaModel;
+import com.github.javaparser.metamodel.JavaParserMetaModel;
+import com.github.javaparser.ast.Generated;
+import static com.github.javaparser.utils.Utils.assertNotNull;
+
+/**
+ * A JavaDoc block tag. Different tag types are defined by the BlockTagType enum.
+ *
+ * @author Simon Sirak & David Johansson
```
matozoid, line 38: I think you're supposed to add two @author tags for two authors :-)

javaparser-core/src/main/java/com/github/javaparser/ast/comments/JavadocDescription.java
```diff
@@ -0,0 +1,152 @@
...
+import com.github.javaparser.metamodel.JavadocDescriptionMetaModel;
+import com.github.javaparser.metamodel.JavaParserMetaModel;
+import com.github.javaparser.ast.Generated;
+
+/**
+ * A JavaDoc description.
+ *
+ * @author Simon Sirak & David Johansson
+ */
+public class JavadocDescription extends Node {
```
matozoid, line 41: Maybe we should have a fake `JavadocNode` so that we get the is/as-methods get generated.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -940,6 +940,15 @@ private static CsmElement typeArguments() {
                 newline()
         ));
 
+        // Javadoc
+        // TODO
+        concreteSyntaxModelByClass.put(JavadocContent.class, sequence());
```
matozoid, line 945: Better create issues for all the TODO's and reference them from the TODO.

matozoid:
This looks very nice! It's the best thing to happen to the Javadoc support since it has been written :-)

Once this is in a reasonably complete state we can deprecate the previous approach.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/JavadocTest.java
simonsirak: The assignment said that we had to specify which tests belonged to which requirements (we specified requirements for this issue internally), so we added references to the requirements as comments for presentation purposes. Sorry, we must have forgotten to remove them for the PR :(

javaparser-core/src/main/java/com/github/javaparser/ast/comments/JavadocDescription.java
simonsirak: I am not sure what you mean, perhaps you could create an issue for this? :)

simonsirak:
Thanks! We are happy to contribute :) It is still in an early stage but hopefully it is a good enough start so that it can be finalized by future contributors. Should I create issues that reference this PR for each of the **not added** items?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/JavadocTest.java
matozoid: Okay, now I know that these can be removed

matozoid:
Yes please, and I'll add my issue :-)

simonsirak:
The issues have now been added! :)

matozoid:
Thanks!

PR #2546, opened 2020-03-01
Added cache size configuration to JavaParserTypeSolver

ReallyLiri:


matozoid:
Hi! Thanks, this is probably useful, I'll see if someone wants to review this. Meanwhile many tests are failing with the same problem?

ReallyLiri:
@matozoid thanks, tests fixed

matozoid:
@ftomassetti - I'd really like to interrupt your day job for an opinion. Are these things caches that can be limited in size, or are they holders of required data for the solver algorithm?

ftomassetti:
I do not have a precise idea on which size would make sense. I would expect that there is some locality in the solving process, so it would make sense to limit the size, but to which level? 10? 100? 1000? I have honestly no idea

matozoid:
@ReallyLiri - can you add that as Javadoc? Something like "the size of various internal symbol solver caches in... bytes I guess? Can be used to reduce excessive memory usage. Beware of setting it too low, it may cause... ehhh... unpredictable behaviour?"

I'll leave the exact phrasing to you ;-)

ReallyLiri:
The reason I added this size feature in the first place is because for very big code bases (thousands of source files), the caches could blow up.
The argument is intended to be used by those who can estimate their size requirements. In any case, using no cache at all would not hurt correctness, only performance. As far as I can tell.
Will add Javadoc :)

matozoid:
@ReallyLiri - what I understand from @ftomassetti (who made the solver) is that some paths in the solver depend on other paths, and they pass information through the cache. So it's not really a cache, and throwing out that information would result in, I guess an exception? To prove or disprove this, we could easily set the cache to 0 and run all the tests :-)

ftomassetti:
Yes, certain times we need two passes to solve types (it really seems an approach defined also in the JLS), a first more "incomplete" (basically being more "loose" on mapping generic arguments) and a second more strict. I really wish it was less complicated... but it seems complicated *by definition*, not just in our implementation

ReallyLiri:
I verified that when running with cache size = 0 all tests are still passing. They take huge time to run however.. 4:46 hours :)
I added Javadoc to emphasize when to use (or not) this feature.

matozoid:
Wow, that's a long time :-D At least we know the caches are useful now! Surprising that everything went fine, but it's also good to have this uncertainty about the functionality gone.

Thanks for this work, let's merge it!

PR #2555, opened 2020-03-08
rename `isSpacesBetweenOperators` to `isSpaceAroundOperators`

MysterAitch:
follow up to #2536


matozoid:
It's still brand new, so no deprecation is fine. The chance of anybody using this already is tiny.

PR #2554, opened 2020-03-08
Generators incremental squashed

MysterAitch:
Here are some updates to the generators.

It's split into two commits for the sake of clean changelog. [This separate branch](https://github.com/javaparser/javaparser/compare/master...MysterAitch:generators-incremental) and the git commit description has some more granular detail.

- The main substantial change is that there is a new `RemoveGeneratorAnnotations` generator which goes through and deletes existing `@Generated` annotations. By running this first, any annotations that are not regenerated (e.g. due to a field rename) will show up in the git diff. 

- There was also a bug(?) where the annotation is not always added in (only where the method is first added, but not when it is being replaced. This means that is not re-added following e.g. a manual removal of the annotation.

- Next is an edit where the files that are touched by the generators are returned from the `generate()` method, and only those files will be saved (helps to limit spurious whitespace-only changes due to the JP printer)

- Progress toward #771

Note that while the diff is rather large, changes are limited to only the core-generator module and changes elsewhere are a result of running the modified generators.

PR #2557, opened 2020-03-09
Remove generated annotation simple

MysterAitch:
Simple, minimal, version of #2554 that doesn't contain any of the follow-up fixes/deprecations etc.

My main hesitation with this specific PR is that there are a substantial number of files which have whitespace-only changes. This is not improved by enabling lexical preservation.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
@@ -18,7 +18,6 @@
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU Lesser General Public License for more details.
  */
-
 package com.github.javaparser;
 
 import com.github.javaparser.ast.CompilationUnit;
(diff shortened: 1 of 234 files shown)


MysterAitch:
Here's a comparison without and without lexical preservation:

https://github.com/MysterAitch/javaparser/compare/MysterAitch:remove_generated_annotation_simple...MysterAitch:remove_generated_annotation_simple_with_lex_pres

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -91,7 +89,7 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
         astParser.setTabSize(configuration.getTabSize());
         astParser.setStoreTokens(configuration.isStoreTokens());
         if (configuration.getLanguageLevel() != null) {
-            switch (configuration.getLanguageLevel()) {
+            switch(configuration.getLanguageLevel()) {
```
matozoid, line 92: Ah, I can see why you're building in that dirty detection. These files shouldn't be touched! The scope for the generators is: the AST nodes and the visitors. If you would like to extend that, please discuss a bit first so I'm not completely surprised :-D

javaparser-core/src/main/java/com/github/javaparser/JavadocParser.java
```diff
@@ -48,33 +47,21 @@ public static Javadoc parse(JavadocComment comment) {
 
     public static Javadoc parse(String commentContent) {
         List<String> cleanLines = cleanLines(normalizeEolInTextBlock(commentContent, EOL));
-        int indexOfFirstBlockTag = cleanLines.stream()
-                .filter(JavadocParser::isABlockLine)
-                .map(cleanLines::indexOf)
-                .findFirst()
-                .orElse(-1);
+        int indexOfFirstBlockTag = cleanLines.stream().filter(JavadocParser::isABlockLine).map(cleanLines::indexOf).findFirst().orElse(-1);
```
matozoid, line 50: Noooooo

MysterAitch:
Closing in favour of #2704 (or whatever comes of that PR).

PR #2564, opened 2020-03-13
Reduce amount of symbol solver jars

### LINKED ISSUE #2563: Same package name in different modules causes error when compile java9 module
Package `com.github.javaparser.symbolsolver.model.typesystem` exists both in `javaparser-symbol-solver-model` and `javaparser-symbol-solver-core`, which causes error when requires module in `module-info.java` 
```
module my.module {
^^^^^^^^^^^^^^^^
    requires com.github.javaparser.symbolsolver.core;
    requires com.github.javaparser.symbolsolver.model;
}
```
Error message: `module my.module reads package com.github.javaparser.symbolsolver.model.typesystem from both com.github.javaparser.symbolsolver.core and com.github.javaparser.symbolsolver.model`

matozoid:
Fixes #2563.

As it seems that nobody knows why the symbols solver is split into three jars, nobody really kept the split clean, and the split is causing trouble with modules (see issue), this PR merges everything into the core jar.


matozoid:
Hi! Can I get some feedback on this?

MysterAitch:
No specific objections from me. 

My only two comments are just what was said on chat re: double checking that this is the only place we have packages duplicated across modues, and whether we should take this as a good time/opportunity to review the module structure more generally.

They can both be done as incremental updates though, no need to block this PR from being included in the next release.

matozoid:
Thanks!

Issue #2537, opened 2020-02-27
Maintenance - Update dependencies


MysterAitch:
A rerun of #1698 -- it's probably worth a review to check we're not running too far behind anything, and to flag up anything that has a 3rd party api change.

MysterAitch:
@sullis has kindly submitted some PRs to update some dependency versions.

Long-term, I would suggest that dependency checks are automated. 

There is a github bot called `dependabot` which scans project build files and submits PRs where it spots an outdated dependency. Alternatively, if wanting platform independence, there are maven plugins which can do update checks as part of the build process (e.g. `versions-maven-plugin`).

My preference is dependabot though :)

MysterAitch:
I've just fired dependabot off on a fork, and it is showing 24 dependencies that are not currently at their latest version. 

Testing will be needed re: if there are any issues or API incompatibilities or similar, but once this first painful update is done it should be a fairly easy maintenance task to keep on top of in the future.

![Untitled](https://user-images.githubusercontent.com/1537214/77019320-456d3b80-6978-11ea-9bef-645d737cc734.png)

matozoid:
Careful with the javacc plugin. It was my goal to stay with the official JavaCC releases, not the branched off ph-javacc. Still, official JavaCC is not moving much (you could join their slack channel to check) so maybe the ph-fork is better.

matozoid:
Note that many of the plugin updates may cause trouble during releasing.

PR #2565, opened 2020-03-15
maven-compiler-plugin 3.8.1

sullis:
Update dependency version 

See also: #2537


MysterAitch:
Included within #2572

PR #2576, opened 2020-03-24
[IDEA] Simple dockerfile for releasing

matozoid:
A simple Dockerfile that runs the release in a simple linux environment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+# a tiny linux distribution:
+FROM alpine
+# get a package list:
+RUN apk update
+# install the packages we need:
+# (can be improved with using fixed versions?)
+RUN apk add openjdk8 git maven
+# copy the source code from our non-docker checkout into /javaparser:
+# (maybe check out master after this?)
+COPY . /javaparser
+# use that as the working directory:
+WORKDIR /javaparser
+# start a release: (can be overridden with --run from the command line)
+# (do something about the version number prompt. You will need to attach a console for that right now)
+CMD mvn release:clean release:prepare release:perform  -DinteractiveMode=false 


MysterAitch:
Looks promising -- assuming that is a complete example, it's much more simple than I expected it to be! :) 

My previous experience is mostly with vagrant which spins up heavier virtual machines rather than more lightweight docker containers.

In terms of expansion / updates, might it be possible to set git ssh keys and other config stuff as environment variables? Or is that not really how it works?

matozoid:
Docker is pretty good at interacting with the environment. The COPY action, for example, takes stuff that is on your normal file system and copies it into the virtual filesystem in Docker.

One improvement I haven't mentioned is that you can magically let your local .m2 repository appear in the container, so it won't spend five minutes downloading the world every time.

MysterAitch:
To skip tests and do a non-interactive `release:prepare`:

```sh
mvn -e --batch-mode -Darguments="-DskipTests" release:clean release:prepare release:perform
```

To build and run the Dockerfile (run from the project root, with the dockerfile within the dev-files subdirectory). No idea if the build/run is the way it's supposed to be done, but it kinda works :P 

- Note the paths to the "volume"s will need to be configured locally. 
- Note the dockerfile will need git config details to be set (I had to do this in the dockerfile, which won't be sustainable if this is shared - is there a better way?)
- Note that there doesn't seem to be a way to do a `release:rollback` once the container closes (other than maybe manually creating the backup poms)

```sh
docker build --tag jpbuild:1.1 --file ./dev-files/docker_trigger_release/Dockerfile . 

docker run \
    -v /home/jp-docker/.m2:/root/.m2 \
    -v /home/jp-docker/.ssh:/root/.ssh \
    jpbuild:1.1
```

Thoughts/comments welcome :)

matozoid:
- Docker can map your local filesystem to the Docker fs, so that's how you can make secret stuff available in the container
- Rollback, in my opinion, is nice when it works, but painful when things have already been pushed to origin. Might be easier to just skip to the next version number :-)

PR #2572, opened 2020-03-22
Update all dependencies

MysterAitch:
This PR updates all dependencies per dependabot on my fork.

There are a couple of breaking changes that have been fixed (e.g. a deprecated mockito method and now needing to pass in a charset when creating a file stream provider).

Oddly one of the bdd tests seemingly wasn't doing case-sensitive comparisons.

Otherwise it compiles and tests pass locally - let's see what travis and appveyor have to say! :) 

Closes #2537 
Supersedes #2565 
Supersedes #2566 
Supersedes #2567

PR #2581, opened 2020-03-28
Dependency updates

MysterAitch:
A re-do of #2572 now that I've managed to pin down that it's not these dependencies that prevented me from being able to push releases :)


MysterAitch:
I'm going to merge - if any unexpected issues arise we can revert them :)

PR #2594, opened 2020-04-06
Issue 2592 -- Fix: Nodes in NodeList losing their parent on replacing a child, with lpp enabled

### LINKED ISSUE #2592: Parent all elements in a NodeList is set to null on replacement when LexicalPreservingPrinter is setup
In the latest version of Javaparser, if we set LexicalPreservingPrinter and replace a parameter in a MethodDeclaration, all parameters have their parent set to null.

    
    String s = "public class A {" + 
			"  public void m(final int a, int b) {" + 
			"  }" +
			"} ";
    CompilationUnit cu = StaticJavaParser.parse(s); 
    Optional<MethodDeclaration> md = cu.findFirst(MethodDeclaration.class);
    Parameter p1 = md.get().getParameter(0);
    Parameter p2 = new Parameter(p1.getModifiers(), p1.getType(), new SimpleName("a1"));
    md.get().replace(p1, p2);
    assertTrue(md.get().getParameters().stream().allMatch(p -> p.getParentNode().isPresent()));

MysterAitch:
Fixes #2592

> Turns out that it's due to the way that list replacements are handled.
> 
> It seems that a brand new nodelist is created on every change, but because this new nodelist hasn't been assigned a parent yet (i.e. the list's parent is null) any children that get copied across (added to this new instance of a nodelist) then get assigned the new list's parent (i.e. null). Hence the clobbering.
> 
> Copying the parent across from the old list before performing the list insertions seems to fix this!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,9 +3,11 @@ Next Release (3.15.19)
 [issues resolved](https://github.com/javaparser/javaparser/milestone/172?closed=1)
 
 * DEPRECATED: Deprecated and documented `JarTypeSolver#getJarTypeSolver(String)`, with a view to later removal.
-  ([#2598](https://github.com/javaparser/javaparser/pull/2598))
+    ([#2598](https://github.com/javaparser/javaparser/pull/2598))
 * FIXED: Fix issue #2552 : UnsupportedOperationException caused by resolving inner annotation
-    ([#2585](https://github.com/javaparser/javaparser/pull/2553))
+    ([#2553](https://github.com/javaparser/javaparser/pull/2553))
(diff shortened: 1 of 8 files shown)


MysterAitch:
Just pushed a commit which makes the three list-manipulating classes be consistent, in addition to some additional comments and minor whitespace fixes.

Happy to pull that out into a different PR if the preference is to keep this PR tightly-focussed (especially given that the actual fix in 4a825e7 is just a single line..).

edefazio (approved):
Looks good to me,
thanks

Issue #2547, opened 2020-03-05
Adding Multiple JarTypeSolver to a CombinedTypeSolver leads to IllegalStateException


matkais:
I upgraded from 3.14.16 to 3.15.14 and cant add multiple JarTypeSolver to my CombinedTypeSolver.
I do something like this:
`combinedTypeSolver.add(JarTypeSolver.getJarTypeSolver(jar));`
where I add multiple jars in a loop to my solver.
In the new version it crashes with a 
> Caused by: java.lang.IllegalStateException: This TypeSolver already has a parent.

Exception occures after I add the 2nd jar to my solver, at this point there is already a ReflectionTypeSolver and a JarTypeSolver added.
Hit me up if you need more information

matozoid:
Hi, thanks for the report! I looked at the code for `getJarTypeSolver` and it uses a static singleton which is extended with a new jar-path on every call. So you get the same solver back each time. That's a rather confusing approach, and it throws off the recent "don't add a typesolver multiple times" detection which throws the exception you have.

So as a workaround you can `new` a `JarTypeSolver` for every path.

`getJarTypeSolver` will need some work because it does not act as its method name suggests. I think it should be documented, deprecated, and the `addPathToJar` should be made public (which is what gets called internally.)

MysterAitch:
Looking into this, the `JarTypeSolver` seems to be inconsistent with the other type solvers but it's an area of the codebase I'm not familiar with.

`JarTypeSolver` has a static method which allows multiple jar files to be included as part of the same typesolver. This method is not available non-statically, however -- the non-static alternative is to create multiple instances of the object per @matozoid 's suggestion above. 

My initial inclination would be to remove the static method and the private `addPathToJar` method (to be consistent with the other typesolvers), but I don't know the code well enough to know if there is a reason to have only a single jar solver (if so, shouldn't the constructor be made private?).

MysterAitch:
Sidenote: here's an example from the book which seems to favour using the static method style:

```java 
TypeSolver myTypeSolver = new CombinedTypeSolver(
    new ReflectionTypeSolver(),
    JarTypeSolver.getJarTypeSolver("jars/library1.jar"),
    JarTypeSolver.getJarTypeSolver("jars/library2.jar"),
    JarTypeSolver.getJarTypeSolver("jars/library3.jar"),
    new JavaParserTypeSolver(new File("src/main/java")),
    new JavaParserTypeSolver(new File("generated_code"))
);
```

[...]

PR #2598, opened 2020-04-08
Deprecated and documented `JarTypeSolver#getJarTypeSolver(String)`

MysterAitch:
See #2547 

My preference would be to remove the singleton stuff completely, rather than just deprecate it.

Given that #2547 shows anybody using a recent version will not be calling this method twice, the only people affected by its removal would be those who use it only once, and their workaround for this API change would be extremely simple (replace `JarTypeSolver.getJarTypeSolver(jar)` with `new JarTypeSolver(jar)`).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@ Next Release (3.15.19)
 ------------------
 [issues resolved](https://github.com/javaparser/javaparser/milestone/172?closed=1)
 
+* DEPRECATED: Deprecated and documented `JarTypeSolver#getJarTypeSolver(String)`, with a view to later removal.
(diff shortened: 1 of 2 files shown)


MysterAitch:
@edefazio / @matozoid - Would you mind sanity checking this please? I'm trying to avoid merging my own PRs if I can :)

edefazio (approved):
:)

edefazio (approved):
:)

matozoid:
You could take it a step further, and indeed remove the "singleton" stuff, and let the deprecated method always return a new solver. That way the code is squeaky clean and only the deprecated method is left to remove someday.

PR #2599, opened 2020-04-09
Fix resolving default interface method calls (fixes #1949)

### LINKED ISSUE #1949: JSS can't resolve super-default method reference

maartenc:
Fixes #1949.


MysterAitch:
It seems reasonable, and the tests pass -- in it goes! :)

Issue #2595, opened 2020-04-06
[Java Symbol Solver] Resolve type from functional interface


arekziobrowski:
**QUESTION**: does `javaparser` support the resolution of types from functional interfaces? Or the `java.lang.UnsupportedOperationException` addresses the lack of this functionality (see description below for more information)?

---
**PROBLEM**:
I have the following code which I try to resolve (*be careful*, `javaparser` types are coincidental, I am parsing a project that utilizes the usage of `javaparser`):
```java
@Override
    public List<Metric<Integer>> compute(CompilationUnit compilationUnit) {
        Set<String> foreignProviders = new HashSet<>();
        ClassMetric<Integer> fdp = c -> {
            List<String> classFieldNames = getAllClassFieldNames(c);
            // DELETED FOR SAKE OF BREVITY
            return foreignProviders.size();
        };
        return getMetricForClass(fdp, compilationUnit);
    }
```
There seems to be a problem with resolution of `getAllClassFieldNames(c);`.  The parameter `c` is of type `ClassOrInterfaceDeclaration` (*be careful*, `javaparser` types are coincidental, I am parsing a project that utilizes the usage of `javaparser`) and it comes from the following functional interface:
```java
@FunctionalInterface
public interface ClassMetric<T> extends Function<ClassOrInterfaceDeclaration, T> {
    T apply(ClassOrInterfaceDeclaration c);
}
```

I get the following exception. There seems to be a problem with resolution of the type from the interface.

```
Exception in thread "main" java.lang.RuntimeException: Unable to calculate the type of a parameter of a method call. Method call: getAllClassFieldNames(c), Parameter: c
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:225)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:240)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:125)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:140)
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:313)
	at Scratch.lambda$main$1(scratch_4.java:36)
	at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
[...]

PR #2604, opened 2020-04-10
WIP - Issue 2595 - lambda parameter resolution

MysterAitch:
ref #2595


maartenc:
Hello @MysterAitch , I've created a new PR in an attempt to fix the issue: #2775 .
The approach there is to fix the getAllMethods() logic so that overriden generied methods are not returned.
Could you take a look?

maartenc:
Fixed by #2775

PR #2611, opened 2020-04-16
Fix UnsupportedOperation Removed with lexical preservation #2610

### LINKED ISSUE #2610: UnsupportedOperation Removed with lexical preservation
With the lastest javaparser version, if we try to replace an expression preceded by a comment, LexicalPreservation throws an UnsupportedOperation

                "public class Bar {\n" + 
                "    public void foo() {\n" + 
                "          // comment\n" +
                "          System.out.print(\"error\");\n" +
                "    }\n" +
                "}"
                );
        // contruct a statement with a comment
        Expression expr = StaticJavaParser.parseExpression("System.out.println(\"error\" + e.toString)");
        // add the statement to the ast
        Optional<MethodCallExpr> mce = cu.findFirst(MethodCallExpr.class);
        mce.get().getParentNode().get().replace(mce.get(), expr);

jlerbsc:
Fixes #2610.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,8 @@ Next Release (3.15.19)
     ([#2594](https://github.com/javaparser/javaparser/pull/2594))
 * FIXED: Fix JavaParserClassDeclaration canBeAssignedTo() to not cause a recursion when a node is its own parent (e.g. `java.lang.Object`)
     ([#2608](https://github.com/javaparser/javaparser/pull/2608))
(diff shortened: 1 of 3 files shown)


MysterAitch:
Thank you for the report and the very quick fix! 

This PR seems reasonable and I don't see any reason it shouldn't be included in the next version, but will leave it open for comment for now for any additional comment. 

The CSM and lexical preservation side of things is still a bit of a mystery though -- would you mind sharing a few comments about how this has helped to fix the issue please?

Issue #2615, opened 2020-04-20
ParserCollectionStrategy and modules


wmdietl:
I'm running into an issue when trying to use `ParserCollectionStrategy` with a directory that contains a module declaration.

When there is no `module-info.java` file, I get the expected behavior. As soon as I add a module declaration, the source root is completely ignored.

I tried building a minimal demo:

https://github.com/wmdietl/javaparser-module-bug

Is this the expected behavior or am I misusing `ParserCollectionStrategy`?
Thanks!

MysterAitch:
Am I right in thinking that `module-info.java` files do not contain a package declaration? And that they are expected to describe the package where that file is located?

If this is the case, it sounds to me like a bug in the collection strategy not ignoring module info files (iirc it tries to reverse engineer the source root by backtracking up the directory listing based on the package name of the first java file it sees in a given directory).

Will take a closer look later tonight or maybe later in the week as it shouldn't be too hard to fix (if my assumptions above are correct), but PRs welcome if you feel so inclined to take a look at it first! :)

MysterAitch:
I've added some test cases in #2619

The source roots all appear to be found correctly (via junit assertions), and I'm getting what I believe to be the expected output (manual inspection of the system out logs -- including references to `module-info.java`).

Could you take a look and tweak them to make them fail please?

wmdietl:
Thanks for writing these tests!
Can you reproduce the issue with https://github.com/wmdietl/javaparser-module-bug ?

One thing I notice about your test setup is that the `module-info.java` file is in the `demo` subdirectory.
Can you try having `module-info.java` in the parent directory? That is where you usually have a `module-info.java` file, in the root of the package hierarchy.

So have a test layout like:

```
.../with_module_info/module-info.java
.../with_module_info/demo/Main.java
```

and then try finding source files in `.../with_module_info/`.

MysterAitch:
Ah, this makes a bit more sense now! 

@wmdietl - Please try setting the language level to at least java 9 and see if that fixes things for you.

--- 

Modules weren't introduced until java 9, but the default parser configuration option is `LanguageLevel.JAVA_8`. This can be overridden either by updating the static config, or by passing in a config object to the `ParserCollectionStrategy` constructor. 

```java
StaticJavaParser.getConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_9);
```

```java
ParserConfiguration parserConfig = new ParserConfiguration();
[...]

PR #2619, opened 2020-04-21
Tests for #2615 and improvements to the collection strategies

MysterAitch:
#2615


MysterAitch:
Oops.. unrelated tests are failing due to the additional files I added to the utils directory... will upload edits to those momentarily...

MysterAitch:
Following exploration of #2615, I noticed that the collection strategies only allow setting of the parser configuration via the static javaparser (collection strategies parsing using the static java parser).

This latest commit remedies this.

MysterAitch:
I've added the next release milestone in anticipation of it being merged.

If it can be looked over before the next release then great, otherwise it'll be be bumped back to the release after.

PR #2621, opened 2020-04-21
WIP - Fix resolving non existing method calls nested classes

### LINKED ISSUE #2495: Method calls to non-existing methods of nested classes are resolved to outer class
Method calls to methods of nested classes are wrongly resolved to methods of the outer class, if the method does not exist on the nested class.

Woah! Explaining time. See the following example code can be used to re-produce the phenomena:

```java
public class OuterClass {

  public void test() {
    System.out.println("Test from OuterClass");
  }

  public void onlyOuter() {
    System.out.println("Only available on the outer");
  }

  public class NestedClass {

    public void test() {
      System.out.println("Test from NestedClass");
    }

  }

  protected void foo() {
    NestedClass n = new NestedClass();
    test(); // method from OuterClass
    n.test(); // method from NestedClass

    onlyOuter(); // method from OuterClass
    n.onlyOuter(); // not valid!
  }
}
```

- The `test()` call refers to `test` of the `OuterClass` and is correctly resolved
- The `n.test()` call refers to `test` of the `NestedClass` (or more correctly `OuterClass.NestedClass`) and is also correctly resolved
- The `onlyOuter()` is the same as `test()`, just to make the test more illustrative.

Now with `n.onlyOuter()` things go wrong. This would be a reference to a method `onlyOuter` of `NestedClass` (or its ancestral classes) and since it does not exist it should throw an exception that it cannot be resolved.

However, it is resolved as the `onlyOuter` function of `OuterClass`. The culprit seem to be `JavaParserTypeDeclarationAdapter#solveMethod` which somehow calls `context.getParent().solveMethod` if it cannot find any `candidateMethods` in the class itself. I guess the idea is to take scope into account but it seems to go wrong here.

maartenc:
Fixes #2495.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
@@ -155,14 +155,6 @@ public class JavaParserTypeDeclarationAdapter {
                 }
             }
         }
-        // We want to avoid infinite recursion when a class is using its own method
-        // see issue #75
-        if (candidateMethods.isEmpty()) {
-            SymbolReference<ResolvedMethodDeclaration> parentSolution = context.getParent().solveMethod(name, argumentsTypes, staticOnly);
-            if (parentSolution.isSolved()) {
-                candidateMethods.add(parentSolution.getCorrespondingDeclaration());
-            }
-        }
 
         // if is interface and candidate method list is empty, we should check the Object Methods
         if (candidateMethods.isEmpty() && typeDeclaration.isInterface()) {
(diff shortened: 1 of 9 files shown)


maartenc:
it looks like I broke something, I'll take a look at it

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
```diff
@@ -155,12 +156,15 @@ private ResolvedTypeDeclaration checkAncestorsForType(String name, ResolvedRefer
...
         }
-        // We want to avoid infinite recursion when a class is using its own method
-        // see issue #75
+
         if (candidateMethods.isEmpty()) {
-            SymbolReference<ResolvedMethodDeclaration> parentSolution = context.getParent().solveMethod(name, argumentsTypes, staticOnly);
-            if (parentSolution.isSolved()) {
-                candidateMethods.add(parentSolution.getCorrespondingDeclaration());
+            // Search for static imports
+            Optional<CompilationUnitContext> cu = findCompilationUnitContext();
```
MysterAitch: Really minor and not something worth blocking this over, but would you mind renaming this to `cuCtx` or similar to indicate that it's not actually a compilation unit please?

MysterAitch:
Really appreciating these PRs @maartenc , thank you! :smile_cat:  :+1: 

Looks good with decent tests so can't immediately see any reason to object to it being merged in :)

maartenc:
thx for the feedback @MysterAitch, I'll rename the variable.

Don't merge it yet, I think something else is broken now, but I'll have to test it first.
I'll post an update in the PR if it's ok.

jlerbsc:

PR #2622, opened 2020-04-21
follow up to #2598 and #2547 -- jartypesolver

MysterAitch:
Per @matozoid 's suggestiong, here's a PR taking the previous PR further by removing the singleton behaviours, instead returning a new instance of the JarTypeSolver whenever the newly-deprecated static method is called, cheers! 

#2547
#2598


MysterAitch:
I've added the next release milestone. 

If it can be looked over before the next release then great, otherwise it'll be be bumped back to the release after.

PR #2626, opened 2020-04-22
WIP - trial of speeding up tests

MysterAitch:
As a starting point:
- travis seems to take around 10mins (includes the slow tests), and
- appveyor takes about 5minutes (excludes the slow tests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: java
 install: mvn -B install -DskipTests=true
-script: mvn -B test -P AlsoSlowTests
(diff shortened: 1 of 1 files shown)


MysterAitch:
Possibly worth running with the github actions option.

https://github.com/MysterAitch/javaparser/pull/50

Here it's taking less than ten minutes to run tests on a full matrix of ubuntu/windows/mac-os, each using both jdk8 and jdk11. 

image

image


It can also run on e.g. jdk13/14 too, but as shown below the test suite currently does not pass on jdk13.

image

image

Issue #1701, opened 2018-07-17
Use less volatile classes for tests


matozoid:
`ReflectionInterfaceDeclarationTest` and `ReflectionClassDeclarationTest` have different results depending on the JDK version used to run it because it analyses java.util.List, which changes a lot every version. Find a good replacement so we don't need all that version logic.

EFregnan:
I would like to start working on this issue if nobody else is doing so.

matozoid:
Cool, consider yourself assigned! Ideas for a solution are: find a class that never changes in the JDK, maybe create a custom class to test, or whatever idea you come up with.

MysterAitch:
Just to add specifics, `ReflectionClassDeclarationTest` currently fails on Java 13.

Here's the test output (with additional spacing added in to highlight the difference):

```
ReflectionClassDeclarationTest.testGetAllFields:294 
expected: <[modCount, serialVersionUID, MAX_ARRAY_SIZE, size, elementData, EMPTY_ELEMENTDATA, DEFAULTCAPACITY_EMPTY_ELEMENTDATA, DEFAULT_CAPACITY]> 
but was:  <[modCount, serialVersionUID,                 size, elementData, EMPTY_ELEMENTDATA, DEFAULTCAPACITY_EMPTY_ELEMENTDATA, DEFAULT_CAPACITY]>
```

MysterAitch:
Java 14:

com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclarationTest.testGetAllMethods

```
2020-04-22T12:58:11.2019592Z [ERROR] com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclarationTest.testGetAllMethods  Time elapsed: 1.508 s  <<< FAILURE!
[...]

PR #2637, opened 2020-04-26
fixed testGetAllFields()

EFregnan:
Hi, I implemented a custom class to fix issue #1701 where test _testGetAllFields()_ failed with the newest versions of Java (13 and 14). 

I tested classes _ReflectionClassDeclarationTest_ and _ReflectionInterfaceDeclarationTest_ with Java versions 8, 10, 13, and 14. 

On the one hand, I think having a custom class might reduce the number of future issues caused by possible changes in Java libraries. On the other hand, I am afraid this made the test overly simple. So, please let me know what you think.

Thank you!


EFregnan:
> org.opentest4j.AssertionFailedError: expected: <[a, b, c]> but was: <[a, b, c, $jacocoData]>
4151
at com.github.javaparser.symbolsolver.reflectionmodel.ReflectionClassDeclarationTest.testGetAllFields(ReflectionClassDeclarationTest.java:293)

Build fails because while running the test the synthetic value _$jacocoData_ is introduced. I am looking into a way to remove synthetic values from the list of resolved fields.

MysterAitch:
> 
> 
> > org.opentest4j.AssertionFailedError: expected: <[a, b, c]> but was: <[a, b, c, $jacocoData]>
> > 4151
> > at com.github.javaparser.symbolsolver.reflectionmodel.ReflectionClassDeclarationTest.testGetAllFields(ReflectionClassDeclarationTest.java:293)
> 
> Build fails because while running the test the synthetic value _$jacocoData_ is introduced. I am looking into a way to remove synthetic values from the list of resolved fields.

I've just added filters to ignore the jacoco syntetic field/method. 

I'll merge this PR now that tests are passing, but further work on #1701 is likely needed to work around the temporary tweaks I also added.

PR #2638, opened 2020-04-27
Revert upgrade of ph-javacc-maven-plugin -- #2627

### LINKED ISSUE #2616: Parse Failed when parsing file with very long comment. Lexical Error.
Hi, I am encountering a problem when trying to parse a file with very long comment line, remove few of those comment line help this parse.

 Version
```
implementation 'com.github.javaparser:javaparser-core:3.15.18'
```
ParseConfig
```java
  final JavaParser javaParser = new JavaParser();
    final ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setAttributeComments(true);
    parserConfiguration.setIgnoreAnnotationsWhenAttributingComments(false);
    parserConfiguration.setDoNotAssignCommentsPrecedingEmptyLines(false);
    parserConfiguration.setLexicalPreservationEnabled(true);
...
javaParser.parse(...)
...
```

Error
```
Parsing failed:
Lexical error at line 111, column 54.  Encountered: <EOF> after : ""
Problem stacktrace : 
  com.github.javaparser.GeneratedJavaParserTokenManager.getNextToken(GeneratedJavaParserTokenManager.java:3007)
  com.github.javaparser.GeneratedJavaParser.jj_ntk_f(GeneratedJavaParser.java:10810)
  com.github.javaparser.GeneratedJavaParser.CompilationUnit(GeneratedJavaParser.java:94)

```

Input:
```java
/*
 * JasperReports - Free Java Reporting Library.
 * Copyright (C) 2001 - 2019 TIBCO Software Inc. All rights reserved.
 * http://www.jaspersoft.com
 *
 * Unless you have purchased a commercial license agreement from Jaspersoft,
 * the following license terms apply:
 *
 * This program is part of JasperReports.
 *
 * JasperReports is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Lesser General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * JasperReports is distributed in the hope that it will be useful,
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
 * GNU Lesser General Public License for more details.
 *
 * You should have received a copy of the GNU Lesser General Public License
 * along with JasperReports. If not, see <http://www.gnu.org/licenses/>.
 */

/*
 * Contributors:
 */
package net.sf.jasperreports.engine.export;

import java.awt.BasicStroke;

import net.sf.jasperreports.engine.DefaultJasperReportsContext;
import net.sf.jasperreports.engine.JRAbstractExporter;
import net.sf.jasperreports.engine.JRException;
import net.sf.jasperreports.engine.JRGenericElementType;
import net.sf.jasperreports.engine.JRPrintPage;
import net.sf.jasperreports.engine.JRPrintText;
import net.sf.jasperreports.engine.JRPropertiesUtil;
import net.sf.jasperreports.engine.JRRuntimeException;
import net.sf.jasperreports.engine.JasperReportsContext;
import net.sf.jasperreports.engine.PrintPageFormat;
import net.sf.jasperreports.engine.export.draw.FrameDrawer;
import net.sf.jasperreports.engine.export.draw.PrintDrawVisitor;
import net.sf.jasperreports.engine.util.JRGraphEnvInitializer;
import net.sf.jasperreports.export.Graphics2DExporterConfiguration;
import net.sf.jasperreports.export.Graphics2DExporterOutput;
import net.sf.jasperreports.export.Graphics2DReportConfiguration;
import net.sf.jasperreports.export.ReportExportConfiguration;
import net.sf.jasperreports.renderers.RenderersCache;


/**
 * Exports a JasperReports document to a <code>Graphics2D</code> object.
 * <p/>
 * JasperReports relies on AWT for text measurements and all sorts of layout calculations
 * during report filling, so documents created using AWT will certainly look perfect when
 * rendered with AWT on a <code>java.awt.Graphics2D</code> context. For this reason, the
 * {@link net.sf.jasperreports.engine.export.JRGraphics2DExporter} is the perfect
 * exporter. The output it produces is considered to be the reference in terms of layout
 * capabilities and element styling.
 * <p/>
 * Generally speaking, the document quality produced by all the other exporters is only an
 * approximation of the perfect output that the Graphics2D exporter can produce. As its
 * name indicates, this exporter is special because it does not produce files or send character
 * or binary data to an output stream. Instead, its only target for rendering the content of a
 * page is a <code>java.awt.Graphics2D</code> object. This exporter is also special because it can
 * export only one page at a time.
 * <p/>
 * This exporter is used by the built-in Swing viewer to render the content of each page, and
 * it is also used when printing the documents. The documents are printed page by page,
 * and the exporter is invoked to draw each document page on the graphic context
 * associated with the selected printer job.
 * <p/>
 * Because we are relying on the same code (same exporter) when viewing the documents
 * using the built-in viewer and when printing them, JasperReports is a perfect WYSIWYG
 * tool. The document quality on paper is the same as on the screen.
 * <p/>
 * In terms of exporter input, note that this exporter does not work in batch mode. If a
 * <code>java.util.List</code> of JasperPrint documents is supplied to it through a
 * {@link net.sf.jasperreports.export.SimpleExporterInput} instance, the exporter
 * considers only the first one for exporting and ignores all the others.
 * <p/>
 * Furthermore, this exporter can export only a single page at a time. The index of the page
 * to be exported can be set using either the
 * {@link net.sf.jasperreports.export.ReportExportConfiguration#getStartPageIndex() getStartPageIndex()}
 * exporter configuration setting or the
 * {@link net.sf.jasperreports.export.implementation 'com.github.javaparser:javaparser-core:3.15.18'
   ReportExportConfiguration#getPageIndex() getPageIndex()}.
 * Note that if present, {@link net.sf.jasperreports.export.ReportExportConfiguration#getPageIndex() getPageIndex()} overrides the value of
 * {@link net.sf.jasperreports.export.ReportExportConfiguration#getStartPageIndex() getStartPageIndex()}.
 * Therefore, this exporter actually exports only the first page from
 * the specified page range, no matter how the page range is specified.
 * <p/>
 * As already mentioned, this exporter needs a target <code>java.awt.Graphics2D</code> object onto
 * which to render the specified page. This Graphics2D object can be set using the special
 * exporter output setting {@link net.sf.jasperreports.export.Graphics2DExporterOutput#getGraphics2D() getGraphics2D()}.
 * If this setting is not present, the exporter will throw an
 * exception signaling to the caller program that no output target was specified for the
 * export process.
 * <p/>
 * By default, the exporter renders the content of the page at normal proportions. However,
 * it can also render it at different proportions if needed. For instance, when used inside the
 * Swing viewer, the Graphics2D exporter must render the page using the user-defined
 * zoom ratio. To set the zoom ratio for the exporter, supply a <code>java.lang.Float</code> value
 * ranging from 0 to 1 as the value for the
 * {@link net.sf.jasperreports.export.Graphics2DReportConfiguration#getZoomRatio() getZoomRatio()} export
 * configuration setting.
 * <p/>
 * The Graphics2D exporter is also used when printing directly from Java. The Java Print
 * Service exporter relies on the Graphics2D exporter
 * and delegates to it all the rendering that needs to be performed on the printer's graphic
 * context. Some of the existing JVM implementations have problems related to the huge
 * size of the printer spool jobs that are created even for small documents. To avoid this, a
 * bug fix was introduced in the Graphics2D exporter to minimize the impact of this
 * problem and reduce the size of print spool jobs, while preserving document quality when
 * printing. However, the bug fix itself is not perfect, and users might experience problems
 * when printing bidirectional writing texts such as Arabic and Hebrew.
 * <p/>
 * This is why the special
 * {@link net.sf.jasperreports.export.Graphics2DReportConfiguration#isMinimizePrinterJobSize() isMinimizePrinterJobSize()}
 * export configuration setting was introduced, along with a configuration property called
 * <code>net.sf.jasperreports.export.graphics2d.min.job.size</code>, to allow users to turn
 * on and off this rendering optimization, depending on their actual needs. The
 * configuration property value is used only in the absence of the export configuration setting.
 *
 * @see net.sf.jasperreports.export.Graphics2DExporterOutput
 * @see net.sf.jasperreports.export.Graphics2DReportConfiguration
 * @see net.sf.jasperreports.export.ReportExportConfiguration
 * @see net.sf.jasperreports.export.SimpleExporterInput
 * @author Teodor Danciu (teodord@users.sourceforge.net)
 */
public class JRGraphics2DExporter extends JRAbstractExporter<Graphics2DReportConfiguration, Graphics2DExporterConfiguration, Graphics2DExporterOutput, JRGraphics2DExporterContext>
{

}


```

[Problem.java.txt](https://github.com/javaparser/javaparser/files/4501053/Problem.java.txt)

### LINKED ISSUE #2627: Wrong node range on very long java file
Hi I encouter a wrong node range (startLine, endLine are shifted up above) in this file for placeholder methodDeclaration node.

https://github.com/tensorflow/java/blob/49b7bac612adce3773bdc1dff27f36542a91e21d/tensorflow-core/tensorflow-core-api/src/gen/annotations/org/tensorflow/op/Ops.java#L3682-L3685

Parsing only a valid subset of this file (about ~2000 lines) still yield the correct result.
Version:
`    implementation 'com.github.javaparser:javaparser-core:3.15.18'`

OS:
Ubuntu 18.04

MysterAitch:
Fixes #2627 
Fixes #2616

Probably also need a new issue to track why the upgraded version leads to this problem.


MysterAitch:
Will need to see what happens with the tests here, but it seems that v4.0.0 passes the tests locally at least, so perhaps not necessary to revert all the way back to v3.0.0 -- will explore which version has problems.

PR #2643, opened 2020-05-01
Fix replace value in String and Integer literals for LexicalPreservingPrinter

### LINKED ISSUE #2382: Cannot set value of StringLiteralExpr when LexicalPreservingPrinter is used
I want to replace all literal values and keep original format of the file. 
According to JavaParserVisited book, I used the `LexicalPreservingPrinter`. 
However, I found that `LexicalPreservingPrinter` keeps all literals, here is the test case:
```java
    @Test
    public void testPrinter() {
        String code = "{\n" +
                "    String str = \"aaa\";\n\n\n\n\n" +
                "}";
        BlockStmt b = JavaParser.parseBlock(code);
        LexicalPreservingPrinter.setup(b);

        b.findAll(StringLiteralExpr.class).forEach(stringLiteralExpr -> {
            stringLiteralExpr.setValue("REPLACEMENT");
        });
        System.out.println(LexicalPreservingPrinter.print(b));
        System.out.println("---");
        System.out.println(b.toString());
    }
```` 

The output is 
```
{
    String str = "aaa";




}
---
{
    String str = "REPLACEMENT";
}
```
But the intended output is 
```
{
    String str = "REPLACEMENT";




}
---
{
    String str = "REPLACEMENT";
}
```

### PR DESCRIPTION
**Pull request by lemoncurry**
Starting from #2382, this fixes two bugs:

Definition of Match for CsmToken
---------------------------------
Most tokens (of type CsmToken) have one unique representation, unique
for their type, and unable to change. For example, the left bracket {
will never be changed without changing type.

Integer and string literals however _can_ change without changing their type.

The DifferenceElementCalculator considered only the token type, and ignored the `content`
and `tokenContentCalculator` fields. By using `.equals()`, these are compared as well. This fixes the code for integer literals (and probably for Booleans and others as well). String literals however are still broken.

Use New Value in CsmToken for String Literals
----------------------------------------------
When building the CSM after a property change, the CsmToken for a string literal should include the new value, not the current one. This is only available in a `PropertyChange` object. If no change has yet been made, the object is an instance of `NoChange` and the current value may be used.

---
For completeness, there is a new test to replace class names. This justs current behaviour and may be removed without impact.

Resolves #2382.

### CODE REVIEW

**Comment by MysterAitch** [comment]
Hey, thanks @lemoncurry !

I don't use the CSM side of things much so really appreciated your write up that goes with this PR -- thank you!

Based on your description and a read through of the changed files this looks good. I don't see any reason it can't be merged and included in the next release, but I'll have a closer look later.

In the meanwhile, would you mind adding some javadoc/comments to the files so that your helpful descriptions can help future readers too please? :)

**Comment by MysterAitch** [comment]
Does this also need to be done with the `CsmChar`? and `CsmTextBlock`? 

I also don't see anything for other literals e.g. int/long or boolean values -- does this also apply to them?

Depending on the scope/size of doing this it would be worth making this fix once for all variable-content nodes rather than doing them in bits, but if it's a sizeable task then we'll need to open up another issue / PR for those. 

Thanks again!

**Comment by lemoncurry** [comment]
I've written similar test cases for `long`, `boolean`, and `double`, and all three of them suffered an exception of the form

```
java.lang.UnsupportedOperationException: removed token(false) vs "false"   <28>   (line 1,col 1)-(line 1,col 5)
```

After looking into the debugger a bit I think the problem lies somewhere else for these cases and `int` and `String` just got 'lucky'. I suspect that `2L`is parsed as an identifier instead of as a long literal when passing it in via `setValue`, and the same thing happens for `true` and `3.0` as well.

I recommend opening another issue to track these problems there.

File: javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java [hunk 1]
```diff
@@ -1334,4 +1335,56 @@ void commentAddedAtTopLevel() {
...
+        final CompilationUnit b = javaParser.parse(code).getResult().orElseThrow(AssertionError::new);
+        LexicalPreservingPrinter.setup(b);
+
+        assertEquals(1, b.findAll(ClassOrInterfaceDeclaration.class).size());
+        b.findAll(ClassOrInterfaceDeclaration.class).forEach(coid -> coid.setName("B"));
+
+        final String expected = "class B {}";
+
+        final String actual = LexicalPreservingPrinter.print(b);
+        assertEquals(expected, actual);
```
MysterAitch, line 1371: This test passes with v3.15.22 before any fixes.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
```diff
@@ -1334,4 +1335,56 @@ void commentAddedAtTopLevel() {
...
         assertEqualsNoEol("package x;class X{}", LexicalPreservingPrinter.print(cu));
     }
+
+    @Test
+    public void testReplaceStringLiteral() {
+        final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLexicalPreservationEnabled(true));
+
+        final String code = "\"asd\"";
+        final Expression b = javaParser.parseExpression(code).getResult().orElseThrow(AssertionError::new);
+        LexicalPreservingPrinter.setup(b);
```
MysterAitch, line 1345: `         final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLexicalPreservationEnabled(true));`

`         LexicalPreservingPrinter.setup(b);`

We're setting up the printer twice here -- once as part of the configuration, and then again afterwards.

MysterAitch: I'll do this in a follow up PR handling the remaining literal values.

PR #2646, opened 2020-05-03
Handle adding a expression after a traling comment

### LINKED ISSUE #2620: LexicalPreservingPrinter print issue when adding a new element
In the latest version, when we add  an element in first position in the ClassOrInterfacedeclaration members list, we are facing to a issue linked to LexicalPreservingPrinter.

                "  public class Foo { //comment\n" +
                "    String a;\n" +
                "  }\n"
                );
        // create a new field declaration
        VariableDeclarator variable = new VariableDeclarator(new ClassOrInterfaceType("String"), "b");
        FieldDeclaration fd = new FieldDeclaration(new NodeList(Modifier.privateModifier(), Modifier.staticModifier()), variable);
        Optional<ClassOrInterfaceDeclaration> cd = cu.findFirst(ClassOrInterfaceDeclaration.class);
        // add the new variable
        cd.get().getMembers().addFirst(fd);
        // should be printed like this
       // but the result is
        System.out.println(LexicalPreservingPrinter.print(cu));

should be printed like this

    public class Foo {
       private static String b;
       String a;
    }

but the result is

     public class Foo { //commentprivate static String b;
        String a;
     }

jlerbsc:
Fixes #2620.


MysterAitch:
@jjbr -- Could you take another look at this please? 

Copying _only_ the text case over to a clone of master results in the test passing -- no need to apply the fix.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -708,6 +708,16 @@ private void applyAddedDiffElement(Added added) {
                 // We want to adjust the indentation while considering the new element that we added
                 originalIndex = adjustIndentation(indentation, nodeText, originalIndex, false);
                 originalIndex++; // Now we can increment
+            } else
```
MysterAitch: Please always include braces.

jlerbsc:
@MysterAitch here issues are related to EOL character used.

MysterAitch:
I'll keep this open, as there is still a problem present when running the original test case.

I've narrowed the difference down to the use of `\n` in #2620 vs `EOL` in the test case here.

MysterAitch:
Seems to fail on Windows, but be fine on Mac/Ubuntu.

Just tweaked the tests to run with all three variations of line separator to see if it's just coincidence that allowed them to pass or if it's the mismatch that is problematic.

image

MysterAitch:
Digging deeper, @jlerbsc 's fix as given seems to work fine for fixing the problem presented in #2620 , but the tests fail due to the injected line endings not matching the parsed text.

For example if the parsed text has `\r\n` line endings, then on Ubuntu the host`s line endings (`\n`) are injected -- ONLY on the injected lines.

I am unsure if this is a separate bug / issue that should be handled separately, or if it should be handled as part of this.

### Ubuntu test failures
https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201141#step:6:1505
_(line breaks / spacing added for clarity)_


```
 [ERROR] Failures: 
[ERROR]   Issue2620Test.testWithCr:49->doTest:110 
expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
 but was: <    public class Foo { //comment\r    private String newField;\n    \n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>

[ERROR]   Issue2620Test.testWithCrLf:59->doTest:110
 expected: <    public class Foo { //comment\r\n    private String newField;\r\n    \r\n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
 but was: <    public class Foo { //comment\r\n    private String newField;\n    \n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>

```


### Mac test failures
https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201231#step:6:1499
_(line breaks / spacing added for clarity)_

```
[ERROR] Failures: 
[ERROR]   Issue2620Test.testWithCr:49->doTest:110 

expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
 but was: <    public class Foo { //comment\r    private String newField;\n    \n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>

[ERROR]   Issue2620Test.testWithCrLf:59->doTest:110 
expected: <    public class Foo { //comment\r\n    private String newField;\r\n    \r\n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
 but was: <    public class Foo { //comment\r\n    private String newField;\n    \n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>


```

### Windows test failures
https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201340#step:6:1505
_(line breaks / spacing added for clarity)_

```
[ERROR] Failures: 
[ERROR]   Issue2620Test.testWithCr:49->doTest:110
expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
 but was: <    public class Foo { //comment\r    private String newField;\r\n    \r\n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>

[ERROR]   Issue2620Test.testWithLf:54->doTest:110 
expected: <    public class Foo { //comment\n    private String newField;\n    \n    private String a;\n        private String b;\n        private String c;\n        private String d;\n    }>
 but was: <    public class Foo { //comment\n    private String newField;\r\n    \r\n    private String a;\n        private String b;\n        private String c;\n        private String d;\n    }>

```

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
MysterAitch: I've just pushed a tidied version. Comment resolved.

jlerbsc:
What is the state of the fix? Can it be integrated in the next version?

MysterAitch:
> What is the state of the fix? Can it be integrated in the next version?

Good question! 

Apologies, I'd gotten distracted with other things. 

IIRC the only thing holding this up is the line endings thing, but that can be split out into a separate issue rather than hold this up.

The assertions here will need to be switched to the equality check that ignores line endings though.

Otherwise, no reason it can't be in the next release :)

PR #2654, opened 2020-05-06
Fix Issue 2445 Pattern Matching for instanceof

### LINKED ISSUE #2445: JEP 305 / JEP 375: Pattern Matching for instanceof (Preview, and 2nd Preview)
https://bugs.openjdk.java.net/browse/JDK-8181287

- grammar change (in addition to the current `x instance Y` we should be able to say `x instance Y y`)
- validation that the second form cannot be used before Java 14.

MysterAitch:
This work in progress is an edit to the grammar to optionally include an identifier within the instanceofexpr

I'm not sure this is the right approach (maybe there should be a `PatternExpr` or similar instead of just adding a name to the `InstanceOfExpr`), but it works so far ;)

#2445

**Parsing**
- [X] Grammar changes
- [X] Generator changes
- [X] JavaDoc for InstanceOfExpr and PatternExpr (new)
- [X] Validator disallowing instanceof patterns before Java 14

**Resolution**
- [X] Have the variables declared as part of the Pattern be in scope for the block immediately following
- [X] Have the variables declared as part of the Pattern be in scope for children of an IfStmt 
- [x] Have the variables declared as part of the Pattern OUT of scope for all other blocks within an if/else if/else chain 
_(e.g. `if(obj instanceof String s) {} else { s.contains("foo"); }`)_
- [x] Have the variables declare as part of the Pattern be made available within the remainder of the condition 
_(e.g. `if(obj instanceof String s && s.length() > 0) {}`)_
- [x] Have the variables declare as part of the Pattern be OUT OF SCOPE for the remainder of the condition everywhere else 
_(e.g. `if(obj instanceof String s || s.length() > 0) {}` and `if(s.length() > 0 && obj instanceof String s) {}`)_
- [x] Correctly handle multiple uses within the same condition 
_(e.g. `if(obj instanceof String s && otherObj instanceof String s2) {}`)_
- [x] Consider cases where the instanceof is a sibling rather than being used in a parent node such as an if condition.
_(e.g. `boolean result = obj instanceof String s; System.out.println(s);`)_
- [x] Consider cases where we're not inside of a block
_(e.g. `if(obj instanceof String s) System.out.println(s)` and `if(obj instanceof String s) {} else System.out.println(s)` )_
- [ ] Figure out if the Javassist stuff and other JavaparserX bits need editing


javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/InstanceOfExprTest.java
```diff
@@ -21,16 +21,97 @@
...
+     *     .. s.contains(..) ..
+     * } else {
+     *     .. s.contains(..) ..
+     * }
+     * }
+     * </pre>
+     *
+     * Allowed / in scope: {@code if (obj instanceof String s && s.length() > 5) {..}}
+     * Not in scope:       {@code if (obj instanceof String s || s.length() > 5) {..}}
+     *
```
matozoid, line 133: I wouldn't add validations for that, I bet they're going to get really complex. Avoid building a compiler :-)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
```diff
@@ -37,10 +37,13 @@
 import java.util.function.Consumer;
 import java.util.Optional;
 import com.github.javaparser.ast.Generated;
+import com.github.javaparser.metamodel.OptionalProperty;
 
 /**
  * Usage of the instanceof operator.
  * <br>{@code tool instanceof Drill}
+ * <br>JEP305 introduced the ability to assign a name to a matched node.
```
matozoid: If you take a look at other Javadocs, you'll see a structure:
1. general description
2. java version
3. specifics for the java version
4. small samples
5. an explanation how the parts of the samples map to the properties of the node
6. goto 2 :-)

I don't think JEP-references are in there, but they seem nice to add. I bet users think in Java versions, not in JEPs :-)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
```diff
@@ -50,15 +53,22 @@
...
+    @OptionalProperty
+    private SimpleName name;
+
     public InstanceOfExpr() {
         this(null, new NameExpr(), new ClassOrInterfaceType());
     }
 
-    @AllFieldsConstructor
     public InstanceOfExpr(final Expression expression, final ReferenceType type) {
         this(null, expression, type);
```
matozoid: In all the other nodes, all the constructors call the most specific constructor to avoid constructor spaghetti. That would mean the this-call gets "null" at the end for example.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
```diff
@@ -50,15 +53,22 @@
 
     private ReferenceType type;
 
+    @OptionalProperty
+    private SimpleName name;
```
matozoid: That seems like a simple and cheap solution. Nice!

javaparser-core/src/main/java/com/github/javaparser/metamodel/InstanceOfExprMetaModel.java
```diff
@@ -30,5 +30,7 @@
 
     public PropertyMetaModel expressionPropertyMetaModel;
 
+    public PropertyMetaModel namePropertyMetaModel;
+
```
matozoid, line 47: Here is a good example that the metamodel also doesn't have the properties in the right order :-( But that's another issue.

javaparser-core/src/main/javacc/java.jj
```diff
@@ -905,40 +917,40 @@ BodyDeclaration<?> ClassOrInterfaceBodyDeclaration():
...
 FieldDeclaration FieldDeclaration(ModifierHolder modifier):
 {
-	Type partialType;
-	NodeList<VariableDeclarator> variables = new NodeList<VariableDeclarator>();
-	VariableDeclarator val;
+    Type partialType;
+    NodeList<VariableDeclarator> variables = new NodeList<VariableDeclarator>();
+    VariableDeclarator val;
 }
 {
```
matozoid: A certain someone mashed the format-file key, it seems :-#

javaparser-core/src/main/javacc/java.jj
```diff
@@ -1470,542 +1501,562 @@ Expression Expression():
...
+    SimpleName name;
 }
 {
-    ret = RelationalExpression() [ "instanceof" type = AnnotatedReferenceType() {
-        ret = new InstanceOfExpr(range(ret, token()), ret, type); 
-    } ]
+    ret = RelationalExpression()
+    [
+        "instanceof"
+        (LOOKAHEAD(AnnotatedReferenceType() SimpleName())
```
matozoid: Hey cool! It took me a year or two to understand how to use lookahead :-)

matozoid:
I think I would have replaced the "type" property with a more generic "expression" and stuck either a "nameexpr" or a... whatever "type+name" is called in there. That would require some precise Javadoc, and more casting for the user though.

MysterAitch:
> I think I would have replaced the "type" property with a more generic "expression" and stuck either a "nameexpr" or a... whatever "type+name" is called in there. That would require some precise Javadoc, and more casting for the user though.


@matozoid  -- It seems I have continued working while you were reading/feeding back -- I've almost replaced the "quick" optional hack with what the JEP refers to as a `Pattern` (is this an expression? or something else?)

```
RelationalExpression:
     ...
     RelationalExpression instanceof ReferenceType
     RelationalExpression instanceof Pattern

Pattern:
     ReferenceType Identifier
```

The `InstanceOfExpr` hasn't been updated yet -- that's where I stopped just now as I'm unsure whether to add in a new `InstanceOfPatternExpr` or similar, or whether to try and bend the existing `InstanceOfExpr` into having both available (instead the type/name are just passed in from the `Pattern`).

javaparser-core/src/main/java/com/github/javaparser/metamodel/InstanceOfExprMetaModel.java
MysterAitch: I'm not sure what you mean here? They seem to be sorted alphabetically, unless that's just coincidence.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
MysterAitch: Done :)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
MysterAitch: The javadocs were just a dumping ground of "stuff" with very little thought given to structure ;)

definitely needs some fleshing out and polishing!! 

JEPs are helpful I think, with something like `(since JDK14, preview since JDK13)` or whatever is applicable in each case.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/InstanceOfExprTest.java
MysterAitch: It will become relevant in the solver part I think, trying to resolve s.length()

But this is perhaps a compilation-time error..

At the very least we should include this as a note in the docs re: known limitations :)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
MysterAitch: Cheers! 

A "proper" fix likely needs to make use of the `Pattern` (defined in the JEP, but not the JLS), but just to get something functional I was chuffed to get this far :)

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -355,6 +355,12 @@ private static CsmElement typeArguments() {
                 child(ObservableProperty.TYPE)
         ));
 
+        concreteSyntaxModelByClass.put(PatternExpr.class, sequence(
+                child(ObservableProperty.TYPE),
```
matozoid, line 359: Here you can see one of the problems with the lexical preserving printer as it is now: shouldn't we say "comment()" in here? In what position? Dunno.

javaparser-core/src/main/javacc/java.jj
MysterAitch: ... If only -- all manual I'm afraid :cry: 

I spent longer than probably sounds reasonable trying to figure out what's what -- sorting out spacing and indentation as I went :D 

Happy to revert these unrelated changes now that I understand the difference between ()/[]/{} blocks and the structure :+1:

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/InstanceOfExprTest.java
matozoid: Ah yes, to keep from going crazy I always tried to delegate the symbol solver changes to someone else :-#

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
MysterAitch: Honestly, I don't have a clue -- at this point I'm aiming for "functional" rather than "correct" :wink: 

I suppose it might be a non-issue if any comments would be attached to the `ReferenceType` or `SimpleName` inside of the `PatternExpr` (if so, then they'll be printed when printing those?)... 

Probably needs some test cases thrown at it to figure out what the intended/actual behaviours are.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
matozoid: That's cool. Here are the latest and pretty extensive Javadocs that I wrote for inspiration: https://www.javadoc.io/doc/com.github.javaparser/javaparser-core/latest/com/github/javaparser/ast/expr/SwitchExpr.html https://www.javadoc.io/doc/com.github.javaparser/javaparser-core/latest/com/github/javaparser/ast/stmt/SwitchEntry.html

Preview status is very interesting - and the whole question of what to do with all these in-between Java versions like 9, 10, 12...

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
matozoid: Once you start digging into this, you'll find that many decisions about how the AST is formed are arbitrary, so don't feel like there is a right way :-) This way it is definitely easy to use for the end user!

javaparser-core/src/main/java/com/github/javaparser/metamodel/InstanceOfExprMetaModel.java
matozoid: Oh, the whole getChildNodes thing. Those nodes should be listed in order of occurrence in source code (so expression goes before type goes before name.) I was thinking the metamodel could help, but that also doesn't have things in the right order. In my opinion the metamodel should be the source of all meta-information about a type of node, so it should have the correct order.

Right now the Concrete Syntax Model has the right order, but that thing should be mostly redundant when you have the metamodel already.

matozoid:
The goal is not to copy the JLS, the goal is to make it easy for the user, using JLS naming where possible :-)

matozoid:
Cool, it's green already!

javaparser-core/src/main/javacc/java.jj
matozoid: Make it as you like it best :-) It just took me some time to find the grammar changes in the github diff. And I'm aware the formatting is horrible in the grammar, so good job

javaparser-core/src/main/java/com/github/javaparser/metamodel/InstanceOfExprMetaModel.java
MysterAitch: I think the order generated here is based on one of the bigger metamodel classes (it's missing the generated tags, but the one where the constructor parameters get added to the metamodel etc, and there's fields are initialised). 

That is where the sorting is applied, and presumably propagates out leading to this side effect we see here with the fields being ordered (they're added here in the order that they exist within the list).

That's probably where I'd start.

javaparser-core/src/main/javacc/java.jj
MysterAitch: For stuff like this I tend to really appreciate the intellij diffs -- there you can set it to ignore whitespace changes.

If you want to see git blame, right click on the gutter with the lite numbers and use "annotate". That can be configured to also ignore whitespace only changes and even moves within the file.

That's been handy a few times in trying to figure out which commit made a particular addition/change, avoiding the noise of whitespace only changes.

MysterAitch:
> The goal is not to copy the JLS, the goal is to make it easy for the user, using JLS naming where possible :-)

I plan to keep the getName() method, just delegate it to the Pattern object instead.

Hopefully this will lead to the best of both worlds! :)

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
MysterAitch: Those are rather nice docs! :1st_place_medal:

MysterAitch:
Symbol resolution is now working, but with one substantial exception -- attempting to resolve `s` within the `else` block discovers the `if()` stmt as a parent thus (wrongly) resolves `s` to be a `String`.

```java
class X {
  public X() {
    boolean result;
    String obj = "abc";
    if (!(obj instanceof String s) && true) {
        result = s.contains("b");
    } else {
        result = s.contains("error");
    }
  }
 }
```

The reason this is problematic is the following example, where the second reference to `s` (i.e. within the `else` block) should resolve `s` to be a `List`, not a `String`).

```java
class X {
  public X() {
    List<Integer> s;
    boolean result;
    String obj = "abc";
    if (!(obj instanceof String s) && true) {
        result = s.contains("b");
    } else {
        result = s.contains("error");
    }
  }
 }
```

MysterAitch:
I've just added a checklist of things that have been completed and some notes on what is left to do. 

Summary: 
- Parsing and validation is complete as far as I'm aware (likely need to do a sweep / tidy up), 
- Resolution is functional for inside if/elseif/else
- There are bunch of other cases that might not yet work:
  - e.g. not "seeing" where it is used within a variable as opposed to an ifstmt condition, 
  - e.g. being too "greedy"/overzealous in assuming that the variable is available EVERYWHERE in an ifstmt (and all children e.g. other branches of the if/elseif/else chain

Some of the scoping stuff can potentially be saved for another PR if I don't crack it soon (e.g. fixing some of the scoping of variables in if/elseif/else).

javaparser-core/src/main/java/com/github/javaparser/metamodel/InstanceOfExprMetaModel.java
matozoid: Yeah, it seems that sorting there wasn't such a good idea in retrospect. I probably made that, but I wasn't aware of these requirements yet :-)

javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java1_0Validator.java
```diff
@@ -124,6 +124,13 @@
             (n, reporter) -> reporter.report(n, "Switch expressions are not supported.")
     );
 
+
+    final Validator noInstanceOfPattern = new SimpleValidator<>(InstanceOfExpr.class,
+            n -> n.getPattern().isPresent(),
+            (n, reporter) -> reporter.report(n, "Use of patterns with instanceof is not supported.")
```
matozoid, line 130: One thing that could be improved for validations is to use a single treewalk that does all validations, instead of a walk for every validation. I wouldn't prioritize that very highly though.

MysterAitch:
Hmm... while the scoping is now mostly working for if/elseif/else, I seem to have broken something else...

javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java1_0Validator.java
MysterAitch: #2658

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
```diff
@@ -46,7 +46,7 @@ public static boolean equals(final Node n, final Node n2) {
     }
 
     private EqualsVisitor() {
-    // hide constructor
+        // hide constructor
```
MysterAitch: @jlerbsc 's fix in #2918 in action - thanks! :smiling_face_with_three_hearts:

MysterAitch:
I need to sweep through re: any todo's and fixme's (and the checklist at the top of this PR) but I think that pattern matching (compilation and solving) is pretty much there! :+1: :tada: :partying_face: 

Even if there are a few solving issues to work through, I'm inclined to merge this and then fix those other cases separately. 

If there's anything not yet tested, it'll be stuff like making sure the scope is correct for sibling blocks e.g. if/else if/else and negated conditionals :+1:

matozoid:
Great work!

MysterAitch:
It's not **absolutely** perfect (details will be added as comments to #2445 e.g. flow scoping), but it is more than "good enough".

I'll squash the commits and merge it momentarily.

PR #2675, opened 2020-05-15
Add symbol solver support for variadic parameters given zero or more than one argument

hfreeb:
Currently, varargs in the symbol solver seem to only really be supported when the number of parameters still equals the number of arguments. This PR fixes that.

~~One caveat is that I don't believe a vararg of arrays will work properly since if an array is the only argument for the variadic parameter, then it is compared with the type of the variadic parameter itself, and not its component type (I found this assumption in other places too).~~

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,8 @@ Next Release (Version 3.15.23)
     (PR [#2679](https://github.com/javaparser/javaparser/pull/2679), by [@MysterAitch](https://github.com/MysterAitch))
 * BREAKING CHANGE: Tokens relating to literal values now have the category of `JavaToken.Category.LITERAL` (previously `JavaToken.Category.KEYWORD`) 
     (PR [#2679](https://github.com/javaparser/javaparser/pull/2679), by [@MysterAitch](https://github.com/MysterAitch))
+* FIXED: Add symbol solver support for variadic parameters given zero or more than one argument, and when an array is given
+    (PR [#2675](https://github.com/javaparser/javaparser/pull/2675), by [@hfreeb](https://github.com/hfreeb))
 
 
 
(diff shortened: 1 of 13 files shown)


MysterAitch:
When I look at PRs I try to run the tests on master before then also apply the fixes.

It seems that the tests all pass even without making the changes to `MethodCallExprContext.java` -- I've just pushed a commit with many of the changes commented out.

@hfreeb - it's an area I don't fully understand what's needed (or not) -- could you take another look and advise please? :smile:

hfreeb:
I've removed the comments and added a new test case which does require the changes to MethodResolutionLogic.
I'm sure that the test case could be generalised, but I'm not sure what determines which of the two MethodResolutionLogic#isApplicable methods is called.

Also, I fixed the caveat I mentioned in the PR text, so I believe this will now work as expected for a vararg of arrays.

MysterAitch:
I got a bit carried away in trying to understand what was going on, and added in a bunch of comments / renamed variables to be more explicit etc., so hopefully it's also now a bit more self-documenting.

Really grateful to @hfreeb for getting this started! This is your PR though -- please do take a look and make any changes you think are appropriate!

hfreeb:
> 
> 
> I got a bit carried away in trying to understand what was going on, and added in a bunch of comments / renamed variables to be more explicit etc., so hopefully it's also now a bit more self-documenting.
> 
> Really grateful to @hfreeb for getting this started! This is your PR though -- please do take a look and make any changes you think are appropriate!

The changes look good to me, the comments should definitely help the next person trying to figure all this out!

MysterAitch:
Cool :)

I'll merge it if you're happy to give the okay to do so!

hfreeb:
> 
> 
> Cool :)
> 
> I'll merge it if you're happy to give the okay to do so!

Yep, sounds good to me!

PR #2685, opened 2020-05-19
Enhanced EOL support

### LINKED ISSUE #2647: Lexical Preservaton -- Injected code uses host's EOL not the source's
Looks like @jlerbsc 's PR in #2646 fixes the issue described (thank you!), but it brings to the fore some other JavaParser behaviours that I am not sure are correct.

It seems that generated code is injected using the host's line endings. This is normally fine, but if the input text has a different line encoding then this leads to a source with mixed line endings. See the quoted post below, and the linked PR, for details.

https://github.com/javaparser/javaparser/pull/2646#issuecomment-623701912

> 
> 
> Digging deeper, @jlerbsc 's fix as given seems to work fine for fixing the problem presented in #2620 , but the tests fail due to the injected line endings not matching the parsed text.
> 
> For example if the parsed text has `\r\n` line endings, then on Ubuntu the host`s line endings (`\n`) are injected -- ONLY on the injected lines.
> 
> I am unsure if this is a separate bug / issue that should be handled separately, or if it should be handled as part of this.
> ### Ubuntu test failures
> 
> https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201141#step:6:1505
> _(line breaks / spacing added for clarity)_
> 
> ```
>  [ERROR] Failures: 
> [ERROR]   Issue2620Test.testWithCr:49->doTest:110 
> expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
>  but was: <    public class Foo { //comment\r    private String newField;\n    \n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
> 
> [ERROR]   Issue2620Test.testWithCrLf:59->doTest:110
>  expected: <    public class Foo { //comment\r\n    private String newField;\r\n    \r\n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
>  but was: <    public class Foo { //comment\r\n    private String newField;\n    \n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
> ```
> 
> ### Mac test failures
> 
> https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201231#step:6:1499
> _(line breaks / spacing added for clarity)_
> 
> ```
> [ERROR] Failures: 
> [ERROR]   Issue2620Test.testWithCr:49->doTest:110 
> 
> expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
>  but was: <    public class Foo { //comment\r    private String newField;\n    \n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
> 
> [ERROR]   Issue2620Test.testWithCrLf:59->doTest:110 
> expected: <    public class Foo { //comment\r\n    private String newField;\r\n    \r\n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
>  but was: <    public class Foo { //comment\r\n    private String newField;\n    \n    private String a;\r\n        private String b;\r\n        private String c;\r\n        private String d;\r\n    }>
> ```
> 
> ### Windows test failures
> 
> https://github.com/javaparser/javaparser/pull/2646/checks?check_run_id=644201340#step:6:1505
> _(line breaks / spacing added for clarity)_
> 
> ```
> [ERROR] Failures: 
> [ERROR]   Issue2620Test.testWithCr:49->doTest:110
> expected: <    public class Foo { //comment\r    private String newField;\r    \r    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
>  but was: <    public class Foo { //comment\r    private String newField;\r\n    \r\n    private String a;\r        private String b;\r        private String c;\r        private String d;\r    }>
> 
> [ERROR]   Issue2620Test.testWithLf:54->doTest:110 
> expected: <    public class Foo { //comment\n    private String newField;\n    \n    private String a;\n        private String b;\n        private String c;\n        private String d;\n    }>
>  but was: <    public class Foo { //comment\n    private String newField;\r\n    \r\n    private String a;\n        private String b;\n        private String c;\n        private String d;\n    }>
> ```

_Originally posted by @MysterAitch in https://github.com/javaparser/javaparser/issues/2620#issuecomment-623703521_

MysterAitch:
Fixes #2647 

This PR adds an enum to represent different types of end-of-line / line-separator characters. 

- It also adds a utility method which allows you to "detect" which style of line-separator a given string uses.

- This is incorporated into the parsing process by reading in the content of the given provider and adding this as a data field within the top-most node. 

- This is then used as part of the CSM and pretty printer where new added nodes will reflect this (defaulting to the system's eol character if it is unknown/ambiguous e.g. it is a single line or there are mixed endings).

Reading in the provider content and then creating a brand new `StringProvider` using this string is a bit "hacky" and I'm sure there must be a better way to do it, I just don't know what it is -- comments and suggestions welcomed!


javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -114,6 +132,18 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
         assertNotNull(start);
         assertNotNull(provider);
 
+        // FIXME: Surely there's a better way...?
+        String x = providerToString(provider);
+        // close the old provider then open a new one...
+        try {
+            provider.close();
```
MysterAitch: Surely there's a better way to achieve this than consuming the whole stream, interpreting it, and then forwarding a brand new stream (`provider`) based on this new string... 

Note that the original stream will be unavailable due to having been fully consumed.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -123,6 +153,10 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
             N resultNode = start.parse(parser);
```
MysterAitch: Note that at this point, the line endings have already defaulted to the system's EOL, meaning that post processors don't seem to have access to the "original" underlying line separators.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmElement.java
```diff
@@ -99,7 +99,11 @@ static CsmElement semicolon() {
     static CsmElement comment() { return new CsmComment(); }
 
     static CsmElement newline() {
-        return new CsmToken(eolTokenKind(), EOL);
+        return newline(LineEnding.SYSTEM);
+    }
+
+    static CsmElement newline(LineEnding lineEnding) {
+        return new CsmToken(eolTokenKind(lineEnding), lineEnding.toString());
```
MysterAitch: Not passing in a parameter here defaults to the system's EOL.

javaparser-core-testing/src/test/java/com/github/javaparser/utils/TestUtils.java
```diff
@@ -194,13 +194,13 @@ public static void assertExpressionValid(String expression) {
      * Assert that "actual" equals "expected", and that any EOL characters in "actual" are correct for the platform.
      */
     public static void assertEqualsNoEol(String expected, String actual) {
-        assertEquals(normalizeEolInTextBlock(expected, EOL), actual);
+        assertEquals(normalizeEolInTextBlock(expected, EOL), normalizeEolInTextBlock(actual, EOL));
```
MysterAitch: My understanding/interpretation of the method name `assertEqualsNoEol` is that it should assert that everything except EOL is equal. Given this, my inclination is that we should normalise to `LineSeparator.ARBITRARY` to explicitly indicate this. 

Note that this would contradict the javadoc which **ALSO** specifies ".. and that any EOL characters in "actual" are correct for the platform.". My preference here would be to have a separate assertion for this.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -100,6 +101,23 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
         return astParser;
     }
 
+    private String providerToString(Provider provider) {
+        // FIXME: Surely there's a better way...?
```
MysterAitch: Commenting this line to flag this up for attention -- this method consumes the whole provider stream and returns a string.

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -722,6 +757,9 @@ protected SymbolResolver getSymbolResolver() {
     public static final DataKey<SymbolResolver> SYMBOL_RESOLVER_KEY = new DataKey<SymbolResolver>() {
     };
 
+    public static final DataKey<LineEnding> LINE_ENDING_KEY = new DataKey<LineEnding>() {
```
MysterAitch: Just noticed this. It might be able to be made private -- originally there wasn't an accessor method on Node. Maybe we can also add in a mutator / setter method.

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
```diff
@@ -51,19 +66,23 @@ public static boolean isComment(int kind) {
...
+        if (lineEnding.equalsString(LineEnding.LF)) {
             return UNIX_EOL;
         }
-        if (EOL.equals("\r\n")) {
+        if (lineEnding.equalsString(LineEnding.CRLF)) {
             return WINDOWS_EOL;
         }
-        if (EOL.equals("\r")) {
+        if (lineEnding.equalsString(LineEnding.CR)) {
             return OLD_MAC_EOL;
```
matozoid, line 77: Would it be a bad idea to put the corresponding tokens in the enum too?

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
matozoid: I think you want to create a preprocessor for this, like the unicode escaping preprocessor. Then add a configuration option so those who don't care about this feature can turn it off :-)

javaparser-core-testing/src/test/java/com/github/javaparser/utils/TestUtils.java
matozoid: It is there to compare a string with "\n"'s in it to some output which contains some kind of line ending, which is normalized to "\n" so they match. If I remember correctly :-)

matozoid:
Random informational remark: the textblock feature of Java 12 or 13 or so is very explicit about the kind of eols it wants. Be careful not to break it :-)

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
MysterAitch: I hesitated there because `WINDOWS_EOL` AND `OLD_MAC_EOL` refer to static ints within the grammar. 

The other aspect that caused hesitation there is to keep the enum relatively "pure" and not needing to (necessarily) ever be aware of the grammar stuff.

Neither are show-stoppers though.

MysterAitch: Unless, of course, I'm misunderstanding you?

javaparser-core-testing/src/test/java/com/github/javaparser/utils/TestUtils.java
MysterAitch: For the purpose of this assert (and how I've ever used it based on the name), the important bit is that they're the **same** line endings as opposed to being any particular line ending.

The javadoc suggests otherwise though -- that it should match the system's line ending.

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
matozoid: Nah, I think you understand right and it's a fine decision. It made me think of this thing by the way: https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/JavaToken.java - it is my attempt to define the definitive "token" class, so that all other weird int-tokens and methods operating on them could be removed. Nothing relevant now, just info.

MysterAitch:
> 
> 
> Random informational remark: the textblock feature of Java 12 or 13 or so is very explicit about the kind of eols it wants. Be careful not to break it :-)

It would be a shame if our test suite wasn't able to highlight any regression like this ;)

But yes -- Just had a read through the JEP and found this: 

https://openjdk.java.net/jeps/378#1--Line-terminators


> The content may include line terminators directly, unlike the characters in a string literal. The use of \n in a text block is permitted, but not necessary or recommended. For example, the text block:
> 
> """
> line 1
> line 2
> line 3
> """
> 
> is equivalent to the string literal:
> 
> "line 1\nline 2\nline 3\n"
> 
> or a concatenation of string literals:
> 
> "line 1\n" +
> "line 2\n" +
> "line 3\n"
> 
> If a line terminator is not required at the end of the string, then the closing delimiter can be placed on the last line of content. For example, the text block:
> 
> """
> line 1
> line 2
> line 3"""
> 
> is equivalent to the string literal:
> 
> "line 1\nline 2\nline 3"

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmTextBlock.java
```diff
@@ -34,7 +35,8 @@ public CsmTextBlock(ObservableProperty property) {
 
     @Override
     public void prettyPrint(Node node, SourcePrinter printer) {
-        printer.print("\"\"\"\n");
+        LineEnding lineEnding = node.getLineEndingStyleOrDefault(LineEnding.SYSTEM);
+        printer.print("\"\"\"" + lineEnding.toString());
```
MysterAitch: As highlighted by @matozoid , textblocks are picky about line endings 

https://github.com/javaparser/javaparser/pull/2685#issuecomment-631008605

https://openjdk.java.net/jeps/378#1--Line-terminators

MysterAitch: Hmm, It has just occurred to me (in the couple mins since posting my reply) that maybe the **VALUE** is always interpreted as \n, but it is unclear to me whether the **source code** within a java file can have different line endings.

Regardless, it needs some more attention +- a comment to explain.

javaparser-core-testing/src/test/java/com/github/javaparser/utils/TestUtils.java
matozoid: Ah yes, the other way around. Well, the important thing is that it is really nice to be able to say "\n" in an expectation and not have to bother with Windows/Unix differences. Yeah, I see what you're doing now. Okay.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmTextBlock.java
matozoid: That should definitely remain "\n", textblocks never generate "\r\n" or so.

MysterAitch: Ah yes, just realised that at this point we're it the CSM dealing with interpreted values (as opposed to source code) so my last comment isn't applicable.

matozoid: And the change we're discussing breaks text blocks, because now they generate platform specific EOL's, which they should not do :-)

MysterAitch: Reverted this and added a line comment to explain

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
MysterAitch: I didn't really understand the unicode preprocessor(!)

What I could determine is that it modifies the stream, but isn't able to store any state or pass any results on..? Or maybe I misunderstood.

The other aspect (I can't remember if I mentioned this in another comment) is that during the pre-processing stage the processors don't seem to have access to either the config or the the resulting nodes.

Presumably it's inappropriate to store this on the processor itself... 

If that's something you can give guidance on, then that would be great! Otherwise I'll ruminate and think on this some more :)

MysterAitch:
Hmm... as far as I can tell, there is no reason for the tests to fail on appveyor (consistently) but then pass on all other places...

As far as I can see, the "expected" and "actual" sections are identical, but there's no information if there are line separator differences... I'll push a revised test in a moment.

```
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   CommentsInserterTest.issue412:114 expected: </* 1
1 */
/*
2
2
 */
package /* a */
/* b */
// c
// d
com.pany/*alma*/
/*k�rte*/
// l�fasz
// j�ska
.experiment;
// z
// x
/*y*/
/*w*/
/*aa*/
/*bb*/
// cc
// dd
public class Main {
    public static void main(String[] args) throws FileNotFoundException, IOException, ParseException {
    // try (FileInputStream fisTargetFile = new FileInputStream(new File(FILENAME))) {
    // final String content = IOUtils.toString(fisTargetFile, "UTF-8");
    // System.out.println(content);
    // }
    }
}
> but was: </* 1
1 */
/*
2
2
 */
package /* a */
/* b */
// c
// d
com.pany/*alma*/
/*k�rte*/
// l�fasz
// j�ska
.experiment;
// z
// x
/*y*/
/*w*/
/*aa*/
/*bb*/
// cc
// dd
public class Main {
    public static void main(String[] args) throws FileNotFoundException, IOException, ParseException {
    // try (FileInputStream fisTargetFile = new FileInputStream(new File(FILENAME))) {
    // final String content = IOUtils.toString(fisTargetFile, "UTF-8");
    // System.out.println(content);
    // }
    }
}
>
[INFO] 
[ERROR] Tests run: 1085, Failures: 1, Errors: 0, Skipped: 6
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for javaparser-parent 3.15.23-SNAPSHOT:
[INFO] 
[INFO] javaparser-parent .................................. SUCCESS [  1.594 s]
[INFO] javaparser-core .................................... SUCCESS [ 27.814 s]
[INFO] javaparser-core-testing ............................ FAILURE [01:02 min]
[INFO] javaparser-core-testing-bdd ........................ SKIPPED
[INFO] javaparser-core-generators ......................... SKIPPED
[INFO] javaparser-core-metamodel-generator ................ SKIPPED
[INFO] javaparser-core-serialization ...................... SKIPPED
[INFO] javaparser-symbol-solver-core ...................... SKIPPED
[INFO] javaparser-symbol-solver-testing ................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:31 min
[INFO] Finished at: 2020-05-22T16:57:37Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.0.0-M4:test (default-test) on project javaparser-core-testing: There are test failures.
[ERROR] 
```

MysterAitch:
It now seems that lexical preserving printer tests are failing with odd placement of line endings (some tests show an extra line break, and some have it missing). 

I'll investigate further tomorrow... 

I still find it odd that the appveyor environment is the only one showing these test failures and that I'm unable to replicate locally. 

EDIT: As a reminder to self for something to investigate tomorrow, maybe the appveyor git configuration is not performing autocrlf changes, meaning that the checked out code is unexpectedly using the repo's line endings (`\n` iirc) as opposed to the system's line endings or something? That might explain why the issues is **only** being seen on appveyor.

MysterAitch:
> EDIT: As a reminder to self for something to investigate tomorrow, maybe the appveyor git configuration is not performing autocrlf changes, meaning that the checked out code is unexpectedly using the repo's line endings (`\n` iirc) as opposed to the system's line endings or something? That might explain why the issues is **only** being seen on appveyor.


Oh dear, I should have followed up on this passing thought for just a couple of minutes... cursory searching turned up many results about this...

> I read on another thread (http://help.appveyor.com/discussions/problems/671-coreautocrlf) the default configuration for the git autocrlf setting is "input". This means AppVeyor is checking out files with LF (\n) endings on Windows. That broke some of my tests. 

https://help.appveyor.com/discussions/problems/1119-allow-changing-git-autocrlf-setting

> The git core.autocrlf setting defaults to true (ie, all text files get
checked out as CRLF on Windows), except on Appveyor where's set to
"input" (ie, all text files get checked out with the upstream
repository's line endings, which for us typically means LF.)

https://gitlab.freedesktop.org/mesa/mesa/commit/9e5e3a8ead5321a6abc0add8f6e59e50052f9698

Adding this to the appveyor.yml has the tests passing :)

```yaml
init:
  # Setup autocrlf -- by default, appveyor uses autocrlf input
  # ... This affects tests which expect resource files to have the systems's line separator.
  - git config --global core.autocrlf true
```

https://ci.appveyor.com/project/ftomassetti/javaparser/builds/33067787

matozoid:
> Adding this to the appveyor.yml has the tests passing :)

Yeeeeessss, but the tests passed before, and shouldn't need an AppVeyor change. Can you reason why this PR broke all those tests in the first place?

One place to look for is files with test content. Those have some kind of EOL, and tests probably depend on that, or better: expect certain EOLs to be in there. That means AppVeyor was doing the right thing - leave the EOLs alone.

Note that the AppVeyor build exists for only one thing: check EOL behaviour on Windows!

matozoid:
I was notified that I was mentioned here or so? I don't see anything? Or was it a notification of a new commit maybe?

MysterAitch:
> > Adding this to the appveyor.yml has the tests passing :)
> 
> Yeeeeessss, but the tests passed before, and shouldn't need an AppVeyor change. Can you reason why this PR broke all those tests in the first place?
> 
> One place to look for is files with test content. Those have some kind of EOL, and tests probably depend on that, or better: expect certain EOLs to be in there. That means AppVeyor was doing the right thing - leave the EOLs alone.
> 
> Note that the AppVeyor build exists for only one thing: check EOL behaviour on Windows!

Basically appveyor checked out the code using the repo's line endings. That meant that the files all have \n endings instead of \r\n.

Once this PR started to respect the file's ACTUAL line endings (e.g. within lexical preservation tests), it broke the tests.


Setting autocrlf to true causes appveyor to checkout the code using the host's line separator again :smile:.

MysterAitch:
From chat: 

> Danny van Bruggen @matozoid 19:02
> Yeah, I think you can simply register it as a pre- and post-processor. Look through the stream while preprocessing, store the eol info inside the processor, and in the post-processor you have the result object so you can store the eol info wherever you're storing it currently.
> 
> The only difficult thing is handling the JavaCC Provider implementation :-# That interface is begging for one-off errors.

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
MysterAitch: I've done removed this and replaced it with a preprocessor - marking as "resolved".

javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
MysterAitch: I've done removed this and replaced it with a preprocessor - marking as "resolved".

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
MysterAitch: resolved -- leaving as public

javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
MysterAitch: Plan -- leaving as-is for now.. it can be improved in a future release.

matozoid:
Well done @MysterAitch ! This is a fundamental improvement!

PR #2689, opened 2020-05-21
Added documentation re: language features at each JDK version

MysterAitch:
Still need to add in some version numbers re: when JavaParser first supported various language features, but it's otherwise pretty complete.


MysterAitch:
Looks like footnotes didn't work as intended...

This suggests an alternative that supposedly does work with github markdown: https://stackoverflow.com/a/32119820/429733

javaparser-core/src/main/java/com/github/javaparser/ast/validator/Java8Validator.java
```diff
@@ -25,7 +25,7 @@
 import com.github.javaparser.ast.validator.chunks.ModifierValidator;
 
 /**
- * This validator validates according to Java 7 syntax rules.
+ * This validator validates according to Java 8 syntax rules.
```
matozoid, line 28: Woops!

PR #2693, opened 2020-05-22
Issue 2669 infinite recursion object javassist for java.lang.Object supertype

### LINKED ISSUE #2669: Infinite recursin on "java.lang.Object"
When the JDK is used with ""JarTypeSolver""  and not with a "ClassLoaderTypeSolver", there is a problem in the class **JavassistClassDeclaration**:

1) the method **getSuperClass()** generates a "StackOverflow" when it try to retrieve the superclass of `java.lang.Object`, because the superclass of `java.lang.Object` is itself.
It is possible to resolve adding and extra check:

```java
@Override
    public ResolvedReferenceType getSuperClass() {
        try {
            if ("java.lang.Object".equals(ctClass.getName())) {
                return null;
            }
            if (ctClass.getClassFile().getSuperclass() == null) ...
    }
```

It is necessary to modify also the method **solveMethod()** to consider that **getSuperClass()** can return null:

```java
@Override
    @Deprecated
    public SymbolReference<ResolvedMethodDeclaration> solveMethod(String name, List<ResolvedType> argumentsTypes, boolean staticOnly) {
        ...
        ResolvedReferenceType rrt = getSuperClass();
        if (rrt == null) {
            return SymbolReference.unsolved(ResolvedMethodDeclaration.class);
        }
        SymbolReference<ResolvedMethodDeclaration> superClassMethodRef =  ...
    }
```

MysterAitch:
Fixes #2669.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
@@ -33,8 +33,10 @@ import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 public interface NodeWithExtends<N extends Node> {
 
     /**
-     * @return All extended types that have been explicitly added.
+     * @return All extended types that have been explicitly added (thus exist within the AST).
+     *   Note that this can contain more than one item if this is an interface.
      *   Note that this will not include {@code java.lang.Object} unless it is explicitly added (e.g. {@code class X extends Object {}})
+     *   If you want the implicitly extended types, you will need a resolved reference.
      */
     NodeList<ClassOrInterfaceType> getExtendedTypes();
 
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
@@ -33,7 +33,8 @@ import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
 public interface NodeWithExtends<N extends Node> {
 
     /**
-     * @return All extended types, through to (but excluding java.lang.Object)
+     * @return All extended types that have been explicitly added.
+     *   Note that this will not include {@code java.lang.Object} unless it is explicitly added (e.g. {@code class X extends Object {}})
      */
     NodeList<ClassOrInterfaceType> getExtendedTypes();
 
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
@@ -32,6 +32,9 @@ import static com.github.javaparser.StaticJavaParser.parseClassOrInterfaceType;
  */
 public interface NodeWithExtends<N extends Node> {
 
+    /**
+     * @return All extended types, through to (but excluding java.lang.Object)
+     */
     NodeList<ClassOrInterfaceType> getExtendedTypes();
 
     void tryAddImportToParentCompilationUnit(Class<?> clazz);
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -293,4 +293,13 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
                 && getQualifiedName().equals(java.lang.Object.class.getCanonicalName());
     }
 
+    /**
+     * @return true, if this represents {@code java.lang.Enum}
+     * @see ResolvedReferenceType#isJavaLangEnum()
+     */
+    default boolean isJavaLangEnum() {
+        return this.isEnum()
+                && getQualifiedName().equals(java.lang.Enum.class.getCanonicalName());
+    }
+
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -274,4 +274,21 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
     }
 
     List<ResolvedConstructorDeclaration> getConstructors();
+
+
+    /**
+     * We don't make this _ex_plicit in the data representation because that would affect codegen
+     * and make everything generate like {@code <T extends Object>} instead of {@code <T>}
+     *
+     * @return true, if this represents {@code java.lang.Object}
+     * @see ResolvedReferenceType#isJavaLangObject()
+     * @see <a href="https://github.com/javaparser/javaparser/issues/2044">https://github.com/javaparser/javaparser/issues/2044</a>
+     */
+    default boolean isJavaLangObject() {
+        return this.isClass()
+                && !isAnonymousClass()
+                && hasName() // Consider anonymous classes
+                && getQualifiedName().equals(java.lang.Object.class.getCanonicalName());
+    }
+
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -88,12 +88,12 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
 
     /**
      * The list of all the ancestors of the current declaration, direct and indirect.
-     * This list does not contains duplicates with the exacting same type parameters.
+     * This list does not contains duplicates with the exact same type parameters.
      */
     default List<ResolvedReferenceType> getAllAncestors() {
         List<ResolvedReferenceType> ancestors = new ArrayList<>();
         // We want to avoid infinite recursion in case of Object having Object as ancestor
-        if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
+        if (!isJavaLangObject()) {
             for (ResolvedReferenceType ancestor : getAncestors()) {
                 ancestors.add(ancestor);
                 for (ResolvedReferenceType inheritedAncestor : ancestor.getAllAncestors()) {
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -247,7 +247,9 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
         if (hasDirectlyAnnotation(qualifiedName)) {
             return true;
         }
-        return getAllAncestors().stream().anyMatch(it -> it.asReferenceType().getTypeDeclaration().hasDirectlyAnnotation(qualifiedName));
+        return getAllAncestors().stream()
+                .filter(it -> it.asReferenceType().getTypeDeclaration().isPresent())
+                .anyMatch(it -> it.asReferenceType().getTypeDeclaration().get().hasDirectlyAnnotation(qualifiedName));
     }
 
     /**
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -194,7 +194,9 @@ public abstract class ResolvedReferenceType implements ResolvedType,
         if (values.contains(tpToReplace)) {
             int index = values.indexOf(tpToReplace);
             values.set(index, replaced);
-            return create(result.getTypeDeclaration(), values);
+            if(result.getTypeDeclaration().isPresent()) {
+                return create(result.getTypeDeclaration().get(), values);
+            }
         }
 
         return result;
@@ -238,13 +240,15 @@ public abstract class ResolvedReferenceType implements ResolvedType,
 
     public final List<ResolvedReferenceType> getAllInterfacesAncestors() {
         return getAllAncestors().stream()
-                .filter(it -> it.getTypeDeclaration().isInterface())
+                .filter(it -> it.getTypeDeclaration().isPresent())
+                .filter(it -> it.getTypeDeclaration().get().isInterface())
                 .collect(Collectors.toList());
     }
 
     public final List<ResolvedReferenceType> getAllClassesAncestors() {
         return getAllAncestors().stream()
-                .filter(it -> it.getTypeDeclaration().isClass())
+                .filter(it -> it.getTypeDeclaration().isPresent())
+                .filter(it -> it.getTypeDeclaration().get().isClass())
                 .collect(Collectors.toList());
     }
 
@@ -298,10 +302,9 @@ public abstract class ResolvedReferenceType implements ResolvedType,
 
     /**
      * Corresponding TypeDeclaration
-     * TODO: Convert to Optional
      */
-    public final ResolvedReferenceTypeDeclaration getTypeDeclaration() {
-        return typeDeclaration;
+    public final Optional<ResolvedReferenceTypeDeclaration> getTypeDeclaration() {
+        return Optional.of(typeDeclaration);
     }
 
     /**
@@ -368,7 +371,11 @@ public abstract class ResolvedReferenceType implements ResolvedType,
         if (typeParameterDeclaration.declaredOnMethod()) {
             throw new IllegalArgumentException();
         }
-        String typeId = this.getTypeDeclaration().getId();
+        if(!this.getTypeDeclaration().isPresent()) {
+            return Optional.empty(); // TODO: Consider IllegalStateException or similar
+        }
+
+        String typeId = this.getTypeDeclaration().get().getId();
         if (typeId.equals(typeParameterDeclaration.getContainerId())) {
             return Optional.of(this.typeParametersMap().getValue(typeParameterDeclaration));
         }
@@ -388,9 +395,17 @@ public abstract class ResolvedReferenceType implements ResolvedType,
      * that have been overwritten.
      */
     public List<ResolvedMethodDeclaration> getAllMethods() {
-        List<ResolvedMethodDeclaration> allMethods = new LinkedList<>(this.getTypeDeclaration().getDeclaredMethods());
-        getDirectAncestors().forEach(a ->
-                allMethods.addAll(a.getAllMethods()));
+        if(!this.getTypeDeclaration().isPresent()) {
+            return new ArrayList<>(); // empty list -- consider IllegalStateException or similar
+        }
+
+        // Get the methods declared directly on this.
+        List<ResolvedMethodDeclaration> allMethods = new LinkedList<>(
+                this.getTypeDeclaration().get().getDeclaredMethods()
+        );
+        // Also get methods inherited from ancestors.
+        getDirectAncestors().forEach(a -> allMethods.addAll(a.getAllMethods()));
+
         return allMethods;
     }
 
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -540,4 +540,14 @@ public abstract class ResolvedReferenceType implements ResolvedType,
                 && getQualifiedName().equals(java.lang.Object.class.getCanonicalName());
     }
 
+    /**
+     * @return true, if this represents {@code java.lang.Enum}
+     * @see ResolvedReferenceTypeDeclaration#isJavaLangEnum()
+     */
+    public boolean isJavaLangEnum() {
+        return this.isReferenceType()
+                && hasName() // Consider anonymous classes
+                && getQualifiedName().equals(java.lang.Enum.class.getCanonicalName());
+    }
+
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -32,10 +32,14 @@ import com.github.javaparser.resolution.types.parametrization.ResolvedTypeParame
 import com.github.javaparser.resolution.types.parametrization.ResolvedTypeParametrized;
 import com.github.javaparser.utils.Pair;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
-
-import static com.github.javaparser.ast.Modifier.Keyword.PRIVATE;
 
 /**
  * A ReferenceType like a class, an interface or an enum. Note that this type can contain also the values
@@ -294,6 +298,7 @@ public abstract class ResolvedReferenceType implements ResolvedType,
 
     /**
      * Corresponding TypeDeclaration
+     * TODO: Convert to Optional
      */
     public final ResolvedReferenceTypeDeclaration getTypeDeclaration() {
         return typeDeclaration;
@@ -504,4 +509,20 @@ public abstract class ResolvedReferenceType implements ResolvedType,
     }
 
     public abstract ResolvedReferenceType deriveTypeParameters(ResolvedTypeParametersMap typeParametersMap);
+
+
+    /**
+     * We don't make this _ex_plicit in the data representation because that would affect codegen
+     * and make everything generate like {@code <T extends Object>} instead of {@code <T>}
+     *
+     * @return true, if this represents {@code java.lang.Object}
+     * @see ResolvedReferenceTypeDeclaration#isJavaLangObject()
+     * @see <a href="https://github.com/javaparser/javaparser/issues/2044">https://github.com/javaparser/javaparser/issues/2044</a>
+     */
+    public boolean isJavaLangObject() {
+        return this.isReferenceType()
+                && hasName() // Consider anonymous classes
+                && getQualifiedName().equals(java.lang.Object.class.getCanonicalName());
+    }
+
 }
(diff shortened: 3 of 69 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
```diff
@@ -32,6 +32,10 @@
  */
 public interface NodeWithExtends<N extends Node> {
 
+    /**
+     * @return All extended types that have been explicitly added.
+     *   Note that this will not include {@code java.lang.Object} unless it is explicitly added (e.g. {@code class X extends Object {}})
```
MysterAitch, line 38: This likely needs to be double checked -- I recall being convinced at the time of writing this, but now I am less sure

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
```diff
@@ -274,4 +274,21 @@ default boolean hasAnnotation(String qualifiedName) {
...
+
+    /**
+     * We don't make this _ex_plicit in the data representation because that would affect codegen
+     * and make everything generate like {@code <T extends Object>} instead of {@code <T>}
+     *
+     * @return true, if this represents {@code java.lang.Object}
+     * @see ResolvedReferenceType#isJavaLangObject()
+     * @see <a href="https://github.com/javaparser/javaparser/issues/2044">https://github.com/javaparser/javaparser/issues/2044</a>
+     */
+    default boolean isJavaLangObject() {
```
MysterAitch, line 289: this is used in several locations throughout the PR, and removes many of the duplicated and slightly differing attempts to compare against `java.lang.Object`

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -360,7 +363,7 @@ public ResolvedType visit(SuperExpr node, Boolean solveLambdas) {
                 if (resolvedTypeName.isInterface()) {
                     return new ReferenceTypeImpl(resolvedTypeName.asInterface(), typeSolver);
                 } else if (resolvedTypeName.isClass()) {
-                    return resolvedTypeName.asClass().getSuperClass();
+                    return resolvedTypeName.asClass().getSuperClass().orElseThrow(() -> new RuntimeException("super class unexpectedly empty"));
```
MysterAitch, line 367: There are many cases throughout this PR which make use of `orElseThrow` -- I consider this to be okay to include as "default" given that the current behaviour in many cases is a value of `null` (and subsequent `NullPointerException` runtime errors further down the line.

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
```diff
@@ -294,6 +298,7 @@ public ResolvedTypeParametersMap typeParametersMap() {
 
     /**
      * Corresponding TypeDeclaration
+     * TODO: Convert to Optional
```
MysterAitch: I plan to do this TODO within this PR so that users have one change to optional to do, as opposed to several smaller changes. 

If anyone has any problems with such a big change, I am more than happy to work with library users to upgrade! :smile:

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -88,6 +88,7 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
         }
 
         // Look among statically imported values
+        // TODO: Is null check required?
```
MysterAitch: Adding a comment here to highlight this and solicit thoughts!

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnonymousClassDeclaration.java
```diff
@@ -142,15 +151,18 @@ public AccessSpecifier accessSpecifier() {
...
-            .build();
-  }
+    @Override
+    public List<ResolvedReferenceType> getAncestors(boolean acceptIncompleteList) {
+        ImmutableList.Builder<ResolvedReferenceType> builder = ImmutableList.builder();
+
+        // Only add the super type if it is present (e.g. java.lang.Object has no super class)
+        getSuperClass().ifPresent(builder::add);
+
+        // All all ancestors of the super type..? TODO: Does this need to be wrapped in a presence check?
```
MysterAitch: Commenting for visibility...

My thinking now as I read through these changes is that this shouldn't need a presence check -- instead `getAncestors()` should return an empty collection if there are zero ancestors.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -238,10 +242,14 @@ public boolean canBeAssignedTo(ResolvedReferenceTypeDeclaration other) {
         if (this.getQualifiedName().equals(other.getQualifiedName())) {
             return true;
         }
-        ResolvedClassDeclaration superclass = (ResolvedClassDeclaration) getSuperClass().getTypeDeclaration();
-        if (superclass != null && superclass != this) {
-            if (superclass.canBeAssignedTo(other)) {
-                return true;
+
+        Optional<ResolvedReferenceType> optionalSuperClass = getSuperClass();
+        if(optionalSuperClass.isPresent()) {
```
MysterAitch: can't use `ifPresent()` here due to the return contained within it.. somebody later might want to edit it if you can see a way to do so, but this is low priority..

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -296,10 +306,11 @@ public boolean isTypeParameter() {
         List<ResolvedReferenceType> ancestors = new ArrayList<>();
 
         // We want to avoid infinite recursion in case of Object having Object as ancestor
-        if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
+        if (!this.isJavaLangObject()) {
             ResolvedReferenceType superclass;
             try {
-                superclass = getSuperClass();
+                // TODO: Meaningful presence check on getSuperClass()
```
MysterAitch: presumably we can just return an empty list here if it's empty..?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -296,10 +306,11 @@ public boolean isTypeParameter() {
         List<ResolvedReferenceType> ancestors = new ArrayList<>();
 
         // We want to avoid infinite recursion in case of Object having Object as ancestor
-        if (!(Object.class.getCanonicalName().equals(getQualifiedName()))) {
+        if (!this.isJavaLangObject()) {
```
MysterAitch: Just spotted this -- I intend to remove this, in favour of this (and then the content can be un-indented)...

```java
if(this.isJavaLangObject()) {
    return ancestors; // empty list
}
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -219,17 +225,26 @@ private String getSuperclassFQN() {
         }
 
         // add the method declaration of the superclass to the candidates, if present
-        SymbolReference<ResolvedMethodDeclaration> superClassMethodRef = MethodResolutionLogic
-                .solveMethodInType(getSuperClass().getTypeDeclaration(), name, argumentsTypes, staticOnly);
-        if (superClassMethodRef.isSolved()) {
-            candidates.add(superClassMethodRef.getCorrespondingDeclaration());
+        if (getSuperClass().isPresent()) {
```
MysterAitch: `.ifPresent()`..?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -309,17 +324,23 @@ public boolean isClass() {
     }
 
     @Override
-    public ResolvedReferenceType getSuperClass() {
+    public Optional<ResolvedReferenceType> getSuperClass() {
         try {
-            if (ctClass.getClassFile().getSuperclass() == null) {
-                return new ReferenceTypeImpl(typeSolver.solveType(Object.class.getCanonicalName()), typeSolver);
+            if ("java.lang.Object".equals(ctClass.getClassFile().getName())) {
```
MysterAitch, line 335: hmm... I remember there being a reason that I left this string comparison in place...

perhaps it can be gotten around by doing something like `javaLangObject().getQualifiedName()` or similar... 

leaving this note here as a reminder to investigate :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -309,17 +324,23 @@ public boolean isClass() {
...
+                // If this is java.lang.Object, ignore the presence of any superclass
+                return Optional.empty();
             }
             if (ctClass.getGenericSignature() == null) {
-                return new ReferenceTypeImpl(typeSolver.solveType(JavassistUtils.internalNameToCanonicalName(ctClass.getClassFile().getSuperclass())), typeSolver);
+                // If generic signature is missing...? TODO: Add explanation here...
```
MysterAitch: comment for visibility/soliciting thoughts

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -309,17 +324,23 @@ public boolean isClass() {
...
+                return Optional.of(new ReferenceTypeImpl(
+                        typeSolver.solveType(JavassistUtils.internalNameToCanonicalName( ctClass.getClassFile().getSuperclass())),
+                        typeSolver
+                ));
             }
 
+            // TODO: Add an explanation here...
```
MysterAitch: comment for visibility/soliciting thoughts

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
```diff
@@ -203,13 +204,36 @@ public ResolvedType transformTypeParameters(ResolvedTypeTransformer transformer)
...
+        ancestors.removeIf(ResolvedReferenceType::isJavaLangObject);
+        ResolvedReferenceTypeDeclaration objectType = typeSolver.getSolvedJavaLangObject();
+        ancestors.add(create(objectType));
+
         return ancestors;
     }
 
+    private boolean hasSuperClassDifferentToSelf(ResolvedReferenceTypeDeclaration typeDeclaration) {
+        if(!typeDeclaration.isClass()) {
+            // If self is not a class, there is no super class
```
MysterAitch: hmm... is this actually true, given that interfaces are able to extend other interfaces?

... but does this create a "superClass" hierarchy?

I'm leaning towards leaving it as-is since, if we do consider this, a list of "super" interfaces will be required as it would allow multiple "parents"...

MysterAitch: Either way, this comment could be expanded to make this thought process more explicit for the benefit of anybody coming across this again in the future.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
```diff
@@ -203,13 +204,36 @@ public ResolvedType transformTypeParameters(ResolvedTypeTransformer transformer)
...
+            return false;
+        }
+
+        if(!typeDeclaration.asClass().getSuperClass().isPresent()) {
+            // If there is no super class, cannot get super class
+            return false;
+        }
+
+        // Note must compare the qualified name strings because the superclass could contain type typeParametersValues (see javadoc for details)
```
MysterAitch: todo: clarify which javadoc to "see"..

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclarationTest.java
```diff
@@ -179,8 +179,8 @@ void getGetAncestors() {
     @Test
     void testGetSuperclassWithTypeParameters() {
         JavaParserClassDeclaration compilationUnit = (JavaParserClassDeclaration) typeSolver.solveType("com.github.javaparser.ast.body.ConstructorDeclaration");
-        assertEquals("com.github.javaparser.ast.body.BodyDeclaration", compilationUnit.getSuperClass().getQualifiedName());
-        assertEquals("com.github.javaparser.ast.body.ConstructorDeclaration", compilationUnit.getSuperClass().typeParametersMap().getValueBySignature("com.github.javaparser.ast.body.BodyDeclaration.T").get().asReferenceType().getQualifiedName());
+        assertEquals("com.github.javaparser.ast.body.BodyDeclaration", compilationUnit.getSuperClass().get().getQualifiedName());
```
MysterAitch, line 182: I've left most of these uses of `get()` within tests as-is for readability... The exception thrown if the optional ends up empty is descriptive enough that it should be obvious why that test fails.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeTest.java
```diff
@@ -246,10 +259,10 @@ void testIsAssignableByGenericsInheritance() {
 
     @Test
     void testGetAllAncestorsConsideringTypeParameters() {
-        assertTrue(linkedListOfString.getAllAncestors().contains(object));
-        assertTrue(linkedListOfString.getAllAncestors().contains(listOfStrings));
-        assertTrue(linkedListOfString.getAllAncestors().contains(collectionOfString));
-        assertFalse(linkedListOfString.getAllAncestors().contains(listOfA));
+        assertThat(linkedListOfString.getAllAncestors(), hasItem(object));
```
MysterAitch, line 260: Converted to the hamcrest matchers here as it gives a more meaningful error message without needing to manually add in a message string (i.e. that a value is missing, as opposed to the junit approach just informs you that a true value was actually false, and vice-versa)

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeTest.java
```diff
@@ -641,36 +658,57 @@ void testDirectAncestorsOfInterfaceExtendingInterface() {
...
-        assertEquals(new HashSet<>(Arrays.asList("java.lang.CharSequence", "java.lang.Appendable",
-                "java.nio.Buffer", "java.lang.Readable", "java.lang.Comparable<java.nio.CharBuffer>")), ancestors);
+        ResolvedReferenceType charbuffer = new ReferenceTypeImpl(new ReflectionClassDeclaration(CharBuffer.class, typeSolver), typeSolver);
+        Set<String> ancestors = charbuffer.getDirectAncestors()
+                .stream()
+                .map(ResolvedReferenceType::describe)
+                .collect(Collectors.toSet());
+
+        assertThat(ancestors, containsInAnyOrder(
+                "java.lang.Object",
```
MysterAitch: FIXME: hmm... Did I muddle things up here by including `java.lang.Object` as an ancestor...?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
```diff
@@ -126,6 +126,7 @@ void resolveReferenceToImportedType() {
         when(compilationUnitDecl.getName()).thenReturn("CompilationUnit");
         when(compilationUnitDecl.getQualifiedName()).thenReturn("com.github.javaparser.ast.CompilationUnit");
         TypeSolver typeSolver = mock(TypeSolver.class);
+        when(typeSolver.getSolvedJavaLangObject()).thenReturn(new ReflectionClassDeclaration(Object.class, typeSolver));
```
MysterAitch, line 129: mocks needed this change so that it returns the "right" thing

MysterAitch:
Ooh interesting, this fails on java 12 -- seemingly due to the addition of `java.lang.constant.Constable`... 

Another one to add to #1701 ?

```
[ERROR] Failures: 
[ERROR]   ReferenceTypeTest.testDirectAncestorsOfClassWithInterfaces:706 
Expected: iterable with items ["java.lang.CharSequence", "java.lang.Object", "java.lang.Comparable<java.lang.String>", "java.io.Serializable"] in any order
     but: not matched: "java.lang.constant.Constable"
[INFO] 
[ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 3
```

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
MysterAitch: done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnonymousClassDeclaration.java
MysterAitch: Removed the todo... the above comment is correct that getAncestor returns an empty list, thus addall of an empty list will not cause an error

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
MysterAitch: Done

MysterAitch:
> 
> 
> Ooh interesting, this fails on java 12 -- seemingly due to the addition of `java.lang.constant.Constable`...
> 
> Another one to add to #1701 ?
> 
> ```
> [ERROR] Failures: 
> [ERROR]   ReferenceTypeTest.testDirectAncestorsOfClassWithInterfaces:706 
> Expected: iterable with items ["java.lang.CharSequence", "java.lang.Object", "java.lang.Comparable<java.lang.String>", "java.io.Serializable"] in any order
>      but: not matched: "java.lang.constant.Constable"
> [INFO] 
> [ERROR] Tests run: 1083, Failures: 1, Errors: 0, Skipped: 3
> ```

Yes, an issue with JDK 12 adding `Constable` and `ConstantDesc` which makes the test invalid. 

https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/lang/String.html


I have added an enum and helper methods to `AbstractSymbolResolutionTest` which can be used to detect the current java version, thus allows tests to be run on specific Java versions.

This feels very "hacky" thus is immediately marked as as deprecated to strongly discourage its use in the future as anything other than a quick workaround, though it may be a potential solution to #1701 .

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
MysterAitch: Done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
MysterAitch: Answer -- Yes, this null check IS required. See #2695

Comments added, so this can now be marked as resolved.

```java
        // null-check <strong>is</strong> required due to definition of {@code getImports()} -- see also #2695
        // Changing this would required editing the generators to return an empty collection as opposed to null
        // Look among statically imported values
        if (wrappedNode.getImports() != null) {
            for (ImportDeclaration importDecl : wrappedNode.getImports()) {
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
MysterAitch: Done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
MysterAitch: done

javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithExtends.java
MysterAitch: Yes, this is correct. 

This is supposed to be indicative of what is within the AST, thus refers to what is explicitly extended (within a class or interface).

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
MysterAitch: Hmm.. something to consider here is that enums implicitly extend `java.lang.Enum`...

Perhaps additional tests are required to figure out the current behaviour versus expected behaviour.

_EDIT: Again, I'm getting muddled between the AST nodes (where enums CANNOT extend anything else) versus the resolved equivalents (where enums ONLY implicitly extend java.lang.Enum and NEVER anything else)._

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeTest.java
MysterAitch: This is still unclear to me.

Other methods such as `getAllMethods()` make use of `getDirectAncestors()` where I would expect methods from implicitly extended classes (i.e. `java.lang.Object`) to be included within the output.

This assumption is contradicted by the javadoc for `getDirectAncestors()`, however, which explicitly states that implicit extends is excluded.

Bottom line: additional testing is required to determine the current behaviour versus this PR's behaviour versus the "desired"/"proper" behaviour (hopefully they are all the same!).

```java
    /**
     * Return direct ancestors, that means the superclasses and interfaces implemented directly.
     * This list should include Object if the class has no other superclass or the interface is not extending another interface.
     * There is an exception for Object itself.
     */
    public abstract List<ResolvedReferenceType> getDirectAncestors();
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
MysterAitch: done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
MysterAitch: personal (very weakly held) conclusion: leave as-is, and if anyone else wants to edit it then this is welcomed

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeImpl.java
MysterAitch: This is an unused new private method, thus can be resolved by removing it...

I still believe that this would be helpful to introduce on the root nodes (ResolvedReferenceType and ResolvedReferenceTypeDeclaration) to avoid the MANY repeated checks against java.lang.Object... but that can be added within a separate PR...

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/model/typesystem/ReferenceTypeTest.java
MysterAitch: Confirmed as a change in behaviour that I caused. 

I've just added the following comments to explain, noting that `getAncestors()` only returns direct ancestors (i.e. ignoring the implicit relationships such as implicitly extending java.lang.Object or java.lang.Enum).

```java 

        List<ResolvedReferenceType> ancestors = typeDeclaration.getAncestors();

        ancestors = ancestors.stream()
                .map(a -> typeParametersMap().replaceAll(a).asReferenceType())

        // Avoid repetitions of Object -- remove them all and, if we removed any, add it back precisely once.
        if(ancestors.removeIf(ResolvedReferenceType::isJavaLangObject)) {
            // We only want to add it back in if we have explicitly extended Object (ignoring implicit supertypes),
            //   thus only do this if it has been removed at least once from the ancestors.
            ResolvedReferenceTypeDeclaration objectType = typeSolver.getSolvedJavaLangObject();
            ResolvedReferenceType objectRef = create(objectType);
            ancestors.add(objectRef);
        }
```

I found this distinction confusing (i.e. the difference between `getAncestors` vs `getAllAncestors` vs `getDirectAncestors` vs what I presume should be named as `getExplicitAncestors` and `getImplicitAncestors`). 

#2696 has been opened to investigate this further.

MysterAitch: I've just edited this PR to undo this inadvertent change of behaviour, so this comment chain can be marked as "resolved"

MysterAitch: I have added a fix to undo this change of behaviour.

See also #2696 and https://github.com/javaparser/javaparser/pull/2693/files/5f06176f074805fcc32a0f18989b031af5aadae0#r429632045

Issue #2671, opened 2020-05-14
It is not possible to use JavaSymbolSolver in a multithreading context


corradomio:
The current implementation of JavaSymbolSolver  can be NOT used in a multithreading application. It is necessary to reimplement some classes and to control the access to some internal data structures.

The main problem is that it uses too memory: this because a type resolver is a dictionary and it is necessary to create the hierarchy of type solvers for each source code analyzed. BUT the content of this dictionary is the same for ALL source codes analyzed in parallel.

It is necessary to separate the dictionary from the typesolver: multiple type solvers, same dictionary. For example the JDK is the same for ALL.

Another problem is **JavaParserFacade.get()**: synchronizid this method we introduce a point of synchronization that can be avoid.

This is only the list of the major problems found.

MysterAitch:
I see that the goal is to deprecate and remove `JavaParserFacade` in favour of using : #1377

More than happy to have improvements made until that happens though :)

Modules and multi-threading are two of my weakest areas, so some extra support from you on these would be very gratefully received.

Personally I'm a fan of the idea supporting multi-threading, I'm just not sure what is needed to make it happen! Hopefully then we can also run the test suites in parallel thus much faster!! :) 

#2671
#2668

MysterAitch:
A reply I got on chat when I asked for comment about this issue is that we would usually advise not using JP across threads whenever similar questions are asked (somewhat paraphrasing here -- note that I'm still quite new and @matozoid has been around a lot longer than I have, so I'm inclined to trust him more than I trust my own views here!). 

My presumption is that it would be a _substantive_ amount of work to do well before we could have "thread-safe" being something that could actually be used to describe JavaParser. Or maybe this is an easy/trivial thing to do for someone who knows what they're doing? At the risk of being annoyingly open _(and repetitive :wink:)_ about pointing this out, personally I don't know enough to say one way or the other.

Given that you have clearly gone further down this route than I have and it sounds like you are already using a modified/patched/extended version that _is_ able to be used across threads (pending #2668), would you mind sharing some details about this and what would be involved if we were to actively pursue a thread-safe JavaParser? 

_perhaps there is a distinction to be made between making JP thread-safe, versus something that actively blocks/breaks other people trying to add a thread-safe wrapper/extension to JP?_

corradomio:
There are some ""philosophical observations"" to consider:

[...]

PR #2694, opened 2020-05-22
Add the modifier synchronized to `JavaParserFacade#get`

### LINKED ISSUE #2668: JavaSymbolResolver used in parallel
In this moment, it is not possible to use **JavaSymbolResolver** in parallel.
The problem is in **JavaParserFacade**, in the static method **get:**

    public static JavaParserFacade get(TypeSolver typeSolver) {
        return instances.computeIfAbsent(typeSolver, JavaParserFacade::new);
    }

**instances** is a **WeakHashMap** but this class is not thread-safe.
It can be enough to convert this method in a "synchronized" method:

    public static **synchronized** JavaParserFacade get(TypeSolver typeSolver) {
        return instances.computeIfAbsent(typeSolver, JavaParserFacade::new);
    }

MysterAitch:
Closes #2668

Hopefully the javadoc is sufficiently clear that this does not indicate javaparser being safe to use within a multi-threaded environment, instead this is to help @migliorabile in his attempts to develop a thread-safe wrapper per #2671 :)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,8 @@ Next Release (Version 3.15.23)
     (PR [#2679](https://github.com/javaparser/javaparser/pull/2679), by [@MysterAitch](https://github.com/MysterAitch))
 * FIXED: Add symbol solver support for variadic parameters given zero or more than one argument, and when an array is given
     (PR [#2675](https://github.com/javaparser/javaparser/pull/2675), by [@hfreeb](https://github.com/hfreeb))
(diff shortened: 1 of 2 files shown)


corradomio:
Indeed: this resolve ONLY a specific problem when used in parallel.
BUT this is not THE solution because this is a SINGLE POINT of synchronization.
This FORCE the application to work in sequential way.
But, in this way, it is possible to use the distributed library and not a customized version.
Thanks a lot.

PR #2704, opened 2020-05-31
WIP -- Generators and Lexical Preserving Printer

### LINKED ISSUE #1301: LexicalPreservingPrinter does not play well with core-generators.

MysterAitch:
This is a bit muddled given that I ended up doing a bunch of printer fixes in addition to the generator stuff... 

Let's see how much I can recall! :+1: :+1: :+1: 

**Fixed:** 
- Any generated methods which are then not re-generated are marked as "stale" (and are reported within the console at the end of running the core generator).
- Generators now run with lexical preservation turned on.
- `Difference.MatchClassification` now has an additional option -- newline. This is used when trying to rank which differences applies (e.g. since JP now respects the line endings in the file, they're not necessarily always going to be equal, thus `\r` is treated differently to `\n` etc and causes problems in the "same before" and "same after" logic).
- A couple of areas of the `ConcreteSyntaxModel` have been fixed/tweaked/reformatted.
- When creating methods/method bodies (e.g. `methodBody.addStatement()`) newlines aren't automatically present, thus the lexical preserving printer retains this code style and doesn't automatically insert them. For this reason, I've added a `prettyPrint` method to the `AbstractGenerator` that is used a few times to rectify this (i.e. where we want the generated nodes to be pretty printed, but the "other" nodes to be preserved).
- Many methods which previously missed the annotation `@Generated` now have it
- Many methods which previously missed the annotation `@Override` now have it
- Fixed setting the annotation of `@SuppressWarnings` (n.b.: side effect of reversing the annotation now that adding that annotation happens prior to adding the `@Generated` annotation)
- Fixed cases where the generated/replaced nodes have non-javadoc comments (previously they'd be removed/lost, now they're retained)
- Added a common ancestor to the printers (not sure why I did this and could probably split out into a separate PR, but it's helpful when we don't really care what kind of printer we have so I'm leaving it)

**Outstanding:** 
- ~~For some reason, the space between the imports and the class/enum header is removed.~~
- Indentation on "entries" (i.e. within an enum and switch) isn't right
   - This can be seen within `JavaToken.java` and `ObservableProperty.java`
- Indentation of the `valueOf` method within `JavaToken.java` increases on every run of the generator.
- When removing all "entries" within an enum, the lines / indentation remain -- they should probably be removed (i.e. when deleting the last entry, the whole line should be removed). The consequence is that when adding the entries back in, these newlines are retained in addition to the newly inserted ones.

_There is probably some tidying that is needed to be done also (offhand: probably leftover commented code, and test code that I'm not sure is in the correct test class)._

Closes #1301

Replaces: #2554 
Replaces: #2557


MysterAitch:
Scanning through, I see that a few reformats slipped through -- my pretty strong preference for code I write is to have methods in alphabetical order but this has led to quite a messy diff.. I normally try to turn that off for JP stuff, sorry!

matozoid:
Nice work man!

MysterAitch:
Hmm, I see now that turns out the stale checks I added here and within #2706 don't play as nicely together as I had hoped -- running one or the other generator ends up marking the "OTHER'S" `@Generated` annotations as stale because they weren't replaced.

Maybe a more nuanced approach is required (e.g. each generator only mark their own generated annotations as stale) but that could get tricky if a generator is ever renamed or removed. That could possibly be worked around by keeping track of which generators were run and then searching for any annotations that don't match those, but this has the same trouble as the current behaviour. 

Alternatively the stale check could be run separately -- run it once at the start, then run each of the generators, and then at the end run a verify/inspect phase to report.

I think that currently I might be leaning towards adding timestamps to the annotations, but that comes with its own questions (e.g. how "old" is stale - seconds/hours/weeks?.. e.g. at what point is it an info vs warning vs error?). It gets a bit easier if we can presume that all runners will run at the same time, and maybe that the generators are run as part of the maven build/verify phases perhaps.


Bottom line is that this needs a bit more thought -- I'll probably do a workaround by more narrowly defining which files each generator should "touch"/"inspect"/ignore (e.g. tell the jp-core generators to exclude the metamodel classes, and tell the metamodel generators to ONLY touch the metamodel classes), then leave the bigger questions and considerations for another issue/PR.

MysterAitch:
Currently has expected / known test failures per the issues described in the comment above -- there are issues with indentation when adding/removing nodes with lexical preservation enabled.

PR #2713, opened 2020-06-03
Add messages for unsupported operation exception

sharath2106:
I've added messages for the exception for some classes. Please do take a look and let me know if these messages looks good to you so that I can continue with others.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeParameterDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeParameterDeclaration.java
@@ -92,7 +92,7 @@ public interface ResolvedTypeParameterDeclaration extends ResolvedTypeDeclaratio
 
             @Override
             public Optional<ResolvedReferenceTypeDeclaration> containerType() {
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException(String.format("Container type not available for %s", this.getClass().getCanonicalName()));
             }
         };
     }
@@ -128,7 +128,7 @@ public interface ResolvedTypeParameterDeclaration extends ResolvedTypeDeclaratio
      * This is unsupported because there is no package for a Type Parameter, only for its container.
      */
     default String getPackageName() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("Unable to get package name for %s", this.getClass().getCanonicalName()));
     }
 
     /**
@@ -136,7 +136,7 @@ public interface ResolvedTypeParameterDeclaration extends ResolvedTypeDeclaratio
      * This is unsupported because there is no class for a Type Parameter, only for its container.
      */
     default String getClassName() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("Unable to get class name for %s", this.getClass().getCanonicalName()));
     }
 
     /**
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typeinference/ControlFlowLogic.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typeinference/ControlFlowLogic.java
@@ -65,7 +65,7 @@ public class ControlFlowLogic {
      * completes normally. In this case, the break target need not be a switch, while, do, or for statement.
      */
     public Statement breakTarget(BreakStmt breakStmt) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("Unable to complete the operation for the statement %s", breakStmt));
     }
 
     /**
@@ -100,11 +100,11 @@ public class ControlFlowLogic {
     }
 
     private boolean contains(Statement container, Statement contained) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("%s is not present in the %s", contained, container));
     }
 
     private List<TryStmt> containedTryStmts(Statement statement) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("%s is not present", statement));
     }
 
     private <P extends Node> boolean parentIs(Node node, Class<P> parentClass) {
@@ -132,7 +132,7 @@ public class ControlFlowLogic {
                 if (!n.isEmpty() && !parentIs(statement, SwitchStmt.class)) {
                     return canCompleteNormally(n.getStatement(n.getStatements().size() - 1));
                 }
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException(String.format("The statement %s is not reachable", statement));
             }
 
             @Override
@@ -140,7 +140,7 @@ public class ControlFlowLogic {
                 // A labeled statement can complete normally if at least one of the following is true:
                 // – The contained statement can complete normally.
                 // – There is a reachable break statement that exits the labeled statement.
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException(String.format("The statement is either incomplete or unreachable"));
             }
 
             @Override
@@ -200,7 +200,7 @@ public class ControlFlowLogic {
         // The then-statement of an if-then-else  statement is reachable iff the if-then-else statement is reachable.
         // The else-statement is reachable iff the if-then-else statement is reachable.
 
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("The statement %s is not reachable", statement));
     }
 
     public boolean isReachable(Statement statement) {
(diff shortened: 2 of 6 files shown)


javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-model/com/github/javaparser/symbolsolver/model/declarations/TypeParameterDeclaration.java
```diff
@@ -23,200 +23,173 @@
...
- * Declaration of a type parameter.
- * For example:
- * <p>
- * class A&lt;E extends String&gt;{}
- * </p>
- * <p>
- * In this case <b>E</b> would be a type parameter.
+ * Declaration of a type parameter. For example:
+ *
+ * <p>class A&lt;E extends String&gt;{}
```
MysterAitch, line 28: If you are editing javadoc, please use `{@code }` and edit it to use valid HTML (e.g. closing `</p>` tags).

For example: 

```java
/**
 * Declaration of a type parameter. For example:
 *
 * {@code class A<E extends String>{}}
 *
 * <p>In this case <strong>{@code E}</strong> would be a type parameter.</p>
 *
 * @author Federico Tomassetti
 */
```

javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-model/com/github/javaparser/symbolsolver/model/declarations/TypeParameterDeclaration.java
```diff
@@ -23,200 +23,173 @@
...
-            @Override
-            public Optional<ReferenceTypeDeclaration> containerType() {
-                throw new UnsupportedOperationException();
-            }
-        };
+  /** Instantiate a TypeParameter defined on a Type with the given data. */
+  static TypeParameterDeclaration onType(final String name, String classQName, List<Bound> bounds) {
+    return new TypeParameterDeclaration() {
+      @Override
+      public String getName() {
```
MysterAitch, line 40: What has been edited here? 

If it is just changes to indentation, please undo these changes.

javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-logic/com/github/javaparser/symbolsolver/logic/InferenceVariableType.java
```diff
@@ -95,7 +95,7 @@ public String describe() {
 
     @Override
     public boolean isAssignableBy(Type other) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("Unable to assign the type %s", other));
```
MysterAitch, line 98: Double check this please -- "Unable to assign..." suggests that we have tried to do an assignment (not the case, here we've just done a check to see if it could be possible).

Could it be that this is one case where this should never be possible, as opposed to being not yet implemented? (note: I don't know the answer!)

Possibly worth taking a look at the documentation:
https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#isAssignableFrom-java.lang.Class-

javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-core/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionParameterDeclaration.java
```diff
@@ -38,7 +38,7 @@ public ReflectionParameterDeclaration(Class<?> type, java.lang.reflect.Type gene
 
     @Override
     public String getName() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException("Unable to get the name");
```
MysterAitch, line 41: It would be helpful to include some reasoning here. 

With the reflection model, it is often the case that names are not available.

javaparser-symbol-solver-testing/src/test/test_sourcecode/javasymbolsolver_0_6_0/src/java-symbol-solver-model/com/github/javaparser/symbolsolver/model/typesystem/Wildcard.java
```diff
@@ -100,7 +100,7 @@ public String describe() {
         } else if (type == BoundType.EXTENDS) {
             return "? extends " + boundedType.describe();
         } else {
-            throw new UnsupportedOperationException();
+            throw new UnsupportedOperationException(String.format("%s is not supported", type));
```
MysterAitch, line 103: This is an improvement - thanks! 

Could we rephrase slightly to not use the phrase "is not supported" -- instead stating _WHAT_ isn't valid please? 

e.g. maybe `String.format("Unsupported BoundType provided: %s" type)` or something like that.

MysterAitch: Ideally error messages should be "actionable" -- i.e. describing how it could be fixed, not just what went wrong.

- "An exception was thrown - here's the stack trace"
- "something went wrong" 
- "the bound type of XXX is invalid" 
- "the provided bound type of XXX is invalid, only BoundType.SUPER and BoundType.EXTENDS is permitted"

MysterAitch:
Note that the tests are failing -- is this intentional? 

Tests must pass before merging any PR :+1:

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeParameterDeclaration.java
```diff
@@ -128,15 +128,15 @@ default boolean declaredOnConstructor() {
...
+        throw new UnsupportedOperationException(String.format("Unable to get package name for %s", this.getClass().getCanonicalName()));
     }
 
     /**
      * The class(es) wrapping the type bound(s).
      * This is unsupported because there is no class for a Type Parameter, only for its container.
      */
     default String getClassName() {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("Unable to get class name for %s", this.getClass().getCanonicalName()));
```
MysterAitch, line 139: Good! 

It would be helpful to also explain **WHY** it is unsupported -- some of this information is included in the javadoc above :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typeinference/ControlFlowLogic.java
```diff
@@ -100,11 +100,11 @@ public boolean continueADoStatement(ContinueStmt continueStmt, DoStmt doStmt) {
     }
 
     private boolean contains(Statement container, Statement contained) {
-        throw new UnsupportedOperationException();
+        throw new UnsupportedOperationException(String.format("%s is not present in the %s", contained, container));
```
MysterAitch, line 103: Be careful here -- the default `toString()` method pretty prints the contents of the node (e.g. if you pass in a method declaration or switch statement or whatever, then there will be a LOT of output). 

Perhaps make use of `container.getClass().getSimpleName()`?

MysterAitch:
> 
> I've added messages for the exception for some classes. Please do take a look and let me know if these messages looks good to you so that I can continue with others.

_I added similar as a line comment, but figured it should probably go here too :+1:_ 


**Error messages should ideally explain WHY it went wrong, AND provide information about how to fix it.**

Note that "bronze standard" is absolutely fine!
  - Bronze standard would be to describe what was provided/tried (where parameters/variables are provided -- not sure what bronze standard would be otherwise)
  - Silver standard describes what was provided/tried and what would actually be supported (e.g. only supporting extends and super)
  - Gold standard would be to differentiate between "unsupported because the java language forbids / doesn't support it" versus "unsupported because we (JavaParser) haven't provided code to handle it yet"


Silver/gold standard is probably a bit more blurry and can't be entirely hierarchical -- probably bits of both!

sharath2106:
Thanks for your inputs @MysterAitch 

It really helped. 

Will incorporate your feedbacks and update the PR.

sharath2106:
I think `mvn clean package` doesn't run the tests marked as @SlowTest

I've investigated the test failures and it points me to a `txt` file which has the expected output. On updating the same, the test passes. Is this the fix for the failing test? Or am I missing something?

MysterAitch:
> I think `mvn clean package` doesn't run the tests marked as @SlowTest
> 
> I've investigated the test failures and it points me to a `txt` file which has the expected output. On updating the same, the test passes. Is this the fix for the failing test? Or am I missing something?

You are correct -- you'll need to enable the slow tests profile using `-P` -- from memory I think it is `mvn test -P alsoSlowTests`).

I haven't looked deeply into the tests that have failed -- if those are the ones I think it is, they might need updating to reflect changes made but I don't recall for sure.

I'm happy to leave that (how to fix it) to your judgement for now and I can review the updated PR once it's uploaded! 😊 👍

PR #2720, opened 2020-06-10
[NOT RDY] Fix stack overflow in methodCallExpr.resolve()

iTakeshi:
Fix `StackOverflowError` in `methodCallExpr.resolve()`.

## test case

```java
// Parent.java
import static main.Parent.MyEnum.*;
public class Parent {
    public enum MyEnum {  ...  }
}
```

```java
// Child.java
public class Child extends Parent {
    public void method() {  ...  }
}
```

```java
// Main.java
public class Main {
    public static void main() {
        Child c;
        c.method();   // <- attempt to resolve this MCE results in StackOverflow
    }
}
```

## cause
`MCE.resolve()` first find for the method in the class itself, but even when there is a candidate method found, [it looks into the ancestors](https://github.com/javaparser/javaparser/blob/69a2e0ed4e0f260ccd61cbb2594b2c3069e47990/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java#L161-L184).
When there are no candidates in the ancestor class (`Parent` in this case), `resolve` looks into the parent context (`CompilationUnit` of Parent.java) and then `import`s in the CU.
However, in this example, there is an `import static` that refers a nested enum in `Parent`.
After `resolve` finds out there are no candidates in `MyEnum`, it looks into the parent context, (i.e., the compilation unit) again.
This causes infinite recursion.

```
Child -> (ancestor) -> Parent
      -> (outer context) -> CU
      -> (import) -> MyEnum
      -> (outer context) -> CU
      -> (import) -> MyEnum
      -> ...
```

## fix
Stop looking into the ancestors if there are candidate methods found in the receiver class itself.


iTakeshi:
I've noticed my solution in this PR does not fix the fundamental cause of the problem. The cause is apparently the handling of `import static`s which refer enums in the same files as the CUs. And because of this, there are other `StackOverflowError`s this PR cannot solve. I'm seeking for a complete solution.

iTakeshi:
Superseded by #2721

PR #2721, opened 2020-06-10
Fix StackOverflowError in CUCtx.solveSymbol() and solveMethod()

iTakeshi:
Supercedes #2720 
This PR fixes `StackOverflowError`s which occur in `CompilationUnitContext.solveSymbol()` and `solveMethod()` under some edge cases.

## test case

```java
// Parent.java
import static main.Parent.MyEnum.*;  // <- refers a member in the same CU
public class Parent {
    public enum MyEnum {  ...  }
}
```

```java
// Child.java
public class Child extends Parent {
    public void method() {  ...  }
}
```

```java
// Main.java
public class Main {
    public static void main() {
        Child c;
        c.method();   // <- attempt to resolve this MCE results in StackOverflow
    }
}
```

## cause
`MCE.resolve()` first looks into the direct receiver `Child`, and its superclass `Parent`.
After inspecting each class, the resolution mechanism checks in the outer context of those classes up to the `CompilationUnit`s.
In this example, there is an `import static` that refers to a nested enum in `Parent`.
After the resolver finds out there are no candidates in `MyEnum`, it looks into the parent context, (i.e., the compilation unit) again.
This causes infinite recursion.

```
Child -> (ancestor) -> Parent
      -> (outer context) -> CU
      -> (import) -> MyEnum
      -> (outer context) -> CU
      -> (import) -> MyEnum
      -> ...
```

## fix
Skip `import`s which refers to a member in the same `CompilationUnit`.


MysterAitch:
Interesting, thanks!

My understanding of going all the way up the inheritance chain is to then figure out which of them is "most specific" - e.g. `void foo(List x) {}` vs `void foo (ArrayList x) {}`.

Not that it is relevant here - just context.

Out of curiosity (I'm on mobile so not able to look thoroughly), does this affect the ability to resolve static members within the enum, when they are referenced from the parent class? 

The whole idea of importing a child member seems a bit weird to me but _kinda_ makes sense in not having to qualify it! :) 

Also, is it relevant that the child is an enum rather than a class?

```java

static import example.Parent.NestedEnum.foo; //or .*;

public class Parent {

    public Parent() {
         foo(); // does this resolve correctly if we skip the static import?
         NestedEnum.foo(); // how about if we qualify the static method?
         NestedEnum c;
         c.bar();
    }


     enum NestedEnum {
           public static void foo() {}
           public void bar() {}
     }
}

```

iTakeshi:
> My understanding of going all the way up the inheritance chain is to then figure out which of them is "most specific" - e.g. void foo(List x) {} vs void foo (ArrayList x) {}.
Not that it is relevant here - just context.

I understood this comment was for what I wrote in #2720 -- Thanks for the important note.

Easier one first...
> is it relevant that the child is an enum rather than a class?

No. Being `enum` is not relevant. The same `StackOverflowError` problem occurs even if the nested structure is a `class`, and this PR fixes this case as well.
```java
// Parent.java
import static main.Parent.Nested.*;
public class Parent {
    // note that both nested class and the method should be `static`
    // in order to be imported with `import static`
    public static class Nested {
        public static void foo() {  ...  }
    }
}
```


Next,
> does this affect the ability to resolve static members within the enum, when they are referenced from the parent class?

*Short answer*: No problem. Both qualified/unqualified MCE can be resolved in your example (I added a test case in a new commit).

*Long answer*:

There is another condition: the `StackOverflowError` problem only occurs for `asterisk` imports (because, the src for non-asterisk imports filters irrelevant ones like [here](https://github.com/javaparser/javaparser/blob/69a2e0ed4e0f260ccd61cbb2594b2c3069e47990/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java#L107) and [here](https://github.com/javaparser/javaparser/blob/69a2e0ed4e0f260ccd61cbb2594b2c3069e47990/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java#L285)).
Thus, my fix as well only affects `asterisk` imports. So, in your example, there is no problem with `import static example.Parent.NestedEnum.foo;` (I mean, there were no `StackOverflowError` problems even before this PR, and there are no "unresolvable" problem with this PR).
This is natural.

What I found surprising is that, even with the asterisk import `import static example.Parent.NestedEnum.*;`, and even though this PR added a filter to avoid infinite recursion, still the unqualified `foo()` CAN be resolved. The qualified `NestedEnum.foo()` is of course okay.
My current understanding by some experiments are:
- When the receiver of `.resolve()` is OUTSIDE the `CompilationUnit` (like in the testcases `00`~`03`), then <<the `wrappedNode` of [the `ref` pointing the `enum`](https://github.com/javaparser/javaparser/pull/2721/files#diff-34bdf2cb3ed7d6f1605770c384f82543R281)>> and <<the `EnumDeclaration` node for the `enum` under `this.wrappedNode`>> are the SAME instance. So, `isAncestorOf` returns `true` and the control flow doesn't go into [the `if` statement](https://github.com/javaparser/javaparser/pull/2721/files#diff-34bdf2cb3ed7d6f1605770c384f82543R283-R287).
- On the other hand, when the receiver of `.resolve()` is INSIDE the `CompilationUnit` (like your example), then <<the `ref.wrappedNode()`>> and <<the `EnumDeclaration` under `this.wrappedNode`>> are DIFFERENT instances (I don't know why at all!). Thus, `isAncestorOf` returns `false`, and the MCE is resolved in the `if` statement.

```
// in CompilationUnitContextResolutionTest/00_receiver
main.Parent.MyEnum
----
EnumDeclaration decl = Navigator.demandEnum(wrappedNode, "Parent.MyEnum");
System.identityHashCode(decl); => 1201632138
System.out.println(decl);
> public enum MyEnum {
> 
>     A("a"), B("b")
> }
----
EnumDeclaration decl = ((JavaParserEnumDeclaration) ref).getWrappedNode();
System.identityHashCode(decl); =>1201632138   --- SAME as above
System.out.println(decl);
> public enum MyEnum {
> 
>     A("a"), B("b")
> }
----
isAncestorOf(ref); => true
```
```
// in CompilationUnitContextResolutionTest/04_review_comment
main.Main.NestedEnum
----
EnumDeclaration decl = Navigator.demandEnum(wrappedNode, "Parent.MyEnum");
System.identityHashCode(decl); => 1431590087
System.out.println(decl);
> public enum NestedEnum {
> ...
> }
----
EnumDeclaration decl = ((JavaParserEnumDeclaration) ref).getWrappedNode();
System.identityHashCode(decl); =>1661048214   --- DIFFERENT from above
System.out.println(decl);
> public enum NestedEnum {
> ...
> }
----
isAncestorOf(ref); => false
```


I'm sorry for the very long and hard-to-follow explanation, but that's what I understand for now.

PR #2730, opened 2020-06-20
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/


MysterAitch:
This is enabled at https://github.com/javaparser/javaparser/blob/master/.github/dependabot.yml

PR #2758, opened 2020-07-15
Fix ArrayType brackets precedence

iTakeshi:
According to [JLS 10.2](https://docs.oracle.com/javase/specs/jls/se11/html/jls-10.html#jls-10.2), brackets pairs after the identifier are preceded to (i.e., become outer of) brackets pairs after the type name.
Hence, the order should be
```
String [<3>][<4>] str[<1>][<2>]
```


iTakeshi:
The test was failed due to github action malfunction. Could you re-run before merging? thank you!

MysterAitch:
Still getting my head around the PR but the description / rationale seems perfectly sound -- thanks @iTakeshi !  :+1:

A few additional notes: 

```java
// Equivalent:
String [<3>][<4>] str[<1>][<2>]
String [<1>][<2>][<3>][<4>] str

// Equivalent:
String @C[]@D[] str@A[]@B[]
String @A[]@B[]@C[]@D[] str
```

Screenshot from the JLS:

image

PR #2774, opened 2020-08-04
The ancestors of the javassist interfaces didn't contain all the inherited type mapping.

maartenc:


MysterAitch:
It seems reasonable, though I don't use this area of JavaParser so I don't fully understand what the "correct" behaviour should be (e.g. should the generics be solved? should this modification also be applied to other classes such as `JavaClassDeclaration.java`?).

Happy to be led by you on this @maartenc re: merging this PR as-is or if any additional modifications would be beneficial :)

maartenc:
Hi @MysterAitch , I also don't know what the correct behaviour should be. I think the generics should be resolved. 

The 'Reflection'-based subclasses already resolve generics, and the `JavassistClassDeclaration` does it as well (I copied the code from there). After taking a quick look in the code, the 'JavaParser'-based sublcasses seems to do this as well (but I could be mistaken).

Maarten

MysterAitch:
I lost track of why I had asked about resolving generics! It might be this line in the test case which refers to `S` (as opposed to the concrete type), but perhaps this is a misunderstanding on my part.

```java
assertEquals("com.github.javaparser.test.GenericChildInterface.S", genericResolvedType.asTypeParameter().getQualifiedName());
```

I'll merge it now and if any additions/modifications are needed they can be added in a separate PR :)

PR #2775, opened 2020-08-04
Finding the functional method doesn't always succeed

### LINKED ISSUE #2595: [Java Symbol Solver] Resolve type from functional interface

maartenc:
Finding the functional method doesn't always succeed when there is more than 1 abstract method in the class hierarchy.

Fixes #2595

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
@@ -54,15 +54,16 @@ public abstract class AbstractTypeDeclaration implements ResolvedReferenceTypeDe
 
         for (ResolvedReferenceType ancestor : getAllAncestors()) {
             List<Pair<ResolvedTypeParameterDeclaration, ResolvedType>> typeParametersMap = ancestor.getTypeParametersMap();
-            for (MethodUsage ancestorMethodUsage : ancestor.getDeclaredMethods()) {
-                MethodUsage methodUsage = ancestorMethodUsage;
+            for (MethodUsage mu : ancestor.getDeclaredMethods()) {
+                // replace type parameters to be able to filter away overridden generified methods
+                MethodUsage methodUsage = mu;
                 for (Pair<ResolvedTypeParameterDeclaration, ResolvedType> p : typeParametersMap) {
                     methodUsage = methodUsage.replaceTypeParameter(p.a, p.b);
                 }
                 String signature = getSignature(methodUsage);
                 if (!methodsSignatures.contains(signature)) {
                     methodsSignatures.add(signature);
-                    methods.add(ancestorMethodUsage);
+                    methods.add(mu);
                 }
             }
         }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
@@ -24,9 +24,13 @@ package com.github.javaparser.symbolsolver.logic;
 import com.github.javaparser.resolution.MethodUsage;
 import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
+import com.github.javaparser.resolution.declarations.ResolvedTypeParameterDeclaration;
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
+import com.github.javaparser.resolution.types.ResolvedType;
+import com.github.javaparser.utils.Pair;
 
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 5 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
```diff
@@ -43,23 +47,43 @@
...
-                    methods.add(mu);
+                    methods.add(ancestorMethodUsage);
                 }
             }
         }
 
         return methods;
     }
 
+    private String getSignature(MethodUsage mu) {
```
MysterAitch: Is there a reason for using a custom private method here rather than using the `getMethodSignature()` method?

My presumption is the answer is "yes" and that there's something I'm currently not seeing :)

MysterAitch:
Looks good, thank you @maartenc ! :) 

Other than the comment re: if that private method belongs in that class (e.g. could it be moved to `MethodUsage`?), happy to merge it in! 

That could probably be done as a separate PR if desired.

maartenc:
Thanks for the feedback @MysterAitch : I'll look at it tomorrow :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
maartenc: I coulnd't find a `getMethodSignature()` method, did you mean why I didn't use the `methodDeclaration.getSignature()` method?

To find overridden generic methods, I need a signature where the generic types have been replaced with the information from the subclass.

For instance, consider the example:

```
public static interface Foo<S, T> extends Function<S, T> {

    T foo(S str);

    @Override
    default T apply(S str) {
        return foo(str);
    }
}
```

If we use the `methodDeclaration.getSignature()` method, we have the following signatures:
`apply(S)`, `foo(S)` (from `Foo`) and `apply(T)` (from `Function`)

But we don't want the `apply(T)` here since it's overridden by `apply(S)`.
For that, I've copied the logic from `methodDeclaration.getSignature()` into a private method and execute it on a `MethodUsage `object (and not on a `MethodDeclaration`).

That method could probably be added as a public method to `MethodUsage`, but I'm not sure it is usable at other places...

maartenc:
I tried to explain it a little more @MysterAitch . I hope it makes sense what I'm saying...
If you think that private method should be a public method of MethodUsage instead, I can update the PR, no problem.

thanks,
Maarten

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractTypeDeclaration.java
MysterAitch: > I coulnd't find a `getMethodSignature()` method, did you mean why I didn't use the `methodDeclaration.getSignature()` method?

Apologies, yes you are correct! Typo on my part!

edit: I had in mind `methodUsage.getSignature()` (as opposed to `methodDeclaration.getSignature()`), but I see now that this method doesn`t yet exist.

MysterAitch:
I've just re-read the comment and realised that I had in mind `methodUsage.getSignature()` (as opposed to `methodDeclaration.getSignature()`), but I see now that this method doesn`t yet exist.

After thinking this through, I think I have come to the conclusion that yes I would prefer to have `MethodUsage#getSignature` and `MethodUsage#getQualifiedSignature` available (see the code and debugger screenshot below). The benefit I see here is that the declared (generic) types are available via `methodUsage.getDeclaration().getSignature()`, and the concrete types are available via `methodUsage.getSignature()`. This isn't a strongly-held view (meaning that I'm happy to be talked out of this) :+1: 

If that's something you're happy to tweak this PR to make happen then go for it, otherwise I'll merge this as-is later this afternoon :) 



```java

    // replacement within MethodUsage.java (note pre-existing TODO re: making use the type parameters)
    public String getQualifiedSignature() {
        return getDeclaration().getQualifiedName() + "." + getSignature();
    }

    
    // added to MethodUsage.java
    public String getSignature() {
        StringBuilder sb = new StringBuilder();
        sb.append(getName());
        sb.append("(");
        for (int i = 0; i < getNoParams(); i++) {
            if (i != 0) {
                sb.append(", ");
            }
            sb.append(getParamType(i).describe());
        }
        sb.append(")");
        return sb.toString();
    }
```

image


Thank you again for the fix! :+1:

maartenc:
Good suggestion @MysterAitch , I've updated this PR :-)

MysterAitch:
Oops, looks like I gave you bad code, sorry! 

It's also visible in the screenshot too (`.apply.apply()`) but I missed that.

```
Expected: is "AnonymousClassDeclarations.DoFn.ProcessContext.innerClassMethod()" 

 but: was "AnonymousClassDeclarations.DoFn.ProcessContext.innerClassMethod.innerClassMethod()"
```

I've stepped away but maybe use the parent's qualified name? Not sure offhand what is available/visible.

maartenc:
Ha no problem, I didn't see it either... :-)
I've committed a new attempt...

maartenc:
Thanks @MysterAitch !
I think you can close PR #2604 as well...

Maarten

PR #2793, opened 2020-09-04
Issue 2283: fix boxing primitive type generics in some method calls

### LINKED ISSUE #2283: Fails to calculate the type of a generic return type constructed from a Primitive type
Given the following code:
```java
public class GenericClass<A> {

    public A val0;

    public GenericClass(final A value0) {
        this.val0 = value0;
    }

    public static <A> GenericClass<A> with(final A value0) {
        return new GenericClass<A>(value0);
    }

    public static void printGenericClass(GenericClass<Integer> p) {
        System.out.println(p.val0);

    }

        GenericClass<Integer> p = GenericClass.with(1);
        printGenericClass(p); // OK
        printGenericClass(GenericClass.with(2)); // Unable to calculate the type of a parameter of a method call.
    }
}
```
It crashes on the specified line with the following stacktrace:
```
Exception in thread "main" java.lang.RuntimeException: Unable to calculate the type of a parameter of a method call. Method call: printPair(Pair.with(2)), Parameter: Pair.with(2)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:225)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:240)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:125)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:140)
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:313)
        ...
Caused by: java.lang.UnsupportedOperationException: com.github.javaparser.resolution.types.ResolvedPrimitiveType
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.matchTypeParameters(MethodCallExprContext.java:356)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.resolveMethodTypeParameters(MethodCallExprContext.java:345)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:86)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:584)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:267)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:44)
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:115)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:448)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:310)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveArguments(JavaParserFacade.java:221)
	... 27 more
```
This is the code segment which throws the original exception:
```java
private void matchTypeParameters(ResolvedType expectedType, ResolvedType actualType, Map<ResolvedTypeParameterDeclaration, ResolvedType> matchedTypeParameters) {
    if (expectedType.isTypeVariable()) {
        if (!actualType.isTypeVariable() && !actualType.isReferenceType()) {
            throw new UnsupportedOperationException(actualType.getClass().getCanonicalName());
        }
        matchedTypeParameters.put(expectedType.asTypeParameter(), actualType);
```
Where, during a debugging session with the code above, actualType is a `PrimitiveTypeUsage{int}` (makes sense) and the expectedType is `TypeVariable`.

We are not sure about how this function is supposed to work, but upon further investigation we think that the issue is that the solver failed to "convert" the int primitive type to Integer, because generics can't use primitives but can indeed use the Integer type (so, the previous bit of code is in fact correct).

We assume this happens with all primitives, but so far we've encountered this problem with both Integer and Double. Testing wether it happens elsewhere shouldn't be too hard.

math4tots:
Fixes #2283.

Treat 'foo()' the same as 'A.foo()' wrt resolving types of static method calls.

---

Simply qualified method calls were being being treated specially and incorrectly. This caused some type parameters to be incorrectly inferred when boxing primitive types, leading to the behavior observed in #2283.

Notably, this special path didn't catch all static method calls, but only when the method name was qualified with a simple name. So `A.foo()` would trigger this path, but `some.pkg.A.foo()` and `foo()` would not trigger this path.

This PR simply deletes this path.

Included test fails before this fix is applied, and passes after.
Note, replacing `A.foo(10)` with just `foo(10)` allows the test to pass even without this fix

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
@@ -81,24 +81,6 @@ public class MethodCallExprContext extends AbstractJavaParserContext<MethodCallE
         ResolvedType typeOfScope;
         if (wrappedNode.getScope().isPresent()) {
             Expression scope = wrappedNode.getScope().get();
-            // Consider static method calls
-            if (scope instanceof NameExpr) {
-                String className = ((NameExpr) scope).getName().getId();
-                SymbolReference<ResolvedTypeDeclaration> ref = solveType(className);
-                if (ref.isSolved()) {
-                    SymbolReference<ResolvedMethodDeclaration> m = MethodResolutionLogic.solveMethodInType(ref.getCorrespondingDeclaration(), name, argumentsTypes);
-                    if (m.isSolved()) {
-                        MethodUsage methodUsage = new MethodUsage(m.getCorrespondingDeclaration());
-                        methodUsage = resolveMethodTypeParametersFromExplicitList(typeSolver, methodUsage);
-                        methodUsage = resolveMethodTypeParameters(methodUsage, argumentsTypes);
-                        return Optional.of(methodUsage);
-                    } else {
-                        throw new UnsolvedSymbolException(ref.getCorrespondingDeclaration().toString(),
-                                "Method '" + name + "' with parameterTypes " + argumentsTypes);
-                    }
-                }
-            }
-
             // Scope is present -- search/solve within that type
             typeOfScope = JavaParserFacade.get(typeSolver).getType(scope);
         } else {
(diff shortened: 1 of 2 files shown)


math4tots:
Oops, broken tests, converting to draft

MysterAitch:
linked issue #2283 closed by PR #2875

PR #2795, opened 2020-09-11
Ensure spaces between annotations and types for lexical-preserving printing

jwaataja:
When using lexical-preserving printing, I found that when adding annotations I got output like `@NullableString` rather than `@Nullable String`. This fix adds spaces between annotations and the base type they apply to for `ClassOrInterfaceType`.


MysterAitch:
Thanks @jwaataja ! :+1:

PR #2813, opened 2020-09-30
Fix issue 2806 EOL separator problems on windows

### LINKED ISSUE #2806: EOL separator problems on windows
Hi, I'm running into a problem on windows when adding and import to an existing source file that uses `\n` line endings.
This happens when using `setLexicalPreservationEnabled(true)` and printing with `LexicalPreservingPrinter`.
I'm currently updating to `javaparser 3.16.1`.

I have a [minimal example](https://github.com/signed/convert-junit4-to-junit5/blob/e5e91942510d8499ca86bc273e0de2b162565fda/src/test/java/jb/convert/ast/LineSeparatorOnWindowsTest.java) that reproduces the problem.

To me it looks like the parser uses [Utils.EOL](https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/utils/Utils.java#L47) that defaults to `\r\n` on windows and during the code modification some assumptions break. I did not find an equivalent to `PrettyPrinterConfiguration` for the `LexicalPreservingPrinter`.

Is there an API that I'm missing?
Am I using the api correctly?

```Java
package testcase;
import com.github.javaparser.ast.ImportDeclaration;
import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;
import org.junit.jupiter.api.BeforeEach;
import static org.hamcrest.CoreMatchers.equalTo;
import static org.hamcrest.MatcherAssert.assertThat;

class LineSeparatorOnWindowsTest {

    @BeforeEach
    void fakeWindows() {
       System.getProperties().setProperty("line.separator", "\r\n");
    }

    @Test
    void importIsAddedOnTheSameLine() {
        String junit4 = "import java.lang.IllegalArgumentException;\n" +
                "\n" +
                "public class A {\n" +
                "}";
        String junit5 = "import java.lang.IllegalArgumentException;\n" +
                "import java.nio.file.Paths;\n" +
                "\n" +
                "public class A {\n" +
                "}";
        JavaParser parser = new JavaParser(new ParserConfiguration().setLexicalPreservationEnabled(true));
        CompilationUnit cu = parser.parse(junit4).getResult().get();
        ImportDeclaration importDeclaration = new ImportDeclaration("java.nio.file.Paths", false, false);
        CompilationUnit compilationUnit = cu.addImport(importDeclaration);
        String out = LexicalPreservingPrinter.print(compilationUnit);
        assertThat(out, equalTo(junit5));
    }
}

```

jlerbsc:
Fixes #2806.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue2806Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue2806Test.java
@@ -22,23 +22,15 @@
 
 package com.github.javaparser.printer.lexicalpreservation;
 
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import org.junit.jupiter.api.Test;
+
 import com.github.javaparser.JavaParser;
-import com.github.javaparser.ParseResult;
-import com.github.javaparser.ParseStart;
 import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.ImportDeclaration;
-import com.github.javaparser.printer.concretesyntaxmodel.CsmElement;
-import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;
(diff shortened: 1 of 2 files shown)


signed:
There are some unused imports in the test file that break the build. After removing them the build worked locally.

PR #2824, opened 2020-10-09
javaparser-symbol-solver-core:add bndtools, build OSGi-Bundle

stbischof:
OSGi Todos:
If you look in the manifest there are several Import Pageges that might be optional.

### DIFF
--- a/javaparser-symbol-solver-core/bnd.bnd
+++ b/javaparser-symbol-solver-core/bnd.bnd
@@ -1,0 +1,29 @@
+# Make the Bundle-SymbolicName fully qualified, not just the artifact id
+Bundle-SymbolicName: com.github.javaparser.javaparser-symbol-solver-core
+
+Export-Package: \
+	com.github.javaparser.symbolsolver,\
+	com.github.javaparser.symbolsolver.core.resolution,\
+	com.github.javaparser.symbolsolver.declarations.common,\
+	com.github.javaparser.symbolsolver.javaparser,\
+	com.github.javaparser.symbolsolver.javaparsermodel,\
+	com.github.javaparser.symbolsolver.javaparsermodel.contexts,\
+	com.github.javaparser.symbolsolver.javaparsermodel.declarations,\
+	com.github.javaparser.symbolsolver.javaparsermodel.declarators,\
+	com.github.javaparser.symbolsolver.javassistmodel,\
+	com.github.javaparser.symbolsolver.logic,\
+	com.github.javaparser.symbolsolver.model.resolution,\
+	com.github.javaparser.symbolsolver.model.typesystem,\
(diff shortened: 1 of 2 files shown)


MysterAitch:
@stbischof -- could you rebase your PR please? It's saying there's a conflict, and the github UI isn't showing the "update" button for me to resolve this.

It's the weekend so I understand if you're not monitoring github and might not see this, but if you do manage to get it done in the next hour or two it'll make it into today's release :+1: :)

MysterAitch:
I've just tried replicating this PR but I can't seem to get it to trigger / do anything. 

Changing the pom per this PR and running compile/package/verify doesn't trigger the creation of `bnd.bnd`. Manually running `bnd:bnd-process` doesn't seem to do it either.

Could you provide some more notes on how this might be run please? 

I would hate to merge this manually produced bnd.bnd file, if it is never to be updated in the future.

MysterAitch:
@stbischof - Please see the comments above.

jlerbsc:
It seems that bnd-process goal generates files from bnd.bnd directives but does not generate bnd.bnd.
https://github.com/bndtools/bnd/tree/master/maven/bnd-maven-plugin
In core target directory there is a MANIFEST.MF i not sure if it is produced by bnd-process but since the manifest file is not declared in the project resources I think it is actually generated.

stbischof:
I changed 2 Plugins here.
bnd to generate osgi related files. Adds this to the bundle and the target folder.
The other is the change in maven-jar-plugin. To pick the Manifest.
Without thi instruction maven-jar-plugin would use its own Manifest

MysterAitch:
I might be being really thick here, but I can't find a way to reproduce or test this PR. This is likely due to my lack of understanding around OSGi and the bnd.bnd file itself!

- How was the `bnd.bnd` file in this PR created? Manually? via running the maven plugin?
- How is it used?

When I run `mvn clean`.. then `mvn bnd:bnd-process` or `mvn package`.. with and without a `bnd.bnd` file. In all combinations it seems that the `MANIFEST.MF` files created at `target/classes/META-INF/MANIFEST.MF` are always the same (just a timestamp difference) and this _seems_ to be true even when manually extracting/inspecting the jar file too.

Is my understanding completely wrong and the `bnd.bnd` file is not generated? Instead the `bnd.bnd` file used as an input in generating the manifest file, perhaps?

My main worry is that if this file is manually created, it won't be kept up to date -- and I really would like to get this done correctly :+1:

MysterAitch:
> 
> 
> @stbischof -- could you rebase your PR please? It's saying there's a conflict, and the github UI isn't showing the "update" button for me to resolve this.
> 

@stbischof -- this still applies.. Even if I wanted to merge this PR I cannot due to the conflict. Please rebase or otherwise update your PR.

image

PR #2828, opened 2020-10-12
action to publish a release to bintray

maartenc:
An attempt to automate publishing a release to maven central using github actions.


matozoid:
Releases are done to Maven Central directly now, so this can be closed.

PR #2834, opened 2020-10-15
Issue 1479 JavaParserFacade.solve(FieldAccessExpr fieldAccessExpr) doesn't return SymbolReference<ResolvedFieldDeclaration>

### LINKED ISSUE #1479: JavaParserFacade.solve(FieldAccessExpr fieldAccessExpr) doesn't return SymbolReference<ResolvedFieldDeclaration>
The used Method of JavaParserFacade is
public SymbolReference\<ResolvedFieldDeclaration> solve(FieldAccessExpr fieldAccessExpr) which should return SymbolReference\<ResolvedFieldDeclaration>.

What I'm looking for is the JavaParserFieldDeclaration to get the initializer value for my fieldAccessExpr. My code is

```
FieldAccessExpr fieldexpr = argument.asFieldAccessExpr();
SymbolReference<ResolvedFieldDeclaration> fieldDeclaration = JavaParserFacade.get(combinedTs).solve(fieldexpr);
JavaParserFieldDeclaration decl = (JavaParserFieldDeclaration) fieldDeclaration.getCorrespondingDeclaration();
```
But I get the following Error:
Exception in thread "main" java.lang.ClassCastException: com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration$1 cannot be cast to com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserFieldDeclaration

I tried to use ResolvedClassDeclaration/JavaParserClassDeclaration as well, but this leads to an compile Error because SymbolReference\<ResolvedFieldDeclaration> could not be cast to SymbolReference\<ResolvedClassDeclaration>.

**Update:** 
I think this problem occures when the fieldexpr is not directly defined in the class but in a parent class. Expample:

```
public class A {
  public static final String AFIELD = "a";
}

public class B extends A {
  public static final String BFIELD = "b";
}

```
The Error occures when resolving the FieldAccessExpr in:

```
doSomething(B.AFIELD);
```

jlerbsc:
Fixes #1479.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -21,6 +21,13 @@
 
 package com.github.javaparser.symbolsolver.javaparsermodel.declarations;
 
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
 import com.github.javaparser.ast.AccessSpecifier;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.body.BodyDeclaration;
@@ -48,13 +55,6 @@ import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
 import com.github.javaparser.symbolsolver.model.typesystem.LazyType;
 import com.github.javaparser.symbolsolver.model.typesystem.ReferenceTypeImpl;
 import com.github.javaparser.symbolsolver.resolution.SymbolSolver;
-
-import java.util.ArrayList;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Optional;
-import java.util.Set;
-import java.util.stream.Collectors;
 
 /**
  * @author Federico Tomassetti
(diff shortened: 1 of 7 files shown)


maartenc:
I think it would be better if the `ResolvedFieldDeclaration` interface also implements the `AssociableToAST<FieldDeclaration>` interface (just like some other ResolvedXXXDeclaration interfaces do). That way, you don't need to cast anything, but you can simply call the `toAst()` method on the `ResolvedFieldDeclaration`. What do you think?

jlerbsc:
Your proposal is interesting I will watch this track. Thanks

jlerbsc:
@maartenc Your solution is really better. I don't know this interface thanks for the tip. Can you review this work and give me an opinion.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -148,9 +147,15 @@ public boolean isStatic() {
                                 public ResolvedTypeDeclaration declaringType() {
                                     return f.declaringType();
                                 }
+                                
+                                @Override
+                                public Optional<FieldDeclaration> toAst() {
+                                    FieldDeclaration fd = JavaParserFieldDeclaration.class.isAssignableFrom(f.getClass()) ?
+                                            ((JavaParserFieldDeclaration)f).getWrappedNode() : null;
+                                    return Optional.ofNullable(fd);
```
maartenc: I think this should be `return f.toAst();` instead of the code here.
In addition, you should implement the `toAst()` method in the class `JavaParserFieldDeclaration` and return the wrappedNode there.

jlerbsc: Yes it's really better. Thanks for the review.

maartenc:
Looks good @jlerbsc  :+1:
If I'm not mistaken, `JavaParserInterfaceDeclaration` also has such an inner class construct. Maybe you should apply the change you made to `JavaParserClassDeclaration` there too?

jlerbsc:
Once again you are right. I apply this change too.

jlerbsc:
It's late in Bordeaux, I'll see the rest tomorrow ;-)

PR #2843, opened 2020-10-18
Add test case for issue #1793 Cloning problems with LexicalPreserving…

### LINKED ISSUE #1793: Cloning problems with LexicalPreservingPrinter (using 3.6.18)
When parsing
``` Java
public class Test {
  public void foo(Bar x, Bar y) {
    x.barf(); // expected to be wrapped
    x.bark(); // expected to be wrapped
    y.barf(); // expected to be wrapped
    y.bark(); // expected to be wrapped
  }
}
```
and reprinting the same instance with the `LexicalPreservingPrinter`, everything works fine. However, when printing a clone of the root node, I get
``` Java
public class Test {

public void foo(Bar x, Bar y) {
    x.barf();
x.bark();
y.barf();
y.bark();
}
}
```
which has lost its indentation and comments. The code I use looks like this, in case I do something dumb:
``` Java
    LexicalPreservingPrinter.setup(sourceASTRoot);
    System.out.println(LexicalPreservingPrinter.print(sourceASTRoot.clone()));
    // ^ good output without the `clone`
```
(FYI, I'm cloning because I don't need to perform a single "path" of mutations, but instead a tree. For example in the above case I would want both one version of the source that has all "x" replaced by "a" and another version that has "y" replaced by "b", but not both at the same time! So unless I wanna undo all my changes or reparse every time, I need to clone, right?)

I saw that the `CloneVisitor` in fact seems to copy the token ranges as well so I'm not sure what goes wrong regarding indentation; but the disappearing comments are the bigger problem, I think without the `LexicalPreservingPrinter` they were moved in front of each line, but at least weren't lost :)

Thanks a ton for this library, it's super easy to use and works great otherwise! 👍

jlerbsc:
…Printer (using 3.6.18)

Fixes #1793.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AbstractLexicalPreservingTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AbstractLexicalPreservingTest.java
@@ -21,20 +21,28 @@
 
 package com.github.javaparser.printer.lexicalpreservation;
 
+import static com.github.javaparser.utils.TestUtils.assertEqualsString;
+import static com.github.javaparser.utils.TestUtils.readResource;
+
+import java.io.IOException;
+
+import org.junit.jupiter.api.AfterEach;
+
+import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.expr.Expression;
 
-import java.io.IOException;
-
-import static com.github.javaparser.utils.TestUtils.assertEqualsString;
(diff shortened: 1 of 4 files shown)


jlerbsc:
@maartenc There are errors in this build which are not related to my work! have you an idea? Perhaps it's related to line separator.
There are also many java.lang.OutOfMemoryError: GC overhead limit exceeded

maartenc:
@jlerbsc : just a guess, but in the testcase you use some static method to configure things, which might have an influence on the other junit tests. Maybe this setup requires more memory?

```
StaticJavaParser.setConfiguration(new ParserConfiguration().setLexicalPreservationEnabled(true));
CompilationUnit cu = StaticJavaParser.parse(src);

```
Could you try to rewrite to testcase to avoid the usage of these static methods? (I don't know if it's possible, I don't have the code in front of me).

PR #2861, opened 2020-10-20
Adding boolean conditional expressions, numeric conditional expressions to conditional expressions resolution

jlerbsc:
Adding boolean conditional expressions, numeric conditional expressions to conditional expressions resolution

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/promotion/NumericConditionalExprHandler.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/promotion/NumericConditionalExprHandler.java
@@ -1,0 +1,98 @@
+package com.github.javaparser.symbolsolver.resolution.promotion;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import com.github.javaparser.resolution.types.ResolvedPrimitiveType;
+import com.github.javaparser.resolution.types.ResolvedType;
+import com.github.javaparser.symbolsolver.resolution.typeinference.TypeHelper;
+
+/*
+ * Numeric conditional expressions are standalone expressions (§15.2).
+ * The type of a numeric conditional expression is determined as follows:
+ * If the second and third operands have the same type, then that is the type of the conditional expression.
+ * If one of the second and third operands is of primitive type T, and the type of the other is the result of applying boxing conversion (§5.1.7) to T, then the type of the conditional expression is T.
+ * If one of the operands is of type byte or Byte and the other is of type short or Short, then the type of the conditional expression is short.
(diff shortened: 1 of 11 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/promotion/NumericConditionalExprHandler.java
```diff
@@ -0,0 +1,98 @@
...
+    ResolvedType elseExpr;
+
+    public NumericConditionalExprHandler(ResolvedType thenExpr, ResolvedType elseExpr) {
+        this.thenExpr = thenExpr;
+        this.elseExpr = elseExpr;
+    }
+
+    @Override
+    public ResolvedType resolveType() {
+        String qnameTypeThenExpr = thenExpr.isPrimitive() ? thenExpr.asPrimitive().describe()
```
matozoid, line 43: I think we have "ifPrimitive" too, which gives an Optional, and then you can use map().orElse()

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/AbstractSymbolResolutionTest.java
```diff
@@ -21,13 +21,24 @@
...
+import org.junit.jupiter.api.AfterEach;
+
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.utils.CodeGenerationUtils;
+
 public abstract class AbstractSymbolResolutionTest {
+    
+    @AfterEach
+    public void reset() {
```
matozoid, line 37: I decided to create a test parser somewhere, I was using that lately. I'm not sure what it is called. Better not use StaticJavaParser for the reason you mention.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/AnnotationsResolutionTest.java
```diff
@@ -21,6 +21,16 @@
 
 package com.github.javaparser.symbolsolver.resolution;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
```
matozoid, line 27: There is a formatter settings file in the repository. (Ignore this if the formatting of this file was incorrect :-) )

jlerbsc: Done.

matozoid:
Why not merge this? The list of PR's is enormous :-)

PR #2891, opened 2020-11-07
Update dependencies

MysterAitch:


MysterAitch:
Looks good to me, but coverage has fallen -- despite no actual code changes (just dependency changes).

Could this be due to some tests no longer being picked up, perhaps..? Or maybe it is more lines of code being picked up that weren't before, perhaps..?

PR #2898, opened 2020-11-10
Fix issue 2489 SymbolSolver choosing wrong method after resolving

### LINKED ISSUE #2489: SymbolSolver choosing wrong method after resolving
Suppose I have three classes: ComponentBase, ObjectContextDecorator and ObjectContext.

```
public class ComponentBase extends ObjectContextDecorator {
}
```

```
public class ObjectContextDecorator extends ObjectContext {

    public void set(String objectTypePath, String characteristicName, Object value) {
        //...
    }
}
```

```
public class ObjectContext {
    public void set(String characteristicName, Object value, boolean forced) {
       //...
    }
}
```

Then, I have the following piece of code. The "set" method called below needs to be resolved by the SymbolSolver:

```
ComponentBase otm4e = _this.getParent().getParent();
otm4e.set(""OTM4E_EFFLEVEL"", ""IE1 / STD"", true);
```

Then, the resolve method points out that the "set" method called belongs to ObjectContextDecorator. In fact, in this situation, it does belong to ObjectContext (that is what the programmer intended to write). Is there any way to change priority to the upper levels of the hierarchy? It seems to me there is a reason why the resolver gave preference to ObjectContextDecorator, even when ObjectContext also has a method with a signature that matches the method called.

jlerbsc:
Fixes #2489 .

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/MethodResolutionLogic.java
@@ -654,6 +654,13 @@ public class MethodResolutionLogic {
                     paramTypeB.isAssignableBy(argType)) {
 
                 return false;
+            // if paramA and paramB are not the last parameters
+            // and the type of paramA or paramB (which are not more specific at this stage) is java.lang.Object
+            // then we have to consider others parameters before concluding
+            } else if (((i + 1) < numberOfArgs)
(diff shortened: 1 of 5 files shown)


jlerbsc:
The Travis CI build is still in progress after 1 hour so i merge the pull request without waiting any longer.

MysterAitch:
Thank you again @jlerbsc for the incredible work you're putting in to closing off issues and submitting pull requests!! :heart:

jlerbsc:
Thank you @MysterAitch It's just a small contribution to this incredible project... and it allows me to work on my English. I need it!

PR #2900, opened 2020-11-10
Fix issue 2062 Type inference issue for method arguments

### LINKED ISSUE #2062: Type inference issue for method arguments
Please consider the following code snippet.

```java
public class Base{
	class Derived extends Base{
	}
	
	public void bar(Optional<Base> o) {
	}

	public void foo() {
		bar(Optional.of(new Derived()));
	}
}
```

For this code the JSS is unable to find a declaration for the method call to `bar` because the argument type is solved as `Optional<Derived>` where Java actually would infer `Optional<Base>`. At least I think that this causes the issue.

jlerbsc:
Fixes #2062.


jlerbsc:
Travis job is queued since more than one hour so i merge the pull request without waiting for the end of the job.

MysterAitch:
> 
> 
> Travis job is queued since more than one hour so i merge the pull request without waiting for the end of the job.

Looks like API issues between travis and github -- this happened back in April too. On the other PRs it's green on travis with the tests passing but github status hadn't updated yet.

Personally, to say this publicly, if the github actions and appveyor are green then ignore travis :)

PR #2903, opened 2020-11-11
Issue 2613 auto update the version in the readme

### LINKED ISSUE #2613: Update versions in the readme when the version changes
My presumption is that there should be a way to do this automatically as part of the release process.

I'm not 100% how best to achieve this, so before running with it I'd like to open it out for discussion.

To start things off, [this SO question](https://stackoverflow.com/questions/34629326/update-versions-contained-in-readme-on-maven-release) has a few answers/suggestions. 

One suggestion there is to have a template in the resources directory with placeholders for version numbers, where these placeholders can be replaced during the build (and/or release) process.

MysterAitch:
Fixes #2613

This PR:
- Has the readme.md file update with the current pom version during compile, ONLY when the `releasing` profile is enabled
- During the release:prepare phase, enable the `releasing` profile (thus update the pom version and the readme)
- Commit the updated readme.md file

This creates an additional commit during the `release:prepare`:
image

(for now, it also has a manual update of the readme.md version -- in the future it should update automatically)

### DIFF
--- a/doc/readme.md
+++ b/doc/readme.md
@@ -31,14 +31,14 @@ Just add the following to your maven configuration or tailor to your own depende
 <dependency>
     <groupId>com.github.javaparser</groupId>
     <artifactId>javaparser-symbol-solver-core</artifactId>
-    <version>3.16.4-SNAPSHOT</version>
+    <version>${project.version}</version>
 </dependency>
 ```
 
 **Gradle**:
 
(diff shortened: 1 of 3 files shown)


MysterAitch:
I'm going to merge as all the testing locally confirms it works - we can always revert it back if needs be! :)

PR #2904, opened 2020-11-11
Fix LOOKAHEAD for ReferenceType; fixes #2879

### LINKED ISSUE #2879: Crash on annotated varargs parameter
JavaParser has helped significantly in a project I'm working on, but I get a crash in one specific edge case. It occurs when a method takes a varargs parameter of primitive type with an annotation. I'm running this code.
```java
package test;

public class AnnotatedVarArgs  {
        StaticJavaParser.parse(
                "import java.lang.annotation.*;" +
                "public class Test {" +
                "    public void method(int @MyAnno ... param) {}" +
                "}" +
                "@Target(ElementType.TYPE_USE) @interface MyAnno {}"
                );
    }
}
```

Here's the crash:
```
com.github.javaparser.ParseProblemException: (line 1,col 78) Parse error. Found "...", expected one of  "["
Problem stacktrace :
  com.github.javaparser.GeneratedJavaParser.generateParseException(GeneratedJavaParser.java:10906)
  com.github.javaparser.GeneratedJavaParser.jj_consume_token(GeneratedJavaParser.java:10752)
  com.github.javaparser.GeneratedJavaParser.ArrayBracketPair(GeneratedJavaParser.java:1786)
  com.github.javaparser.GeneratedJavaParser.ReferenceType(GeneratedJavaParser.java:1739)
  com.github.javaparser.GeneratedJavaParser.Type(GeneratedJavaParser.java:1700)
  com.github.javaparser.GeneratedJavaParser.Parameter(GeneratedJavaParser.java:1462)
  com.github.javaparser.GeneratedJavaParser.Parameters(GeneratedJavaParser.java:1377)
  com.github.javaparser.GeneratedJavaParser.MethodDeclaration(GeneratedJavaParser.java:1225)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBodyDeclaration(GeneratedJavaParser.java:961)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBody(GeneratedJavaParser.java:852)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceDeclaration(GeneratedJavaParser.java:409)
  com.github.javaparser.GeneratedJavaParser.CompilationUnit(GeneratedJavaParser.java:152)
  com.github.javaparser.JavaParser.parse(JavaParser.java:306)
  com.github.javaparser.StaticJavaParser.parse(StaticJavaParser.java:229)
  test.AnnotatedVarArgs.main(AnnotatedVarArgs.java:7)
  org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:254)
  java.base/java.lang.Thread.run(Thread.java:834)

    at com.github.javaparser.StaticJavaParser.handleResult (StaticJavaParser.java:260)
    at com.github.javaparser.StaticJavaParser.parse (StaticJavaParser.java:229)
    at test.AnnotatedVarArgs.main (AnnotatedVarArgs.java:7)
    at org.codehaus.mojo.exec.ExecJavaMojo$1.run (ExecJavaMojo.java:254)
    at java.lang.Thread.run (Thread.java:834)
```

The parsed code compiles with javac, so I assume it's valid. The crash doesn't occur if I change `int` to `String` or `Object` so the problem seems to be unique to primitives.

JavaParser version 3.16.2
OpenJDK version 11.0.9

mernst:
Fixes #2879.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
@@ -308,4 +308,13 @@ class JavaParserTest {
     void xxx(){
         YamlPrinter.print(StaticJavaParser.parse("class X{}"));
     }
+
+    @Test
+    void issue2879() {
+        StaticJavaParser.parse(
+            "public class Test {" +
+            "    public void method(int @MyAnno ... param) {}" +
+            "}" +
+            "@Target(java.lang.annotation.ElementType.TYPE_USE) @interface MyAnno {}");
+    }
 }
--- a/javaparser-core/src/main/javacc/java.jj
+++ b/javaparser-core/src/main/javacc/java.jj
@@ -1249,7 +1249,7 @@ Type Type(NodeList<AnnotationExpr> annotations):
 }
 {
  (
-   LOOKAHEAD(2) ret = ReferenceType(annotations)
+   LOOKAHEAD(PrimitiveType(annotations) Annotations() "[" | ClassOrInterfaceType(annotations)) ret = ReferenceType(annotations)
  |
    ret = PrimitiveType(annotations )
  )


MysterAitch:
Good shout, thanks @mernst ! :) 

How do you feel about using `LOOKAHEAD(ReferenceType(annotations))` instead, to reduce the duplicated logic of "array of primitives or a class/interface name"? 

Depending on the generated parser code, I imagine yours may be a little more performant though?

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
```diff
@@ -308,4 +308,13 @@ void parseTypeDeclaration() {
     void xxx(){
         YamlPrinter.print(StaticJavaParser.parse("class X{}"));
     }
+
+    @Test
+    void issue2879() {
```
MysterAitch, line 313: Should this be in `JavaParserTest`? Perhaps we could move it to `MethodDeclarationTest` or similar?

mernst: I didn't put it in `MethodDeclarationTest` because that file contains no calls to `StaticJavaParser.parse`, only calls to `StaticJavaParser.parseBodyDeclaration`.

I put it here becasue it is adjacent to a different call to `StaticJavaParser.parse`.

I didn't find a README, nor class documentation, explaining the purpose of each test directory and file.

Just let me know where you want it, and I will move it.

mernst:
> How do you feel about using `LOOKAHEAD(ReferenceType(annotations))` instead, to reduce the duplicated logic of "array of primitives or a class/interface name"?

I think that would work.  It does seem less efficient since it requires lookahead through the entire reference type rather than only looking ahead far enough to distinguish the shift conflict.

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
MysterAitch: > I didn't find a README, nor class documentation, explaining the purpose of each test directory and file.

Can't argue there -- very much a free-for-all! ;)

MysterAitch:
Travis is being a pain with queues/delays, so merging without waiting for it :)

mernst:
@MysterAitch Thanks for merging.

MysterAitch:
@mernst Thank you for giving me something to merge!! It really is appreciated :+1:

PR #2905, opened 2020-11-11
Codecov testing thing

MysterAitch:
jacoco doesn't run in the testing submodules, without access to the built sources (copied) from the other submodules


MysterAitch:
Oops, PR against the wrong repo (supposed to be my fork, not this one..)

PR #2911, opened 2020-11-12
Fix issue 2909 Unexpected result when solving an outer class reference

### LINKED ISSUE #2909: Unexpected result when solving an outer class reference
Consider this example:
```java
package test;

public class Program {

    public class OuterClass {
        int field = 0;

        public class InnerClass {
            InnerClass() {
                OuterClass outer = OuterClass.this;
                OuterClass.this.field = 1;
            }
        }
    }
}
```
Using a visitor on the compilation unit, when reaching the `ThisExpr` node and resolving the node with `calculateResolvedType()` the `ResolvedType` corresponds unexpectedly to the `OuterClass.InnerClass`.
Of course, the expected type should've been `OuterClass` as `OuterClass.this` references the outer class.
Looks like the SymbolSolver is erroneously dismissing `ThisExpr`'s typename.

Any thoughts on this? Is it a bug or I am missing something?

jlerbsc:
Fixes #2909 .
Removing previous remote commits on a forked repository generates an unexpected large list of commits.

### DIFF
--- a/.github/workflows/maven_tests.yml
+++ b/.github/workflows/maven_tests.yml
@@ -27,6 +27,9 @@ jobs:
         ### exclude pre-8 (min development version jdk8)
         ### exclude post-12 (changes to jdk causes reflection tests to fail due to added methods #1701 )
         jdk: [8,9,10,11,12,13,14,15]
+    env:
+      OS: ${{ matrix.os }}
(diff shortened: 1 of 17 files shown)


MysterAitch:
@jlerbsc can we close this given #2920 has been merged? :)

PR #2921, opened 2020-11-14
Refactoring of the fix to the issue 2909

jlerbsc:

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
@@ -20,7 +20,32 @@
  */
 package com.github.javaparser.ast;
 
-import com.github.javaparser.*;
+import static com.github.javaparser.JavaToken.Kind.EOF;
+import static com.github.javaparser.Providers.UTF8;
+import static com.github.javaparser.Providers.provider;
+import static com.github.javaparser.Range.range;
+import static com.github.javaparser.StaticJavaParser.parseImport;
+import static com.github.javaparser.StaticJavaParser.parseName;
+import static com.github.javaparser.ast.Modifier.createModifierList;
+import static com.github.javaparser.utils.CodeGenerationUtils.subtractPaths;
+import static com.github.javaparser.utils.Utils.assertNotNull;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -500,6 +504,15 @@ public AnnotationDeclaration addAnnotationDeclaration(String name, Modifier.Keyw
     public Optional<ClassOrInterfaceDeclaration> getClassByName(String className) {
         return getTypes().stream().filter(type -> type.getNameAsString().equals(className) && type instanceof ClassOrInterfaceDeclaration && !((ClassOrInterfaceDeclaration) type).isInterface()).findFirst().map(t -> (ClassOrInterfaceDeclaration) t);
     }
+    
+    /**
+     * Try to get a locally class declaration by its name (top level or inner class)
+     *
+     * @param className the class name (case-sensitive)
+     */
+    public Optional<ClassOrInterfaceDeclaration> getLocaleDeclarationFromClassname(String className) {
```
MysterAitch, line 513: Typo - `getLocalDeclarationFromClassname`

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -500,6 +504,15 @@ public AnnotationDeclaration addAnnotationDeclaration(String name, Modifier.Keyw
     public Optional<ClassOrInterfaceDeclaration> getClassByName(String className) {
         return getTypes().stream().filter(type -> type.getNameAsString().equals(className) && type instanceof ClassOrInterfaceDeclaration && !((ClassOrInterfaceDeclaration) type).isInterface()).findFirst().map(t -> (ClassOrInterfaceDeclaration) t);
     }
+    
+    /**
+     * Try to get a locally class declaration by its name (top level or inner class)
```
MysterAitch, line 509: Typo - "get a local class declaration" (alternative: "get a locally declared class")

MysterAitch:
just a quick note to say that personally I don't have any objections to ignoring codecov and merging anyway for small changes like this

Coveralls says it's a change of -0.0001% coverage -- I wouldn't want the check to get in the way of incremental fixes / updates for the sake of a tiny fraction of a percentage!! :) 

_(looks like I have some configuring to do re: thresholds for codecov)_

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
jlerbsc: Thank you for these corrections because my level in English is relatively low.

PR #2924, opened 2020-11-15
fixed ellipsis and doublecolon to be categorised as separators not operators (fixes #2897)

### LINKED ISSUE #2897: `...` and `::` are (wrongly) categorised as operators, rather than separators
It seems that `...` and `::` are (wrongly) categorised as operators, rather than separators.

Should be a fairly easy one to fix (the grammar in `java.jj` and `JavaToken.Category` in `TokenTypes.java`)

Relevant JLS:
https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.11
https://docs.oracle.com/javase/specs/jls/se15/html/jls-3.html#jls-3.12

Older JLS versions:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-3.html#jls-3.11
https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.11

MysterAitch:
Fixes #2897

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaTokenTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaTokenTest.java
@@ -24,13 +24,19 @@ package com.github.javaparser;
 import com.github.javaparser.ast.expr.Expression;
 import org.junit.jupiter.api.Test;
 
+import java.lang.reflect.Field;
 import java.util.Iterator;
 
 import static com.github.javaparser.GeneratedJavaParserConstants.*;
-import static com.github.javaparser.JavaToken.Category.*;
+import static com.github.javaparser.JavaToken.Category.COMMENT;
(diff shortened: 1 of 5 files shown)


MysterAitch:
It's not immediately clear to me why moving the ellipsis and double colon causes these specific tests to fail. 

Doing diligence prior to merging I had another look at the "fix" I did and wasn't happy - hence the follow up commit that is now failing.

As a result, I'll bump this from the "next release" milestone as it seems like it's a non-trivial change with some other side effects somewhere else.

MysterAitch:
> It's not immediately clear to me why moving the ellipsis and double colon causes these specific tests to fail.

Turns out that `JavaToken` wasn't updated, which meant that it became out of sync with `GeneratedJavaParserTokens` -- reordering the tokens meant that the number index thing didn't align/sync, and led to dodgy behaviours.

Just force pushed a tidier commit which includes a unit test to reduce the chance of this being a problem again in the future :+1:

PR #2931, opened 2020-11-17
Fix issue 2909 try different bottom / up and try with type solver strategies

jlerbsc:

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -620,7 +620,7 @@ public class JavaParserFacade {
      * references an outer class -- as its ancestor, return the declaration corresponding to the class name specified.
      */
     protected Node findContainingTypeDeclOrObjectCreationExpr(Node node, String className) {
-        if (node instanceof ClassOrInterfaceDeclaration && ((ClassOrInterfaceDeclaration) node).getFullyQualifiedName().get().equals(className)) {
+        if (node instanceof ClassOrInterfaceDeclaration && ((ClassOrInterfaceDeclaration) node).getFullyQualifiedName().get().endsWith(className)) {
             return node;
         }
         if (node instanceof EnumDeclaration) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
@@ -28,7 +28,6 @@ import java.util.List;
 import java.util.Optional;
 
 import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.FieldDeclaration;
 import com.github.javaparser.ast.body.Parameter;
 import com.github.javaparser.ast.body.VariableDeclarator;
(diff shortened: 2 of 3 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
```diff
@@ -620,7 +620,7 @@ protected Node findContainingTypeDeclOrObjectCreationExpr(Node node) {
      * references an outer class -- as its ancestor, return the declaration corresponding to the class name specified.
      */
     protected Node findContainingTypeDeclOrObjectCreationExpr(Node node, String className) {
-        if (node instanceof ClassOrInterfaceDeclaration && ((ClassOrInterfaceDeclaration) node).getFullyQualifiedName().get().equals(className)) {
+        if (node instanceof ClassOrInterfaceDeclaration && ((ClassOrInterfaceDeclaration) node).getFullyQualifiedName().get().endsWith(className)) {
```
MysterAitch, line 623: I was thinking about this for #2928 too, which also uses `endsWith`... 

Presumably `SomeOtherObject` would match `endsWith("Object")`, meaning that something more sophisticated like splitting it then iterating right to left would be needed in order to do this robustly?

jlerbsc: I don't see a use case that wouldn't work with this algorithm but I could be wrong. I modified the unit tests by moving a class to test the reliability of the algorithm.

MysterAitch: I was thinking more along the lines of `endsWith` being too greedy, and accidentally matching something it shouldn't. 

e.g.
```
searching for endsWith("c.ClassName") matches:

a.b.c.ClassName
x.y.z.a.b.c.ClassName
a.b.abc.ClassName
```


```
searching for endsWith("List") matches:

java.util.List
java.util.ArrayList
com.github.javaparser.NodeList
com.meeeee.List
```

The "correct" reference for each will depend on the imports used, iirc.

I don't know enough about these tests to know if these are realistic concerns or not though, so please do feel free to tell me I'm wrong / don't need to worry about this! :+1:

jlerbsc: You are right but what we are trying to resolve is a ThisExpr with (or without a Typename) in the compilation unit (or outside it). In case of inner classes, in your examle, c is declared in b that is declared in a ...
So if we search first from the declaration of the expression "this" and we go up the tree until it has no more parent, then we use the symbol solver and infine the name of the class that fully matches that should be correct. In my opinion the last step (if all fails) is useless it would certainly be better to throw an exception but it's actually a part of the implementation.

If you think this is not a good approach I give up and we can integrate the deadlocklogic proposition. No problem ;-)
I'm just trying to find a more elegant solution but it still has to work. Currently the solution correctly performs all the tests.

MysterAitch: What you say seems to make sense, thanks for explaining! 

> If you think this is not a good approach I give up

As long as it works, I really don't care either way :wink: 

Worst case scenario is that either one gets merged and then it then gets improved upon later -- nice, chilled, and fairly low-stakes :)

jlerbsc: Ok i merge it and wait and see if mister deadlocklogic finds a new use case that would not be well handled by this fix.

deadlocklogic:
Yeah, I believe this patch should work because I tested it intensively on my samples. Still would be better to throw an exception rather than ignoring the typeName and deal as it is plain `this`.

PR #2963, opened 2020-11-29
add bndtools, build OSGi-Bundle

MysterAitch:
Rebased copy of #2824


matozoid:
If this works, why not merge it?

MysterAitch:
> If this works, why not merge it?

That's just it -- I couldn't witness it working or making any changes.

iirc there were changes expected that weren't being produced (addition of a package perhaps?), but it might just that I don't fully understand how the OSGi / bnd file side of things works.

See also: https://github.com/javaparser/javaparser/pull/2824#issuecomment-735442899

PR #2974, opened 2020-12-05
Major update for pretty print, adding interfaces for printer configur…

jlerbsc:
…ation, printer, ... and deprecated old PrettyPrinter API

Since the updates included in this PR have potential impacts on JP users, it must be deployed on a major version of JP


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -112,20 +150,29 @@
 import com.github.javaparser.ast.type.WildcardType;
 import com.github.javaparser.ast.visitor.Visitable;
 import com.github.javaparser.ast.visitor.VoidVisitor;
+import com.github.javaparser.printer.configuration.PrettyPrinterConfiguration;
 
 /**
  * Outputs the AST as formatted Java source code.
- *
+ * This class is no longer acceptable to use because it is not sufficiently configurable and it is too tied to a specific implementation
```
MysterAitch, line 157: Please prefix the javadoc note with `@deprecated`

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
```diff
@@ -0,0 +1,122 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.printer;
+
+import com.github.javaparser.Position;
+
+public interface PrintableSource {
```
MysterAitch: Please add JavaDoc explaining what exactly a printable source is. Presumably this is a node, but without examining the rest of the files it's not immediately clear.

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
```diff
@@ -0,0 +1,122 @@
...
+    PrintableSource indent();
+
+    /**
+     * Add to the current indentation until it is reaches "column" and push it on the indentation stack.
+     * Does not actually output anything.
+     */
+    PrintableSource indentWithAlignTo(int column);
+
+    /**
+     * Pop the last indentation of the indentation stack.
```
MysterAitch: "indentation stack" sounds like implementation detail that doesn't belong within an interface. Could this be reworded?

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
```diff
@@ -0,0 +1,122 @@
...
+    /**
+     * Append the source string passed as argument to the buffer.
+     * If this is being appended at the beginning of a line, performs indentation first.
+     * <p>
+     * The source line to be printed should not contain newline/carriage-return characters;
+     * use {@link #println(String)} to automatically append a newline at the end of the source string.
+     * If the source line passed as argument contains newline/carriage-return characters would
+     * impredictably affect a correct computation of the current {@link #getCursor()} position.
+     *
+     * @param arg source line to be printed (should not contain newline/carriage-return characters)
```
MysterAitch: shall we strengthen "should not" to "must not"?

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
```diff
@@ -0,0 +1,122 @@
...
+     * @return this instance, for nesting calls to method as fluent interface
+     */
+    PrintableSource println(String arg);
+
+    /**
+     * Append a newline to the buffer.
+     *
+     * @return this instance, for nesting calls to method as fluent interface
```
MysterAitch: let's add this `@return` to the other methods too? (`indent`/`unindent` etc)

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+        COLUMN_ALIGN_FIRST_METHOD_CHAIN(Boolean.class),
+        /**
+         * Indent the case when it is true, don't if false
+         * switch(x) {            switch(x) {
+         *    case 1:             case 1:
+         *        return y;           return y;
+         *    case 2:             case 2:
+         *        return z;           return x;
+         *}                       }
+         */
```
MysterAitch: General comment applicable to multiple places: 

`<pre>{@code }</pre>` will help with formatting the javadoc, and allowing special characters to be used

Example: 

```java
        /**
         * Indent the case when it is true, don't if false
         * <pre>{@code
         * switch(x) {            switch(x) {
         *    case 1:             case 1:
         *        return y;           return y;
         *    case 2:             case 2:
         *        return z;           return x;
         *}                       }
         *}<pre>
         */
```

MysterAitch:
Based on a quick read-through the github web ui's "changed files", I've no immediate worries! Will have a proper exploration in the near future though :+1: 

For now though, some initial thoughts:

- `PrettyPrintable` reads as if it should be an interface, not a concrete class. e.g. `Iterable` and other `-able` interfaces that describe _what_ an object can do, rather than _how_ it does it. 
   e.g. `String printed = new PrettyPrintable(config).print(cu);`


- If we're doing a substantive change, could we take this opportunity to revisit the naming of the `print()` method please? It doesn't do any "printing" -- _instead_, it returns a string, which can then be passed to something that actually does the printing (e.g. to standard out, or a file, or wherever).

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+import com.github.javaparser.printer.configuration.Indentation.IndentType;
+import com.github.javaparser.utils.Utils;
+
+/**
+ * Configuration options for the {@link Printable}.
+ */
+public class PrinterConfiguration implements ConfigurablePrinter {
+    
+    public enum ConfigOption {
+        ORDER_IMPORTS(Boolean.class),
```
MysterAitch: Let's make this more explicit -- `ORDER_IMPORTS_ALPHABETICALLY`, for instance

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+import com.github.javaparser.utils.Utils;
+
+/**
+ * Configuration options for the {@link Printable}.
+ */
+public class PrinterConfiguration implements ConfigurablePrinter {
+    
+    public enum ConfigOption {
+        ORDER_IMPORTS(Boolean.class), 
+        PRINT_COMMENTS(Boolean.class),
```
MysterAitch: Does `PRINT_COMMENTS` imply `PRINT_JAVADOC`? Or vice-versa? What happens if one is true but the other is not?

I've not read the rest of it yet, maybe `PRINT_COMMENTS` refers to line comments and block comments? If so, let's make this explicit.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+         *        A,
+         *        B,
+         *        C,
+         *        D,
+         *        E,
+         *        F,
+         *        G
+         *     }
+         * </pre>
+         * Set it to a large number to always align horizontally.
```
MysterAitch: Suggestion:
`Set it to a very large number (e.g. {@code Integer.MAX_VALUE} to always align horizontally.`

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+         */
+        DEFAULT_MAX_ENUM_CONSTANTS_TO_ALIGN_HORIZONTALLY(Integer.valueOf(5)),
+        END_OF_LINE_CHARACTER(Utils.SYSTEM_EOL);
+        
+        Object value;
+        
+        Class type;
+        
+        // Option without value
+        <T> ConfigOption(Class clazz) {
```
MysterAitch: I'm rusty with generics -- should this be `Class<T>` ?

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+            value(value);
+        }
+        
+       /* 
+        * Set a value to an option
+        */
+        public ConfigOption value(Object value) {
+            Utils.assertNotNull(value);
+            this.value = value;
+            this.type = value.getClass();
```
MysterAitch: I'm rusty with generics -- If the enum is initialised as being e.g. `Boolean.class` and it turns out that `value` is an `Integer`, will this overwrite the type?

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+     */
+    @Override
+    public ConfigurablePrinter addOption(ConfigOption option) {
+        removeOption(option);
+        options.add(option);
+        return this;
+    }
+    
+    /*
+     * add or update an option
```
MysterAitch: typo - this should be remove an option

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.Set;
+
+import com.github.javaparser.printer.Printable;
+import com.github.javaparser.printer.configuration.Indentation.IndentType;
+import com.github.javaparser.utils.Utils;
+
+/**
+ * Configuration options for the {@link Printable}.
+ */
+public class PrinterConfiguration implements ConfigurablePrinter {
```
MysterAitch: The naming here needs to be tweaked, though I don't have an immediate suggestion -- this is saying that a `Configuration` "is-a" ("has the behaviour of") a `Printer`?

e.g. it would be odd to have a configuration inside of a printer variable:
```java
ConfigurablePrinter printer = new PrinterConfiguration();`
```

Instead of something like this, where the configuration is passed to a printer:
```java
ConfigurablePrinter printer = new PrettyPrinter(new PrinterConfiguration());`
```

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.util.Optional;
+import java.util.Set;
+
+import com.github.javaparser.printer.configuration.PrinterConfiguration.ConfigOption;
+
+/*
+ * This interface defines the api for printer configuration
+ */
+public interface ConfigurablePrinter {
```
MysterAitch: I've just seen this -- the methods (and the javadoc) here are all related to configuration of a printer (i.e. rename this to `PrinterConfiguration` or similar?)

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /*
+     * returns all activated options
+     */
+    Set<ConfigOption> get();
+
+    /*
+     * returns the indentation parameters
+     */
+    Indentation getIndentation();
```
MysterAitch: What is special about indentation, such that it needs its own method here? (as opposed to e.g. `.get(ConfigOption.INDENTATION_STYLE)` or similar)

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrettyPrinterConfiguration.java
```diff
@@ -19,32 +19,30 @@
...
  * Configuration options for the {@link PrettyPrinter}.
+ * This class is no longer acceptable to use because it is not sufficiently configurable and it is too tied to a specific implementation
+ * <p> Use {@link ConfigurablePrinter interface or PrinterConfiguration default implementation } instead.
  */
-public class PrettyPrinterConfiguration {
+@Deprecated
+public class PrettyPrinterConfiguration implements ConfigurablePrinter {
+    
+    
+    ConfigurablePrinter wrapped;
```
MysterAitch: Please rename this to be more explicit - `wrappedConfiguration` for example (`wrapped` elsewhere refers to the wrapped node iirc)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
jlerbsc: I will do this asap

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: i will do this asap

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: In fact I'm not sure this interface is really useful. It's probably an implementation detail

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: I will delete this interface

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: I will delete this interface

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: I will do this asap. it's the legacy code see PrettyPrinterConfiguration

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: It's from the legacy code see methods on PrettyPrinterConfiguration. I don't know exactly what they do.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: It's from the legacy code see methods on PrettyPrinterConfiguration. I don't know exactly what they do.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: It's from the legacy code but i will do it asap.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: I haven't tested but it should throw a ClassCastException exception

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: This class represents the implementation of the configuration of a printer. I am open to a suggestion to improve understanding

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: This interface defines what is a configuration of a printer and PrinterConfiguration is an implémentation (default implementation)

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: Indentation is an other concept than Option so it makes a sens to distinct them.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrettyPrinterConfiguration.java
jlerbsc: I do this asap.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: You are right it is used to order imports alphabetically. I insert javadoc comment on enum. I think that should be explicit enough.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: In fact i think that these options are related to a specific visitor implementation. So i think there is another refactoring decision to take to isolate the option interface and the option implementation. But I would do this soon in another PR.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: ConfigurablePrinter config = new PrinterConfiguration();
config.addOption(ConfigOption.ORDER_IMPORTS.value("1")); // there is no verification here
Boolean b1 = config.get(ConfigOption.ORDER_IMPORTS).get().asBoolean(); // java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Boolean
Boolean b2 = config.get(ConfigOption.ORDER_IMPORTS).get().asValue(); // java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Boolean
Perhaps it could be improved!

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: I propose PrinterConfigurable for the interface (to specify the interface of the configuration of a printer) and PrinterConfiguration or DefaultPrinterConfiguration for the implementation. What do you think about this?

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: I think I will try to take this remark into account but maybe integrate it in another PR (asap)

javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
```diff
@@ -196,7 +199,9 @@ public static void main(String[] args) throws Exception {
                 .setLanguageLevel(ParserConfiguration.LanguageLevel.RAW)
                 .setStoreTokens(false);
         final SourceRoot sourceRoot = new SourceRoot(root, parserConfiguration);
-        sourceRoot.setPrinter(new PrettyPrinter(new PrettyPrinterConfiguration().setEndOfLineCharacter("\n"))::print);
+        ConfigurablePrinter config = new PrinterConfiguration().addOption(ConfigOption.END_OF_LINE_CHARACTER.value("\n"));
+        Printable printer = new PrettyPrintable(config);
```
matozoid: A printer is a printable? I don't want to print the printer?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
```diff
@@ -21,23 +21,29 @@
...
-import static org.junit.jupiter.api.Assertions.*;
+import com.github.javaparser.printer.configuration.ConfigurablePrinter;
+import com.github.javaparser.printer.configuration.Indentation;
+import com.github.javaparser.printer.configuration.Indentation.IndentType;
+import com.github.javaparser.printer.configuration.PrinterConfiguration;
 
 class CommentTest {
 
-    private static final PrettyPrinterConfiguration PRETTY_PRINTER_CONFIG_TWO_INDENT = new PrettyPrinterConfiguration().setIndentSize(2);
+    private static final ConfigurablePrinter PRETTY_PRINTER_CONFIG_TWO_INDENT = new PrinterConfiguration().setIndentation(new Indentation(IndentType.SPACES, 2));
```
matozoid: A printer configuration is a configurable printer? Eh?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -47,19 +49,23 @@
...
 import com.github.javaparser.ast.stmt.Statement;
 import com.github.javaparser.ast.type.PrimitiveType;
+import com.github.javaparser.ast.visitor.VoidVisitor;
+import com.github.javaparser.printer.configuration.ConfigurablePrinter;
 import com.github.javaparser.printer.configuration.Indentation;
 import com.github.javaparser.printer.configuration.Indentation.IndentType;
+import com.github.javaparser.printer.configuration.PrinterConfiguration;
+import com.github.javaparser.printer.configuration.PrinterConfiguration.ConfigOption;
 
 class PrettyPrinterTest {
```
matozoid, line 62: Forgot to rename the class?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -225,9 +234,14 @@ void prettyAlignMethodCallChainsIndentsArgumentsWithBlocksCorrectly() {
 
         CompilationUnit cu = parse("class Foo { void bar() { a.b.c.d.e; a.b.c().d().e(); a.b.c().d.e(); foo().bar().baz(boo().baa().bee()).bam(); foo().bar().baz(boo().baa().bee()).bam; foo().bar(Long.foo().b.bar(), bam).baz(); foo().bar().baz(foo, () -> { boo().baa().bee(); }).baz(() -> { boo().baa().bee(); }).bam(() -> { boo().baa().bee(); }); } }");
         Indentation indentation = new Indentation(TABS_WITH_SPACE_ALIGN, 1);
-        String printed = new PrettyPrinter(new PrettyPrinterConfiguration().setColumnAlignFirstMethodChain(true).setColumnAlignParameters(true).setIndentation(indentation))
-                .print(cu);
 
+        ConfigurablePrinter config = new PrinterConfiguration()
+                .addOption(ConfigOption.COLUMN_ALIGN_FIRST_METHOD_CHAIN)
```
matozoid: Hm, the explicit configuration methods are designed to have the same API approach as ParserConfiguration, and to aid in discoverability (I can look at the methods on the configuration class, instead of looking at the configuration class, then looking at the ConfigOption class - an extra step.)

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AATest.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.ImportDeclaration;
+import com.github.javaparser.ast.NodeList;
+
+public class AATest  extends AbstractLexicalPreservingTest {
```
matozoid: Anti-aliasing test?

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -99,6 +100,9 @@
 
     @InternalProperty
     private Storage storage;
+    
+    // printer used to print the node
+    private Printable printer;
```
matozoid, line 105: A per-node printer? Do we want to specifiy printers at this detailed level? Is this still easy to use?

matozoid: Consider putting this in the "data" map to avoid littering the class with fields.

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
```diff
@@ -755,7 +760,7 @@ public Path getDirectory() {
          * Saves the compilation unit to its original location
          */
         public void save() {
-            save(cu -> printer.print(cu));
+            save(cu -> compilationUnit.getPrinter().print(cu));
```
matozoid, line 773: Is there a reason we want a printer per file, and not a printer per project?

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
```diff
@@ -199,6 +202,37 @@ protected Node(TokenRange tokenRange) {
      */
     protected void customInitialization() {
     }
+    
+    /*
+     * If there is a printer defined in CompilationUnit, returns it
+     * else create a new PrettyPrintable with default parameters
+     */
+    protected Printable getPrinter() {
+        Optional<CompilationUnit> cu = findCompilationUnit();
```
matozoid: You can use map().orElse().

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrintableSource.java
```diff
@@ -0,0 +1,272 @@
...
+import com.github.javaparser.printer.configuration.ConfigurablePrinter;
+import com.github.javaparser.printer.configuration.Indentation;
+import com.github.javaparser.printer.configuration.Indentation.IndentType;
+import com.github.javaparser.printer.configuration.PrinterConfiguration.ConfigOption;
+import com.github.javaparser.utils.Utils;
+
+/**
+ * A support class for code that outputs formatted source code.
+ */
+public class DefaultPrintableSource implements PrintableSource {
```
matozoid: Naming: this is the default for source that is printable? I don't know what that means.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid, line 157: In my opinion all the pretty printing and lexical preservation stuff should be thrown out, and replaced by:
- a library of microformatters (formatting rules) that will format one specific case (like "exactly one space between parameter list and opening brace `a() {`" that work on specific nodes and modify the underlying tokenstream.) which can be selected and configured by the user.
- the AST nodes modifying the underlying tokenstream when nodes and properties are changed/added/removed, after which they will call the user's microformatters.

That would remove an incredible amount of unnecessary complexity currently in JP, and will solve more or less all current lexical preservation and pretty printing issues.

This idea is inspired by Intellij Idea's architecture. Well, copied really ;-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinter.java
```diff
@@ -21,41 +21,60 @@
...
 import com.github.javaparser.ast.visitor.VoidVisitor;
+import com.github.javaparser.printer.configuration.ConfigurablePrinter;
+import com.github.javaparser.printer.configuration.PrettyPrinterConfiguration;
 
 /**
  * Pretty printer for AST nodes.
+ * This class is no longer acceptable to use because it is not sufficiently configurable and it is too tied to a specific implementation
+ * <p> Use {@link Printable interface or PrettyPrintable default implementation } instead.
  */
+@Deprecated
```
matozoid, line 37: This class was added so that people could easily find out how to do pretty printing. Maybe it is no longer wanted, but it would be wise to keep the name "PrettyPrinter" since that is the first thing users will look for when looking for pretty printing functionality.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: This might sound harsh, I know, but think of it like this: it would be really interesting to figure out how to implement this correctly and see it come to life! It is much more rewarding than continuing on the current path. The Pretty Printer visitor approach will never be fun to work on, and will stay very brittle. Just look at all the weirdness that had to be introduced to get stream method calls (map, filter) to print vertically aligned. The lexical preserving printer, an impressive amount of work, has always had trouble because it is a separate component, and keeps its own data about nodes, about token, about node types, which is nowadays all available in the AST and metamodel itself. It always has to sit in its own little corner with its own information, receiving *almost* enough information through the AST listeners to do its work. Putting lexical preservation straight in the middle of the `Node`s will be a big relief!

matozoid: (For completeness: other major improvements to JP would be to move to Antlr, and to rewrite the symbol solver to use a single pass over the whole project instead of reasoning from the inside out)

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
jlerbsc: Yes we lose in semantics but we gain in extensibility.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
jlerbsc: I am open to all naming proposals

jlerbsc:
This PR was just a proposal allowing a user to plug in their own printer which the current API cannot do. But the idea was not to integrate a library of microformatters. I am not opposed to closing this PR (only a little disappointed because it represents several hours of work to make it work with the current code ... it is certainly perfectible). Tell me clearly what you want to do on this subject that I do not spend more time on it if it does not correspond to your vision. ;-)

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
jlerbsc: > That would remove an incredible amount of unnecessary complexity currently in JP,

In my opinion microformatters adds another level of complexity which resides in the management of a multitude of formatting rules because the concept of formatting and especially of preserving the current format is a very complex subject.

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrintableSource.java
jlerbsc: DefaultPrintableSource class is  the same as actual SourcePrinter class (except for the initialization with configuration)

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
jlerbsc: Actually there is a new pretty printer created for each print action because of a state managed in internal class (Sourceprinter). LexicalPreservingPrinter acts as a per project printer, i didn't think it was the preferred orientation. I just tried to integrate a more flexible mechanism into the existing code (but maybe I didn't succeed) :-(

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
jlerbsc: It seems to me that semantically an attribute can make sense and clearly indicate an intention otherwise it would be a questioning of the object-oriented paradigm. But if using a map has a consensus on the project I see no objection.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
jlerbsc: No, this is the actual code. I'm just adapting it to use a PrinterConfiguration (new class) internally.

javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
jlerbsc: I am open to all naming proposals. If you have ideas don't hesitate.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
matozoid: Yes, and you lose consistency and ease of use.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
matozoid: Alright, I guess you're not really interested in my review. I'll go spend my time elsewhere :-)

jlerbsc: No No @matozoid i misspoke. In fact I realize that I should not have to get into this development without sharing with the team a vision of what needed to be done. I arrived last on this project and I have no intention of causing trouble in this team. I was just giving personal advice, you don't have to agree with me. If the vision on this feature is to set up a micro-formatting logic I have nothing to say except that if I had known it earlier I would not have wasted time doing the work I did. Personally I don't feel able to get into microformatting so I'll leave the topic to you and I'll just fix a few bugs.

javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
jlerbsc: Perhaps a Printer interface is more intuitive. PrettyPrinter is a Printer, it make a sense.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
jlerbsc: I will rename ConfigurablePrinter to PrinterConfiguration (configuration of a Printer) and the actual implementation PrinterConfiguration to DefaultPrettyPrinterConfiguration (because PrettyPrinterConfiguration already exists)

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrintableSource.java
jlerbsc: Not sure this is a need for the interface PrintableSource because it's a implementation detail. I will remove it.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrinter.java
jlerbsc: I agree with you and this class is still usable for compatibility purpose. It's why i propose to deprecate this class for the moment. Perhaps this comment "This class is no longer acceptable to use because..." is misinterpreted or poorly worded. I can update this comment if it's the case.

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: I remove this interface because SourcePrinter is an implementation detail. No need for an interface and another implementation.

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: I remove this interface because SourcePrinter is an implementation detail. No need for an interface and another implementation.

javaparser-core/src/main/java/com/github/javaparser/printer/PrintableSource.java
jlerbsc: I remove this interface because SourcePrinter is an implementation detail.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: Now I implements a fail fast assignment
For example if ConfigOption.PRINT_COMMENTS is configured with a Boolean type
config.addOption(ConfigOption.PRINT_COMMENTS.value("1")); // throws an IllegalArgumentException

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: Changed to interface named PrinterConfiguration and implementation named DefaultPrinterConfiguration.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: Changed to interface named PrinterConfiguration and implementation named DefaultPrinterConfiguration.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: An enum can not be defined with generic type. I remove the useless generic type.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: Indentation is integrated in configuration option.

jlerbsc: Indentation is integrated in configuration option

javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
jlerbsc: Renamed Printer

javaparser-core-testing/src/test/java/com/github/javaparser/ast/comments/CommentTest.java
jlerbsc: Configurableprinter interface is renamed PrinterConfiguration

javaparser-core/src/main/java/com/github/javaparser/ast/Node.java
jlerbsc: Updated

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrintableSource.java
jlerbsc: Interface printableSource and class DefautlPrintableSource are removed because there are useless.

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/ConfigurablePrinter.java
jlerbsc: ConfigurablePrinter interface has been renamed in PrinterConfiguration

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: PrinterConfiguration class has been renamed in DefaultPrinterConfiguration

javaparser-core/src/main/java/com/github/javaparser/printer/configuration/PrinterConfiguration.java
jlerbsc: PrinterConfiguration config = new DefaultPrinterConfiguration();
config.addOption(new ConfigurationOption(ConfigOption.PRINT_COMMENTS, "1")); // throws an IllegalArgumentException

jlerbsc:
> * If we're doing a substantive change, could we take this opportunity to revisit the naming of the `print()` method please? It doesn't do any "printing" -- _instead_, it returns a string, which can then be passed to something that actually does the printing (e.g. to standard out, or a file, or wherever).

what do you think? do you have a suggestion?

javaparser-core/src/main/java/com/github/javaparser/ast/CompilationUnit.java
jlerbsc: Property 'printer' is moved to 'data' map.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AATest.java
jlerbsc: I remove this file from the PR

MysterAitch:
Sorry, I spent some time away for a few days so didn't keep up with comments - there is a flurry of new commits and it is now merged? 

With the caveat that I haven't read the updates yet (the GitHub UI is awful - it tells me there are comments but doesn't show where they are?):

Is it possible to keep a backwards-compatible (deprecated) shim of some kind that delegates to the new code?

jlerbsc:
> Is it possible to keep a backwards-compatible (deprecated) shim of some kind that delegates to the new code?

Hi @MysterAitch i'm sorry but i don't understand what you mean. Do you speak about PrettyPrinter?

PR #2976, opened 2020-12-14
Adding method descriptor resolution

### LINKED ISSUE #2059: Method descriptor
Is there a way to get the method descriptor for a method (see http://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.3 ).  JavaSymbolSolver would be great to resolve the fully qualified objects wondering if anyone already has written something like this?

Thanks!

jlerbsc:
Fixes #2059 .
https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.3
The method descriptor for the method: 

     Object m(int i, double d, Thread t) {...}

is 

    (IDLjava/lang/Thread;)Ljava/lang/Object;

Note that the internal forms of the binary names of Thread and Object are used.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/body/MethodDeclaration.java
@@ -20,31 +20,47 @@
  */
 package com.github.javaparser.ast.body;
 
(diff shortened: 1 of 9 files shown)


jlerbsc:
We can get method descriptor from MethodDeclaration.

PR #2997, opened 2020-12-28
Fix issue #2996 Cleanup lexical preservation cache just before the co…

### LINKED ISSUE #2996: LexicalPreservationEnabled causes memory leak
Hi,

In parsing a couple of thousand Java source files, the following happens when `LexicalPreservationEnabled` is enabled, which results in an out of memory error after a while:

Screenshot 2020-12-28 at 15 39 11

Screenshot 2020-12-28 at 15 43 25

<br/><br/>

When I run the same code __without__ `LexicalPreservationEnabled` everything runs fine:

Screenshot 2020-12-28 at 15 36 06

Screenshot 2020-12-28 at 15 36 27

I've tried `JavaParserFacade.clearInstances()` but it doesn't help.

It looks like the GC has issues cleaning up some of the references, but since the GC has no problems when `LexicalPreservationEnabled` is disabled it doesn't seem these objects are still referenced.

Here is the Kotlin code:

```kotlin
File(pathJuliet)
        .listFiles()
        .filterNotNull()
        .filter { it.isDirectory }
        .forEach { directory ->
            val puid = UUID.randomUUID()
            directory
                .walk()
                .filter {
                    it.isFile && it.extension == "java"
                }
                .flatMap { javaFile ->
                    processFile(javaParser, javaFile, puid)
                }
                .let {
                    Json.encodeToString(it)
                }
                .also {
                    File("$pathOutput/${directory.name}.json").writeText(it)
                }
```

Any idea what could be wrong?

jlerbsc:
…mpilation unit is GCed.

Fixes #2996 .

### DIFF
--- a/.github/workflows/maven_tests.yml
+++ b/.github/workflows/maven_tests.yml
@@ -76,7 +76,7 @@ jobs:
         timeout-minutes: 10
         with:
           files: javaparser-core-testing/target/site/jacoco/jacoco.xml,javaparser-core-testing-bdd/target/site/jacoco/jacoco.xml
-          fail_ci_if_error: true # optional (default = false)
+          fail_ci_if_error: false # optional (default = false)
           verbose: false # optional (default = false):
           flags: javaparser-core,AlsoSlowTests,${{ matrix.os }},jdk-${{ matrix.jdk }}
           env_vars: AlsoSlowTests,OS,JDK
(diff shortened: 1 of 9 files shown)


jlerbsc:
In fact there is no need to have a phantom node cache because this attribute could be keep in node data.

PR #3011, opened 2021-01-02
Make the visit order of ModifierVisitor more consistent

mernst:
This pull request makes no logic changes, but it changes the visit order of ModifierVisitor. 

ModifierVisitor works top-down, visiting larger then smaller scopes.

Many visit methods visit annotations and modiifers before members.  This makes sense because the annotations and modifiers apply to the entire node; the entire node is within their scope.  Examples of visit methods that work this way includes those for `Parameter`, `ReceiverParameter`, `VariableDeclarationExpr`, `ArrayCreationLevel`, `VarType`, `PackageDeclaration`, `ModuleDeclaration`, and some other node types.

In other cases, annotations and modifiers are visited later, or the two are visited at different places in the visit order.  This pull request changes such instances.

This change was necessary for a visitor I wrote that needs to know which annotations each construct is in the scope of.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/ModifierVisitor.java
@@ -52,36 +52,36 @@ public class ModifierVisitor<A> implements GenericVisitor<Visitable, A> {
 
     @Override
     public Visitable visit(final AnnotationDeclaration n, final A arg) {
-        NodeList<BodyDeclaration<?>> members = modifyList(n.getMembers(), arg);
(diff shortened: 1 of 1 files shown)


MysterAitch:
Seems reasonable - visit the annotations first, then members? 

Perhaps this is something that we can formalise / make consistent throughout - do you see any other inconsistencies in the order that could be changed?

mernst:
I didn't look for other inconsistencies.  Other visitors could also be changed if desired.  For example, `HashCodeVisitor` could be changed to put annotations and modifiers first, but that would not affect behavior since the hash codes are just added up.

MysterAitch:
Ah... I just realised that the visitor is generated (missing the `@Generated` annotation) which means the ordering gets reset/overridden whenever the generators are run next.

PR #3014, opened 2021-01-02
Remove the receiver parameter if it has no annotations

mernst:
Its only purpose is to write annotations upon.


MysterAitch:
Cheers @mernst -- would you mind adding a test case for it please?

mernst:
It has been 7 weeks and I no longer have a test case handy.  Could you write one to help out?  Thanks!

PR #2512, opened 2020-02-02
[DRAFT] JEP 359: Records (Preview)

matozoid:
Fixes #2446.

PR #3022, opened 2021-01-06
Record support (compilation / parsing only, solving to follow separately)

MysterAitch:
Progress towards the parsing of records #2512

Solving will be done separately

https://openjdk.java.net/jeps/359#Restrictions-on-records

- [x] Records cannot extend any other class,
- [x] cannot declare instance fields other than the private final fields which correspond to components of the state description
- [x] Any other fields which are declared must be static. 
- [x] Records are 
	- [x] implicitly final
	- [x] cannot be abstract
- [x] The components of a record are implicitly final
- [x] they can be declared 
	- [x] top level or 
	- [x] nested,
- [x] they can be generic, 
- [x] they can implement interfaces
- [x] they are instantiated via the new keyword.
- [x] The record's body may declare 
	- [x] static methods, 
    - [x] static fields, 
    - [x] static initializers, 
 	- [x] constructors, `// NOTE: Constructor declarations do not permit arguments/parameters`
  	- [x] instance methods, and 
  	- [x] nested types
- [x] The record, and the individual components in a state description, may be annotated. 
- [x] If a record is nested, then it is implicitly static
- [x] Compact constructors (no parameters)
- [x] Non-Compact constructors (with parameters)
- [x] Local record classes - https://openjdk.java.net/jeps/395#Local-record-classes

Backwards compatability
- [x] Ensure that `record` can be used as an identifier in older versions
- [x] Add validator to prevent `record` being used as a type identifier (e.g. as a class or record or enum name) in recent versions


MysterAitch:
Note that the tests all pass -- it's the lack of test coverage that is giving this PR the red cross :+1:

javaparser-core/src/main/javacc/java.jj
```diff
@@ -330,6 +330,7 @@ TOKEN :
 | < PRIVATE: "private" >
 | < PROTECTED: "protected" >
 | < PUBLIC: "public" >
+| < RECORD: "record" >
```
matozoid, line 333: Maybe you've already done this, but make sure "record" can be used as an identifier, and make sure the token type is identifier when that is the case (and record when it isn't.)

MysterAitch:
The main thing that concerns me at the moment is whether we're going to end up needing a `ClassOrInterfaceOrRecord` class as it will be ambiguous what type of "thing" it will be until it is resolved.

Thoughts on this point specifically would be welcomed :+1:

jlerbsc:
In my opinion, but I am not very aware of the evolutions of the Java language, there seems to be a consensus on the definition of a record. It would be a **kind of class** like an enumeration can be. Unless there are specific features of the records, we can, I think, be considered as a class in the contrary case it will certainly be more appropriate to separate the classes and records as was done for enumerations. In terms of naming convention if a record is some kind of class (without specificity) it is not useful in my opinion to rename the classes ClassOrInterfacexxx.

matozoid:
One suggestion from a long time ago was to split ClassOrInterfaceDeclaration in ClassDeclaration and InterfaceDeclaration. That would answer the question with: a separate RecordDeclaration. If I'm not mistaken they have a shared superclass already, so if you want to talk about "class or interface or record declaration" you can use the superclass.

MysterAitch:
Pretty sure the parsing side of things is sorted now.

A couple of tweaks available -- e.g. in records, the non-static fields are known as "record components", but they're currently only referred to as parameters. Maybe we want to add in a helper method of `getComponents()` or similar (which would just delegate to or replace `getParameters()`, but this is a non-trivial thing given that it returns `Parameter`).

When it comes to solving, these components will have generated getters (but that's separate to the parsing side of things).

MysterAitch:
Most of the patch difference is because of the touches to the visitors btw, which are large files that aren't particularly well tested (they're generated/updated using a generator).

matozoid:
About the record component/parameter mismatch: I would put that in the Javadoc. The Javadoc for every node is supposed to explain what a node stands for, what parts go where, and what is in them.

MysterAitch:
I don't plan to actually do this, but wanted to node down this -- if somebody wants it later, it can potentially be added in.

**NodeWithMembers.java**
```java
    /**
     * Find all components of this node.<br/>
     * <br/>
     * Components are the non-static fields of a record. The declarations for these fields are parameters on the
     * record declaration (record header) - hence the return type.
     *
     * @return The components found. This list is immutable.
     *
     * @deprecated Note this method is subject to change (e.g. could change to be a list of components, or fields) - avoid using it until the final API is determined.
     * @see <a href="https://openjdk.java.net/jeps/359">https://openjdk.java.net/jeps/359</a>
     */
    @Deprecated
    default List<Parameter> getComponents() {
        return unmodifiableList(emptyList());
    }
```

**RecordDeclaration.java**
```java
    /**
     * Find all components of this node.<br/>
     * <br/>
     * Components are the non-static fields of a record. The declarations for these fields are parameters on the
     * record declaration (record header) - hence the return type.
     * <br/>
     * For now, it simply delegates to {@link #getComponents()}.
     *
     * @return The components found. This list is immutable.
     *
     * @deprecated Note this method is subject to change (e.g. could change to be a list of components, or fields) - avoid using it until the final API is determined.
     * @see <a href="https://openjdk.java.net/jeps/359">https://openjdk.java.net/jeps/359</a>
     */
    @Deprecated
    @Override
    public List<Parameter> getComponents() {
        return unmodifiableList(getParameters());
    }
```

MysterAitch:
Just made it draft again -- completely forgot about ensuring `record` can be used as an identifier... currently not possible, with these tests failing with a parse error:

```java
    @Test
    void recordCanBeUsedAsIdentifier() {
        String s = "public class record {}";
        CompilationUnit cu = parseCompilationUnit(s);
    }

    @Test
    void recordCanBeUsedAsIdentifier2() {
        String s = "class X { int record; }";
        CompilationUnit cu = parseCompilationUnit(s);
    }
```

MysterAitch:
https://docs.oracle.com/javase/specs/jls/se15/preview/specs/records-jls.html#jls-3.9

> "var, and yield, and record are not keywords, but rather restricted identifiers"

Turns out that there is a concept of "restricted identifiers". This is different to "keywords".

Therefore, I have just pushed a commit that removes this -- it does mean that (for now) this is a bit more permissive than it should be.

MysterAitch:
I went in circles a little with this -- I couldn't get records to compile at all without record being treated as a token (unexpected identifier token). 

This might be me not fully understanding how to write the grammar though.

I ended up putting it back as a token and adding it here as an acceptable token name:

```
        <MODULE> | <REQUIRES> | <TO> | <WITH> | <OPEN> | <OPENS> | <USES> | <EXPORTS> | <PROVIDES> | <TRANSITIVE> |
        // Make sure older Java versions parse
        <ENUM> | <STRICTFP> | <YIELD> | <RECORD> |
        // An actual plain old identifier
        <IDENTIFIER>
```

MysterAitch:
I think that this is ready to merge in now, but I'll sleep on it and have another look tomorrow.

Outstanding:
- Solving (to be done in a separate PR)
- Double check how we handle the "record as an identifier vs a keyword" (a separate issue)
- Consider adding a note to the javadoc to indicate the way javaparser handles records is subject to change -- "just in case" anything comes up while sorting out the solving (trivial - can be done in this PR)

So.. any last thoughts before it gets merged? :+1:

matozoid:
For token type, check out how "yield" does it. (for records I was trying another approach, but forgot why and whether it is better.)

raghav-deepsource:
Hi, any updates on this? Maybe this can move forward now since Java 16 is now released and records are no longer preview features?

matozoid:
I say merge and make a few issues for the remaining things.

MysterAitch:
I'll get back on this and have it ready for the next weekend's release :+1:

MysterAitch:
> Double check how we handle the "record as an identifier vs a keyword" (a separate issue)

The underlying mechanism might change, but the current implementation works fine (permitted by the grammar, and checked for within a validator).

MysterAitch:
Re-reading through the documentation, there's a compact and non-compact constructor.

A "compact" constructor doesn't include any parameters, but it does also allow the standard style constructors too - looks like I missed that part! 

It's now on the ticklist above :+1: 

Here's an example from the docs: 
https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.10.4.2

```java
record Rational(int num, int denom) {
    private static int gcd(int a, int b) {
        if (b == 0) return Math.abs(a);
        else return gcd(b, a % b);
    }
   
    Rational {
        int gcd = gcd(num, denom);
        num    /= gcd;
        denom  /= gcd;
    }
}
```

> The compact constructor Rational {...} behaves the same as this normal constructor:
```java
record Rational(int num, int denom) {
    private static int gcd(int a, int b) {
        if (b == 0) return Math.abs(a);
        else return gcd(b, a % b);
    }
   
    Rational(int num, int demon) {
        int gcd = gcd(num, denom);
        num    /= gcd;
        denom  /= gcd;
        this.num   = num;
        this.denom = denom;
    }
}
```

MysterAitch:
I forgot that on release there's a new JEP -- here's the one for this which includes a few changes since JEP359 (preview) / JEP384 (2nd preview):
https://openjdk.java.net/jeps/395

PR #3034, opened 2021-01-11
Fixed Resolution of generic types

### LINKED ISSUE #2397: Solving generic type of method paramters or return type fails
Hi, 

For ```MethodDeclaration```s in each ```ClassOrInterfaceDeclaration```, I am trying to obtain resolved return type as well as resolved type of each method parameter. 
I get ```java.lang.UnsupportedOperationException: Unable to get the resolved type of class ResolvedReferenceType from ...``` when calling ```solve()``` on the corresponding return or method parameter type if the type is parametrized. 

The following are examples of code I am trying parse and resolve.

Ex1:
```
static final class ConstantFuture<T> implements Future<T> {
        private final T value;
      
        @Override
        public T get() {
            return value;
        }
}
```

Ex2:
```
    public static <T> T[] toArray(final T... items) {
        return items;
    }
```

Any idea why is this happening?

4everTheOne:
Fixes #2397.


jlerbsc:
Hi @MysterAitch Unless you have another opinion this PR can be merged. Thank you @4everTheOne for this contribution.

PR #3044, opened 2021-01-21
Changed SymbolReference#getCorrespondingDeclaration() to support Optional 

4everTheOne:
When i was working with the code noticed an inconsistency in _SymbolReference_ method comparing with other classes.
Most of the methods that can return null return an optional to allow the programmer to check when the value is present or not.
This method was not following this guideline requiring the use of an additional method _isSolved()_ before the call of _getCorrespondingDeclaration()_.

This Pull Request adds support for the Optional, allowing a more consistent pattern between methods.

## What was changed

In the first commit (**Added Optional to corresponding declaration**) updates the signature to support Optional;
The second commit (**Updated code affected by the change**) refactors code that depended on the old signature;
The third commit (**Update Tests to supports new Optional**) refactors tests that depended on the old signature;
The fourth commit (**Deprecated isSolved method**) deprecated the _isSolved()_ method and suggests the use of the Optional in _getCorrespondingDeclaration()_;
The fifth commit (**Updated tests to match the expected behavior.**) updated the tests to match the new behavior. This change was required, since we don't throw a UnsupportedException anymore because by default when using _Optional#get()_ it throws NoSuchElementException;
The sixth commit (**Code cleanup**) removed unnecessary use of Wildcard;
The seventh (**Added documentation**) updates the documentation for the method _getCorrespondingDeclaration()_.

### **Warning:**
This change may impact the public API.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
@@ -159,7 +159,7 @@ public class JavaSymbolSolver implements SymbolResolver {
             }
         }
         if (node instanceof MethodCallExpr) {
-            Optional<? extends ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
+            Optional<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
                     .solve((MethodCallExpr) node)
                     .getCorrespondingDeclaration();
             if (result.isPresent()) {
@@ -189,7 +189,7 @@ public class JavaSymbolSolver implements SymbolResolver {
             }
         }
         if (node instanceof MethodReferenceExpr) {
-            Optional<? extends ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
+            Optional<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
                     .solve((MethodReferenceExpr) node)
                     .getCorrespondingDeclaration();
             if (result.isPresent()) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
@@ -35,8 +35,9 @@ import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
 import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFactory;
 import com.github.javaparser.symbolsolver.javaparsermodel.declarations.*;
-import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
+
+import java.util.Optional;
 
 /**
  * This implementation of the SymbolResolver wraps the functionality of the library to make them easily usable
@@ -158,51 +159,51 @@ public class JavaSymbolSolver implements SymbolResolver {
             }
         }
         if (node instanceof MethodCallExpr) {
-            SymbolReference<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver).solve((MethodCallExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<? extends ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
+                    .solve((MethodCallExpr) node)
+                    .getCorrespondingDeclaration();
+            if (result.isPresent()) {
+                return resultClass.cast(result.get());
             } else {
                 throw new UnsolvedSymbolException("We are unable to find the method declaration corresponding to " + node);
             }
         }
         if (node instanceof ObjectCreationExpr) {
-            SymbolReference<ResolvedConstructorDeclaration> result = JavaParserFacade.get(typeSolver).solve((ObjectCreationExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<? extends ResolvedConstructorDeclaration> result = JavaParserFacade.get(typeSolver)
+                    .solve((ObjectCreationExpr) node)
+                    .getCorrespondingDeclaration();
+            if (result.isPresent()) {
+                return resultClass.cast(result.get());
             } else {
                 throw new UnsolvedSymbolException("We are unable to find the constructor declaration corresponding to " + node);
             }
         }
         if (node instanceof NameExpr) {
-            SymbolReference<? extends ResolvedValueDeclaration> result = JavaParserFacade.get(typeSolver).solve((NameExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<? extends ResolvedValueDeclaration> result = JavaParserFacade.get(typeSolver)
+                    .solve((NameExpr) node)
+                    .getCorrespondingDeclaration();
+            if (result.isPresent()) {
+                return resultClass.cast(result.get());
             } else {
                 throw new UnsolvedSymbolException("We are unable to find the value declaration corresponding to " + node);
             }
         }
         if (node instanceof MethodReferenceExpr) {
-            SymbolReference<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver).solve((MethodReferenceExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<? extends ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
+                    .solve((MethodReferenceExpr) node)
+                    .getCorrespondingDeclaration();
+            if (result.isPresent()) {
+                return resultClass.cast(result.get());
             } else {
                 throw new UnsolvedSymbolException("We are unable to find the method declaration corresponding to " + node);
             }
         }
         if (node instanceof FieldAccessExpr) {
-            SymbolReference<? extends ResolvedValueDeclaration> result = JavaParserFacade.get(typeSolver).solve((FieldAccessExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<? extends ResolvedValueDeclaration> result = JavaParserFacade.get(typeSolver)
+                    .solve((FieldAccessExpr) node)
+                    .getCorrespondingDeclaration();
+            if (result.isPresent()) {
+                return resultClass.cast(result.get());
             } else {
                 if (((FieldAccessExpr) node).getName().getId().equals("length")) {
                     ResolvedType scopeType = ((FieldAccessExpr) node).getScope().calculateResolvedType();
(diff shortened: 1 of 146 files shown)


4everTheOne:
After some research with the class, found an issue that was reported at #3045.
The last commit (**When not able to solve the parent, throw UnsolvedSymbolException**) fixes this issue by throwing an UnsolvedSymbolException if not found.

jlerbsc:
First of all thank you for this contribution. There is a lot of work and it is necessary to underline your effort to improve the project. Personnally i think this PR is hard to validate because there is a lot of changes and a significant change in the API that can affect a lot of users. Perhaps we can isolate this changes in a new release by specifying that there is an API change. 
@MysterAitch What do you think about this PR?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionMethodResolutionLogic.java
```diff
@@ -59,22 +59,20 @@
...
-                SymbolReference<ResolvedMethodDeclaration> ref = MethodResolutionLogic.solveMethodInType(ancestorTypeDeclaration, name, parameterTypes, staticOnly);
-                if (ref.isSolved()) {
-                    methods.add(ref.getCorrespondingDeclaration());
-                }
-            });
+            ancestor.getTypeDeclaration()
+                    .flatMap(ancestorTypeDeclaration ->
+                            MethodResolutionLogic.solveMethodInType(ancestorTypeDeclaration, name, parameterTypes, staticOnly)
+                                    .getCorrespondingDeclaration()
+                    ).ifPresent(methods::add);
```
MysterAitch: A very minor stylistic thing -- could you move the calls to `ifPresent` to their own line please?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionMethodResolutionLogic.java
```diff
@@ -117,9 +117,11 @@
         if (scopeType.getAncestors().isEmpty()) {
             Optional<ResolvedReferenceTypeDeclaration> optionalObjectClass = new ReferenceTypeImpl(new ReflectionClassDeclaration(Object.class, typeSolver), typeSolver).getTypeDeclaration();
             if (optionalObjectClass.isPresent()) {
-                SymbolReference<ResolvedMethodDeclaration> ref = MethodResolutionLogic.solveMethodInType(optionalObjectClass.get(), name, argumentsTypes);
-                if (ref.isSolved()) {
-                    MethodUsage usage = replaceParams(typeParameterValues, optionalObjectClass.get(), ref.getCorrespondingDeclaration());
+                Optional<ResolvedMethodDeclaration> ref = MethodResolutionLogic
+                        .solveMethodInType(optionalObjectClass.get(), name, argumentsTypes)
+                        .getCorrespondingDeclaration();
+                if (ref.isPresent()) {
```
MysterAitch, line 124: This reads as a change in behaviour -- going from "is solved" to "is present".

Looking at the code, I can see that these methods are equivalent.

Any ideas I come up with to change it are awkward / seem overly verbose / introduce repeated checks (e.g., wrapping in `ifPresent` then doing the `isSolved` inside that (which then again calls `isPresent`) which would be), but if you can think of a quick/easy way around this please go for it!

Just a comment to say "could this be improved?", not a blocker :+1:

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/SymbolSolverWithJavassistClassTest.java
```diff
@@ -83,10 +84,10 @@ void testSolveSymbolInTypeCantResolveNonExistentField() {
...
         try {
-            solvedSymbol.getCorrespondingDeclaration();
+            solvedSymbol.getCorrespondingDeclaration().get();
         } catch (Exception e) {
-            assertTrue(e instanceof UnsupportedOperationException);
-            assertEquals("CorrespondingDeclaration not available for unsolved symbol.", e.getMessage());
+            assertTrue(e instanceof NoSuchElementException);
+            assertEquals("No value present", e.getMessage());
             return;
         }
```
MysterAitch, line 92: This might be unavoidable (I've not gotten to that bit yet) but on the face of it this error message seems like it is (potentially) more vague -- stating that there's no value present, as opposed to directing the reader to a likely cause of the symbol being unsolved.

Can we make this more descriptive - either within this PR or as a follow up?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnonymousClassDeclaration.java
```diff
@@ -80,7 +72,8 @@ public JavaParserAnonymousClassDeclaration(ObjectCreationExpr wrappedNode,
         superTypeDeclaration =
                 JavaParserFactory.getContext(wrappedNode.getParentNode().get(), typeSolver)
                         .solveType(superTypeName)
-                        .getCorrespondingDeclaration();
+                        .getCorrespondingDeclaration()
+                        .orElseThrow(() -> new UnsolvedSymbolException("Unable to solve anonymous class parent"));
```
MysterAitch, line 76: Descriptive error messages are appreciated :+1: 

At the risk of being thick, when might this be thrown? 

My assumptions: 

1. Get the class or interface type of the anonymous class
2. Get the supertype (always `Object` or similar) -- should always exist, never a case where there is no super type.

... meaning that it happens when we've not been able to locate/solve the parent?

Depending on answers to the above, are we sure that throwing an exception within the constructor is what we want to do? (probably yes, just asking the question!)

Given that it's not a change in behaviour (if it's not solved, it would `throw new UnsupportedOperationException("CorrespondingDeclaration not available for unsolved symbol.");`), we can always deal with a TODO / a separate PR rather than this PR.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistClassDeclaration.java
```diff
@@ -227,30 +218,24 @@ private String getSuperclassFQN() {
...
+                MethodResolutionLogic.solveMethodInType(
                         interfaceRef.getTypeDeclaration().get(),
                         name,
                         argumentsTypes,
                         staticOnly
-                );
-                if (interfaceMethodRef.isSolved()) {
-                    candidates.add(interfaceMethodRef.getCorrespondingDeclaration());
-                }
+                ).getCorrespondingDeclaration().ifPresent(candidates::add);
```
MysterAitch: minor comment - could we get these method calls to their own lines please? ditto the other `ifPresent` call above

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistEnumDeclaration.java
```diff
@@ -206,10 +205,10 @@ public String getName() {
...
             CtClass superClass = ctClass.getSuperclass();
             if (superClass != null) {
-                SymbolReference<ResolvedMethodDeclaration> ref = new JavassistClassDeclaration(superClass, typeSolver).solveMethod(name, argumentsTypes, staticOnly);
-                if (ref.isSolved()) {
-                    candidates.add(ref.getCorrespondingDeclaration());
-                }
+                new JavassistClassDeclaration(superClass, typeSolver)
+                        .solveMethod(name, argumentsTypes, staticOnly)
+                        .getCorrespondingDeclaration()
+                        .ifPresent(candidates::add);
```
MysterAitch, line 211: worth extracting out a method here for the shared behaviour in the try blocks?

again, just thinking aloud - can be a separate issue/PR to keep this one simple / focussed just on the introduction of optional

MysterAitch: (see also JavassistInterfaceDeclaration.java -- might be more hassle than it's worth though!)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/JavaSymbolSolver.java
```diff
@@ -158,51 +159,51 @@ public void inject(CompilationUnit destination) {
             }
         }
         if (node instanceof MethodCallExpr) {
-            SymbolReference<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver).solve((MethodCallExpr) node);
-            if (result.isSolved()) {
-                if (resultClass.isInstance(result.getCorrespondingDeclaration())) {
-                    return resultClass.cast(result.getCorrespondingDeclaration());
-                }
+            Optional<ResolvedMethodDeclaration> result = JavaParserFacade.get(typeSolver)
```
MysterAitch: given we're touching these lines already, please rename `result` (here and below) to `declaration` (or `correspondingDeclaration` or `decl` or similar)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -85,9 +77,10 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
         while (itName.contains(".")) {
             String typeName = getType(itName);
             String memberName = getMember(itName);
-            SymbolReference<ResolvedTypeDeclaration> type = this.solveType(typeName);
-            if (type.isSolved()) {
-                return new SymbolSolver(typeSolver).solveSymbolInType(type.getCorrespondingDeclaration(), memberName);
+            Optional<? extends ResolvedTypeDeclaration> type = this.solveType(typeName)
```
MysterAitch: note that this `type` should be named `declaration` now, given that it's the (optional) declaration being returned rather than the type :+1: 

there are multiple places this applies

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
```diff
@@ -152,19 +145,19 @@ public CompilationUnitContext(CompilationUnit wrappedNode, TypeSolver typeSolver
...
-                    solveType(name.substring(0, name.indexOf(".")));
-                if (outerMostRef != null && outerMostRef.isSolved() &&
-                    outerMostRef.getCorrespondingDeclaration() instanceof JavaParserClassDeclaration) {
-                    ref = ((JavaParserClassDeclaration) outerMostRef.getCorrespondingDeclaration())
-                        .solveType(name.substring(name.indexOf(".") + 1));
-                } else if (outerMostRef != null && outerMostRef.isSolved() &&
-                    outerMostRef.getCorrespondingDeclaration() instanceof JavaParserInterfaceDeclaration) {
-                    ref = ((JavaParserInterfaceDeclaration) outerMostRef.getCorrespondingDeclaration())
-                        .solveType(name.substring(name.indexOf(".") + 1));
+                Optional<? extends ResolvedTypeDeclaration> outerMostRef = solveType(name.substring(0, name.indexOf(".")))
```
MysterAitch: here this variable is a(n optional) type declaration, rather than reference -- please rename the variable to reflect this

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AnonymousClassDeclarationContext.java
```diff
@@ -46,165 +42,157 @@
...
-                         .solveType(name.substring(internalType.getName().getId().length() + 1)));
-
-    if (recursiveMatch.isPresent()) {
-      return recursiveMatch.get();
-    }
+        // if is interface and candidate method list is empty, we should check the Object Methods
+        if (candidateMethods.isEmpty() && myDeclaration.getSuperTypeDeclaration().isInterface()) {
+            MethodResolutionLogic.solveMethodInType(
+                    new ReflectionClassDeclaration(Object.class, typeSolver), name, argumentsTypes, false
+            ).getCorrespondingDeclaration().ifPresent(candidateMethods::add);
```
MysterAitch: let's get the method calls on separate lines (here and above) :+1:

MysterAitch:
Looks good to me -- moving towards optional and removing nulls is a good thing :+1: 

The bit I'm less sure about is how to do it -- bit by bit (thus a breaking change multiple releases) versus in one "big bang" (never actually gets done). I think this is an "ideal world" concern though when we're living within the real world :D 


The bulk of the line diffs are adding `.get()` or using `flatMap` etc. -- overall looks good to me!

There are some places where variables need to be renamed to reflect their updated values (e.g., several places where a "symbol reference" is now a(n optional) "declaration").

The rest of the comments are minor and stylistic -- no need to block the merge based on them.

MysterAitch:
barring any objections that come up between now and the release, I have just added it to the "next milestone" release (assuming that we can get the comments above sorted) :+1:

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/SymbolSolverWithJavassistClassTest.java
4everTheOne: I agree with you that the previous message was more descriptive than this one. But i don't think it would be possible to change that message, since the `NoSuchElementException` is create inside of `Optional#get()` method with an hard coded string :/

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnonymousClassDeclaration.java
4everTheOne: > meaning that it happens when we've not been able to locate/solve the parent?

Exacly! That is an issue that was reported at https://github.com/javaparser/javaparser/issues/3045 and this change fixed it (https://github.com/javaparser/javaparser/pull/3044#issuecomment-764857331).

Should I move it to another PR?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistEnumDeclaration.java
4everTheOne: Maybe we could make it in another PR to keep this one focused only on implementing Optional. What do you think?

4everTheOne:
Thank you for you time and for your feedback! I have taken it into consideration and reacted according to it. 

Let me know if you have any more suggestions :)

4everTheOne:
An alternative was suggested at PR #3740 that does not affect public API.

Closing this PR

PR #3049, opened 2021-01-25
Fixes #3048 (`JavaParserSymbolDeclaration#localVar` returning old declaration) and adds tests for `JavaParserSymbolDeclaration`

### LINKED ISSUE #3048: JavaParserSymbolDeclaration#localVar returning old declaration
The static method JavaParserSymbolDeclaration#localVar(VariableDeclarator, TypeSolver) is used in different contexts to create a reference to a variable. During testing, noticed that the declaration was using an old class _JavaParserSymbolDeclaration_ that was later replaced by _JavaParserVariableDeclaration_.

## Test

```java
@Test
void createdLocalVariableShouldBeMarkedAsVariable() {
    VariableDeclarator variableDeclarator = parseVariableDeclarationExpr("int x = 0").getVariable(0);
    ResolvedValueDeclaration localVar = JavaParserSymbolDeclaration.localVar(variableDeclarator, typeSolver);

    assertTrue(localVar.isVariable());
}
```
## Observed behavior
When we call isVariable() of the returned _ResolvedValueDeclaration_ we receive a **false**.

## Expected behavior
When a call to isVariable() of the returned _ResolvedValueDeclaration_ we would expect a **true**.

## Possible solution
A fix for this issue could be replacing the return type of the method JavaParserSymbolDeclaration#localVar(VariableDeclarator, TypeSolver) to be _JavaParserVariableDeclaration_ instead of  _JavaParserSymbolDeclaration_.

4everTheOne:
Fixes #3048 (`JavaParserSymbolDeclaration#localVar` returning old declaration) and adds tests for `JavaParserSymbolDeclaration`


4everTheOne:
Added tests for the static methods of _JavaParserSymbolDeclaration_ to make sure they return the expected result.

For example, in the method **JavaParserSymbolDeclaration#field(VariableDeclarartor,TypeSolver)** the tests make sure the resolved type is a field and can be accessed via **ResolvedValueDeclaration#asField**.

More information about the tests can be found in the comment of the test.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserSymbolDeclaration.java
```diff
@@ -99,95 +78,8 @@ public static int getParamPos(Node node) {
...
-    public boolean isPattern() {
-//        return getWrappedNode() instanceof PatternExpr;
-        return false;
-    }
-
-    @Override
-    public boolean isType() {
-        return false;
+    private JavaParserSymbolDeclaration() {
+        // This private constructor is used to hide the public one
```
MysterAitch, line 82: Removing it immediately makes me a little nervous, but I'm not completely opposed to this.

In either case, `@deprecated` annotations and documentation need to be added to the class / constructor so that users of this class can be guided to what they _should_ be using instead of this.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/TryWithResourceContext.java
```diff
@@ -56,7 +56,7 @@ public TryWithResourceContext(TryStmt wrappedNode, TypeSolver typeSolver) {
             if (expr instanceof VariableDeclarationExpr) {
                 for (VariableDeclarator v : ((VariableDeclarationExpr)expr).getVariables()) {
                     if (v.getName().getIdentifier().equals(name)) {
-                        JavaParserSymbolDeclaration decl = JavaParserSymbolDeclaration.localVar(v, typeSolver);
+                        ResolvedValueDeclaration decl = JavaParserSymbolDeclaration.localVar(v, typeSolver);
```
MysterAitch, line 59: I think this should be `JavaParserVariableDeclaration` given that we're within the `javaparsermodel` here, and `localVar` returns a `JavaParserVariableDeclaration`.

MysterAitch: Not that it particularly matters I suppose, given that it's returned immediately on the next line! :grinning:

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserSymbolDeclaration.java
4everTheOne: Do you think this can be a issue? I was thinking it could be safe to remove, since it already had a private constructor and the only way to create it was using the `localVar` method.

MysterAitch: Ahhh.. yes, you are correct - my mistake!

PR #3058, opened 2021-02-02
Removed definition of method to inherit from super

### LINKED ISSUE #3057: isField and isParameter throwing UnsupportedOperationException
When i try to check if a JavaParserTypeVariableDeclaration is a field or a parameter, it throws an unsupported exception.

https://github.com/javaparser/javaparser/blob/e9af3e337100645141534e3a73004f2b45c2f572/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserTypeVariableDeclaration.java#L137-L145

Looking into ResolvedDeclaration by default it returns False. Should JavaParserTypeVariableDeclaration also return false instead of throwing the exception?

https://github.com/javaparser/javaparser/blob/2903fc918dd87be53192aa1065047392f7adf4d1/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedDeclaration.java#L43-L48

https://github.com/javaparser/javaparser/blob/2903fc918dd87be53192aa1065047392f7adf4d1/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedDeclaration.java#L71-L76

4everTheOne:
Fixes #3057.


jlerbsc:
I think your PR is correct and remains in the spirit of JP and particularly of what a TypeParameter represents. But can you add unit tests please. JavaParserClassDeclarationTest junit test case can be a source of inspiration

jlerbsc:

PR #3059, opened 2021-02-02
Improved JavaParserTypeVariableDeclaration tests

4everTheOne:
This PR contributes with tests for JavaParserTypeVariableDeclaration to increase the line coverage.

The tests were design with the objective of being reusable by other Tests that depend on them to make them consistent. For example, if we now want to add unit tests for JavaParserFieldDeclarationTest we can simple implement ResolvedFieldDeclarationTest and be sure that the shared behavior between all ResolvedFieldDeclaration is consistent between them. 

**Note:**
Some tests are disabled due to additional work. 
As example in the test _declarationMostBeAssociatedToAst_ we would expect that all of the ResolvedDeclaration most implement AssociableToAST. This is not the case, so i left the test disabled.
https://github.com/javaparser/javaparser/blob/c6383311608bd392df8e9801c3ebbe322f7f4f9b/javaparser-core-testing/src/test/java/com/github/javaparser/resolution/declarations/ResolvedDeclarationTest.java#L96-L107

Do you think this can help the project?


jlerbsc:
In my opinion dependency between javaparser-symbol-solver-core and javaparser-core-testing is not desirable because each test project has its own scope. Except that all tests that improve the code coverage is welcome.

4everTheOne:
Thanks for you opinion @jlerbsc! I can see your point and it makes sense.
This two commits fixes it.

4everTheOne:
Note that all the tests have passed but an error occurred while executing the upload of the data.

jlerbsc:
The failing check seems to be related to codecov so there is no reason not to merge this PR. Thanks for this contribution.

PR #3060, opened 2021-02-02
Implemented JavaParserTypeVariableDeclaration getAncestors method 

4everTheOne:
In the PR https://github.com/javaparser/javaparser/pull/3059 i have mentioned that some tests were disabled due to an error in the class JavaParserTypeVariableDeclaration. This PR fixes that issue.

The problem was with the method `JavaParserTypeVariableDeclaration#getAncestors(boolean)`  when the method was called was throwing an UnsupportedException instead of returning an empty list.


4everTheOne:
This last commit fixed the other issue also mention. According to the documentation in`AssociableToAST#toAst()` all the Resolved declaration must be associable to a AST.

This test was disabled because JavaParserTypeVariableDeclaration didn't implemented it.

jlerbsc:
I'm not totally sure of myself but does this pr mean that in this generic type declaration `G <T extends B>`, `T` has no ancestor?
@MysterAitch, @matozoid @ftomassetti  have you an opinion on this?

MysterAitch:
> I'm not totally sure of myself but does this pr mean that in this generic type declaration G T has no ancestor?
> @MysterAitch, @matozoid @ftomassetti have you an opinion on this?


You are correct -- this implementation would mean that the type variable has no implicit or explicitly declared superclass.

Whether this is correct or not I do not know, and I have been unable to find any direct guidance on this within the JLS. The closest I could find is section 8.1.4 (quoted below) which states `For a class other than Object with a normal class declaration, the direct superclass type is Object.` -- in which case, I suppose the distinction is between whether we consider a type parameter to be a class or not.

Given this, I lean towards there being no ancestor for type parameters because there is no class declaration available for this object/type parameter.


---


**4.4 Type Variables**
https://docs.oracle.com/javase/specs/jls/se8/html/jls-4.html#jls-4.4

**4.3.2 The class Object**
https://docs.oracle.com/javase/specs/jls/se16/html/jls-4.html#jls-4.3.2

**8.1.4 Superclasses and Subclasses**
https://docs.oracle.com/javase/specs/jls/se16/html/jls-8.html#jls-8.1.4

> The direct superclass type of a class whose declaration lacks an extends clause is as follows:
> 
> - The class Object has no direct superclass type.
> - For a class other than Object with a normal class declaration, the direct superclass type is Object.
> - For an enum class E, the direct superclass type is Enum<E>.
> - For an anonymous class, the direct superclass type is defined in §15.9.5.
> 
> The direct superclass of a class is the class named by its direct superclass type. The direct superclass is important because its implementation is used to derive the implementation of the class being declared.

MysterAitch:
Given this, I am in favour of merging this PR.

My preference would be to ensure that a documentation comment is added, containing a caveat/warning that this implementation is based on our current understanding of the limited guidance available.

```java
    /**
     * Based on our current understanding of the JLS, type parameters are not a class (or enum/record),
     * therefore do not fit the description within 8.1.4 (Superclasses and Subclasses):
     * <blockquote><pre>
     *  The direct superclass type of a class whose declaration lacks an extends clause is as follows:
     *    - The class {@code Object} has no direct superclass type.
     *    - For a class other than {@code Object} with a normal class declaration, the direct superclass type is {@code Object}.
     *    - For an enum class {@code E}, the direct superclass type is {@code Enum<E>}.
     *    - For an anonymous class, the direct superclass type is defined in §15.9.5.
     * </pre></blockquote>
     * Given this, type parameters do not implicitly extend {@code java.lang.Object} (or any other class),
     * therefore this method returns an empty list and the parameter {@code acceptIncompleteList} has no effect.
     */
    @Override
    public List<ResolvedReferenceType> getAncestors(boolean acceptIncompleteList) {
        return Collections.emptyList();
    }
```

MysterAitch:
Hmm, I have just inspected @jlerbsc 's comment in more detail -- turns out some important parts were being interpreted as HTML (thus hidden)!

Given: 
 `G <T extends B>`

In this example I lean towards classifying `B` as a type bound rather than a super type, though this is (currently) a weakly held view and I wholly agree that this is "fuzzy" / unclear.

4everTheOne:
Looking back at this PR, you are totally right! We must consider the LOWER bounds as ancestors.

I'll update it to match the expected behavior.

jlerbsc:
Since the type variable has no implicit or explicitly declared superclass and we have been unable to find any direct guidance on this within the JLS, we can consider that this PR is correct since it does not introduce regression.

PR #3062, opened 2021-02-03
Improved coverage for resolved declarations and fixed inconsistencies between them

4everTheOne:
First of all, sorry for the long PR but this includes a lot of tests.

I'll try to explain what is going on in each commit to make it easier to understand.
#### 1. Implement the base tests for each ResolvedDeclaration;
The base tests are interfaces that can be reused to validate common behavior .
#### 2. Create tests for javaparsermodel declarations and javassistmodel;
This tests combine the previous tests with additional ones to validate the behavior of the class.
#### 3. Implement AssociableToAST on in classes that missed it. 
With this tests some classes did not implement the AssociableToAST as expected. The classes that required change were:
```
com/github/javaparser/symbolsolver/javaparsermodel/declarations/
  ├── JavaParserEnumDeclaration.java
  ├── JavaParserFieldDeclaration.java
  ├── JavaParserParameterDeclaration.java
  ├── JavaParserPatternDeclaration.java
  ├── JavaParserTypeVariableDeclaration.java
  └── JavaParserVariableDeclaration.java
```
#### 4. Improved code to match the expected behavior.
The issue was discussed at #3057 for different files, but same issue was seen here and fixed accordingly.
By default, some of the methods in ResolvedDeclaration have a default return value. An examples of this is the method `ResolvedDeclaration#isField()`. Some old classes instead of using the default behavior (returning false in the previous example) they throw an UnsupportedException. 
#### 5. JavaParserParameterDeclaration cleanup
Removed unnecessary method definitions since they are already defined in the super class. This helps by improve the line coverage and keep the same behavior.
#### 6. JavaParserPatternDeclaration cleanup
The same as last one but for a different file.


4everTheOne:
Note that all tests have passed. The red cross is there due to lack of coverage of previous tests.

jlerbsc:
There are a lot of changes in this making it difficult and time consuming to validate your work. Thanks for this work.

jlerbsc:
It seems to be correct and the failing check is related only to codecov so i think there is no problem to merge this PR. Thank you for this contribution.

Issue #1940, opened 2018-11-27
Implement AssociableToAST in ResolvedFieldDeclaration


ftomassetti:
See #1935

4everTheOne:
Hello @ftomassetti, i think we can say this issue was fixed here https://github.com/javaparser/javaparser/commit/b96e84c26f49380bf984c1febacf4935aa998fb4.

jlerbsc:
Not sure if we take into account this answer but I let @ftomassetti answer your question.

> However the result could be surprising because a field in the AST can correspond to many resolved fields. One resolvedfield would correspond to a VariableDeclarator inside the AST node for a field declaration

4everTheOne:
I can see what you mean. Do you think PR #3063 can help on that?

jlerbsc:
But cannot validate the impact of your PR #3063 because i'm not experienced enough on JP but I feel that there is a modification of the public API which can impact other users/projects. @MysterAitch @matozoid can you give your opinion on this subject and PR #3063 ?

jlerbsc:
This issue appears to have been resolved if not you can reopen it.

PR #3063, opened 2021-02-04
Simplified usage of class AssociableToAST

4everTheOne:
According to the documentation in [AssociableToAST#toAst()](https://www.javadoc.io/doc/com.github.javaparser/javaparser-core/latest/com/github/javaparser/resolution/declarations/AssociableToAST.html) all the Resolved declaration most be asociable to a AST. 

Currently this doesn't happen, making it hard do access a node from a ResolvedDeclaration. To work around this we need to make some check that requires a lot of boilerplate like in the example below.

https://github.com/javaparser/javaparser/blob/1ce7d85c2b1653a192169ea245c14c537a99132b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java#L353-L366

This PR suggest the remove of the type parameter of the class AssociableToAST and use a helper method to convert to the desired type.
To make it more clear lets take a look at the example mentioned above but this time in the new format:
```java
private boolean isAncestorOf(ResolvedTypeDeclaration descendant) {
    return descendant.toAst()
            .filter(node -> wrappedNode.isAncestorOf(node))
            .isPresent();
}
``` 

Other example, previously we would have to do something like this:
```java
ResolvedDeclaration resolvedDeclaration = ...;
if (resolvedDeclaration instanceof ResolvedFieldDeclaration) {
    ResolvedFieldDeclaration resolvedField = (ResolvedFieldDeclaration) resolvedDeclaration;
    Optional<FieldDeclaration> fieldDeclaration = resolvedField.toAst();
}
```
Now we can do it by simple doing:
```java
ResolvedDeclaration resolvedDeclaration = ...;
Optional<FieldDeclaration> fieldDeclaration = resolvedDeclaration.toNode(FieldDeclaration.class);
```

This also allows add room for more complex examples that were not possible with the previous format, like looking for specific declarations. We can do it now by simply:
```java
Optional<VariableDeclarator> variableDeclarator = resolvedDeclaration.toNode(VariableDeclarator.class);
```


4everTheOne:
After the conversation with jlerbsc at https://github.com/javaparser/javaparser/issues/1940#issuecomment-773414969 that demonstrated the impact of this changes, decided to update this PR to reduce this impact. To reduce this impact a new AssociableToAst was created with a new name AssociatedWithAST and behavior exactly as described in the first comment.

Also AssociableToAst has been marked as Deprecated for not complying with the expected documentation.

jlerbsc:
It would be possible to have a clearer PR by deleting all unnecessary commits.

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/AssociableToAST.java
```diff
@@ -52,7 +51,23 @@
      * In these cases getWrappedNode is particularly nice because it returns the right type of AST node,
      * not just a Node.
      */
-    default Optional<N> toAst() {
-        throw new UnsupportedOperationException();
+    Optional<Node> toAst();
```
jlerbsc: Maybe a default implementation that returns Optional.empty() would simplify the code

jlerbsc:
> After the conversation with jlerbsc at [#1940 (comment)](https://github.com/javaparser/javaparser/issues/1940#issuecomment-773414969) that demonstrated the impact of this changes, decided to update this PR to reduce this impact. To reduce this impact a new AssociableToAst was created with a new name AssociatedWithAST and behavior exactly as described in the first comment.
> 
> Also AssociableToAst has been marked as Deprecated for not complying with the expected documentation.

Given that the next version of JP will introduce a break in the API (this will be a major version certainly 4.0.0.), I send back my opinion, so that you can integrate this change directly into the existing interface.

jlerbsc:
Did you complete your PR?

4everTheOne:
Yes I did. Sorry I forgot to leave a comment.

jlerbsc:

Issue #3038, opened 2021-01-16
Update to JavaParser 3.18.0 from 3.17.0 causes MethodCallExpr::resolve to hang


oxisto:
When analysing the file https://github.com/Fraunhofer-AISEC/codyze/blob/59da2a92a5d6623d0066efb4bbae8d860a8485aa/src/test/resources/java/jca/AESCTR.java we encounter an endless loop in our analysis program. 

The root cause of this seems to be the call to `resolve`  of the expression `System.out.printf("update  %2d: %32s [%s]\n", i, Utils.bytesToHex(ciphertextblock), Utils.bytesToHex(mac));`. Once we call resolve, the program seems to be stuck in an endless-loop, however, it does seem to complete the call to resolve after > 10 mins. 

Running the program with JavaParser 3.17.0 works fine and returns in < 1s. Only analysing the expression on its own also seems to work, but analysing the function as described in the file seems to cause the loop.

To construct JavaParser we use a `CombinedTypeSolver` using a `ReflectionTypeSolver` as well as `JavaParserTypeSolver` based on the root of the file.

jlerbsc:
Hi @oxisto Could you provide a simple test case that reproduces this issue?

Col-E:
Can confirm this occurs, issue seems to be based on some nearly infinite looping [`StementContext.java` around lines 242-255](https://github.com/javaparser/javaparser/blob/2903fc918dd87be53192aa1065047392f7adf4d1/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/StatementContext.java#L242) due to changes introduced in #2654

The code I'm looking at is structured like the following: 
```java
void main() {
    // lots of variable assignments
    // Just do:
    // String s1 = ...
    // String s2 = ...
    // ...
    // String s100 = ...
    Thread t = new Thread(() -> {
       Foo f = Foo.getInstance(); // Resolve the "Foo" as a NameExpression and it will hang
    }.start();
}
```
Attempting to resolve the `NameExpr` (`Foo`) in `Foo.getInstance()` causes the hang. It will attempt to look for instances of `Foo` being declared in a parent context to resolve, then its parent's parent, etc. But this content does not exist, so it does not get resolved. This cycle happens once every iteration in the `while` loop segment linked above. 

The key is that there are lots of assignments in the parent context.

[...]

PR #3072, opened 2021-02-09
Fix Issue #3071 - Hanging when certain names are resolved

### LINKED ISSUE #3071: Update to JavaParser 3.18.0 from 3.17.0 causes NameExpr::resolve to hang
Possibly a duplicate of #3038 as I have only encountered this when explicitly resolving `NameExpr`

Attempting to resolve the `NameExpr` of `Foo` in the following code will take over an hour to resolve _(I stopped it after 75 minutes)_  if the set symbol solver has no way to access `Foo`

```java
public class Foo{
        String s0   = "hello";
        String s1   = "hello";
        String s2   = "hello";
        String s3   = "hello";
        String s4   = "hello";
        String s5   = "hello";
        String s6   = "hello";
        String s7   = "hello";
        String s8   = "hello";
        String s9   = "hello";
        String s10  = "hello";
        String s11  = "hello";
        String s12  = "hello";
        String s13  = "hello";
        String s14  = "hello";
        String s15  = "hello";
        String s16  = "hello";
        String s17  = "hello";
        String s18  = "hello";
        String s19  = "hello";
        String s20  = "hello";
        String s21  = "hello";
        String s22  = "hello";
        String s23  = "hello";
        String s24  = "hello";
        String s25  = "hello";
        String s26  = "hello";
        String s27  = "hello";
        String s28  = "hello";
        String s29  = "hello";
        String s30  = "hello";
        String s31  = "hello";
        String s32  = "hello";
        String s33  = "hello";
        String s34  = "hello";
        String s35  = "hello";
        String s36  = "hello";
        String s37  = "hello";
        String s38  = "hello";
        String s39  = "hello";
        String s40  = "hello";
        String s41  = "hello";
        String s42  = "hello";
        String s43  = "hello";
        String s44  = "hello";
        String s45  = "hello";
        String s46  = "hello";
        String s47  = "hello";
        String s48  = "hello";
        String s49  = "hello";
        String s50  = "hello";
        String s51  = "hello";
        String s52  = "hello";
        String s53  = "hello";
        String s54  = "hello";
        String s55  = "hello";
        String s56  = "hello";
        String s57  = "hello";
        String s58  = "hello";
        String s59  = "hello";
        String s60  = "hello";
        String s61  = "hello";
        String s62  = "hello";
        String s63  = "hello";
        String s64  = "hello";
        String s65  = "hello";
        String s66  = "hello";
        String s67  = "hello";
        String s68  = "hello";
        String s69  = "hello";
        String s70  = "hello";
        String s71  = "hello";
        String s72  = "hello";
        String s73  = "hello";
        String s74  = "hello";
        String s75  = "hello";
        String s76  = "hello";
        String s77  = "hello";
        String s78  = "hello";
        String s79  = "hello";
        String s80  = "hello";
        String s81  = "hello";
        String s82  = "hello";
        String s83  = "hello";
        String s84  = "hello";
        String s85  = "hello";
        String s86  = "hello";
        String s87  = "hello";
        String s88  = "hello";
        String s89  = "hello";
        String s90  = "hello";
        String s91  = "hello";
        String s92  = "hello";
        String s93  = "hello";
        String s94  = "hello";
        String s95  = "hello";
        String s96  = "hello";
        String s97  = "hello";
        String s98  = "hello";
        String s99  = "hello";
        String s100 = "hello";
        new Thread(){
            @Override
            public void run() {
                Foo foo = Foo.getInstance(); // Resolve either Foo here
            }
        }.run();
    }
    static Foo getInstance() {
        return new Foo();
    }
}
```

This is due to what I believe is an oversight in `StatementContext#solveSymbol` where resolving the name checks each prior adjacent statement for a declaration of that name... but each time an adjacent prior statement is checked it tries to look in all of its prior adjacent statements. So if we have 5 adjacent statmeents, we dont visit just the 5. We visit the 5 then the 4 then the 3 then the 2 then the last 1.  This will cause an exponential amount of calls to `solveSymbol`.

Col-E:
Fixes #3071

### How things used to work

1. Attempt to resolve a name in an expression.
2. In `StatementContext` we call `solveSymbol(name)` and the parent node is of type `NodeWithStatements`
3. We get an iterator of previous adjacent statements to attempt to resolve the name in that context
4. We call `solveSymbol` on one of those adjacent statements, and it delegates to `StatementContext#solveSymbol`
5. This then looks at all of the prior adjacent statements before it too...
    - Goto 2 until no more items exist in the iterator
6. We call `solveSymbol` on the parent node

This is horrible for performance, with `o(n!)`

### How this changes that

1. Attempt to resolve a name in an expression.
2. In `StatementContext` we call `solveSymbol(name)` and the parent node is of type `NodeWithStatements`
3. We get an iterator of previous adjacent statements to attempt to resolve the name in that context
4. We call `solveSymbol` on one of those adjacent statements, and it delegates to `StatementContext#solveSymbol`
    - Except these calls prevent the same adjacent statement context lookup 
5. We call `solveSymbol` on the parent node

This basically prevents the adjacent nodes from looking at their adjacent nodes which reduces a massive amount of duplicated calls. This makes the code in the test class go from taking hours to finish to under a second. 

### Scope affected

This patch only applies when the adjacent node is also a `NodeWithStatements`. Running the `javaparser-symbol-solver-testing` module cases yields a 100% pass rate when I ran it locally

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/StatementContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/StatementContext.java
@@ -195,7 +195,7 @@ public class StatementContext<N extends Statement> extends AbstractJavaParserCon
     /**
      * Used where a symbol is being used (e.g. solving {@code x} when used as an argument {@code doubleThis(x)}, or calculation {@code return x * 2;}).
      * @param name the variable / reference / identifier used.
-     * @param iterateAdjacentStmts flag to iterate adjacent statements, default to {@code true} except in cases to prevent repetitive checks.
+     * @param iterateAdjacentStmts flag to iterate adjacent statements, should be set to {@code true} except when calling itself in order to prevent revisiting already visited symbols.
      * @return // FIXME: Better documentation on how this is different to solveSymbolAsValue()
      */
     private SymbolReference<? extends ResolvedValueDeclaration> solveSymbol(String name, boolean iterateAdjacentStmts) {
@@ -258,8 +258,21 @@ public class StatementContext<N extends Statement> extends AbstractJavaParserCon
             while(statementListIterator.hasPrevious()) {
                 Context prevContext = JavaParserFactory.getContext(statementListIterator.previous(), typeSolver);
                 if (prevContext instanceof StatementContext) {
-                    // Do not schedule each adjacent statement to do this exact same thing.
-                    // Since we are doing it here we will eventually cover all adjacent statements anyways.
+                    // We have an explicit check for "StatementContext" to prevent a factorial increase of visited statements.
+                    //
+                    // For example consider the following:
+                    //   String a = "a";
+                    //   String b = "b";
+                    //   String c = get();
+                    //
+                    // If we simply call "prevContext.solveSymbol(name)" we will call the current method with the adjacent statement "prevContext".
+                    // Then "prevContext" will look at its previous adjacent statement. And so on and so forth.
+                    // When there are no more previous statements in this chain of method calls, we come back to here...
+                    // Then we look at the next "prevContext" which causes the entire process to start again.
+                    // This is how we get a factorial increase in calls to "solveSymbol".
+                    //
+                    // So what we do instead with this check is we pass in a flag to say "Do not look at previous adjacent statements".
+                    // Since each visited "prevContext" does not look at its adjacent statements we only visit each statement once in this while loop.
                     symbolReference = ((StatementContext<?>)prevContext).solveSymbol(name, false);
                 } else {
                     symbolReference = prevContext.solveSymbol(name);
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/StatementContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/StatementContext.java
@@ -189,6 +189,16 @@ public class StatementContext<N extends Statement> extends AbstractJavaParserCon
 
     @Override
     public SymbolReference<? extends ResolvedValueDeclaration> solveSymbol(String name) {
+        return solveSymbol(name, true);
+    }
+
+    /**
+     * Used where a symbol is being used (e.g. solving {@code x} when used as an argument {@code doubleThis(x)}, or calculation {@code return x * 2;}).
+     * @param name the variable / reference / identifier used.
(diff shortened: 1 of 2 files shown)


Col-E:
If the main explaination was unclear Ive updated it to include more reasoning and explanation of the issue fixed, while how it retains the same behavior. 

If I made any mistakes please point them out to me so I can update the PR.

Col-E:
I updated the comment in the changed method to include a breakdown of what the issue is, and how the new boolean argument resolves the issue. 

Now I'm also aware that this can likely be improved further. If each checked adjacent statement tries to solve in the parent context, are we not getting N duplicate calls to resolve in that parent context? If we want to fix that here and now I can do that too. Its a one-line fix and all test cases still pass.

I'll wait on your feedback on whether or not to include this as well.

image

oxisto:
I can confirm, that this PR also solves the hanging around method calls in https://github.com/javaparser/javaparser/issues/3038

Col-E:
Ah, I see this was merged. I can open a follow-up PR regarding the second optimization listed above from 13 days ago too. Pretty simple one-line change.

Col-E:
Oh and the change-log for this PR lists the wrong author. @MysterAitch 

> Issue #3038 and Issue #3071 - Hanging when certain names are resolved (PR #3072, by @mernst)

@mernst did the JavaDoc fixes which is the following line in the change-log.

MysterAitch:
> Oh and the change-log for this PR lists the wrong author. @MysterAitch 
> 
> > Issue #3038 and Issue #3071 - Hanging when certain names are resolved (PR #3072, by @mernst)
> 
> 

Oops, sorry!! That's my bad with the copy pasting!

Fancy putting in a PR to fix it? :+1:

Col-E:
> Fancy putting in a PR to fix it? 👍

#3155 Done

MysterAitch:
> Ah, I see this was merged. I can open a follow-up PR regarding the second optimization listed above from 13 days ago too. Pretty simple one-line change.

For this I have no strong opinions at the moment.

I recall a method in the contexts similar to "variables exposed by my children" but suspect this might have been a draft by me or I may be misremembering some code I wrote for pattern expressions.

My gut reaction is that a statement should iterate over previous statements querying if that statement is able to resolve the identifier "inside" it, and then escalate to the parent context (as opposed to "asking the previous statement to solve it"). As above, this might not be possible with the currently available methods though.

Regardless, hopefully this release has removed blocks that were in your way, and future PRs can further  polish/optimise how it does it :+1:

oxisto:
A huge thanks to everyone. This was a major blocker for us at https://github.com/Fraunhofer-AISEC/cpg. I want to do some more testing but it looks like we can once again update JP to the latest releases.

BTW, my original issue #3038 should be able to be closed as well now - or I can probably just do that myself.

Issue #3078, opened 2021-02-10
Question: Uniquely identify a node


4everTheOne:
Imagine the following situation:
```java
class A {
  void method1(int param) {}
  void method2(int param) {}
}
```

Given this example i would like to check if the parameters are the same, but when using the `equals` or` hashCode` they are equals because have identical children. 

Is there an alternative way to check if they are the same, based on the parent path for example or in a unique identifier?

4everTheOne:
There's an alternative way of doing it, by using == instead of equals

PR #3079, opened 2021-02-11
Implemented Node ID

4everTheOne:
Sometimes may be harder to identify if two nodes are the same instance. Imagine the example below:
```java
class A {
  void method1(int param) {}
  void method2(int param) {}
}
```
Given the example above, we would like to check that the parameters are the same.
Using the equals or hash code method, they are considered to be equal (which is true because they have the same name, type and child nodes) but are not the same instance.

This PR adds a new constant property called unique id that is generated at the creation of the node and cannot be changed.

Related to (#3078).


4everTheOne:
There's an alternative way of doing it, by using == instead of equals

Issue #2660, opened 2020-05-10
Remove usages of StaticJavaParser from test cases


MysterAitch:
There seem to be a few different ways of evaluating tests at the moment.

Some variations include using the `StaticJavaParser` directly, some use a custom `TestJavaParser` or similar, others have custom helper methods, and others use `new JavaParser()` directly.

This tripped me up recently with the `instanceof` test which call the statically-imported `parse()` method -- the only way to set it to use JAVA_14 would have been to also do this (presumably) for every other test which uses the static javaparser. 

My thoughts are that ideally we shouldn't use the static javaparser at all within tests, and, long term, it would be ideal to be able to (reliably) run the test suite cases in parallel. Use of static makes it difficult to run tests in parallel, and can introduce tricky scenarios when trying to set configurations etc. and having unexpected/unpredictable side effects depending on the execution order.

This might be a little naive, and other uses of static in the codebase may make it difficult to ensure any kind of thread-safety though. Thoughts/experiences on this are welcomed!

PR #3081, opened 2021-02-12
Added a wrapper for JavaParser

4everTheOne:
Currently there are two alternatives to work with JavaParser. The first one is to work with the JavaParser itself or to work with StaticJavaParser.

Working directly with JavaParser itself, requires a lot of boilerplate making it hard to work with and understand.
On the other hand, StaticJavaParser makes this process very simple, by introducing an API that makes it more user friendly. But unfortunately StaticJavaParser is not perfect, since it works in a static context and if a user wants to parse different projects with different configuration, he is not able to.

This PR adds a new API por JavaParser called `SimpleJavaParser` that works in a non-static way and follow the sames principles of StaticJavaParser but in this the configuration.

#### Examples:

Parse a compilation unit without config:
```java
String sourceCode = "...";
SimpleJavaParser simpleJavaParser = new SimpleJavaParser();
CompilationUnit compilationUnit = simpleJavaParser.parse(sourceCode);
```

Parse a compilation unit with config:
```java
String sourceCode = "...";

ParserConfiguration parserConfiguration = new ParserConfiguration();
parserConfiguration.setTabSize(8);

SimpleJavaParser simpleJavaParser = new SimpleJavaParser(parserConfiguration);
CompilationUnit compilationUnit = simpleJavaParser.parse(sourceCode);
```

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/SimpleJavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/SimpleJavaParser.java
@@ -1,0 +1,484 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2021 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.ImportDeclaration;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.PackageDeclaration;
+import com.github.javaparser.ast.body.BodyDeclaration;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.Parameter;
+import com.github.javaparser.ast.body.TypeDeclaration;
+import com.github.javaparser.ast.expr.*;
+import com.github.javaparser.ast.modules.ModuleDeclaration;
+import com.github.javaparser.ast.modules.ModuleDirective;
+import com.github.javaparser.ast.stmt.BlockStmt;
+import com.github.javaparser.ast.stmt.ExplicitConstructorInvocationStmt;
+import com.github.javaparser.ast.stmt.Statement;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.Type;
+import com.github.javaparser.ast.type.TypeParameter;
+import com.github.javaparser.javadoc.Javadoc;
+
+import java.io.*;
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * A simpler API than {@link JavaParser}.
+ */
+public final class SimpleJavaParser {
+
+	private final JavaParser parser;
+
+    /**
+     * Create a new instance of {@link SimpleJavaParser} with the default configuration.
+     */
+	public SimpleJavaParser() {
+        this( new JavaParser() );
+	}
+
+    /**
+     * Create a new instance of {@link SimpleJavaParser} with a custom configuration.
+     *
+     * @param configuration The configuration to be used in the parser.
+     */
+	public SimpleJavaParser(ParserConfiguration configuration) {
+        this( new JavaParser(configuration) );
+	}
+
+	/**
+	 * Create a new instance of {@link SimpleJavaParser} with the provided {@link JavaParser}.
+	 * <br>
+	 * This constructor is used for testing purposes.
+	 *
+	 * @param parser The parser to be used.
+	 */
+	SimpleJavaParser(JavaParser parser) {
+		this.parser = parser;
+	}
+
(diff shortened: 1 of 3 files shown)


jlerbsc:
I'm not agree with you. A new API introduces more confusion and more maintenance. I do not see what significant improvement brings this new API. Actually we can do the same thing like this

**Parse a compilation unit without config:**

    String sourceCode = "...";
    ParseResult<CompilationUnit> parseResult = javaParser.parse(sourceCode);

The ParseResult class give you the possibility to manage parsing errors. If a project doesn't want to collaborate with this intermediate class, project members can decorate JavaParser.

4everTheOne:
Thanks for your opinion! I will try to show you the improvements we can benefit from this.

In my opinion, the problem with the current JavaParser is the need of boilerplate code to access the parsed content. For example, using the example you mentioned, we would need something like just to get access to the compilation unit:

```java
String sourceCode = "...";
ParseResult<CompilationUnit> parseResult = javaParser.parse(sourceCode);
if (parserResult.isSuccessful()) {
  CompilationUnit compilationUnit = parserResult.getResult().get();
  ....
} else {
  List<Problem> problems = parserResult.getProblems();
  ...
}
```

Note in the previous example on line `CompilationUnit compilationUnit = parserResult.getResult().get();` it very simplified. When this line is analysed by quality assurance tools like sonarlint it is always marked with warnings that we should check if the optional value is present (although it is verified in the function `#isSuccessful()`). To fix this we would need to add another check for the optional which adds more complexity to simply get the compilation unit.

Now in this new API, the same task with the option to handle problems, can be executed by simple doing:
```java
String sourceCode = "...";
SimpleJavaParser simpleJavaParser = new SimpleJavaParser();
try {
  CompilationUnit compilationUnit = simpleJavaParser.parse(sourceCode);
  ...
} catch(ParseProblemException e) {
  List<Problems> problems = e.getProblems();
  ...
}
```

I see this also as an internal issue. Looks like a lot of people prefer to use StaticJavaParser instead of JavaParser (#2660). In my opinion is because it's easier to use and we can see that in tests, where is a need to create a dedicated class to work with JavaParser called `TestParser`.

I understand your concern about confusion and maintenance, but I see it as an improvement to the current JavaParser API.

Did this explanation helped identifying the issue i was mentioning?

4everTheOne:
I forgot to mention, but MysterAitch says in the issue mentioned above:

> My thoughts are that ideally we shouldn't use the static javaparser at all within tests, and, long term, it would be ideal to be able to (reliably) run the test suite cases in parallel. Use of static makes it difficult to run tests in parallel, and can introduce tricky scenarios when trying to set configurations etc. and having unexpected/unpredictable side effects depending on the execution order.

This PR also helps with a step close to this objective by providing a simple easy to use API that can also be used in tests that are run in parallel.

jlerbsc:
I let the other members of the team share their opinions.

jlerbsc:
Oups My bad!

javaparser-core/src/main/java/com/github/javaparser/SimpleJavaParser.java
```diff
@@ -0,0 +1,483 @@
...
+import java.io.*;
+import java.nio.file.Path;
+import java.util.Optional;
+
+/**
+ * A simpler API for {@link JavaParser}.
+ */
+public final class SimpleJavaParser {
+
+	private final JavaParser parser;
```
MysterAitch, line 52: convert to spaces please :)

MysterAitch:
> In my opinion, the problem with the current JavaParser is the need of boilerplate code to access the parsed content. 

I do broadly agree. As a newbie it can be difficult to make the mental leaps around "I want a method that returns a `CompilationUnit`, but this method returns a `ParseResult` -- how do I get from a `ParseResult` to a `CompilationUnit`??" and then spotting "Oh, here's `StaticJavaParser` that can do it!".

That said, the "standard" way is about the same size as the proposed "simple" way -- so perhaps this is a documentation issue?

My _personal_ view is that the proposed `SimpleJavaParser` is perhaps what `StaticJavaParser` should be (with the caveat of not having done a deep side-by-side comparison). But this is perhaps my personal dislike / bias against `static` showing :wink:.

MysterAitch:
Here is an example of accessing the compilation unit where we don't care about the parse errors:
```java
    String sourceCode = "...";

    ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setTabSize(8);

    JavaParser javaParser = new JavaParser(parserConfiguration);
    javaParser.parse(sourceCode).ifSuccessful((CompilationUnit compilationUnit) -> {
        // do stuff with compilation unit
    });
```

Where we do want the parse errors, this new API is just swapping `isSuccessful` for `try`, and the `else` block with `catch`.


Compared with:

```java
    String sourceCode = "...";

    ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setTabSize(8);

    SimpleJavaParser simpleJavaParser = new SimpleJavaParser(parserConfiguration);
    CompilationUnit compilationUnit = simpleJavaParser.parse(sourceCode);
    // do stuff with compilation unit
```

---

Alternatively, if wanting to deal with the parse problems:

```java
    String sourceCode = "...";

    ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setTabSize(8);

    ParseResult<CompilationUnit> parseResult = new JavaParser(parserConfiguration).parse(sourceCode);
    if(parseResult.isSuccessful()) {
        CompilationUnit compilationUnit = parseResult.getResult().get();
        // do stuff with compilation unit
    } else {
        List<Problem> problems = parseResult.getProblems();
        // Handle parse problems
    }
```

Compared with:

```java
    String sourceCode = "...";

    ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setTabSize(8);

    SimpleJavaParser simpleJavaParser = new SimpleJavaParser(parserConfiguration);
    try {
        CompilationUnit compilationUnit = simpleJavaParser.parse(sourceCode);
        // do stuff with compilation unit
    } catch(ParseProblemException e) {
        List<Problems> problems = e.getProblems();
        // Handle parse problems
    }
```

MysterAitch:
Summary: As above, while I do agree that it could be simplified and that there is boilerplate, I m leaning away from merging this PR as-is (in favour of an alternative).

Perhaps we could add methods to `JavaParser` instead?

Let's open an issue to discuss this -- we can always reopen this PR later after further discussion :+1:

PR #3082, opened 2021-02-16
Javadoc fixes

mernst:
This pull request fixes syntax errors in Javadoc comments.


javaparser-core/src/main/java/com/github/javaparser/ast/expr/MethodCallExpr.java
```diff
@@ -49,7 +49,7 @@
 import com.github.javaparser.resolution.types.ResolvedType;
 
 /**
- * A method call on an object or a class. <br>{@code circle.circumference()} <br>In <code>a.&lt;String&gt;bb(15);</code> a
+ * A method call on an object or a class. <br>{@code circle.circumference()} <br>In {@code a.&lt;String&gt;bb(15);}, a
```
MysterAitch: Using the `@code` tag, this could become:

```
{@code a.<String>bb(15);}
```

javaparser-core/src/main/java/com/github/javaparser/ast/type/UnionType.java
```diff
@@ -53,7 +53,7 @@
  * } catch(<b>IOException | NullPointerException ex</b>) {
  * ...
  * }
- * </pre>
+ * </code></pre>
```
MysterAitch, line 56: Double check this -- the closing `</code>` tag has been removed, but the opening one is still there.

mernst: This is the addition of a missing `</code>` tag, not the removal of one.  The `<code>` tag is already there.

MysterAitch: Oops, yes you're right - my bad! 

Would you mind converting it to `{@code }` instead please?

MysterAitch: Scratch that, let's get it merged in and we can do additional tweaks like that in another PR

PR #3114, opened 2021-02-25
[Suggestion] Added unpack methods for JavaParser

### LINKED ISSUE #3108: Simplify the use of JavaParser
Currently working with JavaParser requires some unnecessary boilerplate (See conversation at https://github.com/javaparser/javaparser/pull/3081).

The intention of the issue is to allow a conversation with possible solutions for this problem.

4everTheOne:
Fixes #3108.

Based on the discussion at #3108, there was an agreement that the use of `JavaParser` could be simplified.

This PR proposes new methods that try to simplify the unpacking of `ParseResult`. In a successfull situation the result is returned, otherwise throws an `ParseProblemException` that exposes the problems encountered.


jlerbsc:
Hi @4everTheOne Thank you for this contribution. It seems to me that this PR contains many changes that are not dependent on each other. It is difficult to validate in this state. Maybe you could break it down into different PRs, each containing dependent changes. For exemple
- copywright
- upgrade version
- generator
- maven updates
- unit tests
- ...

4everTheOne:
Hi @jlerbsc, thanks for your feedback.
In this PR i only contribuite with changes in two files (Done in commit with hash e76a2c7):
- JavaParserTest.java
- JavaParser.java

Do you think that the tests should be moved in another PR?

jlerbsc:
I don't knwo why i see more difference. Anyway since the changes are related it is better to keep them in the same commit. I let @MysterAitch validate this PR if he agrees with its content. Thank you.

jlerbsc:
Hi @4everTheOne I'm not a fan of this solution because it multiplies the number of methods in the JavaParser class. Wouldn't it be possible to instead create a new implementation that relies on, or encapsulates, the current parser.
Possibly would it be possible to extract an interface from the parser where the parsing result could be either a ParseResult<? extends Node> or a Node?

4everTheOne:
Looking back at it, i totally agree with you @jlerbsc.
This can be done in a more simple, and cleaner ways! :)

What do you think of this approach? https://github.com/javaparser/javaparser/pull/3747

Thanks for the feedback!

PR #3124, opened 2021-03-03
Improve generated code

### LINKED ISSUE #2708: Improve generated code
There are some improvements that can be made to the generated code:

- Missing `@Override` annotations (not sure offhand where we might check to see if a generated method is overriding something)

- Unnecessary casts within the `PropertyGenerator` (easy enough to add in a check to see if the type being returned matches the current class name)

Maybe some other bits

4everTheOne:
This PR fixes #2708 by removing unnecessary casts and adding `Override` annotation when a generated method overrides another method.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
@@ -70,8 +70,9 @@ public abstract class NodeGenerator extends Generator {
         boolean isOverriding = Arrays.stream(superClass.getMethods())
                 .filter(m -> m.getName().equals(methodDeclaration.getNameAsString()))
                 .anyMatch(m -> m.getParameters().length == methodDeclaration.getParameters().size());
-        if (isOverriding)
+        if (isOverriding) {
             annotateOverridden(methodDeclaration);
+        }
     }
 
     protected void after() throws Exception {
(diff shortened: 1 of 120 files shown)


4everTheOne:
I have run the script `run_core_generators.sh` locally but didn't include in the PR. 
Should I add a commit with the generated content?

MysterAitch:
> I have run the script `run_core_generators.sh` locally but didn't include in the PR. 
> Should I add a commit with the generated content?

Yes, I tend to have a commit for just the change to the generator(s) then a commit after running the generator.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/NodeGenerator.java
```diff
@@ -53,6 +57,23 @@ public final void generate() throws Exception {
...
+    protected void annotateWhenOverridden(BaseNodeMetaModel nodeMetaModel, MethodDeclaration methodDeclaration) {
+        Class<? extends Node> type = nodeMetaModel.getType();
+        Class<?> superClass = type.getSuperclass();
+
+        boolean isOverriding = Arrays.stream(superClass.getMethods())
+                .filter(m -> m.getName().equals(methodDeclaration.getNameAsString()))
+                .anyMatch(m -> m.getParameters().length == methodDeclaration.getParameters().size());
+        if (isOverriding)
+            annotateOverridden(methodDeclaration);
+    }
```
MysterAitch: please do include braces (I mention it here, but it applies to the other PRs too :) )

MysterAitch:
Looks good to me -- will merge once CI checks are complete :+1:

4everTheOne:
Thanks you for your feedback! :)

MysterAitch:
Thank you for the hard work! :+1:

PR #3126, opened 2021-03-04
Optimization to avoid systematically creating a class from javassist when the class has already been created

jlerbsc:


MysterAitch:
Thank you @jlerbsc for your ongoing support in making JavaParser increasingly awesome! ☺️🤗

Issue #2707, opened 2020-06-01
Idea - null / not null annotation


MysterAitch:
As an idea, we could have our own `@Nullable` and `@NotNull` annotations on method parameters.

We could then use a generator to ensure that the first lines of the method include a null check (e.g. `assertNotNull()` or `Objects.requireNonNull()` or similar).

Maybe this could be extended out to fields?

matozoid:
Here's the history, check the oldest issues: https://github.com/javaparser/javaparser/issues?q=is%3Aissue+nullable

PR #3127, opened 2021-03-04
[Suggestion] NonNull generator for parameters

4everTheOne:
One important factor to keep the quality of the software is to make sure is handles unintended behavior. @MysterAitch suggested at #2707 to add `@Nullable` and `@NotNull` to indicate if a given variable can take null values.

In this PR is suggested a way to improve the quality by providing a generator that creates the assertions required for the annotated variables. Lets take a look at an example.

```java
class Foo {
   public void doOperation(String text, String optionalInformation) {
      ...
   }
}
```
To make sure the example above can only take non-null values we simply add a `@NotNull` annotation to the parameter:
```java
class Foo {
   public void doOperation(@NotNull String text, @Nullable String optionalInformation) {
      ...
   }
}
```
By doing this, when the Generator is run, the preconditions will be added automatically making it looks like this:
```java
class Foo {
   public void doOperation(@NotNull String text, @Nullable String optionalInformation) {
      Preconditions.checkNotNull(text, "Parameter text can't be null.");
      ...
   }
}
```

#### How the generator works
At the moment the generator only support method and constructor parameters but can be expanded for other types like variable declarations or even field declaration.

The generator get all methods and constructors in the class, and check if the parameter is annotated with `@NotNull` if it's, then it will generate a call at the top of the method to `Preconditions.checkNotNull` to make sure no null value is present. The logic for the constructor follows the same guidelines as a method, but with a tiny difference. If the the first statement in the constructor is a `ExplicitConstructorInvocationStmt` then it adds the assertion after to avoid compilation issues. 

I would like to hear your opinions! :)


MysterAitch:
Just a quick comment -- how does this behave when running the generators multiple times? e.g., are existing/redundant null checks removed?

4everTheOne:
> Just a quick comment -- how does this behave when running the generators multiple times? e.g., are existing/redundant null checks removed?

At the moment, it doesn't repeat if matches exactly, for example if the assertion is already present. But there's is room for improving. At the moment if we change the error message it will add a new one instead of replacing it.

4everTheOne:
All polished and ready for review! :)

jlerbsc (commented):
comment.getRange().isPresent() can be replaced by comment.hasRange()

4everTheOne:
Thanks for the feedback jlerbsc! I think that could be a improvement for the visitor generators, to use the corresponding "has" method instead of using isPresent() 👍

MysterAitch:
> comment.getRange().isPresent() can be replaced by comment.hasRange()
> 

Those ones are generated iirc

jlerbsc:
Hi @4everTheOne You have a lot of interesting PR waiting with conflicts. Do you have time to resolve these conflicts especially this one so I can merge it? Thanks.

4everTheOne:
Hi @jlerbsc!
Thanks for the feedback! I'll try to sort it out.
I'll ping you when its done ;)

jlerbsc:
I've spent some time looking again at your previous PRs which are generally very good in intent. Often the problem is that the impact is far too great for me to validate your work. It would certainly be necessary to proceed in small steps, trying to impact public apis as little as possible. Thank you for your investment.

4everTheOne:
@jlerbsc, this PR has been updated :)

jlerbsc:
Do you think that this evolution can have an impact on the functioning of applications that use JavaParser? I am asking this question to name the next version of JavaParser.

4everTheOne:
I don't see any impact on it. The most sensitive class here is the _StaticJavaParser_, which is the class that contains the annotation NonNull. But annotation were only added when it makes sense for the parameter to not be null.

But let me know, if you see any impact at _StaticJavaParser_ :)

jlerbsc:
I was thinking a difference at runtime.

jlerbsc:

PR #3141, opened 2021-03-05
Implemented getMethodDeclaration for Annotation declaration

### LINKED ISSUE #1838: Improving annotation support: Implement ResolvedAnnotationDeclaration#getDeclaredMethods()
Currently, all three classes `JavaParserAnnotationDeclaration`, `ReflectionAnnotationDeclaration`, and   `JavassistAnnotationDeclaration` implement the method `getDeclaredMethods()` by throwing an `UnsupportedOperationException`. This method should be implemented properly in all three classes.

My suggestion:

Annotations can have methods, which are then called *elements*. See https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.6.1:

> The body of an annotation type may contain method declarations, each of which defines an element of the annotation type.

Therefore, `getDeclaredMethods()` should return the set of methods (called *elements*) defined by this annotation.

4everTheOne:
Fixes #1838.

According to [JLS](https://docs.oracle.com/javase/specs/jls/se7/html/jls-9.html#jls-9.6
) and malteskoruppa in #1838 the annotation declaration doesn't have methods. 

Since it's an intended behavior i suggest returning a empty set instead of throwing the UnsupportedException.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnnotationDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserAnnotationDeclaration.java
@@ -63,8 +63,9 @@ public class JavaParserAnnotationDeclaration extends AbstractTypeDeclaration imp
 
     @Override
     public Set<ResolvedMethodDeclaration> getDeclaredMethods() {
-        // TODO #1838
-        throw new UnsupportedOperationException();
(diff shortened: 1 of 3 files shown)


jlerbsc:
Returning an empty list and throwing an exception this does not have the same meaning. I'm not sure that returning an empty list is the bets choice.

MysterAitch:
I can't see it in searches where I've commented this before but I think this comes to needing to differentiate between "not _yet_ implemented" versus "not compatible with the java language specification" -- both cases currently are "unsupported".

re: #1838 and this PR in particular, I don't currently have an opinion.

Issue #2762, opened 2020-07-19
JavassistClassDeclaration isAssegnableBy throws NotFoundException when a class/interface is defined in another Jar


corradomio:
There is a "problem" with the current implementation of

JavassistClassDeclaration::isAssegnableBy(ResolvedType)   (line 267)

the problems are in lines:

281:  this.ctClass.getSuperclass()
285: for (CtClass interfaze : ctClass.getInterfaces())

when ctClass is of type:  CtClassType.

For example, the implementation of "CtClassType::getInterfaces()" is

```java
@Override
    public CtClass[] getInterfaces() throws NotFoundException {
        String[] ifs = getClassFile2().getInterfaces();
        int num = ifs.length;
        CtClass[] ifc = new CtClass[num];
        for (int i = 0; i < num; ++i)
            ifc[i] = classPool.get(ifs[i]);                  // <<===

        return ifc;
    }
```

where the implementation of "ClassPool.get(classname)" is

```java
public CtClass get(String classname) throws NotFoundException {
        CtClass clazz;
        if (classname == null)
            clazz = null;
        else
            clazz = get0(classname, true);

        if (clazz == null)
            throw new NotFoundException(classname);
        else {
            clazz.incGetCounter();
            return clazz;
        }
    }
```

Now, you suppose 

1) to have a JarTypeSolver for the JDK and ANOTHER JarTypeSolver for another library LIB.
2) to have a class/interface in LIB (used inside the code) that implements the interface "java.io.Serializable" (or extends/ implements some class/interface in the JDK)

BECAUSE the classes & interfaces are resolved in the SAME classpool of the current class, the code throws the exception "NotFoundException".

The CURRENT SOLUTION is to collect in an UNIQUE classPool ALL external libraries PLUS the JDK.
That is, the classPool must be used like a Java class loader.

Obviously, this is not limited to JDK and LIB, but it is general.

The CORRECT SOLUTION is to resolve classes & interfaces starting ALWAYS from the root type solved.

jlerbsc:
@corradomio Thanks for reporting this issue.

jlerbsc:
Hi @matozoid , @maartenc , @ftomassetti The easiest way to solve this problem is probably to make the JarTypeSolver.addPathToJar (..) method public and offer the possibility to initialize JarTypeSolver with multiple jar files like this. What do you think of this proposal?

    JarTypeSolver jSolver= new JarTypeSolver("src/test/resources/jar1.jar");
[...]

PR #3142, opened 2021-03-05
Implemented getAncestors method for Annotation declaration

### LINKED ISSUE #3140: When resolve a MethodCallExpr, throws 'java.lang.UnsupportedOperationException'
Test case:
```java
import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;

import com.github.javaparser.symbolsolver.resolution.typesolvers.JarTypeSolver;

public class TestIssue3140 {
    public static void main(String[] args) throws IOException {
        Path pathToJar = Paths.get("/Users/liuyuan/Downloads/dataset/traccar-4.6/libs/jakarta.annotation-api-1.3.4.jar");
        CombinedTypeSolver cts = new CombinedTypeSolver();
        cts.add(new ReflectionTypeSolver(false));
        cts.add(new JarTypeSolver(pathToJar));
        config.setSymbolResolver(new JavaSymbolSolver(cts));

        String str = "import java.lang.reflect.Method;\n"
                   + "import javax.annotation.security.PermitAll;\n"
                   + "public class Foo {\n"
                       + "void a() {\n" 
                       + "    Method method = new Method();\n"
                       + "    method.isAnnotationPresent(PermitAll.class);\n"
                       + "}\n" + "}\n";

        CompilationUnit cu = StaticJavaParser.parse(str);
        MethodCallExpr mce = cu.findFirst(MethodCallExpr.class).get();
        
        if (mce.getScope().isPresent()){
            System.out.println(mce.resolve().getQualifiedSignature());
        }
    }
}
```
The code `str` comes from traccar-4.6's `SecurityRequestFilter.java` file, line 108.

The jar is attached here.
[jakarta.annotation-api-1.3.4.zip](https://github.com/javaparser/javaparser/files/6091521/jakarta.annotation-api-1.3.4.zip)

4everTheOne:
Fixes #3140.

Implemented method `getAncestors`  for `JavassistAnnotationDeclaration.java ` . This was the only class missing it, since `ReflectionAnnotationDeclaration` and `JavaparserAnnotationDeclaration` already implement it.


jlerbsc:
Can you correct the configuration of your IDE because the modifications become unreadable.

jlerbsc:
ReflectionFactory class is not a JP class. This PR cannot compile without this class.

jlerbsc:
Idem for JavassistFactory.

4everTheOne:
Thansk for the feedback @jlerbsc!.

I'm not sure if i understood what you meant with:
> ReflectionFactory class is not a JP class. This PR cannot compile without this class.

Correct if i'm wrong, but the files that were affected by this PR belongs to JP Symbol Solver and `JavassistTypeDeclarationAdapter` already uses `JavassistFactory`.
https://github.com/javaparser/javaparser/blob/5099b2edc17d3a4085aead68869a803f32914691/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java#L95-L103

Am I missing something?

jlerbsc:
> Correct if i'm wrong, but the files that were affected by this PR belongs to JP Symbol Solver and `JavassistTypeDeclarationAdapter` already uses `JavassistFactory`

You are right. I don't know why but when i did a search on my ide i couldn't find a reference of this class. Sorry :-)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
```diff
@@ -110,7 +112,9 @@ private void collectDeclaredFields(CtClass ctClass, List<ResolvedFieldDeclaratio
      */
     public List<ResolvedReferenceType> getAncestors() {
         return Collections.singletonList(
-                JavassistFactory.typeUsageFor(ctClass, typeSolver).asReferenceType()
+                JavaParserFacade.get(typeSolver)
+                        .classToResolvedType(Annotation.class)
+                        .asReferenceType()
         );
     }
```
MysterAitch: Potentially silly question -- for what reason are we able to _always_ return `Annotation` here? 

If we were within `JavassistAnnotationDeclaration.java` I could understand (the superclass of an annotation is always `Annotation`), but this is within the more general `JavassistTypeDeclarationAdapter.java`. 

What have I missed? :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
```diff
@@ -100,7 +100,7 @@ public boolean isAssignableBy(ResolvedReferenceTypeDeclaration other) {
 
     @Override
     public List<ResolvedReferenceType> getAncestors(boolean acceptIncompleteList) {
-        throw new UnsupportedOperationException();
+        return javassistTypeDeclarationAdapter.getAncestors();
     }
```
MysterAitch: Would it be possible to have: 

```java
    @Override
    public List<ResolvedReferenceType> getAncestors(boolean acceptIncompleteList) {
        return javassistTypeDeclarationAdapter.getAncestors();
    }
```

If not (or if it's possible but not yet implemented), could we add documentation to explain why?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
```diff
@@ -37,68 +40,81 @@
...
+     * Helper method to get the list of ancestors for the annotation.
+     *
+     * @return The list of ancestors.
+     */
+    public List<ResolvedReferenceType> getAncestors() {
+        return Collections.singletonList(
+                JavaParserFacade.get(typeSolver)
+                        .classToResolvedType(Annotation.class)
+                        .asReferenceType()
+        );
```
MysterAitch: Does `JavassistTypeDeclarationAdapter` only apply to annotations?

Just trying to get my head around how/why `getAncestors()` (plural) can return a singleton list.

I can only assume that getting the ancestor for e.g. `List` doesn't trigger this method?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
4everTheOne: 🤦 You are right! Thanks for pointing it out! For some reason i assumed it was an adapter only for AnnotationDeclaration but it's an adapter for all the TypeDeclaration. I'll make it get all the classes from the `CtClass` :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
4everTheOne: Could you point the difference here please? I'm not understanding it.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistTypeDeclarationAdapter.java
4everTheOne: Looking into the code, looks like other implementations that use it (`JavassistEnumDeclaration`, `JavassistClassDeclaration` and `JavassistInterfaceDeclaration`) implement the same logic. It could probably be simplified by making them call this method.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationDeclaration.java
MysterAitch: Oh... silly typo on my part.... doh!

I meant to ask about passing on `acceptIncompleteList` rather than silently dropping it :+1: 

```java
    @Override
    public List<ResolvedReferenceType> getAncestors(boolean acceptIncompleteList) {
        return javassistTypeDeclarationAdapter.getAncestors(acceptIncompleteList);
    }
```

MysterAitch: .. but it seems you might have dealt with adding that in anyway it seems! :grinning:

4everTheOne:
At the moment i'm trying to see if is possible to make `JavassistClassDeclaration` use the getAncestors from the adapter to make it consistent between them.
I will change this PR to draft to avoid confusions :)

4everTheOne:
I'm having issues when I trie to get the list of interfaces in a CtClass because the behavior is being inconsistent. If I run the tests in the IDE, all of them pass. But when i try to run it with maven it fails due to a weird behavior in `CtClass`.

For example given the `Node` class from the project i would expect it to have 5 interfaces (Cloneable, HasParentNode, Visitable, NodeWithRange and NodeWithTokenRange). But the method is only returning Cloneable.

#### Unit test

```java
@Test
void testCtClass() throws NotFoundException {
    CtClass clazz = ClassPool.getDefault().getCtClass("com.github.javaparser.ast.Node");
    assertEquals(Arrays.asList(
            "java.lang.Cloneable",
            "com.github.javaparser.HasParentNode",
            "com.github.javaparser.ast.visitor.Visitable",
            "com.github.javaparser.ast.nodeTypes.NodeWithRange",
            "com.github.javaparser.ast.nodeTypes.NodeWithTokenRange"
            ),
            Arrays.stream(clazz.getInterfaces()).map(CtClass::getName).collect(Collectors.toList()));
}
```

Does anyone know what can be the cause of this inconsistency?

MysterAitch:
Related to the discussions within #3158 and #2762 perhaps? 

Not sure what would cause this specific problem but concurrency issues are where I would start looking with intermittent / inconsistent problems. 

May also worth doing a maven clean and compile -- it might be that the test modules have old copies of the compiled sibling modules, which the IDE test runner is using instead of the "newer" versions. _(this is something I thought I fixed a while ago but it's slipped through -- iirc the intention is to only copy across compiled core/symbol solver code into the test modules when doing code coverage calculations)_



---

For reference: 

https://github.com/javaparser/javaparser/pull/3142/checks?check_run_id=2058535095#step:6:9582

```

[INFO] Results:
[INFO] 
Error:  Failures: 
Error:    JavassistTypeDeclarationAdapterTest.testGetAncestors:64 expected: <[java.lang.Object, java.lang.Cloneable, com.github.javaparser.HasParentNode, com.github.javaparser.ast.visitor.Visitable, com.github.javaparser.ast.nodeTypes.NodeWithRange, com.github.javaparser.ast.nodeTypes.NodeWithTokenRange]> but was: <[java.lang.Object, java.lang.Cloneable]>
[INFO] 
Error:  Tests run: 1523, Failures: 1, Errors: 0, Skipped: 35
[INFO] 
Warning:  Corrupted STDOUT by directly writing to native stream in forked JVM 1. See FAQ web page and the dump file /home/runner/work/javaparser/javaparser/javaparser-symbol-solver-testing/target/surefire-reports/2021-03-08T17-14-18_085-jvmRun1.dumpstream
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for javaparser-parent 3.20.1-SNAPSHOT:
[INFO] 
[INFO] javaparser-parent .................................. SUCCESS [  0.269 s]
[INFO] javaparser-core .................................... SUCCESS [ 12.928 s]
[INFO] javaparser-core-testing ............................ SUCCESS [ 52.562 s]
[INFO] javaparser-core-testing-bdd ........................ SUCCESS [ 12.609 s]
[INFO] javaparser-core-generators ......................... SUCCESS [  1.019 s]
[INFO] javaparser-core-metamodel-generator ................ SUCCESS [  0.104 s]
[INFO] javaparser-core-serialization ...................... SUCCESS [  1.096 s]
[INFO] javaparser-symbol-solver-core ...................... SUCCESS [  1.550 s]
[INFO] javaparser-symbol-solver-testing ................... FAILURE [06:22 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  07:44 min
[INFO] Finished at: 2021-03-08T17:21:49Z
[INFO] ------------------------------------------------------------------------
```

4everTheOne:
Thank you so much @MysterAitch for those references, it helped a lot to solve the issue!

The issue was related with the class pool. Since JarTypeSolver uses the default class pool it will replace the current Node class with the ones present in other tests Jar causing it behave weird in this situation.

maartenc:
#3167 contains a similar fix
Let's see if the junit tests defined here still work after merging that PR.

maartenc:
Looks good @4everTheOne , but could you change your IDE settings to keep the import statements as they are?

maartenc:
Sorry @4everTheOne , I've deleted the `JavassistInterfaceDeclarationTest.java` file by accident, I wanted to revert the changes you made in that file thinking the 'Delete' action in github was to delete this file from the PR (but I was wrong, it actually deletes the file from version control).

Could you revert that delete commit I've made? I can't find a way to do thay myself from the github UI.

thanks,
Maarten

4everTheOne:
No problem ;)

> I wanted to revert the changes you made in that file

Want me to restore `JavassistInterfaceDeclarationTest.java` to it's original state?

maartenc:
Yes please restore it to it's original state, it has nothing to do with the change you made.
thanks!

maartenc:
Thanks @4everTheOne !

PR #3156, opened 2021-03-08
Fixed generator visitor order to match the expected behavior

### LINKED ISSUE #3016: Make #3011 permanent
The visitors are generated, which means the manual re-ordering within #3011 gets reset/overridden whenever the generators are run next.

4everTheOne:
Fixes #3016.

According to issue #3011 the order of visitors is not consistent, making it not work as expected. In that PR was suggested a fix for it, but it's replaced every time the generator is run. This PR fixes the issue by implementing it directly in the Generator.

Although #3011 only mentioned `ModifiersVisitor`, in this PR the other visitors were also updated to keep the visit order consistent between them.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/VisitorGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/VisitorGenerator.java
@@ -28,9 +28,13 @@ import com.github.javaparser.ast.expr.MarkerAnnotationExpr;
 import com.github.javaparser.ast.expr.Name;
 import com.github.javaparser.metamodel.BaseNodeMetaModel;
 import com.github.javaparser.metamodel.JavaParserMetaModel;
+import com.github.javaparser.metamodel.PropertyMetaModel;
 import com.github.javaparser.utils.Log;
 import com.github.javaparser.utils.SourceRoot;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
 import java.util.Optional;
 
 import static com.github.javaparser.ast.Modifier.Keyword.PUBLIC;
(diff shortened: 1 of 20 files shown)


javaparser-core-generators/src/main/java/com/github/javaparser/generator/VisitorGenerator.java
```diff
@@ -99,5 +103,18 @@ private void generateVisitMethodForNode(BaseNodeMetaModel node, ClassOrInterface
...
+     *
+     * @param baseNodeMetaModel The metamodel to get the properties.
+     *
+     * @return The list of properties.
+     */
+    public List<PropertyMetaModel> getAllPropertyMetaModels(BaseNodeMetaModel baseNodeMetaModel) {
+        List<PropertyMetaModel> propertiesList = new ArrayList<>(baseNodeMetaModel.getAllPropertyMetaModels());
+        Collections.reverse(propertiesList);
+        return propertiesList;
+    }
```
MysterAitch, line 117: This is the only place that I could find which does any sorting (in this case, reversing) of the properties. Is this where the deterministic order happens?

MysterAitch:
I'm struggling to see where the order of the properties is defined. @4everTheOne -- could you help by pointing it out please?

MysterAitch:
Note that the metamodel is generated in alphabetical order by default -- iirc it is specifically done here: 

https://github.com/javaparser/javaparser/blob/bdf0ebd7aac08170a08c9cc3af1d0a501694ce1e/javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java#L259-L260


Rather than sorting alphabetically, we could potentially do a priority sort (e.g. comments first) or something else if we want a particular order.

4everTheOne:
In my point of view, the issue is related with the order `BaseNodeMetaModel#getAllPropertyMetaModels()` fills the list. 

Lets imagine we are in an `AnnotationDeclarationMetaModel` which belongs to `AnnotationDeclaration`. The method will populate the list with the the current properties of the class and then it's super items. To clarify, in the case of a `AnnotationDeclaration` will add properties using the following order `AnnotationDeclaration` -> `TypeDeclaration` -> `BodyDeclaration` -> `Node`.

I think this is the issue, because like was mention in the original [issue](https://github.com/javaparser/javaparser/pull/3011#issue-547773848):
> ModifierVisitor works top-down, visiting larger then smaller scopes.

I think the way it should work should be like this:
`Node` ->  `BodyDeclaration` -> `TypeDeclaration` -> `AnnotationDeclaration`

By using the `Collections#reverse` we can achieve that objective by visiting the lower nodes and finishing at higher nodes. Note that doing this order also makes the visit order consistent since the common nodes are always visited first then the others.

4everTheOne:
Note that the order at `MetaModelGenerator` is generated using a special order to avoid the need of sorting.
https://github.com/javaparser/javaparser/blob/bdf0ebd7aac08170a08c9cc3af1d0a501694ce1e/javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java#L59-L61

With this order, it will match the expected hierarchy.

Issue #1851, opened 2018-09-12
Unable to resolve parent class


ro11ak:
Hi,
I'm working on an analysis tool that would parse through all the javax.security methods in a class and list out all the invoked methods in that class.

I'm trying to parse a class that extends another class and I'm getting the following exception
```java
UnsolvedSymbolException{context='unknown', name='AbstractUsernamePasswordAuthenticationHandler'}
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.toReferenceType(JavaParserClassDeclaration.java:435)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.getSuperClass(JavaParserClassDeclaration.java:166)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.getAncestors(JavaParserClassDeclaration.java:321)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.getAncestors(JavaParserClassDeclaration.java:299)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveType(JavaParserTypeDeclarationAdapter.java:71)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveType(ClassOrInterfaceDeclarationContext.java:90)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:436)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convert(JavaParserFacade.java:496)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convert(JavaParserFacade.java:492)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserFieldDeclaration.getType(JavaParserFieldDeclaration.java:72)
	at com.github.javaparser.symbolsolver.model.resolution.Value.from(Value.java:40)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveSymbolAsValue(ClassOrInterfaceDeclarationContext.java:71)
[...]

Issue #362, opened 2016-07-12
How to get a list of arraylists declared in a java class?


Chatz92:

matozoid:
``` java
        // Cheat by parsing the code we want, then looking at the structure in a debugger.
        CompilationUnit cu = JavaParser.parse(new StringReader("class A {List<ArrayList<String>> b=new ArrayList<>();}"), false);
        System.out.println(cu);


        // First, the left side
        // String
        ReferenceType stringType = new ReferenceType(new ClassOrInterfaceType("String"));
        // ArrayList
        ClassOrInterfaceType arrayListCoi = new ClassOrInterfaceType("ArrayList");
        // ArrayList<String>
        arrayListCoi.setTypeArguments(withArguments(singletonList(stringType)));
        ReferenceType arrayListType = new ReferenceType(arrayListCoi);
        // List
        ClassOrInterfaceType listCoi = new ClassOrInterfaceType("List");
        // List<ArrayList<String>>
        listCoi.setTypeArguments(withArguments(singletonList(arrayListType)));
        ReferenceType listType = new ReferenceType(listCoi);

        // Then the right side
        // ArrayList<>
        ClassOrInterfaceType rightArrayListCoi = new ClassOrInterfaceType("ArrayList");
        rightArrayListCoi.setTypeArguments(withDiamondOperator());
        // new ArrayList<>()
        ObjectCreationExpr newArrayListExpr = new ObjectCreationExpr(null, rightArrayListCoi, null);
        // b = ...
        VariableDeclarator b = new VariableDeclarator(new VariableDeclaratorId("b"));
        // b = new ArrayList<>();
        b.setInit(newArrayListExpr);

        // List<ArrayList<String>> b = new ArrayList<>();
        FieldDeclaration listArrayListStringField = new FieldDeclaration(0, listType, b);

        // Wrap it in a class and print it.
        ClassOrInterfaceDeclaration aClass = new ClassOrInterfaceDeclaration(0, false, "A");
        aClass.getMembers().add(listArrayListStringField);

        List<TypeDeclaration> types= new ArrayList<>();
        types.add(aClass);

        CompilationUnit cu2 = new CompilationUnit(null, null, types);
        System.out.println(cu2);

```

matozoid:
I assume this was sufficient.

PR #3158, opened 2021-03-08
Use a singleton javassist classpool to fix issue #1771 and all relate…

### LINKED ISSUE #1771: fail if a type is defined in a JAR, but implements an interface defined in another JAR
See #362 and #1851

jlerbsc:
…d issues  #362 and #1851 and revert a part of the PR ##3126 that is not compatible with the classpool singleton

Fixes #1771.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
@@ -56,7 +56,9 @@ public class JarTypeSolver implements TypeSolver {
 
     private TypeSolver parent;
     private Map<String, ClasspathElement> classpathElements = new HashMap<>();
-    private ClassPool classPool = new ClassPool(false);
+    // Returns the default class pool. The returned object is always identical since this method is a singleton
+    // factory. The default class pool searches the system search path. This is a difference from the previous class pool instantiation.
+    private ClassPool classPool = ClassPool.getDefault();
     
     /*
      * ResourceRegistry is useful for freeing up resources.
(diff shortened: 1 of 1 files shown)


maartenc:
@jlerbsc , @MysterAitch 

I don't think this is a good fix, although it will probably solve the mentioned issues, it will create new problems.

For instance: I have a use case where I use JavaParser to check the dependencies of multiple Java libraries.
For each library a 'root typeresolver' is created, and external jars are added to this resolver using the javassist typesolver.
When one 'root typesolver' can see classes from another one, strange things can happen.

It's not a good idea to 'leak' javassist classes accross the boundaries of such a typesolver. Since it is static now, this will be the case. Please revert and change the scope of the javasisst classes to the scope of the type solver.

maartenc:
See also my comment in https://github.com/javaparser/javaparser/issues/2762

MysterAitch:
Just commented within #2762 

There are a lot of threads with similar topics, is there an obvious choice available where we can direct discussion to a single issue?

jlerbsc:
> It's not a good idea to 'leak' javassist classes accross the boundaries of such a typesolver. Since it is static now, this will be the case. Please revert and change the scope of the javasisst classes to the scope of the type solver.

@maartenc there are many issues that are related to this problem. Do you have another solution to offer these users who have potentially been blocked for several months or even years.

> When one 'root typesolver' can see classes from another one, strange things can happen.

Can you clarify the "strange things" and propose a test case so that we can work on them as well?

maartenc:
For instance: a problem can be like this:

```
CombinedTypeSolver solver1 = new CombinedTypeSolver();
solver1.add(new JarTypeSolver("guava-12.0.jar"));
solver1.add(new ReflectionTypeSolver());

CombinedTypeSolver solver2 = new CombinedTypeSolver();
solver2.add(new JarTypeSolver("guava-13.0.jar"));
solver2.add(new ReflectionTypeSolver());
```

Which guava version will solver1 and solver2 use?
Will solver1 use 12.0 and solver2 use 13.0 ?
Will both use 12.0? Or both use 13.0? Or will it be a mix?

maartenc:
Another easy solution is your first suggestion in #2762

PR #3161, opened 2021-03-10
Created dedicated class pool for JarTypeSolver to avoid pollution

4everTheOne:
While I was developing test code for the project, noticed that the default `ClassPool` was pointing to old classes. The problem was related with the `JarTypeSolver` using the default class pool to create new classes that override the old ones.

For more information, please see: https://github.com/javaparser/javaparser/pull/3142#issuecomment-793041340

This PR replaces the use of the default `ClassPool` with a dedicated `ClassPool`.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
@@ -21,31 +21,21 @@
 
 package com.github.javaparser.symbolsolver.resolution.typesolvers;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
-
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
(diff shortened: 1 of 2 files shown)


4everTheOne:
Giving an example to make it more clear.
If we ask default `ClassPool` for the class `Node` we would expect to be exactly how it is defined actually. But if we execute the same operation after using `JarTypeSolver` in JP code, the current `Node` class is replaced with the content from the Jar.

jlerbsc:
This subject is already open and not closed for the moment (see https://github.com/javaparser/javaparser/issues/2762). Can you close this PR please?

4everTheOne:
Hello @jlerbsc, sorry but i don't think this PR is directly related with that issue.

In a recent commit the `ClassPool` was changed to a singleton to fix https://github.com/javaparser/javaparser/issues/1771. This ideia does the job, but in the implementation was used the default `ClassPool` instead of creating a singleton just for that class.

This brings the problems that have mentioned above.

jlerbsc:
These 2 PR deal with the same subject. The default classpool acts as a singleton.

4everTheOne:
> These 2 PR deal with the same subject.

Sorry, could you point me it please? The only one I see related to this is already merged https://github.com/javaparser/javaparser/pull/3158. Am II missing something?


>  The default classpool acts as a singleton.

I understand that the default `ClassPool` also works as a singleton. The problem with using the default is that it can cause side effects for the user of JavaParser. Let's assume that the user also needs to use JavaAssist for his project, it is very likely that he uses the default ClassPool. If we are polluting the `ClassPool` with the classes that are being analyzed, we can create problems in the results of the users, for example by replacing classes in the current runtime with others that are being analyzed.

jlerbsc:
> If we are polluting the `ClassPool` with the classes that are being analyzed, we can create problems in the results of the users, for example by replacing classes in the current runtime with others that are being analyzed.

This is probably borderline case that can possibly produce side effects but there are common cases of architecture that currently does not work and that was solved by this modification. Whatever the PR is canceled. Borderline cases work again and the issues  corresponding to the common case remain to be solved.

> These 2 PR deal with the same subject.

I meant issues not PR.

Issue #3138, opened 2021-03-05
Can't find its parameter type declared in another jar file for a method declared in a jar file.


ghost:
I'm trying to get the method signature using this one:
```java
import java.io.IOException;
import java.nio.file.Path;
import java.nio.file.Paths;

import com.github.javaparser.ParserConfiguration;
import com.github.javaparser.StaticJavaParser;
import com.github.javaparser.ast.CompilationUnit;
import com.github.javaparser.ast.expr.MethodCallExpr;
import com.github.javaparser.symbolsolver.JavaSymbolSolver;
import com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver;
import com.github.javaparser.symbolsolver.resolution.typesolvers.JarTypeSolver;
import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;

public class Test1 {
    public static void main(String[] args) throws IOException {
        Path pathToJar1 = Paths.get("/Users/liuyuan/Downloads/dataset/traccar-4.6/libs/jackson-databind-2.9.9.jar");
        Path pathToJar2 = Paths.get("/Users/liuyuan/Downloads/dataset/traccar-4.6/libs/jackson-core-2.9.9.jar");
        ParserConfiguration config = new ParserConfiguration();
        CombinedTypeSolver cts = new CombinedTypeSolver();
        cts.add(new ReflectionTypeSolver(false));
        cts.add(new JarTypeSolver(pathToJar1));
        cts.add(new JarTypeSolver(pathToJar2));
        config.setSymbolResolver(new JavaSymbolSolver(cts));
        StaticJavaParser.setConfiguration(config);

        String str = "import com.fasterxml.jackson.databind.SerializationConfig;\n"
                + "import com.fasterxml.jackson.databind.SerializationFeature;\n" + "public class Foo {\n"
                + "  void a() {\n" + "    SerializationConfig serializationConfig = new SerializationConfig();\n"
                + "    serializationConfig.without(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);\n"
                + "  }\n" + "}\n";

        CompilationUnit cu = StaticJavaParser.parse(str);
        MethodCallExpr mce = cu.findFirst(MethodCallExpr.class).get();
        System.out.println(mce);
        System.out.println(mce.resolve().getQualifiedSignature());
    }
}
```

The test case above is:
```java
import com.fasterxml.jackson.databind.SerializationConfig;
[...]

Issue #2761, opened 2020-07-19
JavassistInterfaceDeclaration unsupports "isAssegnableBy"


corradomio:
The current implementation of "isAssegnableBy" JavassistInterfaceDeclaration throws the exception "UnsupportedOperationException".

I suppose this error can be resolved using the implementation in JavassistClassDeclaration (confirmed!)

PR #3167, opened 2021-03-13
Refactor the javassist implementation to delegate to the typesolver instead of using its own classpool

### LINKED ISSUE #1615: Figure out what JarTypeSolver.getJarTypeSolver is meant to do
It will build a global `JarTypeSolver` for several jars at once, but why global? Why is it unused? Why doesn't `JarTypeSolver` have a public interface for adding jars?

maartenc:
Fixes #1615, #1771, #2761, #2762, #3138, #3140


matozoid:
Hi! I'm not familiar with this code, but it looks really nice, and if it fixes all these issues then I'm all for it :-)

PR #3170, opened 2021-03-14
Implemented logic for isAssignableBy in JavassistInterfaceDeclaration

### LINKED ISSUE #2761: JavassistInterfaceDeclaration unsupports "isAssegnableBy"

4everTheOne:
Fixes #2761.


maartenc:
Thanks!

I'm not quite sure about the type parameters comparison though.
Shouldn't it also take wildcards into consideration?
For instance: 
`List<? extends Foo> foo = new ArrayList<Foo>();`

4everTheOne:
That's a good question!

I created an additional unit test to check for that case and looks like it's resolved properly. Looking into the code surprisingly the isAssignableBy is being solved by `ReferenceTypeImpl` and not by the declaration itself.

This raise the question if we could delegate this task to RefereceTypeImpl instread of declaring in all of the Class/Interface/Enum declaration. For example, if we replace the code in `isAssignableBy` with the code bellow we can get the same behaviour.

```java
if (type.isNull()) {
    return true;
}

if (type instanceof LambdaArgumentTypePlaceholder) {
    return isFunctionalInterface();
}

return type.isAssignableBy(new ReferenceTypeImpl(this, typeSolver));
```

maartenc:
That's an interesting idea indeed.
Let's wait to look into this further untill we merge #3167 which contains some similar changes as the ones in this PR.

4everTheOne:
Thank you @maartenc! That improvement makes the code much easier to work with :)

maartenc:
You have some files with only changes to import statements now.
Could you remove them from this PR?

maartenc:
Shouldn't `JavassistClassDeclaration` and `JavassistEnumDeclaration` be changed to use the adapter class as well?
Otherwise, it's best to put the implementation directly into `JavassistInterfaceDeclaration`.

4everTheOne:
Yes they should, `JavassistAnnotationDeclaration`, `JavassistClassDeclaration` and `JavassistEnumDeclaration` should be also changed. But to make this changes we must create more tests to ensure the behaviors didn't change for each class. Can it be done in another PR to keep this one focused on `JavassistInterfaceDeclaration`?

maartenc:
Thanks!

Issue #3176, opened 2021-03-21
Generate the changelog


MysterAitch:
At the moment I populate the changelog manually each release based on the closed PRs for that version/milestone.

@jlerbsc gave a kind nudge that this kind of work could/should be automated, so here's an issue to track this! :+1: 

It will also help to avoid silly errors on my part such as #3155 :laughing:

MysterAitch:
I have no doubt that there are likely better options available :sweat: but the [Spring Github Changelog Generator](https://github.com/spring-io/github-changelog-generator) is the project/tool that I intend to be run with for this. Specifically, [my fork of it]( https://github.com/spring-io/github-changelog-generator) that is much more configurable and whose output can now be configured to match the current format of our current changelog. I intend to upstream my changes, hopefully then being able to use the official release eventually.

The way that this tool works is that it takes a reference to a milestone (open or closed), extracts details about the PRs/Issues attached to that milestone, and uses the labels attached to those PRs/Issues to make entries under several (configurable) headings. The output is a markdown file containing the generated changelog.

In recent days I have been using it to aid with manually preparing the next version's changelog, and my fork of the project seems to be working really well so far (you might have seen the extra labels I've been adding to PRs - more on this in a moment).

There is potential future scope for fully automating this to:
- generate the milestone changelog then inserting it into CHANGELOG.md
- generate the milestone changelog then adding it to the releases page

Some of this could be done during `mvn release`, and some of it could be done via [github actions](https://github.com/Decathlon/release-notes-generator-action).

Comments? Thoughts?

maartenc:
> At the moment I populate the changelog manually each release based on the closed PRs for that version/milestone.

I didn't specify the milestone when merging PR's last week, does this mean these PR's will not be listed in the changelog?

MysterAitch:
> > At the moment I populate the changelog manually each release based on the closed PRs for that version/milestone.
> 
> I didn't specify the milestone when merging PR's last week, does this mean these PR's will not be listed in the changelog?

It's fine - when preparing a release I go through and add the merged PRs (and associated issues) to the "next release" milestone -- the "recently updated" sort and scanning down for merged PRs that don't have a milestone is easy enough to work through :+1: 

![image](https://user-images.githubusercontent.com/1537214/111904235-c0f74200-8a3d-11eb-953e-17ac3e270dc7.png)

See here for the Issues/PRs next release: https://github.com/javaparser/javaparser/milestone/182?closed=1


[...]

PR #3177, opened 2021-03-21
Changelog generator

MysterAitch:
Per the description within #3176 , this PR enables generating a changelog entry for a given milestone ID (i.e. the one from the milestone URL). 

- It can be configured for a milestone title too, which may be a better option?
- The JAR, shell script, and configuration file are currently in the `dev-files` directory - is there a better place to put it?
- Should it even be in the repo? Alternatively, put instructions for it in the wiki or somewhere for devs to manually download and run?
- Long term it would be great to fully automate this

Comments? Thoughts?


MysterAitch:
Sidenote: Some of the configuration file is a little klunky looking at the moment - these are mostly custom configuration options that I have added to my fork, and may be subject to change. 

For example, all the options for customising the output probably warrant being under their own separate heading/nesting level.

MysterAitch:
Superseded by #3391

Issue #3119, opened 2021-02-28
Partial SymbolSolver for Map/HashMap


thiagodnf:
Hello everyone,

I am trying to use the SymbolSolver to resolve `Map<String, CloneFailedException>`. The problem is because `CloneFailedException` comes from a library that I didn't include (in fact, I don't want to include it), so JavaParser is throwing an exception about the missing of this class.

I would like to know if it is possible to resolve just "Map" and "String"? In other words, is it possible to get the following as output?

```
java.util.Map<java.util.String, CloneFailedException>
```

Thanks in advance.

MysterAitch:
Not to my knowledge.

There might be workarounds like providing a stub for the other library's classes (variable feasibility -- e.g. just a couple of classes vs a whole library), but this isn't something I've personally tried so your feedback here would be appreciated!

jlerbsc:
Hi @thiagodnf if @MysterAitch answered your question can you close this issue please?

Col-E:
If you do end up going with the stubbed classes approach, I'd recommend checking out [JPhantom](https://github.com/gbalats/jphantom). It can automate the process pretty easily for you. I have a fork open with some improvements and the artifact is tossed up on Jitpack if anything in the latest PR interests you.

jlerbsc:
It seems to me that the approach of the JP project leaders has been to avoid as much as possible including external libraries but it is interesting to have this in mind.

thiagodnf:
Thanks everyone for the comments and also @4everTheOne for trying to submit a pull request that tries to address my issue (in fact, you went beyond of my issue)

I think what you did @4everTheOne could be useful to me but, of course, I don't want to break the entire code. Regarding the stubbed classes, it looks also interesting but the problem is I don't know how to create stubs "dynamically" (it means that I want to create those ones just when I have a non-resolved class)

Again, thanks for this discussion

PR #3182, opened 2021-03-22
Efficiency improvement for Symbol Solver

4everTheOne:
When a type has type arguments (example: `List<String>`) the symbol solver will try to solve the outer type (`List`) AND also the type argument (`String`). The type argument don't need to be solved right away, since we may only want to solve the type. This PR suggests a way to "delay" the resolution of the type argument until it's needed.

The way the delay works, is by wrapping the type argument into a `LazyType` resolving it only when it's necessary.

This may also help in situation like have been described at https://github.com/javaparser/javaparser/issues/3119, where the user don't care about the resolution of the argument just the outer type itself.

## Pull-Request structure:
- https://github.com/javaparser/javaparser/commit/31fa100b9793c0b4e5edb1c03ff36cb949fc778c Wrap the type argument in a LazyType;
- All the commits with the following pattern: "Updated XXX to support LazyType" replace the use of `instanceOf` with the equivelent method to support `LazyType`. 
- https://github.com/javaparser/javaparser/commit/2af3112beb07177e46981b4e1baebed2ccbc956e ,https://github.com/javaparser/javaparser/commit/0c8458113d0ee0b209a5602702ebbc0bf58ca6e5 Adds unit tests to make sure the behavior matches the expected.


MysterAitch:
Thank you for the description of the PR structure - it is helpful.

- I don't fully understand the lazy type side of things as a whole, but each individual part seems reasonable.
- Switching away from `instanceof` and casting is generally a good thing :+1: 

Is this an internal only change? Is this a transparent change to users of JavaParser? _(i.e., will users need to be aware of LazyType?)_

Edit: I'm also being slow with this being an efficiency thing. At what point does the full resolution happen? Does it need to be triggered by the user? How does it change the output to the user? e.g. will resolving `List<String>` now return `java.util.List<String>` until the `String` is also resolved? Apologies if I am misunderstanding here.

4everTheOne:
No problem MysterAitch!

> Is this an internal only change? Is this a transparent change to users of JavaParser? (i.e., will users need to be aware of LazyType?)

Yes it's just a internal change, they don't need to be aware of the `LazyType`. This change is completely transparent to the end user, they should not notice that this change happened.

---

> How does it change the output to the user?

This should not affect at all in the user output.

---
> Edit: I'm also being slow with this being an efficiency thing

Let me explain the benefits with an example. (_Note: This example might be silly, but I think it explains the benefits_)

Lets imagine the user wants to solve the type of all variables in a program, and just check if the type is `java.util.List`.
```java
List<List<List<List<List<List<String>>>>>> listOfList= new ArrayList<>();
```
In the current version the JP SS needs to solve ALL the type arguments before providing a resolved type to the user. (Resolving the chain of type arguments)
This takes time to resolve and in the end, the user didn't care about it, he just needed the outer type resolved (The first `List`)

This PR tries to resolve it only when is needed. So in the previous example, the JP SS will only parse the first `List` and mark the type arguments as `Lazy` (which means will only be loaded if needed), avoiding to resolve types that are not used by the final user.

---

> At what point does the full resolution happen? Does it need to be triggered by the user? 

The resolution happen when we need to get information for the type argument. Going back to previous example, if we need to access the types arguments `List<HERE>` using the builtin methods the `LazyType` will solve it, completely transparent to the user.

Did the explanation clarify your doubts?

_If you need a more detailed information about [Lazy loading](https://martinfowler.com/eaaCatalog/lazyLoad.html), I suggest this explanation._

MysterAitch:
What you describe sounds good.

I think my concern is mostly around a scenario where we "successfully" resolving the outer part lazily e.g.: 

```
List<List<List<List<List<List<String>>>>>>
^^^^^                                    ^
```

... but then it failing when we try to get the resolved "other parts" -- (presumably) current behaviour is to fail when any part of the chain cannot be resolved, but this new behaviour is to fail when the user tries to access each part. Is this a reasonable interpretation?


Another example could be e.g. with two classes `a.b.c.ClassA` and `x.y.z.ClassA`, therefore potentially resolving to different "outer" classes -- `List<ClassA<List<String>>> `. 

Presumably the same applies where we have overloaded methods, or perhaps `ClassA.doSomething(List mno)` vs `ClassA.doSomething(ArrayList mno)`. Or perhaps something like `new ArrayList<ClassA>)();`, or `customClass(ClassA mno)`.

With the disclaimer that I haven't downloaded this or thoroughly tested it myself, I do not understand enough to know if these are sensible examples or concerns -- it could easily be that I am worrying about nothing! :grinning:

4everTheOne:
> but this new behavior is to fail when the user tries to access each part

That for sure can be a thing. A work around for this could be the use of flag that the user indicates if he wants to use the lazy parsing or the normal. this allows us to keep the best of both worlds.

In terms of implementation i would imagine something like `#resolve(bool lazyResolution)`

> Another example could be e.g. with two classes a.b.c.ClassA and x.y.z.ClassA, therefore potentially resolving to different "outer" classes -- List<ClassA<List<String>>> .

Don't think that would be a issue, since the same resolution logic is applied. The only diference is just WHEN we call it now or later :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/LazyType.java
```diff
@@ -37,6 +38,34 @@
...
+     * }
+     *
+     * In this example, by calling {@code getConcreteType(typeC)} we would expected the return type to be a
+     * {@link ResolvedVoidType}
+     *
+     * @param type The resolved type to find the concrete type.
+     *
+     * @return The concrete type for the {@link LazyType}.
+     */
+    public static ResolvedType getConcreteType(ResolvedType type) {
```
jlerbsc: I don't understand why this method is static. It seems to me that this method should return the 'concrete type' of the current LazyType

     public ResolvedType getConcreteType() {...}

Besides, why have a new public method that exposes the underlying type. The getType method of the LazyType class cannot return the underlying type?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/LazyType.java
```diff
@@ -165,7 +194,13 @@ public boolean isAssignableBy(ResolvedType other) {
 
     @Override
     public boolean equals(Object o) {
-        return getType().equals(o);
+        if (o instanceof LazyType) {
+            // If is a LazyType then we should find the concrete value of the other before comparing
+            ResolvedType otherType = getConcreteType((LazyType) o);
```
jlerbsc: Assuming o is a LazyType (already casted) we should be able to write 

    ResolvedType otherType = o.getConcreteType();

or probably better

    ResolvedType otherType = o.getType(); // if getType returns the underlying type

4everTheOne:
Thank you for you feedback @jlerbsc !

The PR has been with the suggested changes :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/LazyType.java
```diff
@@ -80,20 +52,32 @@ public LazyType(Supplier<ResolvedType> supplier) {
...
 
-    /**
-     * Get the concrete {@link ResolvedType} for the current {@link LazyType}.
-     *
-     * @return The {@link ResolvedType} if already resolved, otherwise empty.
-     */
-    public Optional<ResolvedType> getConcreteType() {
-        return Optional.ofNullable(concrete);
+        // If the concrete value is a lazy type, then get the inner type.
+        if (concrete instanceof LazyType) {
```
jlerbsc, line 76: Is it possible or autorized to build a LazyType(LazyType(Type))? What could be the meaning of this?

jlerbsc:
> Thank you for you feedback @jlerbsc !
> 
> The PR has been with the suggested changes :)

These are only questions but if you agree then indeed you can modify the code.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/LazyType.java
4everTheOne: > What could be the meaning of this?

To be honest i don't see any point in doing this. But we don't know the type of the wrapped type until we ask for it.
So i think it's better to play safe and assume the user may do this.

jlerbsc: If there is no identified use case, perhaps it is sufficient to protect the constructor against a LazyType nesting.

4everTheOne: The problem is, we don't provide Type in the constructor, we provide a `Supplier` that will create the type only if needed.

If we try to check the type of it in the constructor we are breaking the lazy principle (object is only created when needed).

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/typesystem/LazyType.java
```diff
@@ -80,20 +52,32 @@ public LazyType(Supplier<ResolvedType> supplier) {
...
+     *      LazyType typeA = new LazyType(() -> ResolvedVoidType.INSTANCE);
+     *      LazyType typeB = new LazyType(() -> typeA);
+     *      LazyType typeC = new LazyType(() -> typeB);
+     * }
+     *
+     * In this examples, by calling {@code getType()} we would expected all of them to return {@link ResolvedVoidType}
+     *
+     * @return The concrete type for the {@link LazyType}.
+     */
+    public ResolvedType getType() {
```
jlerbsc: Why are you exposing the underlying type? The LazyType acts as a proxy and delegates behavior to the wrapped type. Is there a special need to do this?

4everTheOne: Essentially just for testing purposes.  Maybe we can reduce the visibility to package instead of public?

jlerbsc: If we consider that the scope of this method is private (it is not useful to test it directly). Isn't invoking the equals method enough to test the behavior of the getType method?

PR #3185, opened 2021-03-23
Further optimization in resolving in StatementContext

### LINKED ISSUE #3181: In some scenarios, version 3.20.0 of type resolution differs greatly from version 3.15.13.
[JavaParserSample3.zip](https://github.com/javaparser/javaparser/files/6181800/JavaParserSample3.zip)

I provided HTML Schema test cases in Java Parser Sample 3. I tested them using 3.20.0 and 3.15.13, and the performance difference was dozens of times.

use 3.20.0 version 
```
208:setPrefix("html")
2785:elementType("<pcdata>", M_EMPTY, M_PCDATA, 0)
2586:elementType("<root>", M_ROOT, M_EMPTY, 0)
3823:elementType("a", M_PCDATA | M_NOLINK, M_INLINE, 0)
6159:elementType("abbr", M_PCDATA | M_INLINE, M_INLINE | M_NOLINK, F_RESTART)
81027:elementType("acronym", M_PCDATA | M_INLINE | M_INLINE | M_INLINE | 
```

use  3.15.13
```
188:setPrefix("html")
21:elementType("<pcdata>", M_EMPTY, M_PCDATA, 0)
9:elementType("<root>", M_ROOT, M_EMPTY, 0)
7:elementType("a", M_PCDATA | M_NOLINK, M_INLINE, 0)
11:elementType("abbr", M_PCDATA | M_INLINE, M_INLINE | M_NOLINK, F_RESTART)
8:elementType("acronym", M_PCDATA | M_INLINE | M_INLINE | M_INLINE | M_INLINE

```

Col-E:
# Symbol resolving optimization

## Preface

This PR is an extension of #3072 

Fixes #3181 

## What's changed

**Old Behavior:**

1. Attempt to resolve a name in an expression
2. If resolution fails try to solve it by looking at all previous nodes
    * If a previous adjacent node fails to resolve the item and the node is also an instance of `NodeWithStatements` it will look at the parent and repeat the process for each prior adjacent node
    * This is wasteful because each prior adjacent statement will look at the parent... but all prior adjacent statements share the same parent, so every call after the first one is pointless since the resolve process will return the same as it did the first time.
3. Try to resolve in the parent node context
    * And to reiterate, at this point we have already done this N times, where N is the number of prior adjacent statements. The result of this resolve for each case will be identical so calling it more than once serves no purpose. Depending on the structure of the code being analyzed this can wreck performance _(See below for an example)_

**New Behavior:**

1. Attempt to resolve a name in an expression
2. If resolution fails try to solve it by looking at all previous nodes
3. If all previous adjacent nodes fails to resolve the item then we will look at the parent and repeat the process
    * This is maintains behavior but only looks are the parent once, saving N wasted calls 

## Validity

image

The `javaparser-symbol-solver-testing` classes all passed.

image

The speed improvement is roughly x235 faster in the demo. It will be more noticeable the more adjacent nodes there are. You can open the test class and fiddle with the parameters to see how the effect scales when you revert the change in `StatementContext`.


Col-E:
Poking 👈 

Any feedback?

maartenc:
Looks good to me.
@matozoid @MysterAitch : any thoughts ?

matozoid:
Not my expertise I'm afraid. All the tests pass so nothing seems broken. I'm okay with it :-)

maartenc:
Merged!
Thanks :)

PR #3187, opened 2021-03-23
Fix issue #3173: Add isAnnotation() and asAnnotation() methods for ResolvedTypeDeclaration

### LINKED ISSUE #3173: Add isAnnotation() and asAnnotation() methods for ResolvedTypeDeclaration
This is a suggestion rather than an issue but it is a completion feature and should be very easy to add.
Add `isAnnotation()` and `asAnnotation()` methods for `ResolvedTypeDeclaration` and implement them accordingly in `ResolvedAnnotationDeclaration`.
In `ResolvedTypeDeclaration`:
```java
/**
 * Is this the declaration of an annotation?
 */
default boolean isAnnotation() {
    return false;
}

/**
 * Return this as a AnnotationDeclaration or throw UnsupportedOperationException.
 */
default ResolvedAnnotationDeclaration asAnnotation() {
    throw new UnsupportedOperationException(String.format("%s is not an annotation", this));
}
```
In `ResolvedAnnotationDeclaration`:
```java
/**
 * This method should always return true.
 */
@Override
default boolean isAnnotation() {
    return true;
}

@Override
default ResolvedAnnotationDeclaration asAnnotation() {
    return this;
}
```
I can submit a short PR if you want.
Thanks.
By the way, I have 2 submitted issues which weren't solved yet, one with a PR and is pretty straight forward, Any feedback?

deadlocklogic:
Fix issue #3173: Add isAnnotation() and asAnnotation() methods for ResolvedTypeDeclaration.

Fixes #3173.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java
@@ -31,5 +31,15 @@ import java.util.List;
 public interface ResolvedAnnotationDeclaration extends ResolvedReferenceTypeDeclaration,
         AssociableToAST<AnnotationDeclaration> {
 
+    @Override
+    default boolean isAnnotation() {
+        return true;
+    }
+
+    @Override
+    default ResolvedAnnotationDeclaration asAnnotation() {
+        return this;
+    }
+
     List<ResolvedAnnotationMemberDeclaration> getAnnotationMembers();
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java
@@ -97,6 +97,13 @@ public interface ResolvedTypeDeclaration extends ResolvedDeclaration {
     }
 
     /**
+     * Is this the declaration of an annotation?
+     */
+    default boolean isAnnotation() {
+        return false;
+    }
+
+    /**
      * Is this the declaration of a type parameter?
      */
     default boolean isTypeParameter() {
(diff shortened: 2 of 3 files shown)


deadlocklogic:
@MysterAitch  Any ideas why `codecov/patch` fails?
I added a test which covers the added methods.

MysterAitch:
> @MysterAitch  Any ideas why `codecov/patch` fails?
> I added a test which covers the added methods.

I suspect it might be that default methods aren't being executed, instead it's the overrides within child classes that are running.

Don't worry too much about patch coverage -- it would be ideal, but it won't block the PR being merged.


Not sure how much is visible to you but if you click on it, here's the detail about specific lines from clicking through to it:


```
 Check warning on line 36 in javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java

￼codecov/ codecov/patch

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java#L36

Added line #L36 was not covered by tests

 Check warning on line 41 in javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java

￼codecov/ codecov/patch

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedAnnotationDeclaration.java#L41

Added line #L41 was not covered by tests

 Check warning on line 103 in javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java

￼codecov/ codecov/patch

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java#L103

Added line #L103 was not covered by tests

 Check warning on line 171 in javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java

￼codecov/ codecov/patch

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java#L171

Added line #L171 was not covered by tests

```

deadlocklogic:
Yes, I can see the details too of course, but it is showing that even the implemented methods aren't covered which is not the case for sure. Anyway, I won't bother myself with this too much. I will try to fix other issues if I can instead.

MysterAitch:
I've just had a closer look and running it through a debugger with breakpoints in each method, both of them within `ResolvedAnnotationDeclaration` gets run.

It might be an issue with codecov or perhaps with the maven surefire plugin. I'll open up a separate issue and get this one merged in :+1:

PR #3188, opened 2021-03-23
Memory optimization for JarTypeSolver (Up to 42% less memory)

4everTheOne:
Currently JarTypeSolver consumes a lot of memory. In this PR suggests an alternative way to get the same behavior in a lighter way with a comparison between both methodologies.

The ideia follows a simple principle, that our `ClassPool` contains ONLY all the types declared in the Jar.
Note that JRE type are not included, if the user want to resolve JRE types he needs to include a ReflectionTypeSolver (Previously the JarTypeSolver also included JRE into the class pool). To check if a type is declared we just need to check if that qualified name is present in the `ClassPool`. This logic is very similar how `ClassLoaderTypeSolver` works.

**Why was field `classpathElements` removed?**

This map was used to keep the reference between the qualified name and the corresponding declaration. Now this functionality was delegated to `ClassPool` to reduce the complexity.

**Why was the class `ClasspathElement` removed?**

This is related to the previous question. Since we have delegated this task to `ClassPool` we don't need this anymore.

**Why was class `ResourceRegistry` removed?**

The `ResourceRegistry` was used to free the resources when the VM finish. Once again, we no longer need this, since we have delegated it to `ClassPool`.

**So if we don't have JRE types here, how does it resolve those types?**

When we have a Javassist declaration we should resolve the fully qualified name using the typesolver instead of relying on javassist itself. For more details see: https://github.com/javaparser/javaparser/issues/2762#issuecomment-798793403

#### Evaluation

To evaluate the algorithms was used the Unit test below, to generate a memory dump to compare the memory usage of each methodology.

```java
@Test
void testMemoryStress() throws IOException {
  
   Path pathToJar = adaptPath("src/test/resources/javaparser-core-2.1.0.jar");
   for (int i = 0 ; i < 10000 ; i++) {
       combinedTypeSolver.add(new JarTypeSolver(pathToJar));
   }
   
}
```

| Name                                                     |Total Memory Usage |
|:---------------------------------------------------------|------------------:|
| Current methodology                                      |          1 037 MB |
| Proposed methodology                                     |            416 MB |
| **Total:**                                               |       **-621 MB** |

As we can see from the results the memory consumption is reduced in approximately **60%**.

The report can be analysed at [Methodology comparison.pdf](https://github.com/javaparser/javaparser/files/6191046/Methodology.comparison.pdf)

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
@@ -21,10 +21,6 @@
 
 package com.github.javaparser.symbolsolver.javassistmodel;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.function.Function;
-
 import com.github.javaparser.ast.expr.BooleanLiteralExpr;
 import com.github.javaparser.ast.expr.CharLiteralExpr;
 import com.github.javaparser.ast.expr.DoubleLiteralExpr;
@@ -34,16 +30,14 @@ import com.github.javaparser.ast.expr.LongLiteralExpr;
 import com.github.javaparser.ast.expr.StringLiteralExpr;
 import com.github.javaparser.resolution.declarations.ResolvedAnnotationMemberDeclaration;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
-import com.github.javaparser.resolution.types.ResolvedPrimitiveType;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
 import com.github.javaparser.symbolsolver.model.typesystem.ReferenceTypeImpl;
-
-import javassist.CtClass;
 import javassist.CtMethod;
-import javassist.NotFoundException;
 import javassist.bytecode.AnnotationDefaultAttribute;
+import javassist.bytecode.BadBytecode;
+import javassist.bytecode.SignatureAttribute;
 import javassist.bytecode.annotation.BooleanMemberValue;
 import javassist.bytecode.annotation.CharMemberValue;
 import javassist.bytecode.annotation.DoubleMemberValue;
@@ -51,6 +45,10 @@ import javassist.bytecode.annotation.IntegerMemberValue;
 import javassist.bytecode.annotation.LongMemberValue;
 import javassist.bytecode.annotation.MemberValue;
 import javassist.bytecode.annotation.StringMemberValue;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
 
 /**
  * @author Malte Skoruppa
@@ -88,16 +86,15 @@ public class JavassistAnnotationMemberDeclaration implements ResolvedAnnotationM
     @Override
     public ResolvedType getType() {
         try {
-            CtClass returnType = annotationMember.getReturnType();
-            if (returnType.isPrimitive()) {
-                return ResolvedPrimitiveType.byName(returnType.getName());
+            String descriptor = annotationMember.getMethodInfo().getDescriptor();
+            SignatureAttribute.MethodSignature signature = SignatureAttribute.toMethodSignature(descriptor);
+            SymbolReference<ResolvedReferenceTypeDeclaration> returnType = typeSolver.tryToSolveType(signature.getReturnType().jvmTypeName());
+            if (returnType.isSolved()) {
+                return new ReferenceTypeImpl(returnType.getCorrespondingDeclaration(), typeSolver);
             }
-            SymbolReference<ResolvedReferenceTypeDeclaration> rrtd = typeSolver.tryToSolveType(returnType.getName());
-            if (rrtd.isSolved()) {
-                return new ReferenceTypeImpl(rrtd.getCorrespondingDeclaration(), typeSolver);
-            }
-        } catch (NotFoundException e) {
-            // nothing to do
+        } catch (BadBytecode e) {
+            // We don't expect this to happen, but we handle it anyway.
+            throw new IllegalStateException("An invalid descriptor was received from JavaAssist.", e);
         }
         throw new UnsupportedOperationException(String.format("Obtaining the type of the annotation member %s is not supported yet.", annotationMember.getLongName()));
     }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
@@ -276,7 +276,7 @@ public class JarTypeSolver implements TypeSolver {
             // The names in stored key should always be resolved.
             // But if for some reason this happen, the user is notified.
             throw new IllegalStateException(String.format(
-                    "Unable to get class witj name %s from class pool." +
+                    "Unable to get class with name %s from class pool." +
                     "This was not suppose to happen, please report at https://github.com/javaparser/javaparser/issues",
                     storedKey));
         }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
@@ -21,31 +21,23 @@
 
 package com.github.javaparser.symbolsolver.resolution.typesolvers;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.jar.JarEntry;
-import java.util.jar.JarFile;
-
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
 import com.github.javaparser.symbolsolver.javassistmodel.JavassistFactory;
 import com.github.javaparser.symbolsolver.model.resolution.SymbolReference;
 import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
-import com.github.javaparser.utils.Log;
-
 import javassist.ClassPool;
-import javassist.CtClass;
 import javassist.NotFoundException;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Path;
+import java.util.Objects;
+import java.util.Optional;
 
 /**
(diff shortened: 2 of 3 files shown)


jlerbsc:
It's very interesting.

maartenc:
Very nice indeed.

One thing I'm a bit worried about is the performance loss. Everytime the JarTypeSolver tries to solve a type, it has to also try the parent names as well (because of the inner classes). In the previous version, the JarTypeSolver knows immediately if the requested class was available in the jar or not. So perhaps, the JarTypeSolver should still know which classes exist in the jar so it can fail fast if a requested type is not present in the jar.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
```diff
@@ -88,16 +86,15 @@ public Expression getDefaultValue() {
...
-                return ResolvedPrimitiveType.byName(returnType.getName());
-            }
-            SymbolReference<ResolvedReferenceTypeDeclaration> rrtd = typeSolver.tryToSolveType(returnType.getName());
-            if (rrtd.isSolved()) {
-                return new ReferenceTypeImpl(rrtd.getCorrespondingDeclaration(), typeSolver);
+            String descriptor = annotationMember.getMethodInfo().getDescriptor();
+            SignatureAttribute.MethodSignature signature = SignatureAttribute.toMethodSignature(descriptor);
+            SymbolReference<ResolvedReferenceTypeDeclaration> returnType = typeSolver.tryToSolveType(signature.getReturnType().jvmTypeName());
+            if (returnType.isSolved()) {
+                return new ReferenceTypeImpl(returnType.getCorrespondingDeclaration(), typeSolver);
```
maartenc, line 93: I think this change has nothing to do with this PR.
Perhaps you should create a new PR for it?

4everTheOne: I had to tweak the JavassistAnnotationMemberDeclaration to support the change, otherwise it will not resolve JRE types.
But if you prefer i can move this fix to another PR.

4everTheOne:
> One thing I'm a bit worried about is the performance loss. Everytime the JarTypeSolver tries to solve a type, it has to also try the parent names as well (because of the inner classes). In the previous version, the JarTypeSolver knows immediately if the requested class was available in the jar or not.

Interesting point @maartenc. I'll think on solution for that :)

4everTheOne:
This commit changes the memory reduction to **42%** instead of the previous 60% making the new results look like:
| Name                                                     |Total Memory Usage |
|:---------------------------------------------------------|------------------:|
| Current methodology                                      |          1 037 MB |
| Proposed methodology                                     |            582 MB |
| **Total:**                                               |       **-455 MB** |

This change adds the verification suggested by @maartenc before solving the type. This verification makes sure the type belongs to this Jar before executing the expensive operation.

The new report can be seen at [New comparison.pdf](https://github.com/javaparser/javaparser/files/6192073/New.comparison.pdf)

MysterAitch:
> > One thing I'm a bit worried about is the performance loss. Everytime the JarTypeSolver tries to solve a type, it has to also try the parent names as well (because of the inner classes). In the previous version, the JarTypeSolver knows immediately if the requested class was available in the jar or not.
> 
> Interesting point @maartenc. I'll think on solution for that :)

Possibly worth a separate issue/PR but one of the things suggested in the past is maintaining a symbol reference table of some kind. This would help to prevent/limit the performance problem of repeated solves/lookups.

Whether this is done proactively or more as a lazy caching system would perhaps be a useful configuration option to include.

4everTheOne:
> Possibly worth a separate issue/PR but one of the things suggested in the past is maintaining a symbol reference table of some kind. This would help to prevent/limit the performance problem of repeated solves/lookups.

I think this could also be integrated here very easily. Since JP SS has Guava has dependency, we can receive a CacheBuilder in constructor which gives us the configuration to be used.

My question is, which configuration should be set by default? Disabled cache?

MysterAitch:
Before running with a cache, can we get some numbers on whether it is needed (i.e. will this PR actually have a substantive performance impact)?

MysterAitch:
Sidenote re: the mention of guava -- I should note that it's on my to-do list to see if we can remove guava as a dependency, given that iirc it is used only in a couple of places within the jss (it's also the only non-test/non-build dependency we have at the moment). 

Whether it _should_ be kept or removed isn't something I personally have a strong opinion of, but I do believe that if we do promote the project as having no runtime dependencies then we should actually follow through and have zero runtime dependencies.

If we do keep it, then let's go the other way and make the most of having it as a dependency and see where else it might be good to use it! :+1:

Again, I know I'm the one bringing it up (hypocrite! :grinning:) but the discussion around this is probably best had in a separate issue.

4everTheOne:
> Before running with a cache, can we get some numbers on whether it is needed (i.e. will this PR actually have a substantive performance impact)?

I will create a performance test tomorrows to compare it :)

maartenc:
Thanks for the update @4everTheOne 

I think the code could be simplified more if you used a HashMap instead of a HashSet. The key in the map is the class name (wich is now contained in the HashSet) and the value is the className with the $ still in it's name.

For instance, the map entries could look like:
org.javaparser.MyClass -> org.javaparser.MyClass
org.javaparser.MyOtherClass -> org.javaparser.MyOtherClass
org.javaparser.MyOtherClass.InnerClass -> org.javaparser.MyOtherClass$InnerClass

Now, when we try to resolve a type, we can look into that map to get the correct classname to load. No more need to try to load a parent class and search for inner types in it.

The memory consumption of a HashSet and HashMap are the same if I'm not mistaken, especially if you take care that for normal classes the key and value point to the same String instance.

Could you give that a try and see how it performs?

maartenc:
> > Possibly worth a separate issue/PR but one of the things suggested in the past is maintaining a symbol reference table of some kind. This would help to prevent/limit the performance problem of repeated solves/lookups.
> 
> I think this could also be integrated here very easily. Since JP SS has Guava has dependency, we can receive a CacheBuilder in constructor which gives us the configuration to be used.
> 
> My question is, which configuration should be set by default? Disabled cache?

I think we should do this in a separate issue to keep this PR focussed. I myself use JavaParser on some pretty large projects (using 500+ jars to solve dependencies). I added such a cache in the CombinedTypeSolver to speed up lookups. So perhaps you could try to do the same in that other PR?

4everTheOne:
Evaluating all of the suggested solutions with the memory usage and the time required to solve the type the results are:
image

In time to solve, i would take in consideration a margin of error of 10ms due to uncontrollable variables. Like @maartenc has said, we could use the HashMap to store the map between the qualified name used in JP and the qualified name for the ClassPool to avoid looking into the parent qualified name. This does not impact the memory consumption at all and may help in some situations.

**Evaluation method**
To evaluate all the methodologies the following test was used: 
```java
@Test
void testMemoryStress() throws IOException {

    List<String> typesToSolve = Arrays.asList(
            "com.github.javaparser.ast.Node",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.NodeList",
            "com.github.javaparser.ast.visitor.GenericVisitor",
            "com.github.javaparser.ast.visitor.HashCodeVisitor",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode",
            "com.github.javaparser.ast.Node.ObserverRegistrationMode"
    );

    Path pathToJar = adaptPath("src/test/resources/javaparser-core-2.1.0.jar");
    for (int i = 0 ; i < 10000 ; i++) {
        combinedTypeSolver.add(new JarTypeSolver(pathToJar));
    }

    for (String type : typesToSolve) {
        combinedTypeSolver.tryToSolveType(type);
    }

}
```
To measure the time to solve was only take into consideration the time it took to run ALL the JarTypeSolver#tryToSolveType(); (Time it took to solve Node + time to take it solves ObserverRegistrationMode + ... = Total time to solve)

One thing to note is that ClassPool has a builtin cache to speed up the process of solving types. This explains why in "without storing known types" the memory usage was bigger then the others. This happens because ClassPool cached additional types while solving.

**Notes on the implementation of HashSet**
To keep the memory low, I used an HashMap<Integer,String> where the keys are the hash code of the qualified name and the value the qualified name to be used in ClassPool. This allows us to keep the same memory footprint as the HashSet, while if we would use HashMap<String,String> the memory usage would increase to +-700MB without performance improvement.

4everTheOne:
Also, it's interestingly where it takes more time in the HashMap methodology is calling JavassistFactory#toTypeDeclaration. This can be decreased by +-72ms with the cache we were talking about,.

Sorting the methods by the amount of time they take (slowest -> faster):
- [41.383 ms] JavassistFactory.ToTypeDeclaration;
- [33.649 ms] java.util.Map.get
- [30.388 ms] SymbolReference.Unsolved
- [9.55 ms] ClassPool.get

maartenc:
> To keep the memory low, I used an HashMap<Integer,String> where the keys are the hash code of the qualified name and the value the qualified name to be used in ClassPool. This allows us to keep the same memory footprint as the HashSet, while if we would use HashMap<String,String> the memory usage would increase to +-700MB without performance improvement.

I'm not sure this is a good idea since it's possible that 2 different qualified names have the same hash.
But I don't understand why a HashMap<String,String> would take more memory. Did you put the same String instance in the map when key and value are equal?

4everTheOne:
> Did you put the same String instance in the map when key and value are equal?

That was it, i forgot to add a check if they were equals, duplicating the Strings when not needed!

Adding that additional check the results for the `HashMap` are: 676 MB (+63 MB) / 153 ms (Inside margin error)

maartenc:
Thanks!
For what it's worth: I tried your test with some different scenario's and looked at the memory usage with JProfiler:

original code: 878MB
Map<Integer,String> code: 531MB
Map<String,String> code: 528MB

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
```diff
@@ -191,27 +261,24 @@ public void setParent(TypeSolver parent) {
...
-                return SymbolReference.unsolved(ResolvedReferenceTypeDeclaration.class);
-            }
-        } catch (IOException e) {
-            throw new RuntimeException(e);
+            return SymbolReference.solved(JavassistFactory.toTypeDeclaration(classPool.get(storedKey), getRoot()));
+        } catch (NotFoundException e) {
+            // The names in stored key should always be resolved.
+            // But if for some reason this happen, the user is notified.
+            throw new IllegalStateException(String.format(
+                    "Unable to get class witj name %s from class pool." +
```
maartenc: typo here...

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistAnnotationMemberDeclaration.java
maartenc: ok no problem to leave it here in this PR.

maartenc:
This PR looks good to me.
What do you think @MysterAitch ?

MysterAitch:
> This PR looks good to me.
> What do you think @MysterAitch ?

I think you're all too clever for me ;)


Reading through the diff online, I see no obvious problems and it sounds like there will be benefits (memory usage comes up often enough).

I'm up for it being merged in whenever you think it's ready! :+1:

PR #3197, opened 2021-03-25
Implemented isAssignableBy for VoidType - now return `false` instead of throwing `UnsupportedOperationException`

4everTheOne:
According to https://docs.oracle.com/javase/specs/jls/se16/html/jls-14.html#jls-14.8:

> Note that the Java programming language does not allow a "cast to void" - void is not a type - so the
> traditional C trick of writing an expression statement such as:
>
> (void)... ;  // incorrect!
>
> does not work.

In short, nothing can be assign to "void".


MysterAitch (approved):

MysterAitch:
Looks good to me.

PR #3206, opened 2021-03-27
Implemented isReferenceType in `ResolvedTypeDeclaration` and isTypeParameter in `ResolvedTypeParameterDeclaration`

4everTheOne:
`ResolvedTypeDeclaration` was missing a method isReferenceType and `ResolvedTypeParameterDeclaration` did not implement the methods isTypeParameter and asTypeParameter.


javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
```diff
@@ -47,6 +47,11 @@ default ResolvedReferenceTypeDeclaration asReferenceType() {
         return this;
     }
 
+    @Override
+    default boolean isReferenceType() {
+        return true;
```
jlerbsc, line 49: I'm not sure a declaration can be considered as a type.

4everTheOne: At least, the `ResolvedReferenceType` and `ResolvedType` implement the method "asReferenceType".
https://github.com/javaparser/javaparser/blob/dfbd466829fca9047b995b5bf9a28343b81d7c5a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java#L45-L53

jlerbsc:

PR #3209, opened 2021-03-27
Improve tests on `getAllAncestors` method

jlerbsc:


jlerbsc:
@MysterAitch Can you tell me why there is a compilation error on StringProvider.

PR #3213, opened 2021-03-30
[Fixes #3099] Added ability to solve type with a list of expected type arguments

### LINKED ISSUE #3099: IllegalArgumentException when solving ClassOrInterfaceType
JP Symbol Solver throws a `IllegalArgumentException` when we have a type definition that matches the name of a implemented type.

#### File structure:
```
.
└── com/example
    ├── Alpha.java
    └── Beta.java
```
#### Files content
Alpha.java
```java
package com.example;

interface Alpha {
  interface CustomInterface {}
}
```

Beta.java
```java
package com.example;

class Beta implements Alpha.CustomInterface {
  
  public interface CustomInterface<T> {}

  private final CustomInterface<Object> instanceOfBetaInnerClass;

}
```

### Explanation
The solver solves `CustomInterface` as `Alpha.CustomInterface` instead of  `Beta.CustomInterface`.

### Possible fix
I think this issue is related with the logic in method `ClassOrInterfaceDeclarationContext#solveType(String)` that gives priority to implemented and extended types over the local logic. I think this could be fixed by prioritizing the local search over the others.
https://github.com/javaparser/javaparser/blob/be4e455f228eaf6500e6aa47c8472b708826ee27/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContext.java#L108-L144

### Stack trace
```
java.lang.IllegalArgumentException: expected either zero type arguments or has many as defined in the declaration (0). Found 1

	at com.github.javaparser.resolution.types.ResolvedReferenceType.<init>(ResolvedReferenceType.java:78)
	at com.github.javaparser.symbolsolver.model.typesystem.ReferenceTypeImpl.<init>(ReferenceTypeImpl.java:76)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:646)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.convertToUsage(JavaParserFacade.java:606)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.toResolvedType(JavaSymbolSolver.java:281)
	at com.github.javaparser.ast.type.ClassOrInterfaceType.resolve(ClassOrInterfaceType.java:314)
```
### Unit test
```java
@Test
void illegalArgumentExceptionWhenSolvingName() throws IOException {

  // Setup symbol solver
  JavaParserTypeSolver javaParserTypeSolver = new JavaParserTypeSolver(adaptPath("src/test/resources/issueXXX/"));
  StaticJavaParser.getConfiguration()
		  .setSymbolResolver(new JavaSymbolSolver(
				  new CombinedTypeSolver(new ReflectionTypeSolver(), javaParserTypeSolver))
		  );
  
  // Parse the File
  Path filePath = adaptPath("src/test/resources/issueXXX/com/example/Beta.java");
  CompilationUnit cu = StaticJavaParser.parse(filePath);
  
  // Get the expected inner class
  List<ClassOrInterfaceDeclaration> classes = cu.findAll(ClassOrInterfaceDeclaration.class);
  assertEquals(2, classes.size());
  ResolvedReferenceTypeDeclaration innerInterface = classes.get(1).resolve();
  assertTrue(innerInterface.isInterface());
  
  // Check if the value is present
  Optional<ResolvedReferenceType> resolvedType = cu.findFirst(VariableDeclarator.class)
		  .map(VariableDeclarator::getType)
		  .map(Type::resolve)
		  .filter(ResolvedType::isReferenceType)
		  .map(ResolvedType::asReferenceType);
  assertTrue(resolvedType.isPresent());
  assertEquals(innerInterface, resolvedType.get().getTypeDeclaration().orElse(null));
}
```

4everTheOne:
Currently, when JP SS is trying to solve a type, it doesn't take into consideration the number of type arguments it expects to find. This have been an issue that was reported in #3099 where we can have two different classes/interfaces with the same name. In this scenario JP SS will return the first one it finds, and not the one that fits the expected type.

This PRs adds a new method in context to solve a type with the expected type arguments. If the type arguments is provided it will only find a type where the type arguments matchs the expected.

#### Examples
Lets imagine we have two declarations of `Foo`. We have com `Alpha.Foo` and `Beta.Foo<T>` where `Foo` is a inner class of `Alpha ` and `Foo<T>` is a inner class of `Beta`. If we want to solve `Alpha.Foo` without arguments we can solve it like:
```java
context.solveType("Foo", Collections.emptyList());
```
if we want to resolve it as `Beta.Foo<T>`  we execute:
```java
List<ResolvedType> expectedTypeArguments = ...;
context.solveType("Foo", expectedTypeArguments);
```

If we just want it to be resolved and don't care how many arguments it has we can use:
```java
context.solveType("Foo"); // or context.solveType("Foo", null);
```

#### Implementation notes
- The logic that was previously in `ClassOrInterfaceDeclarationContext` was moved to `JavaParserTypeDeclarationAdapter`, because `AnnotationDeclarationContext` should also follow the same rules;
- `JavaParserTypeDeclarationAdapter` was reordered to match the expected behavior (the same order that was previously in `ClassOrInterfaceDeclarationContext`).

#### Related Issues
- Fixes #3099 by checking if the type parameters match the expected.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
@@ -22,12 +22,15 @@
 package com.github.javaparser.symbolsolver.javaparsermodel.contexts;
 
 import com.github.javaparser.ast.AccessSpecifier;
+import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.BodyDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
 import com.github.javaparser.ast.nodeTypes.NodeWithExtends;
 import com.github.javaparser.ast.nodeTypes.NodeWithImplements;
+import com.github.javaparser.ast.nodeTypes.NodeWithTypeArguments;
 import com.github.javaparser.ast.nodeTypes.NodeWithTypeParameters;
 import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.TypeParameter;
 import com.github.javaparser.resolution.declarations.*;
 import com.github.javaparser.resolution.types.ResolvedReferenceType;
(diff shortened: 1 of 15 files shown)


4everTheOne:
This PR is not ready to be merged. This needs more tests to make sure the intended behavior is covered.

4everTheOne:
Ready to be reviewed! :)

jlerbsc:
Can you reorganize your commits?

4everTheOne:
Updated commit history to simplify reading

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
```diff
@@ -65,53 +68,63 @@ public JavaParserTypeDeclarationAdapter(com.github.javaparser.ast.body.TypeDecla
...
                 if (implementedType.getName().getId().equals(name)) {
                     return context.getParent()
                         .orElseThrow(() -> new RuntimeException("Parent context unexpectedly empty."))
-                        .solveType(implementedType.getNameWithScope());
+                        .solveType(implementedType.getNameWithScope(), typeArguments);
                 }
             }
         }
 
+        // Check if the node implements other types
```
jlerbsc: Probably "Check if the node extends other types"

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    void testSolveWithoutTypeArguments() {
+        CompilationUnit alphaCU = parse("class Alpha { class Foo {} }");
+        ClassOrInterfaceDeclaration alpha = Navigator.demandClass(alphaCU, "Alpha");
+        ClassOrInterfaceDeclarationContext alphaContext = new ClassOrInterfaceDeclarationContext(alpha, typeSolver);
+
+        assertTrue(alphaContext.solveType("Foo").isSolved());
+        assertTrue(alphaContext.solveType("Foo", Collections.emptyList()).isSolved());
+        assertFalse(alphaContext.solveType("Foo", Collections.singletonList(ResolvedPrimitiveType.INT)).isSolved());
```
jlerbsc, line 61: Why is the type "Foo" resolved in this context?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    void testSolveWithTypeArguments() {
+        CompilationUnit betaCU = parse("class Beta { class Foo<T> {} }");
+        ClassOrInterfaceDeclaration beta = Navigator.demandClassOrInterface(betaCU, "Beta");
+        ClassOrInterfaceDeclarationContext betaContext = new ClassOrInterfaceDeclarationContext(beta, typeSolver);
+
+        assertTrue(betaContext.solveType("Foo").isSolved());
+        assertFalse(betaContext.solveType("Foo", Collections.emptyList()).isSolved());
```
jlerbsc, line 71: I'm not sure if this is the expected result?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
```diff
@@ -0,0 +1,79 @@
...
+        assertTrue(betaContext.solveType("Foo", Collections.singletonList(ResolvedPrimitiveType.INT)).isSolved());
```
jlerbsc, line 72: I'm not sure if this is the expected result?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
4everTheOne: In this assertions, there are three things being confirmed:

`assertTrue(alphaContext.solveType("Foo").isSolved());`
We are making sure, that we are retro-compatible, so if someones is using the API without type parameters, the type is solved using the previous rules.

`assertTrue(alphaContext.solveType("Foo", Collections.emptyList()).isSolved());`
Looking at the class definition, the class Foo does not have type parameters. In this test, we make surethat trying to resolve it with a empty list it's resolved. Note that a empty list in type arguments is the same as saying it does not have type parameter.

`assertFalse(alphaContext.solveType("Foo", Collections.singletonList(ResolvedPrimitiveType.INT)).isSolved());`
Test to make sure, if we try to resolve Foo with a type parameter, it's not resolved since the declaration says it does not have one.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
4everTheOne: Considering now that Foo has a type argument (T), this test should Fail.
Since this assertion, is trying to solve a type "Foo" that does not have any type parameters.

This would be equivalent of doing this in IDE, which is not possible:
```java
class Foo<T> {}

Foo<> fooObject;
```

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/ClassOrInterfaceDeclarationContextTest.java
4everTheOne: Considering now that Foo has a type argument (T), this test should be valid, since this assertion, is trying to solve a type "Foo" that does have a type parameters of integer type.

This would be equivalent of doing this in IDE, which is possible:
```java
class Foo<T> {}

Foo<Integer> fooObject;
```

jlerbsc:

PR #3221, opened 2021-04-05
Fix Javadoc comment * escaping problem.

### LINKED ISSUE #3219: The stars have been mangled in the JavadocComment Javadoc
`&#42;` doesn't seem to be something Javadoc understands: https://javadoc.io/doc/com.github.javaparser/javaparser-core/latest/com/github/javaparser/ast/comments/JavadocComment.html

matozoid:
Fixes #3219. A rather silly fix for a silly problem. Replacing `<code>` with `{@code` was certainly the right choice, except for in this one place where `*/` was escaped so that it wouldn't end the Javadoc. But that used a HTML escape sequence for the `*`, and `{@code` doesn't seem to be able to escape anything.

The obvious solution is to revert this comment to use `<code>`, so of course I didn't do that and replaced the asterisk with other unicode asterisks that look more or less the same but aren't recognized by Java as asterisks. That makes the comment readable in the source code without any escapes, stays consistent with `{@code` usage, and introduces a tiny risk that someone will paste these unicode characters in their code and gets all confused.


MysterAitch:
I have no strong feelings about this, though do lean mildly towards encoding the slash as described here:

https://stackoverflow.com/a/24696014/429733
```java
/**
 * Returns true if the specified string contains "*{@literal /}".
 */
```

Or perhaps:
https://stackoverflow.com/a/631839/429733

```java
/**
 * Returns true if the specified string contains "*&#47;".
 */
// returns true if the specified string contains "*/"
public boolean containsSpecialSequence(String str)
```

or 
```java
/**
 * Returns true if the specified string contains "*&#47;".
 */
public boolean containsSpecialSequence(String str)
```

matozoid:
@MysterAitch @jlerbsc - please pick an option and I'll update the PR :-)

jlerbsc:
Personally I think that &#42; is not really readable. So whatever the solution, we have to keep the code readable. I'm not sure this will advance the 'Schmilblick' (in french Schmilblick~subject )

matozoid:
@jlerbsc did you type an escape code and did it change into a *? :-D

jlerbsc:
No, it's just an opinion.

matozoid:
Guys, this is a really small PR, just tell me how you want it to be and I'll make it so :-)

matozoid:
Good to see you @MysterAitch !

Issue #2714, opened 2020-06-04
ADRs - Architecture Design Records


MysterAitch:
ADR - Architecture Design Records

@matozoid talked about these and they do seem like a really good idea!

Here are some ideas off the top of my head:

- [ ] Metamodel - WIP #2710
- [ ] Generators for the metamodel - WIP #2710
- [ ] Naming conventions (e.g. `Statement` when used on its own, and abbreviated when part of a longer name such as `BreakStmt`)
- [ ] Places where we have deviated from the Java Language Specification (JLS) (e.g. `NodeList`) or where our grammar differs (I can't think of any right now, but I'm sure there are some cases of this where we favour making this library usable and using JLS naming wherever possible, rather than sticking religiously to the JLS)
- [ ] How NodeList works/why (e.g. there is interesting (good) logic re: parents)
- [ ] The API of `StaticJavaParser` differing from `JavaParser`
- [ ] Favouring / embracing mutability rather than immutability
- [ ] The generators -- core generators vs metamodel generators etc (additional info / recent discussion in #2705)
- [ ] Avoiding methods ever returning null -- instead favouring optional

Each of these probably warrants a separate issue with an ADR label, but to get the ball rolling let's stick to the one issue for now (at least until the first PR is merged).

PR #3229, opened 2021-04-15
Add checkstyle

MysterAitch:
This PR adds checkstyle to the build, to help minimise the amount of churn of e.g. import statements. The config is adapted from the default Sun style, with changes made to match our current codebase.

- Where I have been able to get a "quick win" to adopt a consistent style then that is included within this PR (e.g. adding comments to `public static final` fields and making them `UPPER_CASE`, tweaking whitespace around operators, tweaking imports).

- Where it's not feasible to apply the rule yet or the rule is in default sun style but might not be relevant / applicable to our project, I have added it to the suppressions (exceptions) file. Preferably these suppressions are limited to the relevant files, but there are some rules added as a global supression.

It's a TODO for future PRs to investigate whether the remaining "suppressed" checks are applicable or if they definitely don't apply / we don't care. 

For example, there are lots of warnings about magic numbers where we use `31` in the `hashCode()` methods - maybe one day we will convert them all to use `Objects.hashCode(...)` but perhaps not right now and certainly not within this PR. Ditto for ensuring there is javadoc for all public methods/classes - this is something we probably should be doing, but there are too many places we don't have this that it's a substantial piece of work (assuming we want to do it - again a separate discussion / PR :+1: ).


MysterAitch:
I suspect the `avoidStarImports` rule might be a potentially controversial one.

Personally I have a mild preference to avoid start imports simply because 
- it is objective (otherwise, where is the cut-off where we want star imports?) and
- diffs will show when/where uses of classes are added/removed (e.g. when introducing new AST nodes)

...but this is a weakly held view -- any advocates for star imports are welcome to comment and we can roll that bit back! :+1: :smile:

koppor:
Nitpick 😇: There should be an ADR (refs #2714) on using custom checkstyle rules versus [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). Besides the [two spaces rule](https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation), I personally like it, because there is good tooling: [Google Java Format](https://github.com/google/google-java-format): Formats the code, ...

PR #3233, opened 2021-04-22
Optimizations for Node class (CPU time and Memory usage)

4everTheOne:
# Why replacing `LinkedList` with `ArrayList`?

Lets take a look at the table below which compares `LinkedList` with the `ArrayList`.

**Comparison**
| Operation             | LinkedList time complexity | ArrayList time complexity                              | Preferred  |
|-----------------------|----------------------------|--------------------------------------------------------|------------|
| Insert at last index  | O(1)                       | O(1) (If array copy operation is Considered then O(N)) | LinkedList |
| Insert at given index | O(N)                       | O(N)                                                   | LinkedList |
| Search by value       | O(N)                       | O(N)                                                   | ArrayList  |
| Get by index          | O(N)                       | O(1)                                                   | ArrayList  |
| Remove by value       | O(N)                       | O(N)                                                   | LinkedList |
| Remove by index       | O(N)                       | O(N)                                                   | LinkedList |

Source: [Performance Analysis of ArrayList and LinkedList in Java by Anant Mishra](https://dzone.com/articles/performance-analysis-of-arraylist-and-linkedlist-i)

There are some situations where is preferred to use the `LinkedList`. This happens when the list is updated more time than it's read.

## Benefits
In our case `LinkedList` isn't the best choice since the fields are read more time then they are written to (ex: `childNodes` is read every time someone calls `Node#findAll(...)`). In this scenario, the `ArrayList` fits better, since we can iterate it with a complexity of O(1) instead of O(N), reducing the CPU time required to execute the operation.

Another benefit of using `ArrayList` instead of `LinkedList` is the use of less memory overhead. ` LinkedList` requires additional information, pointing where the next node is in memory while the `ArrayList` uses an sequential Array to store all the values.

## Downsides
The insertions of nodes in a tree, now has a complexity of O(N) instead of the previous O(1).

# Why was `HashSet` replaced with `ArrayList`?
 The ideia behind of `Node#observers` is to keep record of the list of observers to be notified when something happens on the node. The problem with the use of the `HashSet` in this scenario it uses a lot of memory and it's functionality is not properly used. An `HashSet` it worth to be used when we need to use the HashSet#contains() that has complexity oif O(1). In this case we are not taking any advantage of this case property, since the method that is used more often is `HashSet#foreach`. In this case, we can replace the use of a `HashSet` with an `ArrayList` that is more CPU and memory friendly.

# Why was `ArrayList#trimToSize()` called after removing items from the list?
This call tries to reduce the memory footprint when an item is removed from a node. When the items is removed from the `ArrayList` the Array size is the same. If we imagine the Array has space allocated for 30 elements but only stores 2, we are wasting precious memory space. Calling this method makes sure that we are only using the space we really need. 

# Evaluation

The evaluation was made using the tools [VisualVM](https://visualvm.github.io/) and JavaParser Symbol Solver tests. In this evaluation, the execution time was reduced the execution time by +- 30 seconds.

**Current:**
0-Original

**Proposed:**
3-Final-Optimazation


maartenc:
Merged :-)
Thanks!

PR #3235, opened 2021-04-24
Fixes issue #3113 -- Arrow missing in Switch Expression + jumbled up in LexicalPreservingPrinter

### LINKED ISSUE #3113: Arrow missing in Switch Expression + jumbled up in LexicalPreservingPrinter
Hi,
I stumbled across a problem that is probably related to issue #2636 - I have the same problem with parsing Java 12 switch expressions, the arrows are replaced with colons. When the resulting CompilationUnit is later passed to the LexicalPreservingPrinter, the arrows are there, but the expression after "case" is appended to the end of the line.
Here's a minimal code example:
```kotlin
StaticJavaParser.getConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_12)

val originalSourceCode = "public class HelloWorld {\n" +
        "  public static void main(String[] args) {\n" +
        "      final int value = 2;\n" +
        "      String numericString;\n" +
        "    \n" +
        "      switch (value)\n" +
        "      {\n" +
        "        case 1 -> numericString = \"one\";\n" +
        "        case 2 -> numericString = \"two\";\n" +
        "        case 3 -> numericString = \"three\";\n" +
        "        default -> numericString = \"N/A\";\n" +
        "      }\n" +
        "    \n" +
        "      System.out.println(\"value:\" + value + \" as string: \" + numericString);\n" +
        "  }\n" +
        "}\n"

val originalCompilationUnit = StaticJavaParser.parse(originalSourceCode)
LexicalPreservingPrinter.setup(originalCompilationUnit)

// ...

val modifiedSourceCode = LexicalPreservingPrinter.print(originalCompilationUnit)
```

After executing, originalCompilationUnit looks like this...
```java
public class HelloWorld {

        final int value = 2;
        String numericString;
        switch(value) {
            case 1:
                numericString = "one";
            case 2:
                numericString = "two";
            case 3:
                numericString = "three";
            default:
                numericString = "N/A";
        }
        System.out.println("value:" + value + " as string: " + numericString);
    }
}

```

... and modifiedSourceCode like this:
```java
public class HelloWorld {
      final int value = 2;
      String numericString;
    
      switch (value) { 
        case  -> numericString = "one";1
        case  -> numericString = "two";2
        case  -> numericString = "three";3
        default -> numericString = "N/A";
      }
    
      System.out.println("value:" + value + " as string: " + numericString);
  }
}
```

I'm not sure if this is caused by the issue mentioned above or a problem on its own (I'm new to JavaParser). I'd be really grateful if someone could take a look at it and/or give some hints on a workaround :)

WanliZhong:
Fixes #3113.


MysterAitch:
Thanks @Zoom1111 :partying_face:

PR #3238, opened 2021-04-25
Issue 3064 conditional nested lambda

### LINKED ISSUE #3064: Failed to parse lambda expression
Parsing the following class:
```java
package foo.bar;

public class Fizz extends A {

   public Fizz(Properties properties, Buzz buzz) {
      this.table = buzz != null ? () -> buzz : properties.supp != null ? properties.supp : () -> new Fuzz(this.getA().getB(), "path/" + this.getA().getPath());
   }
}
```
generates a Problem: `(line 6,col 158) Failed to parse lambda expression! Please create an issue at https://github.com/javaparser/javaparser/issues`

Parsing code:
```java
package test;

import java.io.StringReader;

import com.github.javaparser.ParseResult;

public class Parsing {
	private static final String classCode = "package foo.bar;\r\n"
			+ "\r\n"
			+ "public class Fizz extends A {\r\n"
			+ "\r\n"
			+ "   public Fizz(Properties properties, Buzz buzz) {\r\n"
			+ "      this.table = buzz != null ? () -> buzz : properties.supp != null ? properties.supp : () -> new Fuzz(this.getA().getB(), \"path/\" + this.getA().getPath());\r\n"
			+ "   }\r\n"
			+ "}";
	
		JavaParser parser = new JavaParser();
		ParseResult<CompilationUnit> unitOpt = parser.parse(new StringReader(classCode));
		unitOpt.getProblems().stream().forEach(p -> System.err.println(p.toString()));
		CompilationUnit unit = unitOpt.getResult().orElseThrow(() -> new IllegalStateException("Could not parse file"));
		System.out.println(unit.toString());
	}
}
```

Printing the the parsed CompilationUnit outputs:
```java
package foo.bar;

public class Fizz extends A {

    public Fizz(Properties properties, Buzz buzz) {
        this.table = buzz != null ? () -> buzz : properties.supp != null ? properties.supp : () -> {
        };
    }
}
```
As you can see the last lambda is not parsed correctly.

Testet with `javaparser-core` version `3.18.0`

si-e:
Fixes #3064 .

Lambda is given a body only when Expression is reached.
When multiple ternary operators are nested, it is not good to check only one level.

Change the syntactic definition of ConditionalExpression.
At the same time, we can delete the original judgment in Exp.

Add Tests.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/Issue3064Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/Issue3064Test.java
@@ -1,0 +1,43 @@
+package com.github.javaparser;
+
+import com.github.javaparser.ast.CompilationUnit;
+import org.junit.jupiter.api.Test;
+
+import java.io.StringReader;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class Issue3064Test {
+
+    @Test
+    public void test0() {
+        String str = "import java.util.function.Supplier;\n" +
+                "\n" +
+                "public class MyClass {\n" +
+                "\n" +
+                "    public MyClass() {\n" +
+                "        Supplier<String> aStringSupplier = false ? () -> \"\" : true ? () -> \"\" : () -> \"path\";\n" +
+                "    }\n" +
+                "}\n";
+
+        JavaParser parser = new JavaParser();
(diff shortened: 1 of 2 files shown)


ramidzkh (approved):

matozoid:
Nice!

PR #3241, opened 2021-04-27
Issue 3200 this exp in anonymous class

### LINKED ISSUE #3200: The type is parsed in an anonymous internal class. The method calling with this is parsed. The parsing fails.
Consider this scenario

```java
public class Test {
    private void bad() {
        Test test = new Test();
        test.setRunnable("", new Runnable() {
            @Override
            public void run() {
                getContext(Test.this);
            }
        });
    }

    private void getContext(Test test) {
    }

    private void setRunnable(String str, Runnable runnable) {

    }
}
```
Invoking `resolve` on `getContext(Test.this);` I get the following error

```
UnsolvedSymbolException{context='null', name='We are unable to find the method declaration corresponding to getContext(Test.this)', cause='null'}
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:167)
	at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:317)
	at JavaParserStuck1.lambda$main$0(JavaParserStuck1.java:30)
	at java.util.ArrayList.forEach(ArrayList.java:1257)
	at JavaParserStuck1.main(JavaParserStuck1.java:28)

```

si-e:
Fixes #3200.

Problem: The main class "this" expression in the anonymous class does not get the type correctly as a function argument.

Reason: When parsing the type of "this" argument, it will loop through to get the parent node. If it is a declaration statement, it will determine whether the name of its type is the same as the name of type "this", but it will exit when the object creation statement is obtained.
So it exits in the anonymous class creation statement, resulting in retrieving the type of the anonymous class.
Personally, I don't think it's right.

Modification: Removes the code that gets the object creation statement to exit.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -628,17 +628,10 @@ public class JavaParserFacade {
      */
     protected Node findContainingTypeDeclOrObjectCreationExpr(Node node, String className) {
         Node parent = node;
-        boolean detachFlag = false;
         while (true) {
             parent = demandParentNode(parent);
             if (parent instanceof BodyDeclaration) {
                 if (parent instanceof TypeDeclaration && ((TypeDeclaration<?>) parent).getFullyQualifiedName().get().endsWith(className)) {
-                    return parent;
-                } else {
-                    detachFlag = true;
-                }
-            } else if (parent instanceof ObjectCreationExpr) {
-                if (detachFlag) {
                     return parent;
                 }
             }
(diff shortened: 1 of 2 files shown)


raghav-deepsource:
I think you are missing the point of this method.

It is trying to retrieve the first body declaration parent, which may be a type declaration or an anonymous class. Rather than deviating from the stated contract of this method, it would be better to replace the usage of this method with a different one which accounts for this particular edge case, or modify this method itself to do so.

si-e:
> I think you are missing the point of this method.
> 
> It is trying to retrieve the first body declaration parent, which may be a type declaration or an anonymous class. Rather than deviating from the stated contract of this method, it would be better to replace the usage of this method with a different one which accounts for this particular edge case, or modify this method itself to do so.

Yes, I agree with you.
I really feel bad about my rude revision.
I'll cancel this PR and try to modify the method call above it.

PR #3243, opened 2021-05-01
fix Log.error() throwing NullPointerException

CD4017BE:
Playing around with JavaParser, I noticed that Log.error(Throwable) always throws a NullPointerException with the StandardOutStandardErrorAdapter assigned because it expects a non null message supplier, but Log.error() would pass it null.

I fixed the problem by letting Log.error() pass a null-returning supplier instead of null, since the API javadoc says message suppliers are allowed to return null. According to the StandardOutStandardErrorAdapter implementation this should work.


jlerbsc:
It seems to be correct. Thank you for your contribution

PR #3247, opened 2021-05-05
Recover wrong statements no only to ';', but to '}' (not including) as well

32kda:
I've read discussions for #581 and #589, but current implementation seems not enough for me. For example, user can be typing smth like
```
class A { 
  int a(int b) {
     if (b > 5) <caret here> 
  }
}
```

In this case, recovery will try to find ; - and fail, since there's no ; till the end of class.
I suggest another approach - still simplified, but giving us at least all member signatures, when brace structure is correct.
In this approach, we not only check for ;, but check for } as well, haing proper level. Of course, we do not "eat" }, letting parse to finish.


MysterAitch:
Thanks @32kda  :partying_face:

PR #3248, opened 2021-05-07
Improved `RemoveMethodGenerator` and `ReplaceMethodGenerator` to only override super when needed.

4everTheOne:
The current version of the generators always generate the remove and replace method even when there's no need for it making the code harder to read.

Some examples of this can be seen in `PrimitiveType`, `ImportDeclaration` and many more. 
https://github.com/javaparser/javaparser/blob/7a8796f7334feb9563014c7da71868cfd4c49798/javaparser-core/src/main/java/com/github/javaparser/ast/type/PrimitiveType.java#L197-L203
https://github.com/javaparser/javaparser/blob/78098954b97381b2549adb957e60350ecc3d28c4/javaparser-core/src/main/java/com/github/javaparser/ast/ImportDeclaration.java#L146-L152

This PR adds a check to see if there's any statements added, if there's none then we don't need to override it.


MysterAitch:
Looks good to me :+1:

PR #3265, opened 2021-05-17
Improved performance when resolving types for big source files

maartenc:
The fix for #2995  (PR #3019) introduces a huge performance impact when resolving big source files.
I have for instance a (generated) sourcefile containing lots of inner types and consist of about 70.000 lines of code. Resolving all methods/params/types in this class takes more than 24h! 

In addition, I think the #3019 is not correct, because the `ClassOrInterfaceDeclarationContext` shouldn't search the complete tree when trying to solve a type. It shouldn't have knowledge of nested inner types.

This PR addresses these 2 issues, together with some other performance improvements.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeDeclaration.java
@@ -23,7 +23,6 @@ package com.github.javaparser.resolution.declarations;
 
 import com.github.javaparser.resolution.UnsolvedSymbolException;
 
-import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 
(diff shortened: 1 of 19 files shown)


maartenc:
@matozoid , @MysterAitch : one of the performance changes is to change the return type of the `ResolvedTypeDeclaration#internalTypes()` method from `Set` to `List`. The reason is that the `hashCode()` implementation of `Node` is very costly for big node trees, so changing it to `List` avoids calculating that hashCode. But this implies an API change, what do you think?

matozoid:
@maartenc - these should be helpful: VisitorSet, ObjectIdentityHashCodeVisitor, and ObjectIdentityEqualsVisitor.

maartenc:
Cool, thanks for the tip!
I'll try to use such a VisitorSet instead of a HashSet for the Nodes and see if that works.
If so, we don't have to change the API :)

matozoid:
Hey, `IdentityHashMap`! Cool, I've learned something today :-)

maartenc:
:-)
Me too... I already knew IdentityHashMap, but I didn't knew there was a way to create an 'IdentitySet' from it...
But luckily StackOverflow knows everything :-)

PR #3268, opened 2021-05-21
Issue #3200 `this` exp in anonymous class

### LINKED ISSUE #3200: The type is parsed in an anonymous internal class. The method calling with this is parsed. The parsing fails.

si-e:
Fixes #3200 .
Problem:
In method findContainingTypeDeclOrObjectCreationExpr, a value is returned without class name matching when an object creation expression is found.
Fixed:
The class name match is also performed when the object creation expression is found.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -628,17 +628,10 @@ public class JavaParserFacade {
      */
     protected Node findContainingTypeDeclOrObjectCreationExpr(Node node, String className) {
         Node parent = node;
-        boolean detachFlag = false;
         while (true) {
             parent = demandParentNode(parent);
             if (parent instanceof BodyDeclaration) {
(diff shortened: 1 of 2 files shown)


WanliZhong:
What can we do to solve the failure of codecov？

MysterAitch:
Thanks @si-e :tada:

PR #3273, opened 2021-05-22
Issue #3272 resolve lambda exp type

### LINKED ISSUE #3272: Resolve type of lambda exp
When a lambda expression is used as the argument to a function call expression, the Javaparser can properly resolve its type.
However, it is only in this case that the type of the lambda expression can be resolved; in other cases, an exception will be thrown.

```java
UnsupportedOperationException("The type of a lambda expr depends on the position and its return value")
```

But I don't think this makes sense, since lambda expressions can be resolved as function arguments, there are other cases where we should be able to resolve their types as well.

For example, when we initialize an interface variable with a lambda expression, we can easily know its type, which is the type of the variable.

Is this an area for improvement, or is it normal?

si-e:
Fixes #3272 .


jlerbsc:
This implementation seems to improve the processing of lambda expressions. Thanks for your contribution.
Can you update your pr with the base branch?

PR #3274, opened 2021-05-22
Refactored NodeListTest to use RepeatedTests

eas5:
Problem:

No problem is found in this test. This is a refactoring suggestion that simplifies test design by eliminating repeated steps that can be replaced by an annotation.

Solution:

The @RepeatedTest annotation provides the total number of repetitions desired. Each invocation of a repeated test behaves like the execution of a regular @test method with full support for the same lifecycle callbacks and extensions, which improves test design, eases future maintenance activities when needed, and better informs which repetition failed.


maartenc:
I did try it myself, with the @RepeatedTest annotation, the `hasNext()` method will be invoked on a new iterator instance each time, which is not the purpose of this test (which tests if `hasNext()` keeps returning true on successive calls on the same instance).

So I'll close this PR because it modifies the behaviour of the test.

PR #3299, opened 2021-06-16
Issue 3296

### LINKED ISSUE #3296: LexicalPreservation bug for array brackets
Hello all, thank you for developing such a great tool. However, I found a bug when using `LexicalPreservation` tool when I try to add an annotation to a class field which has the type `array` and the brackets `[]` are behind the parameter's name as shown in the example below.

```java
class Test{
   String[] allTest;
}
```
Please note that the example below works fine and the exception happens only when the brackets are behind field name. 

```java
class Test{
   String allTest[];
}
```

I also provided the code below that will generate the same exception mentioned above when executed:
```java
String code = "public class Test { String[] allTest; }";
CompilationUnit cu = LexicalPreservingPrinter.setup(StaticJavaParser.parse(code));
Optional<ClassOrInterfaceDeclaration> clazzOptional = cu.getClassByName("Test");
assertTrue(clazzOptional.isPresent());
ClassOrInterfaceDeclaration clazz = clazzOptional.get();
clazz.getMembers().forEach(
        bodyDeclaration ->
                bodyDeclaration.ifFieldDeclaration(
                        fieldDeclaration -> {
                            NodeList<VariableDeclarator> vars =
                                    fieldDeclaration.asFieldDeclaration().getVariables();
                            for (VariableDeclarator v : vars) {
                                if (v.getName().toString().equals("allTest")) {
                                    fieldDeclaration.addMarkerAnnotation("Nullable");
                                    break;
                                }
                            }
                        }));
String changed = LexicalPreservingPrinter.print(cu);
```

Here is the error I get when executing the code above:

```java
java.lang.UnsupportedOperationException: kept child(VariableDeclarator) vs "["   <101>   (line 1,col 27)-(line 1,col 27)

	at com.github.javaparser.printer.lexicalpreservation.Difference.applyKeptDiffElement(Difference.java:513)
```

With best regards, 

Nima Karimipour

nimakarimipour:
Fixes #3296

### DIFF
--- a/javaparser-core-generators/pom.xml
+++ b/javaparser-core-generators/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
-        <version>3.24.2-SNAPSHOT</version>
+        <version>3.22.2-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/javaparser-core-generators/pom.xml
+++ b/javaparser-core-generators/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
-        <version>3.22.2-SNAPSHOT</version>
+        <version>3.24.2-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/javaparser-core-metamodel-generator/pom.xml
+++ b/javaparser-core-metamodel-generator/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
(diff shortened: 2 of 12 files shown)


nimakarimipour:
@jlerbsc I believe this is ready for a review !

javaparser-core-generators/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <artifactId>javaparser-parent</artifactId>
         <groupId>com.github.javaparser</groupId>
-        <version>3.24.2-SNAPSHOT</version>
+        <version>3.22.2-SNAPSHOT</version>
```
jlerbsc: Why did you modify the java parser version?

jlerbsc:
Can you amend your commits to have a clean history?

javaparser-core-generators/pom.xml
nimakarimipour: This was for having two versions on my local maven, in the final version of this PR every version is reverted to its original one.

nimakarimipour:
> Can you amend your commits to have a clean history?

Sure, I will.

PR #3301, opened 2021-06-18
Java 11 is now the most used version

matozoid:
https://snyk.io/jvm-ecosystem-report-2021/

This value is supposed to get updated every now and then, so here it goes!


javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -218,7 +218,7 @@ public boolean isYieldSupported() {
     private boolean preprocessUnicodeEscapes = false;
     private SymbolResolver symbolResolver = null;
     private int tabSize = 1;
-    private LanguageLevel languageLevel = JAVA_8;
+    private LanguageLevel languageLevel = POPULAR;
```
matozoid, line 221: I'm not sure why I ever set this to Java 8, it seems more logical to have the most popular Java version as the default.

javaparser-core/src/main/java/com/github/javaparser/utils/CollectionStrategy.java
```diff
@@ -48,26 +48,25 @@
...
-                        final Optional<CompilationUnit.Storage> storage = compilationUnit.getStorage();
-                        if (storage.isPresent()) {
-                            final Optional<Path> sourceRootPath = storage.map(CompilationUnit.Storage::getSourceRoot);
-                            return sourceRootPath;
-                        } else {
-                            Log.info("Storage information not present -- an issue with providing a string rather than file reference?");
+                if (parseResult.getResult().isPresent()) {
+                    final Optional<CompilationUnit.Storage> storage = parseResult.getResult().flatMap(CompilationUnit::getStorage);
+                    if (storage.isPresent()) {
+                        if (storage.get().getFileName().equals("module-info.java")) {
```
matozoid, line 56: Ran into an older issue with module-info.java confusing the ParserCollectionStrategy. The issue was marked as fixed, but it sure failed badly here :-)

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -175,7 +175,7 @@
         /**
          * The most used Java version.
          */
-        public static LanguageLevel POPULAR = JAVA_8;
+        public static LanguageLevel POPULAR = JAVA_11;
```
matozoid, line 178: Woohoo!

matozoid:
Alright, these tests run fine locally, I have no idea why they fail here.

MysterAitch:
> Alright, these tests run fine locally, I have no idea why they fail here.

They run (and pass) locally here too. 

I've just pushed an additional check for `module-info.java` :+1:

matozoid:
Thanks!

PR #3302, opened 2021-06-18
Fix Java 11+ AST postprocessing

matozoid:
Postprocessor configuration was missing, so the "var" type was only working in Java 10.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/validator/postprocessors/Java11PostProcessor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/validator/postprocessors/Java11PostProcessor.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2021 The JavaParser Team.
+ *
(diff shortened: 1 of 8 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/validator/postprocessors/Java11PostProcessor.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.VarType;
+
+import static com.github.javaparser.ParseResult.PostProcessor;
+
+/**
+ * Processes the generic AST into a Java 11 AST and validates it.
+ */
+public class Java11PostProcessor extends Java10PostProcessor {
```
matozoid, line 32: These things are set up in the same way as the validators: newer versions extend from all older versions, adding and removing some features. No new postprocessing has been needed for version 11+, so all of those are empty ("the same as 10")

matozoid:
18 builds for every PR 8-O

matozoid:
Thanks!

PR #3303, opened 2021-06-18
Fixes Issue 3296 -- LexicalPreservation bug for array brackets

### LINKED ISSUE #3296: LexicalPreservation bug for array brackets

nimakarimipour:
Fixes #3296

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
@@ -751,12 +751,12 @@ public class Difference {
                 originalIndex++; // Now we can increment.
             } else if (currentIsNewline && addedTextElement.isChild()) {
                 // here we want to place the new child element after the current new line character.
-                // Except if indentation has been inserted just before this step (in the case where isPreviousElementNewline is true)
+                // Except if indentation has been inserted just before this step (in the case where isPreviousElementNewline is true) 
                 // or if the previous character is a space (it could be the case if we want to replace a statement)
                 // Example 1 : if we insert a statement (a duplicated method call expression ) after this one <code>  value();\n\n</code>
                 // we want to have this result <code>  value();\n  value();\n</code> not <code>  value();\n  \nvalue();</code>
-                // Example 2 : if we want to insert a statement after this one <code>  \n</code> we want to have <code>  value();\n</code>
-                // not <code>  \nvalue();</code> --> this case appears on member replacement for example
+                // Example 2 : if we want to insert a statement after this one <code>  \n</code> we want to have <code>  value();\n</code> 
+                // not <code>  \nvalue();</code> --> this case appears on member replacement for example 
                 if (!isPreviousElementNewline && !isFirstElement && !previousIsWhiteSpace) {
                     originalIndex++; // Insert after the new line
                 }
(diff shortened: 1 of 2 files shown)


nimakarimipour:
@jlerbsc as requested, I amended the commits. I think this is ready for a review.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -510,7 +510,7 @@ private void applyKeptDiffElement(Kept kept, TextElement originalElement, boolea
                     originalIndex++;
                     diffIndex++;
                 } else {
-                    throw new UnsupportedOperationException("kept " + kept.getElement() + " vs " + originalElement);
+                    originalIndex++;
```
jlerbsc, line 513: Can you be more specific here before jp threw an exception but with your solution it always keep the token?

MysterAitch:
Thank you @nimakarimipour for the report and PR! 

I agree that this fix works to solve the given test cases and doesn't _seem_ to break anything else... but it makes me nervous that I can't rationalise _why_ it works or when it might break :sweat_smile::sweat_smile:

This is a broader issue with `Difference.java` though, and I have added #3313 to tackle that more broadly.

MysterAitch:
Ideally I would like to see branch conditions that specifically allow this case where the array brackets are in a different location. Perhaps it can be something similar to the handling of type arguments a little above it?

If this isn't forthcoming, given that tests pass, I lean towards merging this in as an incremental improvement with a `// TODO` comment linking to the issue and/or indicating that it needs to be verified.

MysterAitch:
@jlerbsc / @nimakarimipour 

Given that #3316 has been merged, is this PR still needed? :)

jlerbsc:
Hi @MysterAitch there is no need to merge this PR. Thank you @nimakarimipour for your proposal.

PR #3310, opened 2021-06-28
Add test case for issue #2210 Resolution of Method References

### LINKED ISSUE #2210: Resolution of Method References

jlerbsc:
Fixes #2210. It seems that this issue is resolved now


jlerbsc:
Codecov failures "Unable to locate build via Github Actions API"

Issue #2471, opened 2019-12-25
Exception in thread "main" java.lang.StackOverflowError


qingdujun:
Seems like a recursive endless loop. in here [#122](https://github.com/javaparser/javaparser/blob/ee50a828535b5853e15011a6b7ea755a5c50d9e7/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java#L122)  [#FYI](https://github.com/javaparser/javaparser/issues/2050)

example: https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcFile.java

```
/Users/qingdujun/repos/workspace/data/code2track/repos/hive/ql/src/java/org/apache/hadoop/hive/ql/io/orc/OrcFile.java
Exception in thread "main" java.lang.StackOverflowError
	at java.io.UnixFileSystem.getBooleanAttributes0(Native Method)
	at java.io.UnixFileSystem.getBooleanAttributes(UnixFileSystem.java:242)
	at java.io.File.exists(File.java:819)
	at sun.misc.URLClassPath$FileLoader.getResource(URLClassPath.java:1334)
	at sun.misc.URLClassPath.getResource(URLClassPath.java:249)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:366)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:363)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:362)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at com.github.javaparser.symbolsolver.resolution.typesolvers.ClassLoaderTypeSolver.tryToSolveType(ClassLoaderTypeSolver.java:72)
	at com.github.javaparser.symbolsolver.resolution.typesolvers.ClassLoaderTypeSolver.tryToSolveType(ClassLoaderTypeSolver.java:90)
	at com.github.javaparser.symbolsolver.resolution.typesolvers.ClassLoaderTypeSolver.tryToSolveType(ClassLoaderTypeSolver.java:90)
[...]

PR #3324, opened 2021-07-20
Check if ancestor also for super class

ReallyLiri:
This is an improvement of the fix done in #2956
It also might be fixing #2471 , I didn't get the chance to test.

I encountered this bug and tested it against https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/search/suggest/phrase/PhraseSuggestion.java , where the following class snippet gets into an infinite loop over ancestors:

```java
public static class Option extends Suggestion.Entry.Option {
...
}
```


MysterAitch:
Thanks, looks good!

PR #3335, opened 2021-08-01
Fixed prettyprinting new switch-statements (and switch-expressions).

kozsik:
If the type of a SwitchEntry is STATEMENT_GROUP then it is an entry in an old switch-statement,
otherwise it is an entry in a new switch, and " ->" shall be used instead of ":".


jlerbsc:

PR #3337, opened 2021-08-02
Fix String Index out of range in TextBlockLiteralExpr

134ARG:
Fix calling stripIndentOfLines() on textblocks will produce
StringIndexOutOfBoundsException if there are white lines shorter than
the minimum common white space prefix.
For example (star * denotes a single white space):

"""*
**Hello
**World"""

The reason is the calculation of commonWhiteSpacePrefixSize excludes the
white lines but the final return still tries to get substring of all
lines. Once the commonWhiteSpacePrefixSize is greater than the length of
some white lines, StringIndexOutOfBoundsException will occur.


jlerbsc:

PR #3343, opened 2021-08-19
Configurable caching system for type solvers

4everTheOne:
_JavaParser_ implements a caching system for some of the type solver.
For example, in _JavaParserTypeSolver_ there are 3 different caches being used.
The first cache, stores the list of files that have been parsed (A association between a java file and it's compilation unit),
the second cache stores the list of compilations units in a given directory and finally the third item stores the symbol references that have been found.

This does it purpose as intended, but does not allow the use to choose if he want's to use it or not.
For example, let's imagine the user has a lot of computation power, but has low memory resources. In this situation, he may want to prefer to use more time instead of using the precious memory.

This PR refactors the  caching system by extracting the caching logic from the solvers. With this change the users, can provide the logic they want to use for each location.

For example, imagine that the user want's to create a _JavaParserTypeSolver_ that only stores the symbol references that were found and does care about the cached compilation units. This can be accomplished by:
```java
Path sourceCodePath = ...;
JavaParserTypeSolver javaParserTypeSolver = new JavaParserTypeSolver(
    sourceCodePath, // The source root where to locate the files
    javaParser, // The javaparser to use when parsing the files
    NoCache.create(), // How should files be cached
    NoCache.create(), // How should directories be cached
    InMemoryCache.create() // How should symbols found be cached
);
```

To improve the efficiency of the symbol solver an optional cache was added to _CombinedTypeSolver_ where the user can
also choose to use cache in it. An example for this can be seen below:
```java
List<TypeSolver> childSolvers = ...;
Cache<String, SymbolReference<ResolvedReferenceTypeDeclaration>> symbolCache = InMemoryCache.create();
CombinedTypeSolver combinedTypeSolver = new CombinedTypeSolver(
  ExceptionHandlers.IGNORE_NONE, // How exceptions should be handled by the solver
  childSolvers, // The list of type solver to be added as child
  symbolCache // How should symbols found be cached
);
```

**Note**: This refactor does not changes the current API.

Let me know what are your thoughts about this change and/or any suggestion you may have.


jlerbsc:
Hi @4everTheOne  For my part I think it is an improvement of the current implementation. Thank you for your suggestion.

4everTheOne:
Thank you for your comment @jlerbsc.

jlerbsc:
More precisely, what is interesting in your solution is the introduction of a cache management interface and the possibility of using another implementation (the code is more scalable). Perhaps to add a separation of responsibilities between the type solver and the cache it would have been better to implement the cache as a proxy. On the other hand, I am not convinced that there is a need to make the use of the cache configurable because I think it is an internal operation and not a feature that we want to expose to users.

4everTheOne:
> Perhaps to add a separation of responsibilities between the type solver and the cache it would have been better to implement the cache as a proxy

I'm not sure if i fully understood what you were saying. 
The figure, the cache logic is is handled by the "CachedTypeSolver" and when there's no hit's it delegates the solving to the underlying type solver.
tmp_jp (2)

Is this related with your sentence?

> On the other hand, I am not convinced that there is a need to make the use of the cache configurable because I think it is an internal operation and not a feature that we want to expose to users.

That's right, that's an internal operation. But this operation may be resource hungry, specially in the _JavaParserTypeSolver_.
For example, when trying to solve a type, it may try to parse and cache a entire directory and it can be an issue when the memory is limited.
The idea of allow the more advanced users to provide a cache object, that is going to be used as a cache. For example, bringing back the example from above, the user may want to avoid caching the information on _JavaParserTypeSolver_ and only cache the types that have been resolved. 
This operation may not be efficient when _JavaParserTypeSolver_ is queried frequently, but at least the user has the possibility to choose what he think suits the best for him.

jlerbsc:
> I'm not sure if i fully understood what you were saying.
The figure, the cache logic is is handled by the "CachedTypeSolver" and when there's no hit's it delegates the solving to the underlying type solver.
Yes it's exactly want i mean. In your picture CachedTyeSolver acts as a proxy on JavaParserTypeSolver and add the cache responsability.

> On the other hand, I am not convinced that there is a need to make the use of the cache configurable because I think it is an internal operation and not a feature that we want to expose to users.
This is from a end user perspective and not directly related to the previous remark. There are a lot of people who have trouble initializing the type resolver and this new API may add even more confusion. Maybe we should come up with a factory to build the solver type.

This is only my opinion and it deserves to be challenged by the other members of the project.

4everTheOne:
Thanks for clarifying! 

And yes, I agree with you. A builder could make it easier for the end user to setup a new type solver. 
Would be great to simply do something like:
```java
TypeSolver typeSolver= new TypeSolverBuilder()
    .useJavaRuntime()
    .addJar("path/to/file.jar")
    .addSourceCode("path/to/project")
    .build()
```

Should we a create an thread/issue to discuss about that?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JavaParserTypeSolver.java
```diff
@@ -144,24 +176,28 @@ public void setParent(TypeSolver parent) {
 
     private Optional<CompilationUnit> parse(Path srcFile) {
         try {
-            return parsedFiles.get(srcFile.toAbsolutePath(), () -> {
-                try {
-                    if (!Files.exists(srcFile) || !Files.isRegularFile(srcFile)) {
-                        return Optional.empty();
-                    }
+            Optional<Optional<CompilationUnit>> cacheValue = parsedFiles.get(srcFile.toAbsolutePath());
```
MysterAitch: minor comment -- could we rename `cacheValue` to avoid the word `value` and, instead, describe the content please?

I'm not set on any specific alternative, maybe `cachedParsedFile` or `parsedFilesCacheEntry` or similar?

MysterAitch:
Looks good, thank you @4everTheOne !

It looks like no new caches have been added (only where caching happened before).

There are two parts I am nervous about:

1. What happens where the type (i.e. the key) is not a string with a fully-qualified type -- i.e. could it accidentally give an inappropriate cache hit for a short name? 
This is not so much an issue if it is non-public internal-only code, but if we are now exposing caching publicly via the API then this increases the chance of it being misused.

2. Referring to a `symbolCache` - this is only a cache of types, correct? If so, perhaps we could rename it to `typeCache` or similar to make this explicit? 
_(caveat: I do need to double check my understanding of terminology/phrasing but, to me, `symbols` would also include variables such as local variables, fields, and parameters -- per "symbol solver")_

MysterAitch:
> Thanks for clarifying!
> 
> And yes, I agree with you. A builder could make it easier for the end user to setup a new type solver.
> Would be great to simply do something like:
> 
> ```java
> TypeSolver typeSolver= new TypeSolverBuilder()
>     .useJavaRuntime()
>     .addJar("path/to/file.jar")
>     .addSourceCode("path/to/project")
>     .build()
> ```
> 
> Should we a create an thread/issue to discuss about that?

A builder sounds good to me btw!

4everTheOne:
Thanks for you review @MysterAitch!

> 1. What happens where the type (i.e. the key) is not a string with a fully-qualified type -- i.e. could it accidentally give an inappropriate cache hit for a short name?
This is not so much an issue if it is non-public internal-only code, but if we are now exposing caching publicly via the API then this increases the chance of it being misused.

I can't identify a situation where we can accidentally get an inappropriate cache hit for a short name. 
The implementation uses the string that was given to the type solver as the key to store the type that have been found into the cache.
The only way that we can "pollute" the cache is by using it outside of the type solver. 
In that case, the user needs to fill the key and the expected value manually. 
Given the circumstances, I would consider it as an intentional "pollution".

> 2. Referring to a symbolCache - this is only a cache of types, correct? If so, perhaps we could rename it to typeCache or similar to make this explicit?
(caveat: I do need to double check my understanding of terminology/phrasing but, to me, symbols would also include variables such as local variables, fields, and parameters -- per "symbol solver")

Yes you are right! Maybe renaming the identifiers that use "symbolCache" to "typeCache" would be easier to understand the purpose of the caching system. :)

MysterAitch:
Let's get it merged in! :+1:

PR #3361, opened 2021-09-02
Try to fix yield with expression containing dots

koppor:
Details at https://github.com/javaparser/javaparser/issues/3364.

This PR adds a test case for the case that after yield a method is called. Example: `yield Collections.emptyList()`.

```java
        List<Integer> result = switch (s) {
            case "Foo":
                yield List.of(1);
            case "Bar":
                yield List.of(2);
            default:
                System.out.println("Neither Foo nor Bar, hmmm...");
                yield Collections.emptyList();
        };
```

Error:

    org.opentest4j.AssertionFailedError: [(line 3,col 15) Parse error. Found ".", expected one of  "," ";" "=" "@" "["

Expected: No error


koppor:
On the CI, I get following error:

    ParseErrorRecoveryTest.labeledStatementSemicolonRecovery:62 » IllegalState ???...

In IntelliJ, this test is green. Locally, with `./mvnw`, I also get this error. Not sure, how to proceed.

Longer error:

```
>
[ERROR]   JavaRecordNameTest.recordIsAValidVariableName:32 [(line 1,col 8) 'record' cannot be used as an identifier as it is a keyword.]
[ERROR]   TransformationsTest.exampleParam5:169->AbstractLexicalPreservingTest.assertTransformed:79  -- failed due to line separator differ

ences -- Expected: \n, but actual:  (system eol: \r\n) ==> expected: <// This is my class, with my comment
class A {
    int foo(int p1, char p2) {
        return p1;
    }
}> but was: <// This is my class, with my comment
class A {
    int foo(int p1, char p2) {
        return p1;
    }
}>
[ERROR] Errors:
[ERROR]   ParseErrorRecoveryTest.labeledStatementSemicolonRecovery:62 ; IllegalState ???...
```

jlerbsc:
Hi @koppor you have to use TestUtils.assertEqualsStringIgnoringEol to compare 2 string without the test being sensitive at the end of the line.

koppor:
I added the string comparison.

The other error still is there - maybe my grammar "fix" messed up some special case. Currently, I have no clue how to continue here.

```
 [INFO] Running com.github.javaparser.ParseErrorRecoveryTest
Error:  Tests run: 7, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.05 s <<< FAILURE! - in com.github.javaparser.ParseErrorRecoveryTest
Error:  com.github.javaparser.ParseErrorRecoveryTest.testBodyRecoverLevel  Time elapsed: 0.003 s
Error:  com.github.javaparser.ParseErrorRecoveryTest.labeledStatementSemicolonRecovery  Time elapsed: 0.006 s  <<< ERROR!
java.lang.IllegalStateException: ???; is not LabeledStmt, it is UnparsableStmt
	at com.github.javaparser.ParseErrorRecoveryTest.labeledStatementSemicolonRecovery(ParseErrorRecoveryTest.java:62)

```

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AbstractLexicalPreservingTest.java
```diff
@@ -76,22 +77,22 @@ protected void assertTransformed(String exampleName, Node node) throws IOExcepti
         String actualCode = LexicalPreservingPrinter.print(node);
 
         // Note that we explicitly care about line endings when handling lexical preservation.
-        assertEqualsString(expectedCode, actualCode);
+        TestUtils.assertEqualsStringIgnoringEol(expectedCode, actualCode);
```
jlerbsc: You are changing behavior that is intentional!

> Note that we explicitly care about line endings when handling lexical preservation.

koppor: Then I did not get your comment https://github.com/javaparser/javaparser/pull/3361#issuecomment-911969873 right. Reverted bbfc0d7987e9659bef6b7206c.

koppor:
I updated the PR to contain two commits:

1. Add test case for valid Java Code: [`4a0123f` (#3361)](https://github.com/javaparser/javaparser/pull/3361/commits/4a0123ffc3eca2f056e201a5393373c9bd1e76d9)
2. Naive try to fix the issue: [`693079d` (#3361)](https://github.com/javaparser/javaparser/pull/3361/commits/693079dcaa929eef8b19623955fc550c0e948b83)

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/AbstractLexicalPreservingTest.java
MysterAitch: Apologies, without context (note the comment on line 79) it wasn't obvious from the quoted errors that this was a place where line endings were important.

jlerbsc:
@MysterAitch The PR goal was to fix a grammar issue. Can you approve the @koppor 's proposal?

javaparser-core-testing/src/test/java/com/github/javaparser/JavaParserTest.java
```diff
@@ -121,6 +123,31 @@ void rangeOfArrayCreationLevelWithExpressionIsCorrect() {
...
+    }
+
+    @Test
+    void parseLYieldExpression() {
+        String code = "yield List.of(1);";
+        YieldStmt yieldStmt = (YieldStmt) parseStatement(code);
+        MethodCallExpr methodCallExpr = (MethodCallExpr) yieldStmt.getExpression();
+        IntegerLiteralExpr firstArgument = (IntegerLiteralExpr) methodCallExpr.getArguments().get(0);
+        assertEquals(1, firstArgument.asNumber());
+    }
```
matozoid, line 149: Hmmm, these ought to go to YieldStmtTest

matozoid:
I find it weird that the BlockStatement grammar has to be changed for this - it has been fine for years. Can you explain the fix, @koppor ?

MysterAitch:
> I find it weird that the BlockStatement grammar has to be changed for this - it has been fine for years. Can you explain the fix, @koppor ?

Within a `SwitchEntry`, the element to the right of a colon is one-or-more `BlockStatement`.

I am still investigating, but it appears that our definition of a `BlockStatement` isn't wholly correct per #3364 whereby the `Yield.of(1)` is first being interpreted/parsed as a `VariableDeclaration` before being considered as a `Statement` -- hence (I presume) the change in order/precedence suggested within @koppor 's PR.

Consider the following grammar entries from the JLS: 

```
SwitchBlockStatementGroup:
    SwitchLabel : {SwitchLabel :} BlockStatements 
```

```
BlockStatements:
    BlockStatement {BlockStatement}
BlockStatement:
    LocalVariableDeclarationStatement
    ClassDeclaration
    Statement
```

```
Statement:
    StatementWithoutTrailingSubstatement
    LabeledStatement
    IfThenStatement
    IfThenElseStatement
    WhileStatement
    ForStatement
StatementNoShortIf:
    StatementWithoutTrailingSubstatement
    LabeledStatementNoShortIf
    IfThenElseStatementNoShortIf
    WhileStatementNoShortIf
    ForStatementNoShortIf
StatementWithoutTrailingSubstatement:
    Block
    EmptyStatement
    ExpressionStatement
    AssertStatement
    SwitchStatement
    DoStatement
    BreakStatement
    ContinueStatement
    ReturnStatement
    SynchronizedStatement
    ThrowStatement
    TryStatement
    YieldStatement
```

MysterAitch:
The PR appears to fix the yield problem, but breaks the error recovery code (multiple tests no longer pass).

@matozoid -- Perhaps you are able to advise on this, please?

matozoid:
What's the error recovery problem you see, @MysterAitch ?

koppor:
> I find it weird that the BlockStatement grammar has to be changed for this - it has been fine for years. Can you explain the fix, @koppor ?

I was trying to use the `LOOKAHEAD` feature to ensure that `yield` is properly captured. However, it does not work.

Maybe, the added test cases are useful for a "proper" fix?

jlerbsc:
This PR is closed because JavaParser 3.24.7 correctly parse yield statement.

PR #3362, opened 2021-09-02
Fix "record" as non-type identifier in Java 16

### LINKED ISSUE #3366: "record" is a valid variable name in Java 16
Following statement is a valid statement in Java:

```java
    Object record;
```

When parsing that, following error is output:

    org.opentest4j.AssertionFailedError: [(line 1,col 8) 'record' cannot be used as an identifier as it is a keyword.]

Expected output: no error.

This seems to be https://github.com/javaparser/javaparser/issues/3261, which should be resolved (https://github.com/javaparser/javaparser/issues/3261#issuecomment-841089780), but did not work here.

At https://github.com/javaparser/javaparser/pull/3362, I tried to craft a PR fixing that. However, the CI does not like it. (Locally, in IntelliJ, it runs)

Refs https://github.com/javaparser/javaparser/issues/2446 (requesting for record support), https://github.com/javaparser/javaparser/pull/3022 (implementing records), https://github.com/javaparser/javaparser/issues/3260#issuecomment-839869071 (proposing some fix on `record` handling)

koppor:
Details at https://github.com/javaparser/javaparser/issues/3366.

Following statement is a valid statement in Java:

```java
    Object record;
```

This PR creates a test for it and tries to provide a fix.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java
@@ -25,21 +25,17 @@ import com.github.javaparser.JavaParser;
 import com.github.javaparser.ParseResult;
 import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.stmt.Statement;
 import com.github.javaparser.utils.TestUtils;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 
 import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
-import static com.github.javaparser.ParseStart.STATEMENT;
 import static com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_15_PREVIEW;
-import static com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_16;
 import static com.github.javaparser.Providers.provider;
-import static com.github.javaparser.utils.TestUtils.assertProblems;
 
 class Java15PreviewValidatorTest {
 
-    private final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLanguageLevel(JAVA_16));
+    private final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLanguageLevel(JAVA_15_PREVIEW));
 
     /**
      * Records are available within Java 14 (preview), Java 15 (2nd preview), and Java 16 (release).
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java
@@ -49,22 +49,45 @@ class Java15PreviewValidatorTest {
     class Record {
 
         @Nested
-        class RecordAsIdentifierForbidden {
+        class RecordAsTypeIdentifierForbidden {
             @Test
-            void recordUsedAsClassName() {
+            void recordUsedAsClassIdentifier() {
                 String s = "public class record {}";
                 ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT, provider(s));
-                TestUtils.assertProblems(result, "(line 1,col 14) 'record' cannot be used as an identifier as it is a keyword.");
+                TestUtils.assertProblems(result, "(line 1,col 14) 'record' is a restricted identifier and cannot be used for type declarations");
             }
 
             @Test
-            void recordUsedAsFieldName() {
+            void recordUsedAsEnumIdentifier() {
+                String s = "public enum record {}";
+                ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT, provider(s));
+                TestUtils.assertProblems(result, "(line 1,col 13) 'record' is a restricted identifier and cannot be used for type declarations");
+            }
+
+            @Test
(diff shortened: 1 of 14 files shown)


koppor:
I added `ReservedKeywordValidatorRecord` to handle the `record` special case and adapted test cases accordingly.

MysterAitch:
Tests appear to be failing due to the Java 15 (preview) validator tests -- note that records were in preview for both Java 14 and Java 15 before being available in Java 16 :+1: 

- `src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java`.

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/Java14PreviewValidator.java
```diff
@@ -39,7 +39,7 @@ public Java14PreviewValidator() {
         {
             // first preview within Java 14 - https://openjdk.java.net/jeps/359
             remove(noRecordDeclaration);
-            add(recordAsTypeIdentifierNotAllowed);
+            add(recordAsClassAndInterfaceIdentifierNotAllowed);
```
MysterAitch: Please revert this back to `TypeIdentifier` - `record` is not permitted e.g. as an enum`s identifier either.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java16ValidatorTest.java
```diff
@@ -57,7 +57,7 @@ void yieldAllowed() {
     class Record {
 
         @Nested
-        class RecordAsIdentifierForbidden {
+        class RecordAsIdentifierSometimesForbidden {
```
MysterAitch: Please keep the tests within the `Java16ValidatorTest` consistent with the ones within the Java14/15 preview validator tests.

I fear I misled you within the issue -- my mistake. Perhaps rename this to `RecordAsTypeIdentifierForbidden`?

MysterAitch: Additionally, you have correctly pointed out to me that records used as e.g. field and method identifiers is fine -- move the `recordUsedAsFieldName()` test method outside of this?

javaparser-core/src/main/java/com/github/javaparser/ast/validator/ReservedKeywordValidatorRecord.java
```diff
@@ -0,0 +1,78 @@
...
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
+import com.github.javaparser.ast.body.RecordDeclaration;
+import com.github.javaparser.ast.expr.Name;
+import com.github.javaparser.ast.expr.SimpleName;
+
+import static com.github.javaparser.utils.CodeGenerationUtils.f;
+
+/**
+ * Validates that "record" cannot be used as class name.
```
MysterAitch: It's not just classes / interfaces - using `record` as the name of enums and records are forbidden too.

javaparser-core/src/main/java/com/github/javaparser/ast/validator/ReservedKeywordValidatorRecord.java
```diff
@@ -0,0 +1,78 @@
...
+ * For details, see <a href="https://openjdk.java.net/jeps/395">JEP 395</a>
+ */
+public class ReservedKeywordValidatorRecord extends VisitorValidator {
+    private final String keyword;
+    private final String error;
+
+    public ReservedKeywordValidatorRecord() {
+        this.keyword = "record";
+        error = f("'%s' cannot be used as an identifier as it is a keyword.", keyword);
```
MysterAitch: "as it is a restricted identifier."

javaparser-core/src/main/java/com/github/javaparser/ast/validator/ReservedKeywordValidatorRecord.java
```diff
@@ -0,0 +1,78 @@
...
+        super.visit(n, arg);
+    }
+
+    private boolean validUsage(Node node) {
+        if (!node.getParentNode().isPresent()) {
+            return false;
+        }
+        Node parent = node.getParentNode().get();
+
+        if (parent instanceof ClassOrInterfaceDeclaration) {
```
MysterAitch: It's not just classes / interfaces - using `record` as the name of enums and records are forbidden too.

MysterAitch:
I think this is good to merge, just a few bits about naming to update -- thank you @koppor  :+1:

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java15PreviewValidatorTest.java
```diff
@@ -25,21 +25,17 @@
...
-import static com.github.javaparser.ParseStart.STATEMENT;
 import static com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_15_PREVIEW;
-import static com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_16;
 import static com.github.javaparser.Providers.provider;
-import static com.github.javaparser.utils.TestUtils.assertProblems;
 
 class Java15PreviewValidatorTest {
 
-    private final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLanguageLevel(JAVA_16));
+    private final JavaParser javaParser = new JavaParser(new ParserConfiguration().setLanguageLevel(JAVA_15_PREVIEW));
```
MysterAitch, line 38: Good spot, thanks! :+1:

PR #3363, opened 2021-09-02
Reduce mvn verbosity on GitHub actions (and switch to mvnw)

koppor:
When reading error output, it is time consuming to scroll through all download progress.

Example:

grafik

This PR a) reduces verbosity when running on GitHub actions and B) uses the contained Maven wrapper to ensure that always the same Maven version is used.


MysterAitch:
Thanks!

PR #3398, opened 2021-10-26
Fix resoure leak due to `File.walk`

lujiefsi:
Files.walk  will open stream, we should close it.

see jdk:
the {try} -with-resources construct should be used to ensure that the
stream's

{[@link|https://github.com/link] Stream#close close}
method is invoked after the stream
operations are completed.


MysterAitch:
Thanks @lujiefsi !

PR #3407, opened 2021-11-13
Issue #3406 `ParseProblemException` when parsing char `\u005cn`

### LINKED ISSUE #3406: ParseProblemException when parsing char '\u005cn'
Here is a minimal test case to reproduce it 

```
public class Test {

    char c = '\u005cn';

}



```
The grammar currently supports unicode characters somewhat but not unicode escape sequences inside characters without pre-processing.

Stacktrace is 
```
Exception in thread "main" com.github.javaparser.ParseProblemException: Lexical error at line 3, column 21.  Encountered: "n" (110), after : "\'\\u005c"
Problem stacktrace : 
  com.github.javaparser.GeneratedJavaParserTokenManager.getNextToken(GeneratedJavaParserTokenManager.java:3010)
  com.github.javaparser.GeneratedJavaParser.jj_ntk_f(GeneratedJavaParser.java:13611)
  com.github.javaparser.GeneratedJavaParser.VariableInitializer(GeneratedJavaParser.java:1824)
  com.github.javaparser.GeneratedJavaParser.VariableDeclarator(GeneratedJavaParser.java:1752)
  com.github.javaparser.GeneratedJavaParser.FieldDeclaration(GeneratedJavaParser.java:1712)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBodyDeclaration(GeneratedJavaParser.java:1640)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBody(GeneratedJavaParser.java:1183)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceDeclaration(GeneratedJavaParser.java:500)
  com.github.javaparser.GeneratedJavaParser.CompilationUnit(GeneratedJavaParser.java:152)
  com.github.javaparser.JavaParser.parse(JavaParser.java:199)
  com.github.javaparser.StaticJavaParser.parse(StaticJavaParser.java:138)

```

apixandru:
Fixes #3406 

Currently there is some support for unicode but not unicode escape sequences. The only option is to pre-process the unicode characters but the original characters are lost or just fail to parse the sequence entirely.

This PR addresses this issue by having the tokens parsed with their original meaning at their original positions without unicode pre-processing.


apixandru:
@jlerbsc any concerns related to this PR?

jlerbsc:
Hi @apixandru I am not very comfortable with this subject. @MysterAitch what do you think about this topic/PR?

apixandru:
rebased on master

javaparser-core-testing/src/test/resources/com/github/javaparser/EscapeSequences.java
```diff
@@ -0,0 +1,32 @@
...
+                "---",
+                '\r', '\u005cr', '\u005Cr',
+                "---",
+                '\t', '\u005ct', '\u005Ct',
+                "---",
+                '\b', '\u005cb', '\u005Cb',
+                "---",
+                '\f', '\u005cf', '\u005Cf',
+                "---",
+                '\'', '\u005c'', // '\u005C'', // that's weird, this last one won't compile
```
MysterAitch, line 19: ~`\u005C` is a back-slash character~

~This means that `'\u005C'` becomes `'\'` which, reasonably, does not compile.~

~If you change it to `\u005C\u005C` then this does compile (equivalent to `'\\'`).~

javaparser-core-testing/src/test/resources/com/github/javaparser/EscapeSequences.java
```diff
@@ -0,0 +1,32 @@
+package com.github.javaparser;
+
+public class EscapeSequences {
```
MysterAitch, line 3: Perhaps rename this class/file to make it explicitly obvious that this is a manually-run test rather than a unit test file.

javaparser-core-testing/src/test/resources/com/github/javaparser/EscapeSequences.java
apixandru: I agree that `'\u005C'` becomes `'\'` and that would be invalid, however the sequence that i'm escaping has a double single quote.

Just like the lowercase version `'\u005c''` gets interpreted as `\'` I would have expected the upper case version to do the same but something about this sequence won't compile.

`\u005C\u005C` is already covered a few lines above

MysterAitch: Actually, reading more into it, I'm starting to realise why this is odd -- the Unicode escape sequences should be case insensitive (as demonstrated in the other examples) so it's strange that this specific example doesn't compile.

MysterAitch:
Like @jlerbsc , the Unicode side of things is not an area I am comfortable with and not something I have really explored.

Without (yet) reading more into the JLS or how this PR changes things, as a guiding principles:

- I would hope to be able to use JavaParser to parse any valid Java input AND then be able to print it back out in the same way it was written/read in. 
- Additionally, the position information is important too to retain (e.g. the Unicode escapes should move anything later in the line by the the number of chars in the escape).

Assuming these two are met, I'm up for merging this in :)

apixandru:
You might want to check the ExpectedTokensTest test case included in the PR.

It verifies that the escape sequences are properly preserved in the parsed tokens and that the position of elements are where you would expect them to be in the source file.

javaparser-core-testing/src/test/resources/com/github/javaparser/EscapeSequences.java
MysterAitch: Scratch this idea... I have come back with fresh eyes and noticed it's within the `resources` directory, so my comment is not applicable! :raised_eyebrow: :+1:

MysterAitch:
With a fresh pair of eyes after sleeping on it, the comments I made yesterday don't seem to apply and it seems everything is in order! 

Thank you for your contribution @apixandru ! :+1:

Issue #3355, opened 2021-08-26
[Suggestion] Builder to construct type solvers


4everTheOne:
In the comment https://github.com/javaparser/javaparser/pull/3343#issuecomment-902648574, was suggested to created builder to make it easier to setup the type solvers.

A simple example could be something like:
```java
TypeSolver typeSolver= new TypeSolverBuilder()
    .useJavaRuntime()
    .addJar("path/to/file.jar")
    .addSourceCode("path/to/project")
    .build()
```

Let's try to group suggestion of what would be nice to have in it in this thread.

4everTheOne:
I think it could be also interesting to add android archives (aar) and manual symbols (memory type solver) with something like:
```java
TypeSolver typeSolver= new TypeSolverBuilder()
    .addAndroidArchive("path/to/file.aar")
    .addManualType("com.example.Foo", fooTypeSymbol)
    .build()
```

jlerbsc:
I am closing this issue since a pr was merged.

PR #3421, opened 2021-11-26
Created TypeSolverBuilder

4everTheOne:
Hi team!

In this PR I suggest a Type Solver builder. This class tries to simplify the process of creating a new type solver.
Using the current method, to create a type solver that uses the current JRE (ReflectionTypeSolver), a JAR file (JarTypeSolver) and a Aar file (AarTypeSolver) we need to execute something like this:
```java
combinedTypeSolver.add(new ReflectionTypeSolver());
combinedTypeSolver.add(new JarTypeSolver("foo.bar"));
combinedTypeSolver.add(new AarTypeSolver("bar.aar"));
```

With this builder, the process of creating a new type solver looks like this:
```java
TypeSolver combinedTypeSolver = new TypeSolverBuilder()
                .withCurrentJRE()
                .withJAR("foo.jar")
                .withAAR("bar.aar")
                .build();
```

Do you think it may be useful? Or any ideias to include in this builder?

---

Related with #3355.


MysterAitch:
Looks good to me!

jlerbsc:
Thank you for this improvement.

PR #3424, opened 2021-11-27
Minor refactoring on `LexicalreservingPrinter` especially in the class `Difference`

jlerbsc:


MysterAitch:
Note there's currently an ongoing incident with Github Actions, which is likely why they're not running here.

jlerbsc:
@MysterAitch what can we do?

MysterAitch:
Looks like it's back up. 

Maybe push it again +- a new commit to trigger checks again?


https://www.githubstatus.com/

PR #3430, opened 2021-12-05
Call `orElse()` instead of `orElseGet()`

msridhar:
If an `Optional` is not present, `orElseGet()` will fail with an NPE if passed `null`:

https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html#orElseGet-java.util.function.Supplier-

Probably `orElse()` is what was desired here.


msridhar:
BTW, I found this using [Error Prone](https://errorprone.info).  If there is interest in running it as part of the build I could contribute that; it does slow down compile times.

jlerbsc:

PR #3432, opened 2021-12-05
Add a missing `hashCode()` method

msridhar:
This was also caught by Error Prone.  Adds a `hashCode()` method to match the `equals()` method in the class.  Went for simplicity, but if this `hashCode()` is performance-critical I can optimize.


MysterAitch:
Thanks @msridhar :+1:

PR #3462, opened 2022-01-09
Updated release script to be non-interactive, and added option to use a manually-triggered GitHub Action to build a release

MysterAitch:
fyi @jlerbsc 

Currently, to perform a release, process performed is per the first code block -- which is interactive and requires a user to respond to prompts for information (e.g., version numbers, git tag).

1. This PR adds a non-interactive equivalent to the `release:prepare` (`./run_release_prepare.sh`) script, where the relevant versions are passed as script arguments (instead of being interactively prompted for).

2. This PR also adds a GitHub action that requires the version number of the release and the version number of the next development release. The git tag is derived from the release version.

The intention for this GitHub Action is that it can be run from a fork where that developer's credentials ("github secrets") can be added to that fork (to avoid one person's credentials being added as secrets to the main repo)
  - Once this code is merged, I will test this and verify the GitHub Action works as intended by releasing v3.24.0
  - If triggering a release via GHA on a fork fails, I will then try adding my credentials to the main repo and trigger it from here
  - As a last resort, I'll just run it locally in order to get the release out today :+1: 

----

## Current (manual) process

```sh

## This is currently interactive, prompting for the versions and tags to be used as part of the release
## This is also the only area that changes between this (current) process and the proposed (new) process described within the next code block
##
## NOTE: Pre-Requisites:
## 1. SSH access and keys setup for GitHub
## 2. GPG setup with signing key created
## 3. (performing the release i.e., pushing it to Sonatype OSSRH / Maven Central also requires OSSRH credentials and authorisation setup)
## 4. All of the above to be configured within your `.m2/settings.xml` file
##     (or, advanced usage, for the above to be defined within environment variables and you instruct maven to use the settings file/template within `dev-files/settings.xml`)
./run_release_prepare.sh

################################
## If wanting to abort the release at this stage, we must rollback the changes made
## Note that `release:rollback` exists, but 
## - This does not always delete the tag and 
## - It is additive to the git history (creating a revert commit) 
## ... while the steps below erase the changes as if they never happened.
##
## Steps:
### 1. Remove the (local) tag (for example: `git tag -d javaparser-parent-3.24.0`)
git tag -d <tag_name>
### 2. Hard-reset the (local) commits -- currently three commits (1) update readme, (2) set pom release version number, (3) set pom new development/snapshot version
git reset --hard HEAD~3
### 3. Reset the `release:prepare` (it creates `release.properties` and backups of the pom files)
###    - Note that release:revert
./mvnw release:clean
################################


## The release:prepare is configured to create the relevant commits and tags locally - these need to be pushed
git push --follow-tags

################################
## If wanting to abort at this stage: 
## - The remote tag must be removed. 
## - Resetting the git history is unlikely to be possible due to branch protection rules re: force pushes (therefore the release:rollback option is more appropriate at this stage).
################################


## Once the release is ready, this step pushes it to Sonatype OSSRH
./run_release_perform.sh

################################
## If wanting to abort at this stage: 
## - follow the steps above
## - login to the Sonatype OSSRH site and "drop" the staged release
################################


## At this point it is uploaded to OSSRH as a draft release -- visible within "Staging Repositories" at https://oss.sonatype.org/
## Once this draft release is (manually) closed and promoted, it will be uploaded to Maven Central

################################
## If wanting to abort at this stage: 
## - As a rule, removing artefacts from Maven Central is not possible (though can be done in exceptional circumstances)
## - If a mistake is discovered, update the changelog/release notes and perform a new release with a higher version nurmber
################################

```

## Updated (manual / GitHub Actions) process

```sh
## Note the new script which does not require interactive user input.
## These variables which define the `release version` and `next development/snapshot version` will be passed either: 
## - Manually via a developer running these scripts via local machine
## - Manually via the GitHub action's input variables web UI
./run_release_prepare_non-interactive.sh 3.24.0 3.24.1-SNAPSHOT

################################
## If wanting to abort the release at this stage, we must rollback the changes made.
## How this is achieved depends on whether the script is run on a developer's local machine, versus via GitHub Actions runner
## 
## - On the developer's local machine
## -- ... in which case, follow the guidance in the "old" code block above
## 
## - Within the GitHub actions runner - in which case the changes will be pushed to the GitHub master branch
## -- ... in which case, follow the steps re: rollback and follow the OSSRH steps to drop the release 
##        because (assuming no errors) the tag/commits will already exist in upstream master 
##        and the release will already be staged on OSSRH.
## 
################################


## If running locally, will need to push the git changes to upstream and perform the release to push it to Sonatype OSSRH,
## and rollback/abort instructions are per the notes above.
git push --follow-tags
./run_release_perform.sh


## At this point it is uploaded to OSSRH as a draft release -- visible within "Staging Repositories" at https://oss.sonatype.org/
## Once this draft release is (manually) closed and promoted, it will be uploaded to Maven Central

################################
## If wanting to abort at this stage: 
## - As a rule, removing artefacts from Maven Central is not possible (though can be done in exceptional circumstances)
## - If a mistake is discovered, update the changelog/release notes and perform a new release with a higher version nurmber
################################

```


MysterAitch:
Note that #3464 removes the github action code because there are problems faced when running from a fork (mismatches between checked out repo/branch/head and the SCM reference) and problems faced when running directly on the main repo (the master branch contains protections to restrict pushes to master).

Note the actual release prepare (and presumably perform) run absolutely fine when the relevant secrets are passed to it -- the problems above are related to interacting with the github repo.

Release 3.24.0 will be done manually using the non-interactive script added in this PR (+- a few tweaks) and the steps outlined above.

PR #3473, opened 2022-01-13
Bug in ArrayCreationExpr constructors

### LINKED ISSUE #3467: new ArrayCreationExpr(Type elementType) creates element with invalid syntax
A test to reproduce:
```java
@Test
void invalidSyntaxCreatedWhenUsingConstructorWithOnlyTypeProvided() {
    var expr = new ArrayCreationExpr(PrimitiveType.byteType());

    assertEquals("new byte[]", expr.toString());
}
```

Additionally, this expression resolved type is wrong. It resolves to the `ResolvedPrimitiveType`, rather to `ResolvedArrayType`:
```java
    @Test
    void resolvesToTypeUsedInArrayNotArrayType() {
        var expr = new ArrayCreationExpr(PrimitiveType.byteType());
        insertIntoCompilationUnit(expr); // My custom method

        assertInstanceOf(ResolvedArrayType.class, expr.calculateResolvedType());
    }
```

The actual output of `expr.toString()` is `new byte {}` which is an invalid array definition. I think I can prepare a PR with a fix, but I need guidance as this is a BC-breaking change.

sergekukharev:
Fixes #3467.

Some notes:
* Removed validator test as this case won't happen anymore
* Added braces to if statement for consistency

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/ArrayCreationExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/ArrayCreationExprTest.java
@@ -1,0 +1,37 @@
+package com.github.javaparser.ast.expr;
+
+import com.github.javaparser.ast.ArrayCreationLevel;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.type.PrimitiveType;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class ArrayCreationExprTest {
+
+    @Test
+    void correctlyCreatesExpressionWithDefaultConstructor() {
+        ArrayCreationExpr expr = new ArrayCreationExpr();
+
+        assertEquals("new empty[] {}", expr.toString());
+    }
+
+    @Test
+    void correctlyCreatesExpressionWithSimpleConstructor() {
+        ArrayCreationExpr expr = new ArrayCreationExpr(PrimitiveType.byteType());
+
+        assertEquals("new byte[] {}", expr.toString());
+    }
+
+    @Test
+    void correctlyCreatesExpressionWithFullConstructor() {
+        ArrayInitializerExpr initializer = new ArrayInitializerExpr(new NodeList<>(
+                new IntegerLiteralExpr("1"),
+                new IntegerLiteralExpr("2"),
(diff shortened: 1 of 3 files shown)


jlerbsc (commented):
It seems to me that it's more appropriate to modify the default constructor like this.

> public ArrayCreationExpr() {
        this(null, new ClassOrInterfaceType(), new NodeList<>(new ArrayCreationLevel()), new ArrayInitializerExpr());
    }

sergekukharev:
That was the first thing that came to my mind, then I realized the problem can still appear in the future in case there's a new constructor. It's a weak preference - I'm happy to change the code.

Additionally, people can still call a public `ArrayCreationExpr.setLevels(new NodeList<>())` and thus break AST validity. 

Let me know if you still prefer the option you suggested, I'll update my branch.

jlerbsc:
Yes but you can protect this method  `ArrayCreationExpr.setLevels(new NodeList<>())` too

matozoid:
I'd say if the validator is there, the test should stay, and the validator should stay because there is still a way to get into the invalid state :-) (But the validators are only triggered while parsing, so it is useless unless you apply the validators to your modified AST manually. Unless this functionality has changed?)

sergekukharev:
@jlerbsc take a look please, I hope I understood you correctly. I kept changes in `setLevels()` and explicitly added ArrayCreationLevel to constructors as you suggested.

@matozoid I would love to keep the test, but now since both constructors and the setter of levels make sure it's not empty, I don't think there's a way to make this test work.

matozoid:
@sergekukharev alright, and in that case the validation itself can be removed too.

sergekukharev:
I'm not very familiar with this part of the project. Would be changing metamodel for levels enough?

_src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java:_
```java
arrayCreationExprMetaModel.levelsPropertyMetaModel = new PropertyMetaModel(arrayCreationExprMetaModel, "levels", com.github.javaparser.ast.ArrayCreationLevel.class, Optional.of(arrayCreationLevelMetaModel), false, false, true, false);
```

matozoid:
Sorry @sergekukharev - I thought there would be some code in the Java1_0Validator, but it turns out that yes, it is validated with the metamodel, and the metamodel is already correct, so no need to change. Everything is fine then :-)

jlerbsc:

PR #3485, opened 2022-02-02
Fix issue #3218 GetSourceRoots() does not return all source roots

### LINKED ISSUE #3218: GetSourceRoots() does not return all source roots
This happened only with apache's log4j (https://github.com/apache/log4j), I have never experienced this problem before. Here's the code I'm executing:

```java
ProjectRoot projectRoot = new SymbolSolverCollectionStrategy().collect(Paths.get(projectRootStr));
List<SourceRoot> sourceRoots = projectRoot.getSourceRoots();
```
Where projectRootStr is the absoulute path to log4j's directory (Example: C:/Users/user/Desktop/log4j-trunk). 

The size of list (sourceRoots) returned is 5, whereas the original size is much bigger. Am I missing something? 

Thank you for your time and help!

jlerbsc:
Fixes #3218

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/HasParentNode.java
@@ -83,23 +83,6 @@ public interface HasParentNode<T> extends Observable {
         return findAncestor(predicate, type);
     }
     
-//    /**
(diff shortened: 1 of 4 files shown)


jlerbsc:
some errors with the following type "Failed to write uploader binary: EPERM: operation not permitted"

PR #3499, opened 2022-02-20
Adding constructor descriptor

### LINKED ISSUE #3498: Constructor descriptor
Similar to https://github.com/javaparser/javaparser/issues/2059, is there a way to obtain the descriptor of the constructor?  

I noticed that com.github.javaparser.ast.body.ConstructorDeclaration lacks a `toDescriptor` method.

kanghj:
Fixes #3498.

The code uses an implementation identical to ASM's `getConstructorDescriptor`: https://github.com/consulo/objectweb-asm/blob/master/asm/src/main/java/org/objectweb/asm/Type.java#L546

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/DescriptorTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/DescriptorTest.java
@@ -21,7 +21,7 @@ class DescriptorTest extends AbstractResolutionTest {
     void descriptorTest() {
         String code = 
                 "public class A {\n" +
-                "  A(int i, double d, Thread t) {return new Object();}\n" +
+                "  A(int i, double d, Thread t) {}\n" +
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/DescriptorTest.java
```diff
@@ -20,6 +21,7 @@
     void descriptorTest() {
         String code = 
                 "public class A {\n" +
+                "  A(int i, double d, Thread t) {return new Object();}\n" +
```
jlerbsc: Even if the test is correct/pass, this is not a correct implementation for a constructor. A constructor cannot return a value. Can you change it please?

kanghj: Good catch. I've removed the return statement.

jlerbsc:

PR #3509, opened 2022-03-03
add the EnumExpression

bmnizar:
Fixes #9999.
Hi ,

I have added an EnumExpression class to handle Enumeration .
For example:Suppose we have this part of code:
private MyEntity entity;

We need to add this annotation @Transactional(propagation = MANDATORY)  to the entity object.
The actual Expression does not allow to add an enumeration (Mandatory for example) .
For that .I have added the EnumExpression to handle it


jlerbsc:
Hi @bmnizar you can add this annotation to the FieldDeclaration just like this.

> AnnotationExpr annotationExpr = new NormalAnnotationExpr(new Name("Transactional"), new NodeList<MemberValuePair>()).addPair("propagation", "MANDATORY");        
        fd.addAnnotation(annotationExpr);

jlerbsc:
Thank you for your proposal but this PR does not seem useful because your problem can be solved with the current code.

PR #3515, opened 2022-03-13
Issue #3405 thread safety of pre/postprocessors (incl. breaking change to `Processor` with pre/post processor methods).

### LINKED ISSUE #3405: Why is Javaparser not thread-safe?
I have to analyze 15.000/20.000 files.
The idea is to use Javaparser in parallel.
Then the problem is: why is javaparser not thread safe?
Which classes have problems?

I analyzed all "static" keywords present in the source code by I have not found the class that can have problems.
However this class there exists because using the current implementation in parallel, sometimes I found strange "Lexical error" during the analysis, and sometimes the AST is not complete.

matozoid:
Fixes #3405.


javaparser-core/src/main/java/com/github/javaparser/JavaParser.java
```diff
@@ -114,17 +118,20 @@ private GeneratedJavaParser getParserForProvider(Provider provider) {
         assertNotNull(start);
         assertNotNull(provider);
 
-        for (PreProcessor preProcessor : configuration.getPreProcessors()) {
-            provider = preProcessor.process(provider);
+        List<Processor> processors = configuration.getProcessors().stream().map(Supplier::get).collect(toList());
```
matozoid, line 121: Since some processors use preprocessing and postprocessing, and communicate between these two stages, we can't go instantiating all the preprocessor instantiating all the postprocessors. Some of them need to be the same instance between pre and postprocessing! This PR merges `PreProcessor` and `PostProcessor` into one `Processor` class and instantiates them all here, just before parsing.

matozoid: It is more awkward to use, since you can't use lambda syntax anymore.

matozoid: This is an API change. I don't feel to bad about it though, because I doubt many people have taken advantage of pre and postprocessors. Preprocessing is a highly specialized and rather awkward thing, and postprocessing can just be done in user code after parsing, so why use a postprocessor?

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -196,7 +195,7 @@
                 JAVA_17, JAVA_17_PREVIEW
         };
 
-        LanguageLevel(Validator validator, ParseResult.PostProcessor postProcessor) {
+        LanguageLevel(Validator validator, PostProcessors postProcessor) {
```
matozoid, line 198: The validators still talk about post-processing, because that's what they do.

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -287,43 +283,51 @@ public void process(ParseResult<? extends Node> result, ParserConfiguration conf
...
         }
 
-        UnicodeEscapeProcessor unicodeProcessor = new UnicodeEscapeProcessor();
-        preProcessors.add(unicodeProcessor);
-        postProcessors.add(unicodeProcessor);
-
-        LineEndingProcessor lineEndingProcessor = new LineEndingProcessor();
-        preProcessors.add(lineEndingProcessor);
-        postProcessors.add(lineEndingProcessor);
+        processors.add(UnicodeEscapeProcessor::new);
```
matozoid: So it looks like this now. You don't add a processor, but you add a supplier that creates a processor. ::new is such a thing.

MysterAitch:
I can't comment first-hand on the thread safety element but (assuming it works) I've no objection to making the change to make all pre/post processors be a processor :+1:

MysterAitch:
No other objections raised, merging in :+1: 

Thanks @matozoid !

matozoid:
Yaaay!

PR #3527, opened 2022-03-22
Default pretty printer should print inner-class receiver parameters on constructors

kelloggm:
I copied the receiver parameter printing code from the logic for other kinds of methods.


MysterAitch:
Good spot, thank you @kelloggm :+1:

PR #3535, opened 2022-03-25
Code refactoring

meghzman:
**Move Method refactoring**
getExpressionBody method of class LambdaExpr was more interested in the methods of the class Statement, so I have done the move method refactoring to solve this 'Feature Envy' design smell.  I have tested no code is breaking because of this change.


meghzman:
**Rename Variable refactoring**
in class ManipulationSteps, method MethodDeclaration variable 'n' was not a proper name. I changed it to methodDeclaration

meghzman:
**Push Down Refactoring**
In class ReferenceType, method toDescriptor was abstract and caused a Rebellious hierarchy design to Smell because it was opposing the Type super class. I changed the super type class method to abstract and then pushed it to child classes.

meghzman:
**Extract Method**
A Complex Conditional smell was observed in class Difference because of the method applyRemovedDiffElement which was very large. I extracted and introduced a new method out of it.

meghzman:
**Extract Class**
Usually, a class should not have multiple responsibilities, otherwise, it would cause a Multifaceted Abstraction smell. 
Class StringEscapeUtils was found to be with Multifaceted Abstraction Smell, in order to solve this, I create a new class Constant and moved all the constant fields in that class, and smell is solved.

javaparser-core/src/main/java/com/github/javaparser/ast/type/IntersectionType.java
```diff
@@ -108,6 +108,11 @@ public IntersectionType setAnnotations(NodeList<AnnotationExpr> annotations) {
         return (IntersectionType) super.setAnnotations(annotations);
     }
 
+    @Override
+    public String toDescriptor() {
+        return null;
```
jlerbsc: The expected value when calling this method is a string. I think it's better to return an empty string because if we return a null value it may throw exceptions for users.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -386,44 +386,48 @@ private void extractReshuffledDiffElements(List<DifferenceElement> diffElements)
...
-                // We expected to remove a proper node but we found a comment in between.
-                // If the comment is associated to the node we want to remove we remove it as well, otherwise we keep it
-                Comment comment = (Comment) originalElementChild.getChild();
-                if (!comment.isOrphan() && comment.getCommentedNode().isPresent() && comment.getCommentedNode().get().equals(removed.getChild())) {
-                    nodeText.removeElement(originalIndex);
-                } else {
-                    originalIndex++;
-                }
-            } else {
+    private void applyRemovedDiffElementConditionOneforIf( RemovedGroup removedGroup,Removed removed, TextElement originalElement) {
```
jlerbsc: Can you give an understandable name to this method?

meghzman:
Yes, sure I will refactor that now.
Thanks much for having a look at it

javaparser-core/src/main/java/com/github/javaparser/ast/type/IntersectionType.java
meghzman: Hello, thanks for recommending, I have made this change. Now, toDescriptor is returning empty String.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
meghzman: Yes, I have refactored it again. The method name is comapreRemovedDiffElement.

matozoid:
I have no idea what the point of all these refactorings are. It does not seem like you are trying to fix an issue with JP, but are only running a refactoring tool on the code. Since they are making the code worse on some points, I strongly advise against merging this. Please start with identifying a real issue.

jlerbsc (commented):
why make this change?

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/Statement.java
```diff
@@ -481,4 +482,13 @@ public YieldStmt asYieldStmt() {
     @Generated("com.github.javaparser.generator.core.node.TypeCastingGenerator")
     public void ifYieldStmt(Consumer<YieldStmt> action) {
     }
+
```
jlerbsc, line 485: why did you put this in statement?

meghzman:
Super class method was not generic,
making it abstract and pulling down in all clssses solved the problem of
Rebelious Design smell.

javaparser-core-testing-bdd/src/test/java/com/github/javaparser/steps/ManipulationSteps.java
```diff
@@ -270,15 +270,15 @@ public void thenMethodInClassParameterIsTypeIntCalled(int methodPosition, int cl
 
     private static class ChangeMethodNameToUpperCaseVisitor extends VoidVisitorAdapter<Void> {
         @Override
-        public void visit(MethodDeclaration n, Void arg) {
-            n.setName(n.getNameAsString().toUpperCase());
+        public void visit(MethodDeclaration methodDeclaration, Void arg) {
+            methodDeclaration.setName(methodDeclaration.getNameAsString().toUpperCase());
```
jlerbsc, line 274: It seems to me that this modification brings more readability to the code

javaparser-core/src/main/java/com/github/javaparser/ast/type/IntersectionType.java
```diff
@@ -108,6 +108,11 @@ public IntersectionType setAnnotations(NodeList<AnnotationExpr> annotations) {
         return (IntersectionType) super.setAnnotations(annotations);
     }
 
+    @Override
```
jlerbsc, line 111: There is a lot of confusion, sometimes the methods return a null value and sometimes it returns an empty string

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
jlerbsc: What is the meaning of this method name?

javaparser-core/src/main/java/com/github/javaparser/utils/StringEscapeUtils.java
```diff
@@ -32,6 +32,7 @@
  * Unescapes escaped chars in strings.
  */
 public final class StringEscapeUtils {
+    private static Constant constant;
```
jlerbsc, line 35: Personally I think that this commit does not bring anything more.

jlerbsc:
Finally I agree with @matozoid. There are too many approximations for us to integrate this PR. Thank you anyway for your suggestion. Next time I advise you to propose PRs by dissociating each of the problems. Maybe you could actually contribute to this project by proposing solutions to the problems that are already identified or those that you could identify yourself. Anyway thanks for your time.

jlerbsc:
Definitely i do not agree to integrate this PR because not only does it push inappropriate modifications but in addition it seems to me that it does not bring anything except confusion. Moreover, it pushes for a so-called improvement without highlighting the benefits it will bring or the problems it solves.

jlerbsc (commented):
In my opinion, it makes no sense to make this type of modification. This amounts to producing technical debt. Design patterns should only apply when they add value.

MysterAitch:
I mostly agree with the comments above.

There are quite a few places where some refactoring is needed in the codebase, but reading through this PR I find that the benefits of _these specific changes_ are mostly unclear (and, in places, incompatible / misleading). 

A few examples, sharing some thought process:

1. An interface `JavaParser` and a factory class `JavaParserFactory` have been added -- the names of these are ambiguous/misleading. Consider, for example, the meaning of `implements JavaParser` and how this should be interpreted by others reading this code. I suggest that they should likely contain the word `Context` to indicate that they are specific to solver contexts and not "globally used" / relevant.
2. The method `getContext()` has been added as an interface method (likely reasonable), but multiple `return null` implementations have been added -- instead, these could/should inherit this from the abstract context.
3. I am currently unsure how/why a context would have a context of its own -- my intuition suggests that it would return either itself (no need for this method) or the parent context (methods for the parent context already exist). Maybe there is a good reason to do this, but it is currently not obvious to me.
4. There are other examples where the benefit is unclear to me (introducing a singleton for the grammar letter generator, renaming a variable in one location but not others, adding the method `toDescriptor` with repeated empty implementations, etc) but, for brevity, I will stop the list here and not comment in detail on those.

In summary, introducing patterns and doing refactorings to improve the codebase are good to have and I really do welcome your contributions in this area -- however, I am closing this PR because it contains lots of small, unclear, and unrelated changes.

Do feel free to resubmit single changes with a bit more detail about the use-cases which the changes add, ideally with a brief motivation / description beyond "implements pattern x" (noting that patterns should be used to solve a problem, rather than being the goal itself).

Thank you again for spending the time to help improve JavaParser -- it really is appreciated and I look forward to seeing further PRs from you @meghzman :)

PR #3595, opened 2022-05-31
Fix lambda generic types that are always resolved to the first type param

johannescoetzee:
This PR fixes an issue where generic types in lambdas would always resolved to the type of the first parameter. For example, given a method (used in the unit test for this PR)
```
public static void sink(Float f, String s) {}

public static BiConsumer<Float, String> foo() {
    return (i1, i2) -> sink(i1, i2);
}
```
the types of `i1` and `i2` in the context of the lambda would both resolve to `java.lang.Float`, so the call to `sink` cannot be resolved. If, instead, we had
```
public static void sink(Float f, Float s) {} // <-- the types here don't match BiConsumer anymore

public static BiConsumer<Float, String> foo() {
    return (i1, i2) -> sink(i1, i2);
}
```
then the `sink` call would incorrectly resolve to the `sink` declaration provided.

For interest's sake, this fixes the problem I mentioned in my long comment on https://github.com/joernio/joern/pull/1437.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/LambdaExprContext.java
```diff
@@ -66,9 +67,10 @@ public LambdaExprContext(LambdaExpr wrappedNode, TypeSolver typeSolver) {
 
     @Override
     public Optional<Value> solveSymbolAsValue(String name) {
+        int index = -1;
         for (Parameter parameter : wrappedNode.getParameters()) {
+            index++;
```
johannescoetzee, line 72: I chose to initialize index to -1 and do the update immediately to make it more obvious what the index is referring to. Since this method is quite long, updating index at the end made it difficult to connect the value of index to what was happening in code.

jlerbsc: Hi @johannescoetzee I don't see what fixes the bug you are trying to solve.

johannescoetzee: The only thing this PR really changes is that it moves the declaration and initialization of `index` out of the `for(Parameter ...)` loop.

Before, `index` was always set to 0 inside the loop, so when it got to the statement
```
ResolvedType lambdaType = functionalMethod.get().getParamType(index);
```
on line 154, `lambdaType` was always set to the type of param 0 (hence the bug). By initializing `index` before the loop, it's now correctly incremented as parameters are checked. 

Edit: My first comment above was just to explain why I chose to initialize `index` to -1 and increment it immediately, instead of just moving it up a few lines.

jlerbsc:

PR #3611, opened 2022-06-10
Prevent Exception type Erasion

yufanyufan:
If UnsolvedSymbolException is thrown, the exception type should be preserved.


jlerbsc:
Thank you for your suggestion but it is possible that the problem you raised is more important. There are a lot of runtime exceptions thrown by JP so maybe we should fully manage this problem rather than providing a solution to a problem. For now, as a workaround, the root cause of the exception is passed so you can analyze that root cause.

yufanyufan:
What is the proposed solution to manage it fully?
This commit makes it consistent with line 298 in the same file.

jlerbsc:
Maybe an aspect based solution.

jlerbsc:
Thank you for proposing a solution to this problem and alerting us to this weakness in exception handling. A first PR has just been published to centrally manage this type of exception.

PR #3634, opened 2022-07-13
Flexible symbol resolution for declaration types

Col-E:
I wanted to create a custom declaration type to source type information from a source that isn't based in Javassist and this came to my attention. In `com.github.javaparser.symbolsolver.resolution.SymbolSolver` the method `solveSymbolInType` uses a large if chain with specific implementation types to solve for symbols. This PR refactors that to instead use a common interface. The implementation of each is is the same, just moved into the interface method implementation for each type.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
@@ -392,6 +392,9 @@ public abstract class ResolvedReferenceType implements ResolvedType,
         return Optional.empty();
     }
 
+    /**
+     * @return A copy of the current reference type, without type parameters.
+     */
     public abstract ResolvedType toRawType();
 
     /**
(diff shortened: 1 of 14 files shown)


javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedReferenceType.java
```diff
@@ -170,6 +170,11 @@ public String describe() {
      */
     public abstract ResolvedType transformTypeParameters(ResolvedTypeTransformer transformer);
 
+    /**
+     * @return A copy of the current reference type, without type parameters.
+     */
+    public abstract ResolvedType asRawType();
```
jlerbsc: toRawType() method does not do what you want?

Col-E: I'm blind, let me fix that.

jlerbsc:
Very interesting! Thank you for your proposition. Is it possible to add unit tests?

Col-E:
Since its only consolidation of existing logic, the existing Javassist and JavaParser symbol solver declaration unit tests already cover the code modified in this PR. What sort of additional testing were you thinking of adding?

jlerbsc:
It seems to be ok for me. Tnank you for your contribution.

PR #3655, opened 2022-08-09
Faster line-separator handling when reading files

abego:
In the old code every character read from a file was checked for
"line separator" in a quite expensive procedure, involving:

- creating a string from the char
- calling the method LineSeparator.lookup(String)
  - calling up to three time "...asRawString()"
  - doing up to three String comparisons
  - creating an Optional
- checking the Optional if it holds a "line separator"

These operations were executed for *every* character read.

In the new solution this detailed/expensive line separator code
is only executed when the char read from the buffer could actually
be part of a line separator (i.e. is a '\n' or '\r')

*Background*

This problem was detected when profiling the execution of
JavaParser code when analysing a large code base and running
into performance (and memory) issues.

*Test*

No extra test code required as the functionality is not changed
but the performance is improved.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/LineEndingProcessingProvider.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/LineEndingProcessingProvider.java
@@ -123,27 +123,35 @@ public class LineEndingProcessingProvider implements Provider {
                     break;
                 }
             } else {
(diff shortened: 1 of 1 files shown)


jlerbsc:
Hi @abego Can you share with us a screenshot from your profiler that demonstrates the issue you observed.

abego:

PR #3656, opened 2022-08-09
Speedup identifier validation (against 'strictfp', 'enum')

abego:
JavaParser does a very good job in validating a given source code against
the Java standard. E.g. it also validates every identifier if it is `strictfp` or
`enum` which are no legal identifier in Java (anymore).

However, this extra validation slows down the parsing, especially when
using the JavaParser to analyze large code bases that are known to be
legal Java code. This effect is that large that this validation was identified
as a significant performance issue when profiling the code in such a
setting.

To disable this identifier validation against `strictfp` and `enum` set the
system properties:
- com.github.javaparser.performance.no-strictfp-identifier-validation
- com.github.javaparser.performance.no-enum-identifier-validation

to the value `true` before loading the JavaParser, or more specifically,
before loading the ParserConfiguration class.


matozoid:
A nicer approach, I would think, is to subclass the validator in your own code and remove these validations from it.

jlerbsc:

abego:

PR #3657, opened 2022-08-09
Name Node string caching

abego:
The method `Name.asString()` is frequently used. However, calculating
its value can be quite expensive, especially when used with qualified
identifiers. In that case calling asString will always create a new
String object from the concatenation of the different parts of the
qualified name and the "." strings. This takes a considerable
amount of time and also puts pressure on the memory used.

On the other hand, recalculating the asString() value is not
necessary when we cache its value and re-use the value the
next time it is needed (and invalidate the cache when required.)

To enable this cache set the system property

- com.github.javaparser.performance.cache-node-name-string

to `true` before loading the JavaParser, or more specifically,
before loading the Name class.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
@@ -33,7 +33,12 @@ import com.github.javaparser.metamodel.JavaParserMetaModel;
 import com.github.javaparser.metamodel.NameMetaModel;
 import com.github.javaparser.metamodel.NonEmptyProperty;
 import com.github.javaparser.metamodel.OptionalProperty;
+
+import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.WeakHashMap;
+
 import static com.github.javaparser.utils.Utils.assertNonEmpty;
 
 /**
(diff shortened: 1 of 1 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/expr/Name.java
```diff
@@ -105,13 +115,22 @@ public Name setIdentifier(final String identifier) {
         }
         notifyPropertyChange(ObservableProperty.IDENTIFIER, this.identifier, identifier);
         this.identifier = identifier;
+        if (asStringOfName != null) {
+            asStringOfName.remove(this); // clear cached value for asString
+        }
```
matozoid, line 120: I'm afraid this isn't possible since it is part of generated code. The next run of the generators will wipe it out.

abego: I didn't know this was generated code. From reading the code I couldn't tell. Typically in those cases I see a special comment or so to warn not to modify generated code. 

Where do I learn more about the generator and how you build your code? Rather than subclassing and overriding extending the generator may be a different solution to make sure the enhancements keep included in the code.

jlerbsc:

abego:

PR #3658, opened 2022-08-09
Fast Name Node equals

abego:
When analyzing Java code using the JavaParser comparing
Name nodes using the equals method is a very frequent operation.
This comparison is done through the generic approach implemented in
JavaParser via the EqualsVisitor and HashCodeVisitor. In that process
the nullable field values of Name (identifier and qualifier) are wrapped
in an Optional before the actual comparison takes place. (*)

Creating multiple Optionals for every comparison between Name nodes
has a significant performance impact and also puts pressure on the
memory used (the Optionals will only have a very short live time but
will trigger frequent garbage collections, especially when working
on large code bases).

To avoid these extra Optionals the comparison between Name nodes
is now implemented by directly working on the field values, and checking
for null if necessary. This implementation in Name also speeds up the code
by avoiding overhead introduced by the Visitor pattern.

---
(*) A similar wrapping in Optional also occurs for the comment property,
inherited from Node. In this refactoring, however, the comment property
is not considered. First, this way less code is involved in the refactoring.
Secondly, it can be assumed that most Name nodes that are not equal
will already differ in the identifier and qualifier, so the comment check
is rarely executed.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
@@ -771,14 +771,7 @@ public class EqualsVisitor implements GenericVisitor<Boolean, Visitable> {
 
     @Override
     public Boolean visit(final Name n, final Visitable arg) {
-        final Name n2 = (Name) arg;
-        if (!objEquals(n.getIdentifier(), n2.getIdentifier()))
-            return false;
-        if (!nodeEquals(n.getQualifier(), n2.getQualifier()))
-            return false;
-        if (!nodeEquals(n.getComment(), n2.getComment()))
-            return false;
-        return true;
+        return n.equals(arg);
     }
 
     @Override
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/HashCodeVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/HashCodeVisitor.java
@@ -267,7 +267,7 @@ public class HashCodeVisitor implements GenericVisitor<Integer, Void> {
     }
 
     public Integer visit(final Name n, final Void arg) {
-        return (n.getIdentifier().hashCode()) * 31 + (n.getQualifier().isPresent() ? n.getQualifier().get().accept(this, arg) : 0) * 31 + (n.getComment().isPresent() ? n.getComment().get().accept(this, arg) : 0);
+        return n.hashCode();
     }
 
     public Integer visit(NodeList n, Void arg) {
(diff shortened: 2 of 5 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
```diff
@@ -771,14 +771,7 @@ public Boolean visit(final ObjectCreationExpr n, final Visitable arg) {
...
     public Boolean visit(final Name n, final Visitable arg) {
-        final Name n2 = (Name) arg;
-        if (!objEquals(n.getIdentifier(), n2.getIdentifier()))
-            return false;
-        if (!nodeEquals(n.getQualifier(), n2.getQualifier()))
-            return false;
-        if (!nodeEquals(n.getComment(), n2.getComment()))
-            return false;
-        return true;
+        return n.equals(arg);
```
matozoid, line 774: This too is generated code. You could however subclass it and override this method.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/HashCodeVisitor.java
```diff
@@ -267,7 +267,7 @@ public Integer visit(final NameExpr n, final Void arg) {
     }
 
     public Integer visit(final Name n, final Void arg) {
-        return (n.getIdentifier().hashCode()) * 31 + (n.getQualifier().isPresent() ? n.getQualifier().get().accept(this, arg) : 0) * 31 + (n.getComment().isPresent() ? n.getComment().get().accept(this, arg) : 0);
+        return n.hashCode();
```
matozoid, line 270: Generated code.

javaparser-core/src/main/java/com/github/javaparser/ast/visitor/EqualsVisitor.java
abego: I could subclass and overriding this method (and the others) however I don't know how I can make this code being used by the JavaParser. 

From what I see there is a private SINGLETON in EqualsVisitor, that is used in the static "equals" method of EqualsVisitor and that equals method is used in Node.equals. As the AST is build from code subclassing Node I would need to make this Node.equals method use "my" EqualsVisitor subclass, so the Node subclasses inherit the new behaviour. How do I do this?

jlerbsc:

abego:
> Can you share with us a screenshot from your profiler that demonstrates the issue you observed.

I didn't saved the profiler results while I was working on the performance of my analyser tool that is using JavaParser. I tried to reproduce these profiler results after I read your request but it turns out to be difficult. 

I tried various approaches to get you the profiling results, with no convincing results, at least not as convincing as I recall them when I implemented the improvements. I guess it has to do with the way I worked on this. In general I used the "most obvious hotspot" indicated by the profiler(s) to look for a faster implementation. As typical with this approach this makes the one hotspot go away and a new hotspots moves to the top. And later changes may make earlier changes less relevant. I guess this happened here: with a very late (and large) change to my analyser tool I could drastically reduced the memory requirements of the application, leading to a good-enough performance. With less memory pressure the previous changes no longer seem to be that relevant. This became very obvious when I compared the performance of my tool using the original JavaParser code and my patched one:  14:50 min vs. 12:36 min. I.e. my enhancements to JavaParser speed up the execution not even by 20%.

As I have such a hard time to reproduce the scenarios that justify the changes, applying the changes is not easy because because of your generator approach and the performance of the official JavaParser version is sufficient for my tool I withdraw my pull requests.

PR #3712, opened 2022-09-27
Remove white space after a comment during comment deletion.

eak24:
This change is to remove all whitespace tokens after a comment removal. Without this, if you removed a comment from this:
``` java
public class Test {
  /**
  * Hello
  */
  @MyAnnotation
  public void foo(Bar x, Bar y) {
  }
}
```
you would get:
``` java
public class Test {
    @MyAnnotation
  public void foo(Bar x, Bar y) {
  }
}
```
with the `@MyAnnotation` shifted over by the amount of whitespace between the end of the comment and the start of the next non-whitespace token. With this change, all that whitespace is 'eaten up' and the `@MyAnnotation` shows up with the correct indentation.


eak24:
I think this may be a duplicate of #3694... Feel free to close if that is the case. Seems like that solution is more generic.

jlerbsc:
This problem seems to be fixed in this version

PR #3713, opened 2022-09-27
Accept final in instanceof pattern

PPazderski:
Javaparser does not accept final in an instanceof pattern but from my understanding of JLS it should be possible.
The following snippet is rejected by javaparser with an unexpected token "final" error but the same code can be compiled with javac and Eclipse compiler.

```
Object o = null;
if (o instanceof final String str) {
   // use string
}
```

### DIFF
--- a/javaparser-core/src/main/javacc/java.jj
+++ b/javaparser-core/src/main/javacc/java.jj
@@ -3278,13 +3278,15 @@ Expression EqualityExpression():
  */
 PatternExpr PatternExpression():
 {
+    ModifierHolder modifier;
     ReferenceType type;
     SimpleName name;
 }
 {
-    type = AnnotatedReferenceType()
+    modifier = Modifiers()
+    type = ReferenceType(modifier.annotations)
     name = SimpleName()
-    { return new PatternExpr(range(type, token()), type, name); }
+    { return new PatternExpr(range(type, token()), modifier.modifiers, type, name); }
 }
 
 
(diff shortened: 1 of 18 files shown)


javaparser-core/src/main/javacc/java.jj
```diff
@@ -3278,13 +3278,15 @@ Expression EqualityExpression():
...
 PatternExpr PatternExpression():
 {
+    ModifierHolder modifier;
     ReferenceType type;
     SimpleName name;
 }
 {
-    type = AnnotatedReferenceType()
+    modifier = Modifiers()
+    type = ReferenceType(modifier.annotations)
```
jlerbsc, line 3287: Why did you change the type?

jlerbsc: Looks like a good job to me. Maybe @MysterAitch will take a look at this PR.

PPazderski: At first, because `InstanceOfExprTest.annotationsOnTheType_patternExpression` failed otherwise. From my understanding AnnotatedReferenceType is just a helper to parse annotation and then create a ReferenceType with those annotations.
```
ReferenceType AnnotatedReferenceType():
{
    NodeList<AnnotationExpr> annotations;
    ReferenceType type;
}
{
    annotations = Annotations()
    type = ReferenceType(annotations)
    { return type; }
}
```
But since `Modifiers()` also parses annotations the `Annotations()` in the later `AnnotatedReferenceType()` couldn't find anything because annotations where already parsed by `Modifiers()`. That's why I took the annotations from the `ModifierHolder` instead.

jlerbsc: For me it's good. Good work. I'll give @MysterAitch a chance to look at your proposal as well.

jlerbsc:

PR #3714, opened 2022-09-29
Fix some yield expressions not recognized

PPazderski:
Existing parser fails to recognize some yield expressions. Especially yielding the result of a method call or the result of a conditional expression.
E.g. the following snippet currently fails:
```
int x = switch (5) {
    default -> {
        yield getX();
    }
};
```

The problem I found is that yield is not a keyword but just a restricted identifier or since Java 17 called contextual keyword. Javaparser currently tries to parse the statement as `VariableDeclarationExpression`.

This pull request is a minimal change to fix the problem of yield statements not recognized. For a better solution I think javaparser would need to better different between Identifier and TypeIdentifier according to https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.8
Then the `VariableDeclarationExpression` would start with a type which cannot be the name "yield".
If a class could be named "yield" a statement like `yield x = null;` is ambiguous and could be a local variable declaration plus initialization or a yield statement whose expression is an assignment.
But I expect this to be a notably larger and riskier change.

PS: if someone wants to try the `SwitchExprTest.yieldYield` example, the following code compiles:
```
public class Main {

    public enum X {
        yield,
    }

        X yield = X.yield;
        yield = switch (yield) {
        default -> {
            yield yield = yield;
        }
        case yield -> {
            yield yield == yield ? yield : yield;
        }
        };
    }
}
```

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
@@ -24,7 +24,8 @@ package com.github.javaparser.ast.expr;
 import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.stmt.BlockStmt;
 import com.github.javaparser.ast.stmt.SwitchEntry;
-import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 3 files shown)


jlerbsc:

PPazderski:
Thank you for merging my pull requests.

PR #3717, opened 2022-10-06
Fix maven wrapper not found in generator scripts

PPazderski:
I assume since commit 54fe277ca4e5b9ce5b342a3a2aef6335e70a6e02 the two helper scripts `run_core_generators.sh` and `run_core_metamodel_generator.sh` do not work correctly anymore but produce the following error:
```
./run_core_generators.sh: line 10: ./mvnw: No such file or directory
```
The other helper scripts should be fine.


jlerbsc:

Issue #3415, opened 2021-11-22
align StaticJavaParser and SourceRoot parser config APIs


ericprud:
``` java
ParserConfiguration pc1 = StaticJavaParser.getConfiguration();
ParserConfiguration pc2 = new SourceRoot(...).getParserConfiguration();
```
Perhaps call them both `getParserConfiguration()` (would imply an update to this javaparser-visit book and git repo).

MysterAitch:
Sounds like a good idea! 

I've marked it as an Easy first change -- anybody looking at this, I would request marking `getConfiguration` as deprecated rather than removing it.

nelson-ng-96:
Hi, I would like to work on this issue for my university project. Is it possible for this issue to be assigned to me?

jlerbsc:
This issue has been resolved.

ericprud:
I lost my copy somewhere in the bowels of my file system, but should the authors of [JavaParser: Visited](https://leanpub.com/javaparservisited) be informed?

jlerbsc:
> I lost my copy somewhere in the bowels of my file system, but should the authors of [JavaParser: Visited](https://leanpub.com/javaparservisited) be informed?

I'm not sure whether this book is regularly updated by its author(s), who are no longer contributing to the project.

PR #3722, opened 2022-10-19
Issue #3415

nelson-ng-96:
Added deprecated for the method getConfiguration to fix issue #3415.


nelson-ng-96:
Added a new method in StaticJavaParser which aligns with the method `getParserConfiguration()` naming in SourceRoot. 
Made the deprecated method in StaticJavaParser to call the new method in StaticJavaParser.

jlerbsc:

PR #3723, opened 2022-10-19
Fix for ReflectionAnnotationDeclaration getClassName()

Blackgen:
I've noticed that the returned ClassName for ReflectionAnnotationDeclaration contains a dot.
In the initial commit for the PR (#1833) the dot was stripped, in the merged commit however the method was changed. And the classname was not returned correctly.

I've adjusted the method getClassName() to strip the dot.

I've added the following testcase:
```java
  @Test
  void getClassName_shouldReturnCorrectValue() {
    ReflectionAnnotationDeclaration annotation =
        (ReflectionAnnotationDeclaration) typeSolver.solveType(
            "com.github.javaparser.symbolsolver.reflectionmodel.WithField");
    assertEquals("WithField", annotation.getClassName());
  }
```

Previously the method returned ".WithField" instead of "WithField"


jlerbsc:

Issue #3721, opened 2022-10-18
Removing a node with LexicalPreservingPrinter causes UnsupportedOperationException, Part 2


adrianus1:
Using Java Parser 3.24.4
There are several issues with node removal and LexicalPreservingPrinter:

1. removing last variable in filed declaration (see below) causes UnsupportedOperationException
2. removing nodes causes UnsupportedOperationException
3. removing statements does not actually remove anything
All three issues do not occur for Pretty Printing.

All three issues reproduce all the time, but I could only reproduce the issue 1. in a somewhat small example, see code below.

```
public class ReproduceLfException {
    final static String source =
            "public class Temp {\n" +
            "    String message;\n" +
            "}\n";

    public static void main(String[] args) {
        CompilationUnit cu = StaticJavaParser.parse(source);
        LexicalPreservingPrinter.setup(cu);

        // remove the field variable
        List<ClassOrInterfaceDeclaration> classList = cu.findAll(ClassOrInterfaceDeclaration.class);
        for(ClassOrInterfaceDeclaration clazz : classList) {
            List<Node> nodes = clazz.getChildNodes();
            for(Node node : nodes) {
                if (node instanceof FieldDeclaration) {
                    FieldDeclaration fieldDeclaration = (FieldDeclaration)node;
                    NodeList<VariableDeclarator> variableDeclarators = fieldDeclaration.getVariables();
                    ListIterator<VariableDeclarator> variableDeclaratorListIterator =
                            variableDeclarators.listIterator();
                    while (variableDeclaratorListIterator.hasNext()) {
                        VariableDeclarator variableDeclarator = variableDeclaratorListIterator.next();
                        if (variableDeclarator.getNameAsString().equals("message")) {
                            variableDeclaratorListIterator.remove();
                        }
                    }
                }
            }
        }

        System.out.println(LexicalPreservingPrinter.print(cu));
    }
}
```
Issue 1. disappears when in source the "String" is replaced by "int", or when another variable is added, for instance "String, message, log;"

I will report the other issues if I manage to isolate them into small examples.

There is already another issue (#3440) but it should have been solved in 3.24.4.

Is this a bug or is the usage of Java Parser not correct?

herbert-venancio:
I found the same issue while trying to modify the type of a variable. Here is a minimum code to reproduce the exception:
```java
package bug;

import org.junit.jupiter.api.Test;

import com.github.javaparser.JavaParser;
import com.github.javaparser.ParseResult;
[...]

PR #3726, opened 2022-10-24
Fix part of issue #3721 UnsupportedOperationException while trying to modify the type of a variable

jlerbsc:
By mistake this PR also contains some refactoring which should have been the subject of another PR


jlerbsc:
Fix part of issue #3721 UnsupportedOperationException while trying to modify the type of a variable

Issue #820, opened 2017-03-04
Increase code coverage


ftomassetti:
We have currently a code coverage of ~60%. We could look into increasing it, perhaps focusing on low hanging fruits. I do not think it makes sense to aim at a very high value but I suppose that with a limited effort we could move the needle.

matozoid:
Shall I tag this "easy"?

ftomassetti:
Yes, I think many tests would be easy to write. Maybe we should just point to where to read the code coverage.

matozoid:
Some thoughts:
- I'm not sure the code generated by JavaCC is included correctly. We never seem to get much above 60%
- a lot of code in src/main/java is generated by us. I don't feel a need to test all of it, so coverage will never be high

JulienVig:
Hi! Is this issue still open?

matozoid:
Yeah, a few more tests are always welcome!

AnnaSkantz:
Hi,
We are a group of students who would like to work on this issue as part of a group assignment.

jlerbsc:
Hi @AnnaSkantz you can submit PR as you wish!

ghost:
Hello, some other students and I would like to work on this for a university project. What would be some good code for us to test and can we get this issue assigned to us? Or should we just submit a PR with our additions without this issue getting assigned?

jlerbsc:
You can indeed propose improvements. Thank you for asking the problem you are encountering or trying to solve.

YinyinChencr7:
Hi, can I work on this issue?

jlerbsc:
Your help will be appreciated.

soybeaam:
Hello!

I was wondering if my team and I could still work on this issue.

jlerbsc:
Hi @soybeaam Thank you for your proposition, but there is no immediate need at this time. First, the coverage ratio should be analysed and the cases with the highest added value identified.

PR #3729, opened 2022-10-25
More unit tests for VisitorList, VisitorMap, and VisitorSet

ghost:
Unit tests to contribute to #820. 

The `toString` test for VisitorSet currently fails - this was intentional. From the looks of things, the current implementation deletes the last character for the last element unintentionally. I have just written the test based on how `toString` works for the other classes - that is, there should be a space after the commas separating the elements. I figured having that space would fix the problem of the missing character since that space would get deleted instead. 

If you would like this test changed however (or any for that matter) I'd be happy to.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
@@ -162,7 +162,7 @@ class VisitorSetTest {
         CompilationUnit x2 = parse("class Y{}");
         set.add(x2);
 
-        String[] validStrings = new String[]{"[" + x1 + "," + x2 + "]", "[" + x2 + "," + x1 + "]"};
+        String[] validStrings = new String[]{"[" + x1 + ", " + x2 + "]", "[" + x2 + ", " + x1 + "]"};
         boolean valid = false;
         String setString = set.toString();
         for (String s : validStrings) {
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
@@ -142,7 +142,7 @@ class VisitorSetTest {
     }
 
     @Test
-    void containsAllFalse() {
+    void visitorSetContainsOnlyPart() {
         List<CompilationUnit> list = new ArrayList<>();
         list.add(parse("class X{}"));
         list.add(parse("class X{}"));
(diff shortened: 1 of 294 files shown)


javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
```diff
@@ -138,4 +137,42 @@ void visitorSetToArray() {
...
+                new ObjectIdentityEqualsVisitor());
+        assertTrue(set.isEmpty());
+
+        set.addAll(list);
+        list.add(parse("class X{}"));
+        assertFalse(set.containsAll(list));
+    }
+
+    @Test
+    void visitorSetToString() {
```
jlerbsc, line 159: Maybe you can use Parameterized  tests here

javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
```diff
@@ -138,4 +137,42 @@ void visitorSetToArray() {
         for (CompilationUnit u : set.toArray(new CompilationUnit[2]))
             assertTrue(set.contains(u));
     }
+
+    @Test
+    void containsAllFalse() {
```
jlerbsc: Please can you keep the method naming convention used in this test class?
Can you rename this method for example visitorSetContainsOnlyPart.

javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
```diff
@@ -162,7 +162,7 @@ void visitorSetToString() {
         CompilationUnit x2 = parse("class Y{}");
         set.add(x2);
 
-        String[] validStrings = new String[]{"[" + x1 + "," + x2 + "]", "[" + x2 + "," + x1 + "]"};
+        String[] validStrings = new String[]{"[" + x1 + ", " + x2 + "]", "[" + x2 + ", " + x1 + "]"};
```
jlerbsc: I think it is this pattern that should be modified by using for example the join functionality in streams.

> for (EqualsHashcodeOverridingFacade facade : innerList) {
            sb.append(facade.overridden.toString() + ", ");
        }
        return sb.replace(sb.length() - 2, sb.length(), "]").toString();

ghost: To clarify sir, when you say the pattern that should be modified, are you referring to the `StringBuilder` for loop in the `toString` method or the way `validStrings` are created in the test? Or perhaps both should be modified?

jlerbsc: Hi @flanbino both should be modified in particular i refer to this code that add a separator and finally replace the last separator.

ghost: Hello again, I've made the changes requested

jlerbsc:
Sounds good to me.

PR #3733, opened 2022-10-29
3586 method declaration duplicate comments

### LINKED ISSUE #3586: foramt code has issue when method declaration has comment
there is a single line comment("**// this is a test**") in the method declaration
图片

`
public class TestClass {`
`    public void test(String p1, String p2) // this is a test`
`    {`
`        System.out.println("test");`
`    }`
`}
`

use below code to format:(the content of a is above code):
`CompilationUnit cu = StaticJavaParser.parse(a);`
`System.out.println(cu.toString());`

the output is:
图片

the line of method declaration split to 4 lines and have 3 same comments

my questions is: how to fix this issue

Bash-09:
Fixes #3586.
Moved from https://github.com/comp2120-workshop11-group1/javaparser/pull/2

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser;
 
+import static com.github.javaparser.StaticJavaParser.parse;
 import static com.github.javaparser.utils.TestUtils.assertEqualToTextResource;
 import static com.github.javaparser.utils.TestUtils.assertEqualToTextResourceNoEol;
 import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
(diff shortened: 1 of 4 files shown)


Bash-09:
@jlerbsc
> Hi @Bash-09 it seems to be correct but the unit test doesn't test comment insertion. You can learn from the class CommentsInserterTest.

I've added another test which is hopefully in line with what you had in mind, let me know if not.
I also rebased and moved to a PR on the main repo for if you think my solution is suitable.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
```diff
@@ -1757,4 +1757,23 @@ void testRemovingInlinedAnnotation_alternate_case() {
 
     }
 
+    @Test
+    void testAtypicalCommentAssignment() {
```
jlerbsc, line 1761: I think this test is not very useful because it just prints a class that has not been modified.

javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
```diff
@@ -112,4 +113,24 @@ void issue412() throws IOException {
         CompilationUnit cu = parseSample("Issue412").getResult().get();
         assertEqualToTextResourceNoEol(makeExpectedFilename("Issue412"), cu.toString());
     }
+
+    @Test
+    void lineCommentAssignedToLastNode() {
```
jlerbsc, line 118: Instead, the test should verify that the comment is backed to the correct node. Here the test is not sufficient because it tests that the pretty printer displays the comment on the last node. In addition, the test class is badly chosen to carry out this test which concerns the pretty printer.

jlerbsc:
Can you reorganize your commits because we no longer understand the process.

jlerbsc:
I am closing this PR because another solution has been found to solve this problem. Thanks for your proposition.

Issue #2681, opened 2020-05-17
Verify the behaviours of `Position`


MysterAitch:
Reading through `Position.java`, there seem to be some strange behaviours and areas for improvement. 

- The constructor allows a column of 0, but the first column is explicitly stated to be 1 in multiple places
- `orIfInvalid()` -- when both `this` and the "other" position are both invalid, it returns "this" position.. is this intentional? My interpretation is that it should return the "other" position, even if that is invalid.
- Introducing a static final field for the first line and first column would allow other places to use this (this already exists with HOME, but)
- Why is `Node.ABSOLUTE_END_LINE` within the class `Node` as opposed to `Position`? Ditto `Node.ABSOLUTE_BEGIN_LINE`? It seems like Node pre-dates Position so it might just not have moved across yet.

Possibly some other thoughts/comments to add too.

joakimAbdinurIusuf:
Hi, we are a group of students from KTH in Sweden and would like to work on this issue as part of a group assignment :)

ameliagenova:
Hello, my team and I would like to help with this issue, could I please be assigned to this issue? Thank you!

ameliagenova:
Hello, regarding this issue, my team and I wanted to clarify and check our understanding:  

- With the first dot point, we believe that in the constructor, the conditional statements should check if the input line and column are greater than FIRST_LINE and FIRST_COLUMN instead of the ABSOLUTE_END_LINE since this will prevent permission for negative column and line numbers as well as ensure they start from 1. Would this be the correct understanding? 

- With the last three dots points, the forked branch AnnaSkantz/javaparser has already addressed these issues; however, there is no open pull request for this. Would you like us to develop further from this branch and improve on their changes or start from the master branch?

Hope to hear back from you soon, thank you!

jlerbsc:
Hi @ameliagenova ,
> * With the first dot point, we believe that in the constructor, the conditional statements should check if the input line and column are greater than FIRST_LINE and FIRST_COLUMN instead of the ABSOLUTE_END_LINE since this will prevent permission for negative column and line numbers as well as ensure they start from 1. Would this be the correct understanding?

I confess that I do not understand what the ABSOLUTE_BEGIN_LINE value means but it is possible that it represents an uninitialized position. To know what the impact of your modification is, you should run all the unit tests and see in which case there may be errors.

jlerbsc:
[...]

PR #3734, opened 2022-10-29
Position

ameliagenova:
Attempted to fix 3 out of the four dot points for issue #2681, and also added unit tests for the position class to test the changes made and to test for code in the position class that was not yet tested. We could not complete the first dot point as it caused errors in the rest of the existing codebase, and resolving this would require an in-depth understanding of the codebase. From our analysis, a suitable solution could not be found to fix the first dot point for the above-mentioned issue.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
@@ -3,6 +3,7 @@ package com.github.javaparser;
 import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
 
 public class PositionTest {
 
@@ -18,6 +19,23 @@ public class PositionTest {
 
         assertEquals(p1, invalid.orIfInvalid(p1));
         assertEquals(invalid2, invalid2.orIfInvalid(invalid));
+    }
+
+    @Test
+    public void testPositionExceptionMessage() {
+        try {
+            Position p = new Position(-10, 1);
+            fail("Created " + p + " without exception.");
+        } catch (IllegalArgumentException e) {
+            assertEquals("Can't position at line -10", e.getMessage());
+        }
+
+        try {
+            Position p = new Position(1, -10);
+            fail("Created " + p + " without exception.");
+        } catch (IllegalArgumentException e) {
+            assertEquals("Can't position at column -10", e.getMessage());
+        }
     }
 
 }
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/Position.java
```diff
@@ -134,31 +135,26 @@ public boolean invalid() {
...
     public Position orIfInvalid(Position alternativePosition) {
         assertNotNull(alternativePosition);
-        // TODO: Why the || ?
-        //  ... It seems that if both this and the alternative are invalid, then we return this..?
-        if (valid() || alternativePosition.invalid()) {
+        if (this.valid()) {
             return this;
         }
-        return alternativePosition;
+        return alternativePosition.valid() ? alternativePosition : this;
```
jlerbsc, line 143: @MysterAitch comments "orIfInvalid() -- when both this and the "other" position are both invalid, it returns "this" position.. is this intentional? My interpretation is that it should return the "other" position, even if that is invalid.". Here the method returns always this if the alternative position is not valid. Is it your intention? Can you add the note in comment.

javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class PositionTest {
+
+    @Test
+    public void testOrIfInvalid() {
+        Position p1 = new Position(1, 1);
```
jlerbsc, line 12: There are only tests for the orIfInvalid method does this mean that the other modified methods are sufficiently tested?

javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
```diff
@@ -20,4 +21,21 @@ public void testOrIfInvalid() {
         assertEquals(invalid2, invalid2.orIfInvalid(invalid));
     }
 
+    @Test
+    public void testPositionExceptionMessage() {
```
jlerbsc: Can you use assertThrows instead of try/catch?
Is the intent of the test to verify the error message or to ensure that an incorrect instance of the Position class cannot be created.
The method name should describe the intent of the test.

javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
ameliagenova: Yes, the IntelliJ code coverage tools show that tests hit the other methods in position. If there is any particular you would like tested we are happy to write more tests.

javaparser-core-testing/src/test/java/com/github/javaparser/PositionTest.java
ameliagenova: Yes sure, this change has been made. The intention of this test is to verify the exception message format to alleviate concerns from this comment  https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/Position.java#L64

jlerbsc:

PR #3736, opened 2022-10-29
More unit tests for JavaToken and CodeGenerationUtils

ameliagenova:
Unit tests to contribute to #820. 

Hello, I'm one of the students in the group @flanbino was talked about on this issue.

I have created more unit tests for JavaToken and CodeGenerationUtils classes. In the JavaTokenTest, for testFindLastToken, I am unsure of the intended output.  It does not appear to be currently used in the implementation of JavaParser, findLastToken() is used in one of the unused methods in TokenRange, toTokenRange(), however, I am unsure if it's supposed to reflect the getEnd() JavaToken in TokenRange class or just always be EOF, hence it is commented out as the test proposed will fail since it is based that findLastToken would return the same value as getEnd() which is currently not the case.

If you would like this any of these tests to be changed I'd be happy to make any changes from the suggested reviews.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/JavaTokenTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/JavaTokenTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser;
 
+import com.github.javaparser.ast.CompilationUnit;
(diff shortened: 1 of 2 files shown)


jlerbsc:
This is OK you have just to refresh your branch with the base branch.

jlerbsc:

PR #3737, opened 2022-10-29
#2625: Add messages to thrown exceptions

### LINKED ISSUE #2625: Add messages to thrown exceptions
We have quite a few calls to `new UnsupportedOperationException()` (and similar) without a message to describe why it is unsupported. 

In most cases this will be because an inherited method is not appropriate for that particular sub-class based on the rules/logic of Java / the AST.

In other cases this will be because it's not yet implemented or similar.

Adding in messages to these calls will make the distinction more visible / apparent to any users that come across this, and future developers who are looking at those bits of code.

nelson-ng-96:
Fix #2625. I added about 15 messages to `new UnsupportedOperationException()` calls. This is an incremental improvement to the issue.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
@@ -1040,8 +1040,7 @@ public class Difference {
                 return childTextElement.getChild() == csmChild.getChild();
             }
         } else {
-            throw new UnsupportedOperationException("CsmElement is neither an instance of CsmToken or CsmChild. Throw" +
-                    " in isCorrespondingElement method in Difference class.");
+            throw new UnsupportedOperationException("CsmElement is " + csmElement.getClass());
         }
 
         return false;
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
@@ -1040,7 +1040,8 @@ public class Difference {
                 return childTextElement.getChild() == csmChild.getChild();
             }
         } else {
-            throw new UnsupportedOperationException("CsmElement is " + csmElement.getClass());
+            throw new UnsupportedOperationException("Operation only supported for instance of CsmToken and CsmChild. " +
+                    "The CsmElement is " + csmElement.getClass());
         }
 
         return false;
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
@@ -1040,7 +1040,8 @@ public class Difference {
                 return childTextElement.getChild() == csmChild.getChild();
             }
         } else {
-            throw new UnsupportedOperationException();
+            throw new UnsupportedOperationException("CsmElement is neither an instance of CsmToken or CsmChild. Throw" +
+                    " in isCorrespondingElement method in Difference class.");
         }
 
         return false;
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
@@ -170,7 +170,7 @@ public class LexicalPreservingPrinter {
                             nodeText.removeElement(index);
                         }
                     } else {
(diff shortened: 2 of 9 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Difference.java
```diff
@@ -1040,8 +1040,7 @@ private boolean isCorrespondingElement(TextElement textElement, CsmElement csmEl
                 return childTextElement.getChild() == csmChild.getChild();
             }
         } else {
-            throw new UnsupportedOperationException("CsmElement is neither an instance of CsmToken or CsmChild. Throw" +
-                    " in isCorrespondingElement method in Difference class.");
+            throw new UnsupportedOperationException("CsmElement is " + csmElement.getClass());
```
jlerbsc: This new message is certainly not clear enough.

javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedTypeParameterDeclaration.java
```diff
@@ -93,12 +93,14 @@ public String toString() {
...
             public Optional<ResolvedReferenceTypeDeclaration> containerType() {
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException("containerType method is unsupported in " +
+                        "ResolvedTypeParameterDeclaration class.");
             }
 
             @Override
             public ResolvedReferenceType object() {
-                throw new UnsupportedOperationException();
+                throw new UnsupportedOperationException("object method is unsupported in " +
```
jlerbsc, line 101: Object method is **not supported**
There is no added value to specify the class because it is already in the stacktrace

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -331,7 +331,8 @@ public void concreteListChange(NodeList<?> changedList, ListChangeType type, int
             } else if (type == AstObserver.ListChangeType.ADDITION) {
                 differenceElements = LEXICAL_DIFFERENCE_CALCULATOR.calculateListAdditionDifference(findNodeListName(changedList), changedList, index, nodeAddedOrRemoved);
             } else {
-                throw new UnsupportedOperationException("ListChangeType is " + type.name());
+                throw new UnsupportedOperationException("ListChangeType is not listed in ENUM of Removal and " +
```
jlerbsc, line 277: Or "Unknown change type " + type.name()

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -453,7 +453,7 @@ public ResolvedType visit(UnaryExpr node, Boolean solveLambdas) {
     @Override
     public ResolvedType visit(VariableDeclarationExpr node, Boolean solveLambdas) {
         if (node.getVariables().size() != 1) {
-            throw new UnsupportedOperationException();
+            throw new UnsupportedOperationException("The number of variables in VariableDeclarationExpr node is not 1");
```
jlerbsc, line 452: ??? This message is visible to the user. It should be clearer.

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedWildcard.java
```diff
@@ -106,7 +106,7 @@ public String describe() {
         } else if (type == BoundType.EXTENDS) {
             return "? extends " + boundedType.describe();
         } else {
-            throw new UnsupportedOperationException("BoundType is " + type.name());
+            throw new UnsupportedOperationException("The BoundType is not null, Super or Extends. BoundType is " + type.name());
```
jlerbsc, line 107: The BoundType must be  null, Super or Extends

jlerbsc:
I cannot validate all the messages but in general they must be clear for the user and sufficiently precise.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
jlerbsc: A clearer message would be for example "TypeExtractor support only one variable on a variable declaration expression".

javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedWildcard.java
jlerbsc: Please can you check that the messages are clearer.

PR #3738, opened 2022-10-29
Inserting at specified layout locations in NodeWithMembers

### LINKED ISSUE #2458: Is there a layout api?
Hello, by default, the new field is written to the end of the class, the code writing specifications are not friendly. Is there a similar layout API? For example, the class is divided into two parts, the upper part is the field area, and the lower part is the function area. All fields are added to the end of the field area, and the function is the same

6000j:
Fixes https://github.com/javaparser/javaparser/issues/2458 using a more generic solution, as requested recently.

Implements addFieldAtLocation and addMethodAtLocation methods in NodeWithMembers that allow for inserting a new FieldDeclaration/MethodDeclaration at a specified point relative to the other pre-existing elements in the NodeList. Uses a new enum NodeWithMembers.LocationType to specify what sublist to position relative to.

Includes unit tests in NodeWithMembersTest for the two methods.

(apologies for second PR, github does not allow for changing the branch to pull from in a PR)

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithMembers.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithMembers.java
@@ -1,6 +1,6 @@
 /*
  * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
- * Copyright (C) 2011, 2013-2021 The JavaParser Team.
+ * Copyright (C) 2011, 2013-2022 The JavaParser Team.
  *
  * This file is part of JavaParser.
  *
@@ -32,6 +32,7 @@ import com.github.javaparser.ast.type.VoidType;
 import java.util.List;
 import java.util.Optional;
 
+import static com.github.javaparser.StaticJavaParser.parse;
 import static com.github.javaparser.StaticJavaParser.parseType;
 import static com.github.javaparser.ast.Modifier.Keyword;
 import static com.github.javaparser.ast.Modifier.Keyword.*;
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/ast/nodeTypes/NodeWithMembers.java
```diff
@@ -275,6 +338,48 @@ default MethodDeclaration addMethod(String methodName, Keyword... modifiers) {
         return methodDeclaration;
     }
 
+    /**
+     * Add a field to this at the specified location, with options for before/after.
```
jlerbsc, line 342: A field or a method?

jlerbsc:
Now it's clearer we can start discussing features.

First i don't see any point in providing an api to place a field or method at a particular position. To my knowledge no one has ever asked for this type of functionality. In addition, it seems difficult to implement for a user. The possibility of indicating a particular position should be deleted.

Second, rather than statically defining a way to place the fields and methods (for example addMethodAtLocation/addFieldAtLocation), you could perhaps come up with a more generic way (e.g. with a functional interface - a consumer). The client could implement its own strategy or use one of the default strategies that would be proposed (before/after a field or a method). It also requires the current method to rely on a default strategy.
If my explanation is too confusing, I can send you the signature of the method that would implement these cases or i can offer you a solution but I'm not sure if this is what is expected.


Thirdly the strategy must carry the implementation instead of having a switch (open close principle). For example

    enum LocationType {
        AFTER {
            // implementation of the add after method
        },
        BEFORE {
            // implementation of the add before method
        };
        // define an abstract method implemented by all members
   }

After all your work I will ask you again to organize your commits so that your work can be reviewed easily.

PR #3740, opened 2022-10-30
Added Optional method in SymbolReference

4everTheOne:
When i was working with the code noticed an inconsistency in SymbolReference method comparing with other classes.
Most of the methods that can return null return an optional to allow the programmer to check when the value is present or not.
This method was not following this guideline requiring the use of an additional method isSolved() before the call of _getCorrespondingDeclaration()_.

This Pull Request adds support for the Optional, allowing a more consistent pattern between methods.

This PR is similar to #3044 BUT does not affect the Public API.

Let me know your thoughts!

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
@@ -65,11 +65,6 @@ public class SymbolReference<S extends ResolvedDeclaration> {
         return new SymbolReference<>(null);
     }
 
-    @Override
-    public String toString() {
-        return "SymbolReference{" + correspondingDeclaration + "}";
-    }
-
     /**
      * Get the declaration associated with the Symbol.
      *
@@ -104,6 +99,11 @@ public class SymbolReference<S extends ResolvedDeclaration> {
     @Deprecated
     public boolean isSolved() {
         return getDeclaration().isPresent();
+    }
+
+    @Override
+    public String toString() {
+        return "SymbolReference{" + correspondingDeclaration + "}";
     }
 
     public static <O extends ResolvedDeclaration> SymbolReference<O> adapt(SymbolReference<? extends O> ref, Class<O> clazz) {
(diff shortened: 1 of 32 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
```diff
@@ -33,54 +34,81 @@
...
+    private SymbolReference(@Nullable S correspondingDeclaration) {
         this.correspondingDeclaration = correspondingDeclaration;
     }
 
     /**
      * Create a solve reference to the given symbol.
      */
     public static <S extends ResolvedDeclaration, S2 extends S> SymbolReference<S> solved(S2 symbolDeclaration) {
-        return new SymbolReference<S>(Optional.of(symbolDeclaration));
+        return new SymbolReference<>(symbolDeclaration);
```
jlerbsc, line 41: Can this method declaration be simplified?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
```diff
@@ -33,54 +34,81 @@
...
+    public static <S extends ResolvedDeclaration> SymbolReference<S> unsolved() {
+        return new SymbolReference<>(null);
     }
 
     /**
      * Create an unsolved reference specifying the type of the value expected.
      */
     public static <S extends ResolvedDeclaration, S2 extends S> SymbolReference<S> unsolved(Class<S2> clazz) {
-        return new SymbolReference<>(Optional.empty());
+        return new SymbolReference<>(null);
```
jlerbsc: Can this method declaration be simplified?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
4everTheOne: It's a good question. Simplifying is not easy, since it impacts the api. For example, tried to do that in commit c4e42df but it may break the current API.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/model/resolution/SymbolReference.java
4everTheOne: It's a good question. Simplifying is not easy, since it impacts the api.
I think the best approach would be leave it as it is, and consider it as deprecated :)

PR #3744, opened 2022-11-01
Updated workflow to only run one job per PR

4everTheOne:
At the moment, if a contributor opens a PR and is constantly pushing changes to the branch multiple actions will be created for that PR, making the build process slow for everyone.

This PR suggests a change in this behavior, by only building the latest commit. If a build is in progress and the author submits a new change, the current build will be aborted and a new build will start with all the changes.

### DIFF
--- a/.github/workflows/maven_tests.yml
+++ b/.github/workflows/maven_tests.yml
@@ -18,6 +18,12 @@ on:
   # Enable manual triggering (important for contributors to enable a check on their fork)
   workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


4everTheOne:
An example of the build being cancelled can be seen at: https://github.com/javaparser/javaparser/actions/runs/3369330764
The previous action was cancelled due to a new commit being publish: https://github.com/javaparser/javaparser/actions/runs/3369332925

(This was trigger by updating this branch with the master)

PR #3747, opened 2022-11-04
Suggestion to restructure APIs

### LINKED ISSUE #3108: Simplify the use of JavaParser

4everTheOne:
## Motiviation

By looking into thee documentation provided @Matozoid at issue #3108, it's clear that we have two ways of interacting with JavaParser

- Quick API
- Full API

Both APIS have strong and weak points regarding. Looking at Full API (JavaParser) we can reuse the same instance for each parse, allowing to improve the performance and in some cases reduce resources footprint. But this comes with a cost, of having a API that may difficult the on-boarding of new members to it's complexity.

Looking on the other hand, we have Quick API (StaticJavaParser) that is really straight forward to use, and makes the process of on-boarding new members really simple. Unfortunately StaticJavaParser has some downsides, for example each time we whant to parse something a new parser has to be created, maybe due to concurrency.

 **What if we could have the best of both worlds?**

Having the flexibility of JavaParser but being able to use API like StaticJavaParser... 🤔

That's what this PR suggests!

---

## What was done

Previously if the use want to use Quick API the only option he had was to use StaticJavaParser and to use Full API was to use JavaParser.
The change started here, by allowing the user to decide which API he wants to use JavaParser with. For example:
```java
// FullJavaParser
JavaParser fullParser = new JavaParser();
ParseResult<CompilationUnit> fullCu = fullParser.parse("class A {}");

// QuickJavaParser
QuickJavaParser quickParser = new QuickJavaParserAdapter(fullParser);
CompilationUnit quickCu = quickParser.parse("class A {}");

// StaticJavaParser
CompilationUnit staticCu = StaticJavaParser.parse("class A {}");
```

Suggestion and feedback are welcome!

Let me know your thoughts!

---

Fixes #3108.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/StaticJavaParser.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/StaticJavaParser.java
@@ -51,10 +51,7 @@ import java.nio.file.Path;
 public final class StaticJavaParser {
 
     // use ThreadLocal to resolve possible concurrency issues.
-    private static ThreadLocal<ParserConfiguration> localConfiguration = ThreadLocal.withInitial(() -> new ParserConfiguration());
-
-    private StaticJavaParser() {
-    }
+    private static final ThreadLocal<ParserConfiguration> localConfiguration = ThreadLocal.withInitial(ParserConfiguration::new);
 
     /**
      * Get the configuration for the parse... methods. Deprecated method.
@@ -80,10 +77,6 @@ public final class StaticJavaParser {
     public static void setConfiguration(@NotNull ParserConfiguration configuration) {
         Preconditions.checkNotNull(configuration, "Parameter configuration can't be null.");
         localConfiguration.set(configuration);
-    }
-
-    private static JavaParser newParser() {
-        return new JavaParser(getConfiguration());
     }
 
     /**
@@ -113,7 +106,7 @@ public final class StaticJavaParser {
      */
     public static CompilationUnit parse(@NotNull final InputStream in) {
         Preconditions.checkNotNull(in, "Parameter in can't be null.");
-        return handleResult(newParser().parse(in));
+        return newParserAdapted().parse(in);
     }
 
     /**
@@ -145,7 +138,7 @@ public final class StaticJavaParser {
      */
     public static CompilationUnit parse(@NotNull final File file) throws FileNotFoundException {
         Preconditions.checkNotNull(file, "Parameter file can't be null.");
-        return handleResult(newParser().parse(file));
+        return newParserAdapted().parse(file);
     }
 
     /**
@@ -177,7 +170,7 @@ public final class StaticJavaParser {
      */
     public static CompilationUnit parse(@NotNull final Path path) throws IOException {
         Preconditions.checkNotNull(path, "Parameter path can't be null.");
-        return handleResult(newParser().parse(path));
+        return newParserAdapted().parse(path);
     }
 
     /**
@@ -192,7 +185,7 @@ public final class StaticJavaParser {
      */
     public static CompilationUnit parseResource(@NotNull final String path) throws IOException {
         Preconditions.checkNotNull(path, "Parameter path can't be null.");
-        return handleResult(newParser().parseResource(path));
+        return newParserAdapted().parseResource(path);
     }
 
     /**
@@ -244,7 +237,7 @@ public final class StaticJavaParser {
      */
     public static CompilationUnit parse(@NotNull final Reader reader) {
         Preconditions.checkNotNull(reader, "Parameter reader can't be null.");
-        return handleResult(newParser().parse(reader));
+        return newParserAdapted().parse(reader);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


4everTheOne:
**Question:** Should this deprecated methods be removed? Has been 4 years since marked as deprecated.
https://github.com/javaparser/javaparser/blob/5ccfef80bd2b6bf32901cfe98ece61e14672d44a/javaparser-core/src/main/java/com/github/javaparser/StaticJavaParser.java#L165-L181

jlerbsc:
> **Question:** Should this deprecated methods be removed? Has been 4 years since marked as deprecated.

We can do that (in another PR) but in this case the version number must take this into account to alert users of a modification that may have an impact on their codes.

javaparser-core/src/main/java/com/github/javaparser/QuickJavaParserAdapter.java
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Helper function to handle the result in a simpler way.
+     *
+     * @param result The result to be handled.
+     *
+     * @param <T> The return type.
+     *
+     * @return The parsed value.
+     */
+    private <T extends Node> T handleResult(ParseResult<T> result) {
```
jlerbsc, line 58: It seems to me that this method is no longer used.

javaparser-core/src/main/java/com/github/javaparser/StaticJavaParser.java
```diff
@@ -528,6 +533,23 @@ public static TypeParameter parseTypeParameter(@NotNull String typeParameter) {
...
+        return QuickJavaParser.create(getParserConfiguration()).parseMethodDeclaration(methodDeclaration);
+    }
+
+    // Private methods
+
+    private static JavaParser newParser() {
+        return new JavaParser(getConfiguration());
+    }
+
+    private static <T extends Node> T handleResult(ParseResult<T> result) {
```
jlerbsc, line 531: It seems to me that this method is no longer used.

4everTheOne: At this point, method handleResult is only used by Deprecated methods in order to provide retro-compatibility.
https://github.com/javaparser/javaparser/blob/e545c2a11c062096b96284b3a263184eb1cbe7d8/javaparser-core/src/main/java/com/github/javaparser/StaticJavaParser.java#L101-L116
But later, when this deprecated methods have been removed, this function can also be removed :)

jlerbsc:
Hi @4everTheOne I think this approach is much more interesting than the first. It is easier to use and understand. However, I have a few minor remarks on the implementation and on the naming of the classes.

- First, I agree that defining an interface for the parser would decouple the implementation and the API. I'm wondering if creating this "JavaParser" interface is actually necessary as it assumes one could override the implementation. I don't know if this is really realistic. On the other hand, it seems to me that adding an interface for an adapter is not necessary.
- Secondly the names of classes and interfaces can certainly be improved "Quick...", "Full....". For example, since JavaParser is already in use, I would opt for a Parser interface (although I would have preferred JavaParser which indicates that it is a parser for the Java language). I know that this point is in contradiction with the previous point but I share my feelings with you.

I'm also not a fan of interfaces that are bound with implementations. For example the QuickJavaParser interface is linked with the QuickJavaParserAdapter class.

Ultimately I think the new parser would actually be an adapter so a name JavaParserAdapter seems to me to make enough sense.

I also find it interesting to integrate the adapter in the StaticJavaParser class because it allows to homogenize the development without breaking the implementations.

I suggest you leave the java parser interface aside for the moment and create a discussion thread (issue) to evaluate the pros and cons. 

I know that all these remarks are a bit disjointed but I give you my feeling at the moment. What do you think about this?

4everTheOne:
Hi @jlerbsc, thanks for the feedback! It is always welcome!

Sorry, but I'm not sure if I understood your comment. What i understood was:

You agree of keeping in this PR:
- The idea of creating an "JavaParserAdapter", that is similar to the API used by "StaticJavaParser", so users can wrap a "JavaParser" to use with the static API;
- The ideia of "StaticJavaParser" in consuming the "JavaParserAdapter" to reduce the amount of code to maintain;

What you suggest, to be removed from this PR:
- Create a discussion (issue) with the ideia of moving the APIs to an interface do decouple the implementations. (For both JavaParser API and StaticJavaParser API)

Is this correct?

jlerbsc:
> Is this correct?

You understood well.
What I also suggest is to rename the Quick... and Full... classes.

4everTheOne:
Without the API, the previous code looks like this:
```java
// JavaParser
JavaParser fullParser = new JavaParser();
ParseResult<CompilationUnit> fullCu = fullParser.parse("class A {}");

// JavaParserAdapter
JavaParserAdapter quickParser = new JavaParserAdapter(fullParser);
CompilationUnit quickCu = quickParser.parse("class A {}");

// StaticJavaParser
CompilationUnit staticCu = StaticJavaParser.parse("class A {}");
```

jlerbsc:
Your work is very interesting.
Can you reorganize the commits to make reading the changes easier.
What do you think the factory brings? Personally, I'm not sure it's necessary in this context.

4everTheOne:
The ideia behind the factory was to remove the logic for creating a parser from the parsers. But without the interfaces that define the API, they are no longer needed.
I'll remove them

jlerbsc:

PR #3762, opened 2022-11-24
Fix: #3761 Lexical preserving corrupts source when adding a modifier in first position

### LINKED ISSUE #3761: Lexical preserving corrupts source when adding a modifier in first position
The attached test case shows how the activation of lexical preseving correpts source, i.e. it will not compile any more.
As you yesterday fixed #3750 (which is at least related from the effect on the source),
I checked it with the current version 3.24.9-SNAPSHOT, but it still fails.
Great to see you working on lexical preseving - I think it could be a very helpful feature, but it seems unfortunately not really production ready.

Output:
```
lexicalPreserving=false: class C {

    protected static String S = "s";
}

lexicalPreserving=true: class C { 
    staticprotected  String S = "s";
}
```

Source:
```
public class ShowBugLexicalCorruptsSource2 {

	static String src = "class C { \n"
			+ "    static String S = \"s\";\n"
			+ "}";

		test(false);
		test(true);
	}

	static void test(boolean useLexicalPreserving) {
		TypeDeclaration<?> cu = StaticJavaParser.parseTypeDeclaration(src);

		if (useLexicalPreserving) {
		}

		List<FieldDeclaration> fields = cu.findAll(FieldDeclaration.class);
		FieldDeclaration field = fields.get(0);

		List<Keyword> kws = field.getModifiers().stream().map(Modifier::getKeyword).collect(Collectors.toList());
		kws.add(0, Keyword.PROTECTED);
		field.setModifiers(kws.toArray(new Keyword[] {}));

		String str;
		if (useLexicalPreserving) {
			str = LexicalPreservingPrinter.print(cu);
		} else {
			str = cu.toString();
		}
		System.out.println("lexicalPreserving=" + useLexicalPreserving + ": " + str + "\n");
	}
}
```

jlerbsc:
Fixes #3761 .


jlerbsc:
It seems that there is a bug in the java 10 compiler but since the other builds did not reveal this error we can consider that this error does not prevent to merge the PR.

PR #3763, opened 2022-11-24
Fix: #2137 ClassOrInterfaceDeclaration addMember using index

### LINKED ISSUE #2137: ClassOrInterfaceDeclaration addMember using index

jlerbsc:
Fixes #2137.
Adding a test case. It seems that this issue has already been fixed.


jlerbsc:
It seems to be a bug in the compiler (10.0.2) so we can bypass this problem

PR #3768, opened 2022-11-27
Fix: #3725 JavaParserFacade var type in for-each loop cannot be resolved

### LINKED ISSUE #3725: Cannot resolve `var` which has no initializer.
The problem is that javaparser cannot parse types declared with 'var'. 

Here's the test sample:

```
import java.io.File;
import java.io.FileInputStream;
import com.github.javaparser.ast.visitor.VoidVisitor;
import com.github.javaparser.symbolsolver.model.resolution.TypeSolver;
import com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver;

class ParseTest {
    public void ParseMethods(File Path) throws FileNotFoundException {

        JavaCodeParser p = new JavaCodeParser(Path);
        p.getMethodDetails();

    }
}

public class Main {

        ParseTest parseTest = new ParseTest();
        parseTest.ParseMethods(new File("src\\MyClass.java"));
    }
}

class JavaCodeParser {

    // Root of AST; and StaticJavaParser is a simple parser class.
    private CompilationUnit cu;

    JavaCodeParser(File JavaSourceFilePath) throws FileNotFoundException {

        CombinedTypeSolver combinedSolver = new CombinedTypeSolver();

        TypeSolver reflectionTypeSolver = new ReflectionTypeSolver();
        TypeSolver javaParserTypeSolver = new JavaParserTypeSolver(
                "src");

        combinedSolver.add(reflectionTypeSolver);
        combinedSolver.add(javaParserTypeSolver);

        JavaSymbolSolver symbolSolver = new JavaSymbolSolver(combinedSolver);
        StaticJavaParser.getConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_15);
        StaticJavaParser.getConfiguration().setSymbolResolver(symbolSolver);

        cu = StaticJavaParser.parse(new FileInputStream(JavaSourceFilePath));
    }

    public void getMethodDetails() {

        // This visitor will collect all method names into a list.
        VoidVisitor<Void> methodNameCollector = new MethodDetailCollector();

        // Execute visitor; note that instead of 'null' now its a 'method
        methodNameCollector.visit(cu, null);

    }

    private class MethodDetailCollector extends VoidVisitorAdapter<Void> {

        @Override
        public void visit(MethodDeclaration md, Void arg) {

            super.visit(md, arg);

            try {

                // Get information on method calls in body
                md.accept(new MethodCallVisitor(), arg);

            } catch (Exception e) {
                System.out.println(e);
            }
        }
    }

    private class MethodCallVisitor extends VoidVisitorAdapter<Void> {

        @Override
        public void visit(MethodCallExpr n, Void arg) {

            super.visit(n, arg);

            String MethodCallObjectType;
            String MethodCallName;

            try {
                if (n.hasScope()) {
                    MethodCallObjectType = n.getScope().get().calculateResolvedType().describe();
                } else {
                    MethodCallObjectType = "this";
                }
                MethodCallName = n.getNameAsString();
            } catch (Exception e) {
                System.out.println("Exception occurred: " + n.getName() + ": " + e);
            }
        }
    }
}
```

The files used for parsing:

**Account.java**

```
import java.util.ArrayList;
import java.util.List;

public class Account {
    
    public Account(int id, String accountName, double accountValue) {
        this.id = id;
        this.accountName = accountName;
        this.accountValue = accountValue;
    }

    private int id;
    private String accountName;
    private double accountValue;
    
    private List<Account> accountList = new ArrayList<Account>();

    void insertAccountIntoList(Account account) {
        accountList.add(account);
    }
    
    List<Account> getAccountList() {
        return accountList;
    }
    
    void printInfo() {
        System.out.println("id: " + id);
        System.out.println("accountName: " + accountName);
        System.out.println("accountValue: " + accountValue);
    }
}
```

**MyClass.java**

```
public class MyClass {
    
        Account a = new Account(1, "AD1", 545.7);
        Account b = new Account(2, "AD2", 656.6);
        
        var accounts = a.getAccountList();
        accounts.add(a);
        accounts.add(b);
        
        for (var account : accounts) {
            account.printInfo();
        }
    }
}
```

The error is following:

`Exception occurred: printInfo: java.lang.IllegalStateException: Cannot resolve 'var' which has no initializer.`

The error occurs when I try to parse method body of main function for method calls inside it. How can I solve it?

abego:
Before this fix it was not possible to resolve the type of the variable used in a for-each loop when the type of the variable was defined as `var`. This was because the original code only took care of the `var with initializer` case (e.g. `var i = "hello";`).

The fix now also covers the usage of a `var` in a for-each loop, e.g. as in `for (var s: names) {...}`. The expression at the right side of the `:` must either be an array or an Iterable.

The bug was originally reported in https://github.com/javaparser/javaparser/issues/3725

Fixes #3725 .

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -28,6 +28,7 @@ import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.*;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.stmt.ExplicitConstructorInvocationStmt;
+import com.github.javaparser.ast.stmt.ForEachStmt;
 import com.github.javaparser.ast.type.*;
 import com.github.javaparser.resolution.MethodAmbiguityException;
 import com.github.javaparser.resolution.MethodUsage;
@@ -72,9 +73,9 @@ public class JavaParserFacade {
     };
 
     private static final Map<TypeSolver, JavaParserFacade> instances = new WeakHashMap<>();
-    
+
     private static final String JAVA_LANG_STRING = String.class.getCanonicalName();
-    
+
     /**
      * Note that the addition of the modifier {@code synchronized} is specific and directly in response to issue #2668.
      * <br>This <strong>MUST NOT</strong> be misinterpreted as a signal that JavaParser is safe to use within a multi-threaded environment.
@@ -115,11 +116,11 @@ public class JavaParserFacade {
     }
 
     // End of static class
-    
+
     private final TypeSolver typeSolver;
     private final TypeExtractor typeExtractor;
     private final SymbolSolver symbolSolver;
-    
+
     private FailureHandler failureHandler;
 
     private JavaParserFacade(TypeSolver typeSolver) {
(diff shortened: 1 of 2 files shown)


jlerbsc:

PR #3774, opened 2022-12-02
Refactoring convert to usage

jlerbsc:
The main objective of this PR is to refactor the methods for converting a type into a "resolved" type. To achieve this, it was necessary to move interfaces from the "symbol solver" module to the "core" module, delete relations between interfaces and concrete classes, delete relations with the type solver...


jlerbsc:
There is an unknown compilation problem occurred probably a bug in the compiler.

PR #3781, opened 2022-12-10
Basic CheckStyle validation

4everTheOne:
This PR pretends to help improving the quality of the code, by providing a code style validation while the feature is in development.

This is a simple [CheckStyle ](https://checkstyle.sourceforge.io/) based at [Coding Guidelines](https://github.com/javaparser/javaparser/wiki/Coding-Guidelines).

What do you think of this feature? Would it help delivering value to the project?


jlerbsc:
There are 2419 errors reported by Checkstyle 8.45.1 with dev-files/JavaParser-CheckStyle.xml ruleset. The approach is good but it is unmanageable.

4everTheOne:
Hi @jlerbsc , thanks for your feedback. I totally agree with you when you say it's unmanageable.
I think we can improve the approach of implementation by gradually enforcing the rules.

Breaking down the 2419 results we have the following cases:

- LineLength: 1628 (The line length exceeds 120 characters per line)
- Indentation: 292 (Code is wrongly indented)
- CustomImportOrder: 499 (Importers are placed in the wrong order)

One way of doing this, would be to changing the severity of "LineLength" from error (which means mandatory) to a warning. By doing this, it will not break the build but for developers will have a warning notifying about a improvement to be done, like shown below:
image
Later, when in a more manageable stage we can revert it back to mandatory :)


About the topic Indentation and CustomImportOrder, this can be easily fixed with the IDE tools and if you want i can include that fix on another PR, or if you prefer on this one :)

Do you think it can be manageable this way?

jlerbsc:
It actually seems a lot more manageable to me. But above all, it must not block a PR for the moment.

4everTheOne:
Updated the Checkstyle to report every violation as Warning.
When we think is time to enforce, we call change it back to error

PR #3807, opened 2022-12-18
Source printer import ordering strategy

4everTheOne:
Hello!

This PR adds support for custom import ordering.
Previously, the user could not define the "template" to be used for imports, only having the option to have imports like this:
```java
import java.util.List;
import com.github.javaparser.ast.NodeList;
import com.github.javaparser.ast.ImportDeclaration;
```

What if the user wants to separate the static imports from the others? What if the user wants to group all java imports into a single group?

Example:
```java
import com.github.javaparser.ast.NodeList;
import com.github.javaparser.ast.ImportDeclaration;

import java.util.List;
```

With this PR this can now be achieve Included in this PR, there are 3 implementation of import ordering.
- DefaultImportOrderingStrategy: The logic that existed previously in JavaParser.
- IntelliJImportOrderingStrategy: Implements the ordering like IntelliJ does;
- EclipseImportOrderingStrategy: Implements the ordering like Eclipse does;

An example of code that makes the printer use a custom formatitng can be seen below:
```java
CompilationUnit cu = ...;

IntelliJImportOrderingStrategy strategy = new IntelliJImportOrderingStrategy();
DefaultConfigurationOption option = new DefaultConfigurationOption(ConfigOption.SORT_IMPORTS_STRATEGY, strategy);
printerConfiguration.addOption(option);

Printer printer = new DefaultPrettyPrinter(printerConfiguration);
String actualCode = printer.print(cu);
```

This design is flexible, so if the user wants to create their own implementation they can also do it.

Feedback is welcome! :)


jlerbsc:
As usual it is a very nice contribution.
Nevertheless, would it be possible to rename the "strategy" classes so as not to refer to a solution (eclipse, intelliJ) but rather to an order. If it's too complex to describe the order other than by the name of the solution, that's okay.

4everTheOne:
Hi @jlerbsc. Thanks for your feedback!
I'm not sure how to name it, because they are a little complex.

For example, in intellij the order is:
- Other Imports;
- Java and Javax:
- Static Imports

On eclipse the order is:
- Static Imports
- Java
- JavaX
- org
- com
- Other

If you have any suggestion, please let me know! I would be happy to integrate it.

jlerbsc:
No I have no idea that's why I asked you for yours. Have you finished this PR?

4everTheOne:
Yes I have finished it :)
Thanks

PR #3814, opened 2022-12-24
Handle nested records

mernst:
This pull request handles nested records.
It also updates some outdated comments about the Java grammar.


jlerbsc:
Can you give us the reference in the Java specification?

mernst:
@jlerbsc The reference in the Java specification is in the pull request: https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10 .

At https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10.2 you can see that a `RecordBody` can contain anything a class body can, and classes (and records, enums, etc.) can be nested within classes.

jlerbsc:

PR #3815, opened 2022-12-26
#3808 toDescriptor for ResolvedMethodDeclaration

### LINKED ISSUE #3808: Get method descriptor for ResolvedMethodDeclaration / MethodCallExpr
## Description
While using javaparser I realized that it is cumbersome to get the corresponding method descriptor for a MethodCallExpr or for the ResolvedMethodDeclaration you get when resolving the MethodCallExpr.

Similar to #3559, #3615 and #2059, #3498, I would imagine that it would be useful to easily get the descriptor for a method call (or rather the respective method declaration).

Thus for a method call like `Math.pow(4.2, 3)` you would quickly get `(DD)D` (meaning (double, double) -> double).

### Problem
A MethodCallExpr resolves to a ResolvedMethodDeclaration, but that can actually be an instance of:
1. JavaParserMethodDeclaration
2. ReflectionMethodDeclaration
3. JavassistMethodDeclaration
4. maybe even more? The 3 above were the ones I encountered.

Of those 3, the JavaParserMethodDeclaration seemed to be the only one where it was somewhat straightforward get the descriptor.

### Possible Improvement?
In the end, I was able to get the descriptors for all the encountered cases. 
For ReflectionMethodDeclaration and JavassistMethodDeclaration, it was necessary to access the private fields `method` and `ctMethod`.
Now, I am wondering if it would be appreciated if I made a pull request adding a straightforward way to get the descriptor in all cases.

## More detailed use case example
Couple of months ago I had to match nodes in call tree xml files exported from JProfiler to methods and method calls in source code.  Each node basically looks like this (simplified):
```
<node leaf="true" class="java.lang.Math" methodName="pow" methodSignature="(DD)D" />
```

Take for example:
```
public void doMath() {
    DecimalFormat df = new DecimalFormat(".00");
    double dblResult = Math.pow(4.2, 3);
}
```
The corresponding (sub) tree in the xml file could look like this:
```
<node leaf="false" class="main.java.components.MyClass" methodName="doMath" methodSignature="()V">
    <node leaf="true" class="java.lang.Math" methodName="pow" methodSignature="(DD)D" />
</node>
```

So given the xml file (alongside other things to be done), I needed to get the full qualified path and the method descriptor (method signature) of the methods that were called  to reliably pair them to the right nodes.

In this simple example `Math.pow` and `<node leaf="true" class="java.lang.Math" methodName="pow" methodSignature="(DD)D" />`.
Of course this example could effortlessly be solved in a simpler way, but imagine for instance a scenario where a method is overloaded.

0verhead:
Fixes #3808.

Adds `toDescriptor()` to ResolvedMethodDeclarations with all the needed implementations. It was also necessary to add `toDescriptor()` for ResolvedType.

For convenience, a TypeUtils class was added that contains utility methods for getting the descriptors for `java.lang.reflect.Method` and primitive classes (compare to [ASM](https://gitlab.ow2.org/asm/asm/-/blob/master/asm/src/main/java/org/objectweb/asm/Type.java#L580)). 

### Found issue:
Inner classes(/enums) do not have the correct fully qualified name. You get Class.InnerClass instead of Class$InnerClass. This also effects my test cases for the `valueOf` and `values` method of enums.
For now, it should be fine, but I think this needs to be addressed in a future issue and PR.


jlerbsc:
First of all we thank you for this contribution. To make your work easier to read and understand, you can submit a clean PR that focuses only on the subject of the PR and the final version of the code.

0verhead:
> First of all we thank you for this contribution. To make your work easier to read and understand, you can submit a clean PR that focuses only on the subject of the PR and the final version of the code.

I have undone all the auto formatting (stuff from the guidelines). The changes should now be only related to the PR subject.

jlerbsc:
There are 23 commits in this PR. It's hard to validate it in this state. Please create a new one containing only the finished development (without reverting, without reformatting code...).

0verhead:
I am going to close this PR as it is replaced by #3819.

PR #3819, opened 2022-12-27
Add toDescriptor to ResolvedMethodDeclaration

### LINKED ISSUE #3808: Get method descriptor for ResolvedMethodDeclaration / MethodCallExpr

0verhead:
Fixes #3808.

Adds `toDescriptor()` to ResolvedMethodDeclarations with all the needed implementations. It was also necessary to add `toDescriptor()` for ResolvedType.

For convenience, a TypeUtils class was added that contains utility methods for getting the descriptors for `java.lang.reflect.Method` and primitive classes (compare to [ASM](https://gitlab.ow2.org/asm/asm/-/blob/master/asm/src/main/java/org/objectweb/asm/Type.java#L580)). 

### Found issue:
Inner classes(/enums) do not have the correct fully qualified name. You get Class.InnerClass instead of Class$InnerClass. This also effects my test cases for the `valueOf` and `values` method of enums.
For now, it should be fine, but I think this needs to be addressed in a future issue and PR.


javaparser-core/src/main/java/com/github/javaparser/utils/TypeUtils.java
```diff
@@ -0,0 +1,86 @@
...
+        } else {
+            stringBuilder.append("L").append(currentClass.getName().replace(".", "/")).append(";");
+        }
+    }
+
+    public static String getPrimitiveTypeDescriptor(final Class<?> clazz) {
+        String descriptor;
+        if (clazz == Void.TYPE) {
+            descriptor = new VoidType().toDescriptor();
+        } else if (clazz == Integer.TYPE) {
```
jlerbsc: Maybe you can simplify this structure by using ResolvedPrimitiveType.byBoxTypeQName(..)

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/DescriptorTest.java
```diff
@@ -17,18 +22,35 @@
...
-                "  void m() {}\n" +
+                "  void m() {\n" +
+                "    System.out.println(\"a\");\n" +
+                "    ClassPool cp = ClassPool.getDefault();\n" +
+                "    TestEnum.valueOf(\"TEST_ENUM\");\n" +
+                "    TestEnum.values();\n" +
+                "  }\n" +
                 "}";
+
+        Path javassistJar = adaptPath("src/test/resources/issue3808/javassist-3.29.2-GA.jar");
```
jlerbsc: Do not include the javassists library in the project resources

javaparser-core/src/main/java/com/github/javaparser/utils/TypeUtils.java
0verhead: I think such a method only exists for ResolvedPrimitiveType and not PrimitiveType. Unfortunately, `ResolvedPrimitiveType.byBoxTypeQName` won't be useful here as it only returns Resolved(Primitive)Types.

But I can implement a similar method for PrimitiveTypes. Maybe like this:
```java
public static Optional<Primitive> getPrimitiveByBoxedTypeName(String name) {
    for (Primitive prim : Primitive.values()) {
        if (prim.nameOfBoxedType.equals(name)) {
            return Optional.of(prim);
        }
    }
    return Optional.empty();
}
```

And then in `TypeUtils.getPrimitiveTypeDescriptor` I can do:
```java
// void case:
...
// primitve cases:
Optional<Primitive> optPrim = PrimitiveType.getPrimitiveByBoxedTypeName(clazz.getCanonicalName());
if (optPrim.isPresent()) {
    return new PrimitiveType(optPrim.get()).toDescriptor();
}
```

jlerbsc: > I think such a method only exists for ResolvedPrimitiveType and not PrimitiveType. Unfortunately, `ResolvedPrimitiveType.byBoxTypeQName` won't be useful here as it only returns Resolved(Primitive)Types.

You are right I meant ResolvedPrimitiveType.
Why do you say it's not useful? You can do something like this (or something better with a little more work).

    @Test
	void test() {
		Optional<ResolvedType> rt = ResolvedPrimitiveType.byBoxTypeQName(Integer.class.getCanonicalName());
		String type = toDescritor(rt.get().asPrimitive().name());
		assertEquals("I", type);
	}
	
	private String toDescritor(String primitiveTypeName) {
		return getPrimitiveType(primitiveTypeName)
				.map(p -> new PrimitiveType(p).toDescriptor())
				.orElseThrow(() -> new IllegalArgumentException("Unknown primitive"));
	}
	
	private Optional<Primitive> getPrimitiveType(String name) {
		return Optional.of(Primitive.valueOf(name));
	}

0verhead: Ah, that's clever.
What are your thoughts on adding a (getter) method `toDescriptor` to `Primitive` and then
```java
public String getPrimitiveTypeDescriptor(final Class<?> clazz) {
    String className = clazz.getCanonicalName();
    Optional<Primitive> optPrim = Optional.of(Primitive.valueOf(className));
    return optPrim.map(p -> p.toDescriptor())
            .orElseThrow(() -> new IllegalArgumentException("Unknown primitive"));
}
```

Wouldn't that be even cleaner?

jlerbsc: Why not but in this case it is also necessary to adapt the toDescriptor method in the PrimitiveType class.

0verhead: The new commit should contain the mentioned improvements. Additionally I realized, that in my original code only for the 'real' primitive classes (int.class, byte.class etc.) the correct descriptor was returned. Passing a wrapper class (like Integer.class) into the method resulted in an error.

I fixed that. Now you should get the correct descriptor for all primitive classes as well as their wrapper classes.
The solution looks like this:
```java
public static String getPrimitiveTypeDescriptor(final Class<?> clazz) {
    if (clazz == Void.TYPE || clazz == Void.class) {
        return new VoidType().toDescriptor();
    }
    String className = clazz.getCanonicalName();
    Optional<Primitive> prim = getPrimitive(className);
    if (prim.isPresent()) {
        return prim.get().toDescriptor();
    }
    Optional<ResolvedType> rt = ResolvedPrimitiveType.byBoxTypeQName(className);
    if (rt.isPresent()) {
        return getResolvedPrimitiveTypeDescriptor(rt);
    }
    throw new IllegalArgumentException("Unknown primitive: " + className);
}

private static String getResolvedPrimitiveTypeDescriptor(Optional<ResolvedType> rt) throws AssertionError {
    String typeName = rt.get().asPrimitive().describe();
    Optional<Primitive> prim = getPrimitive(typeName);
    return prim.map(p -> p.toDescriptor())
            .orElseThrow(() -> new AssertionError(
                    String.format(
                            "ResolvedPrimitiveType name \"%s\" does not match any Primitive enum constant identifier.",
                            typeName.toUpperCase())));
}

private static Optional<Primitive> getPrimitive(String name) {
    try {
        return Optional.of(Primitive.valueOf(name.toUpperCase()));
    } catch (IllegalArgumentException e) {
        return Optional.empty();
    }
}
```

But now I am honestly wondering, if it would'nt  be better to just do:
```java
if (clazz == Void.TYPE || clazz == Void.class) {
    descriptor = new VoidType().toDescriptor();
} else if (clazz == Integer.TYPE || clazz == Integer.class) {
    descriptor = Primitive.INT.toDescriptor();
} else if (clazz == Boolean.TYPE || clazz == Boolean.class) {
    ...
```

javaparser-core/src/main/java/com/github/javaparser/utils/TypeUtils.java
```diff
@@ -59,28 +62,36 @@ private static void appendDescriptor(final Class<?> clazz, final StringBuilder s
...
-        } else if (clazz == Float.TYPE) {
-            descriptor = PrimitiveType.floatType().toDescriptor();
-        } else if (clazz == Long.TYPE) {
-            descriptor = PrimitiveType.longType().toDescriptor();
-        } else {
-            throw new AssertionError("Unknown primitive: " + clazz.getName());
+        if (clazz == Void.TYPE || clazz == Void.class) {
+            return new VoidType().toDescriptor();
+        }
+        String className = clazz.getCanonicalName();
```
jlerbsc: It still seems a bit too complex to me.

If you add 2 methods in Primitive

    public static Optional<Primitive> byBoxedType(String qName) {
            for (Primitive primitive : values()) {
                if (primitive.nameOfBoxedType.equals(qName)) {
                    return Optional.of(primitive);
                }
            }
            return Optional.empty();
        }
        
        public String toDecriptor() {
        	return descriptor;
        }

then you can do it just like that.

    @Test
	void test0() {
		Optional<Primitive> primitive = Primitive.byBoxedType(Integer.class.getSimpleName());
		assertEquals("I", primitive.map(type -> type.toDecriptor()).orElseThrow(() -> new IllegalArgumentException("Unknown primitive type")));
	}

Also don't throw an AssertionException and an optional parameter is bad practice.

0verhead: The newest commit https://github.com/javaparser/javaparser/pull/3819/commits/eaf1f0d06727b9262f3189c01e1f2951e4fc5bc4 basically did what you suggest using the unboxMap and also adding `Primitive.byTypeName` for all the inputs like int.class, boolean.class etc.

jlerbsc:
Can you add unit test for all primitive type and remove the unit test dependency with the javassist library.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/DescriptorTest.java
0verhead: I have removed the jar and used the javassistmethoddecl.jar from the resources.

0verhead:
> Can you add unit test for all primitive type and remove the unit test dependency with the javassist library.

Did that in the latest commits https://github.com/javaparser/javaparser/pull/3819/commits/e6901c31067ce2a4b560164311c8a0c14d7da8f9 and https://github.com/javaparser/javaparser/pull/3819/commits/358e5364688514d15f4ea175f97d8234e6abd452.

jlerbsc:
Thank you for your contribution which will be the last of the year. It will be available in the next version (end of January) because version 3.24.10 was released this afternoon.

0verhead:
> Thank you for your contribution which will be the last of the year. It will be available in the next version (end of January) because version 3.24.10 was released this afternoon.

Very nice, thanks!

Seems like maven_test (windows-latest, 8) failed the second time (even though it was successful the first time).
I have also fixed the import order now which caused the checkstyle test to fail.

jlerbsc:
> Seems like maven_test (windows-latest, 8) failed the second time (even though it was successful the first time).
> I have also fixed the import order now which caused the checkstyle test to fail.

Please stop modifying the PR so that I can validate it.

PR #3821, opened 2022-12-27
explicit use asString for performance

dencat:
Replaced implicit `toString` call with explicit `asString` to improve performance (`toString` uses `DefaultPrettyPrinter`, which slows down, `asString` immediately produces the desired result)


jlerbsc:

PR #3830, opened 2023-01-03
Fix: update codecov setting to not fail ci on error

jlerbsc:
To avoid having errors related to codecov during builds we have disabled the fail_ci_if_error attribute by hearing that the following issue is resolved.
https://github.com/codecov/codecov-action/issues/678


jlerbsc:
This update does not seem to have an impact on the problem

PR #3836, opened 2023-01-06
feat: Implement addRecord & getRecordByName for CompilationUnit

marcluque:
Hi there,

while working on a project, I noticed missing implementations for `getRecordByName` and `addRecord`. I added their implementations and some simple tests for the implementations (similar to the existing ones for enum, etc.).

Let me know if something could be improved or is missing.


jlerbsc:
Hi @marcluque Thank you for this suggestion but there is already an addType method. You can adapt the test cases and let the getRecordByName method.

marcluque:
Hi @jlerbsc, thanks for the quick feedback! 
I pushed a new commit that removes `addRecord`. However, I left the test case and adapted it to use `addType`, let me know if that's desired.

jlerbsc:

PR #3867, opened 2023-01-23
Add support for ZipFileSystem

freya022:
This PR makes `CodeGenerationUtils#packageAbsolutePath` resolve the package against the original `Path`, preserving the filesystem the path comes from, instead of using the default filesystem with `Paths.get`.

My use case was to use a `ZipFileSystem` to read sources directly from my `.m2` directory

Edit: To my knowledge, supporting *every* filesystem would require taking into account path separators, affecting resolving/relativizing paths, and, for example, getting the package name from the `Path`. This is why this PR is only going to add support for Java's ZipFileSystem


javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
```diff
@@ -54,4 +61,23 @@ void testGetterToPropertyName() {
         assertEquals("Unexpected getterName 'value'", thrown.getMessage());
     }
 
+    @Test
+    void testPackageToAbsolutePath() throws IOException {
+        final Path pathToZip = classLoaderRoot(CodeGenerationUtilsTest.class)
+                .resolve("com/github/javaparser/utils/MyActivity.zip");
```
jlerbsc: What is this file? If you need a compressed file for your test case there are already resources available for example in the source_zip directory

javaparser-core/src/main/java/com/github/javaparser/utils/CodeGenerationUtils.java
```diff
@@ -112,17 +113,28 @@ public static String packageToPath(String pkg) {
...
 
+    /**
+     * Calculates the path of a package.
+     * <p><b>Note:</b> Unlike {@link #packageAbsolutePath(String, String)}, the {@link Path} will use the same filesystem as the given path.
+     *
+     * @param root the root directory in which the package resides
+     * @param pkg  the package, like "com.laamella.parser"
+     * @return the {@link Path} of the package's directory
+     */
     public static Path packageAbsolutePath(Path root, String pkg) {
```
jlerbsc, line 135: Is there a change in behavior compared to the previous version of the method?

javaparser-core/src/main/java/com/github/javaparser/utils/CodeGenerationUtils.java
```diff
@@ -112,17 +113,28 @@ public static String packageToPath(String pkg) {
...
-        return packageAbsolutePath(root.toString(), pkg);
+        pkg = packageToPath(pkg);
```
jlerbsc, line 147: Is there a change in behavior compared to the previous version of the method?

freya022: No, as per the docs of [`Paths.get`](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Paths.html#get(java.lang.String,%20java.lang.String...)):
>  Note that while this method is very convenient, using it will imply an assumed reference to the default FileSystem and limit the utility of the calling code. Hence it should not be used in library code intended for flexible reuse. A more flexible alternative is to use an existing Path instance as an anchor, such as:

     Path dir = ...
     Path path = dir.resolve("file");
 
Replacing `Paths.get` by `Path#resolve` is the same, except the `FileSystem` being used is preserved

javaparser-core-testing/src/test/java/com/github/javaparser/utils/CodeGenerationUtilsTest.java
freya022: Sure, will replace

javaparser-core/src/main/java/com/github/javaparser/utils/CodeGenerationUtils.java
freya022: I will leave this open as a draft for the moment, my code isn't giving the same results when using the ZIP and I need to figure out why

freya022: Fixed with the next commits

jlerbsc:
There is far too much change for what seemed to be a minor evolution. I do not master all these changes and their impacts for other users and in addition the PR becomes unreadable so I will not follow up on this modification.

freya022:
True, there was more changes than I expected unfortunately, I'll just apply my workaround for the moment then (simply extract the ZIP into a directory)

PR #3873, opened 2023-01-24
fix #3859 UnsupportedOperationException when trying to resolve a type…

### LINKED ISSUE #3859: `UnsupportedOperationException` when trying to resolve a type in an argument which is an EnclosedExpr
# Problem

Given the following code

    class Demo {
        void foo(Consumer<String> arg) {}
        void print(Object arg) {}
        public void bar() {
            foo(s->print(s));
        }
        public void baz() {
            foo((s->print(s)));
        }
    }

Resolving the type of  the lambda expression used as an argument in `bar()` works fine. Also resolving the name used in the lambda expression  `bar()` works fine. 

However both operations fail with `java.lang.UnsupportedOperationException`s when attempted for the lambda expression defined `baz`.

Notice the only difference between both cases is the extra `(...)` around the lambda expression in `baz`.

# Analysis

The problems are caused by missing handling of EnclosedExpr in the code related to arguments in an MethodCallExpr.

abego:
This fixes the issue https://github.com/javaparser/javaparser/issues/3859

This commit contains just the fix. Additional refactorings are planned for a later commit/PR.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/Navigator.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/Navigator.java
@@ -31,6 +31,7 @@ import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.ast.stmt.SwitchStmt;
 
 import java.util.Optional;
+import java.util.function.Predicate;
 
 /**
  * This class can be used to easily retrieve nodes from a JavaParser AST.
@@ -145,6 +146,19 @@ public final class Navigator {
 
     public static Node demandParentNode(Node node) {
         return node.getParentNode().orElseThrow(() -> new IllegalStateException("Parent not found, the node does not appear to be inserted in a correct AST"));
+    }
+
+    /**
+     * Traverses the parent chain starting at {@code node} and returns the
+     * first Node that returns make {@code isAcceptedParentNode} evaluate to
+     * {@code true}.
+     */
+    public static Node demandParentNode(Node node, Predicate<Node> isAcceptedParentNode) {
+        Node parent = node;
+        do {
+            parent = demandParentNode(parent);
+        } while (!isAcceptedParentNode.test(parent));
+        return parent;
     }
 
     public static ReturnStmt demandReturnStmt(MethodDeclaration method) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
@@ -54,11 +54,17 @@ import com.google.common.collect.ImmutableList;
 
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Predicate;
 
 import static com.github.javaparser.resolution.Navigator.demandParentNode;
 
 public class TypeExtractor extends DefaultVisitorAdapter {
 
+    /**
+     * Returns {@code true} when the Node to be tested is not an
+     * {@link EnclosedExpr}, {@code false} otherwise.
+     */
+    private static final Predicate<Node> IS_NOT_ENCLOSED_EXPR = n -> !(n instanceof EnclosedExpr);
     private static final String JAVA_LANG_STRING = String.class.getCanonicalName();
     private final ResolvedType stringReferenceType;
 
@@ -449,12 +455,12 @@ public class TypeExtractor extends DefaultVisitorAdapter {
 
     @Override
     public ResolvedType visit(LambdaExpr node, Boolean solveLambdas) {
-        if (demandParentNode(node) instanceof MethodCallExpr) {
-            MethodCallExpr callExpr = (MethodCallExpr) demandParentNode(node);
+        if (demandParentNode(node, IS_NOT_ENCLOSED_EXPR) instanceof MethodCallExpr) {
+            MethodCallExpr callExpr = (MethodCallExpr) demandParentNode(node, IS_NOT_ENCLOSED_EXPR);
             int pos = getParamPos(node);
             SymbolReference<ResolvedMethodDeclaration> refMethod = facade.solve(callExpr);
             if (!refMethod.isSolved()) {
-                throw new UnsolvedSymbolException(demandParentNode(node).toString(), callExpr.getName().getId());
+                throw new UnsolvedSymbolException(demandParentNode(node, IS_NOT_ENCLOSED_EXPR).toString(), callExpr.getName().getId());
             }
             Log.trace("getType on lambda expr %s", ()-> refMethod.getCorrespondingDeclaration().getName());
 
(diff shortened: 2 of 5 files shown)


javaparser-core/src/main/java/com/github/javaparser/resolution/Navigator.java
```diff
@@ -147,6 +148,19 @@ public static Node demandParentNode(Node node) {
         return node.getParentNode().orElseThrow(() -> new IllegalStateException("Parent not found, the node does not appear to be inserted in a correct AST"));
     }
 
+    /**
+     * Traverses the parent chain starting at {@code node} and returns the
+     * first Node that returns make {@code isAcceptedParentNode} evaluate to
+     * {@code true}.
+     */
+    public static Node demandParentNode(Node node, Predicate<Node> isAcceptedParentNode) {
```
jlerbsc, line 156: This can be done by recursion instead of using a loop.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -54,11 +54,17 @@
...
 
 import static com.github.javaparser.resolution.Navigator.demandParentNode;
 
 public class TypeExtractor extends DefaultVisitorAdapter {
 
+    /**
+     * Returns {@code true} when the Node to be tested is not an
+     * {@link EnclosedExpr}, {@code false} otherwise.
+     */
+    private static final Predicate<Node> IS_NOT_ENCLOSED_EXPR = n -> !(n instanceof EnclosedExpr);
```
jlerbsc, line 72: if n is an expression you can write n.isEnclosedExpr() instead of using instanceof.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/LambdaExprContext.java
```diff
@@ -255,6 +263,9 @@ protected final Optional<Value> solveWithAsValue(SymbolDeclarator symbolDeclarat
     private int pos(MethodCallExpr callExpr, Expression param) {
         int i = 0;
         for (Expression p : callExpr.getArguments()) {
+            while (p instanceof EnclosedExpr) {
+                p = ((EnclosedExpr) p).getInner();
```
jlerbsc, line 267: Here p is an expression so you can write n.isEnclosedExpr() and use p.asEnclosedExpr() instead of casting the expression.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
```diff
@@ -638,10 +644,14 @@ public ResolvedType visit(FieldDeclaration node, Boolean solveLambdas) {
...
     private static int getParamPos(Node node) {
-        if (demandParentNode(node) instanceof MethodCallExpr) {
-            MethodCallExpr call = (MethodCallExpr) demandParentNode(node);
+        if (demandParentNode(node, IS_NOT_ENCLOSED_EXPR) instanceof MethodCallExpr) {
+            MethodCallExpr call = (MethodCallExpr) demandParentNode(node, IS_NOT_ENCLOSED_EXPR);
             for (int i = 0; i < call.getArguments().size(); i++) {
-                if (call.getArguments().get(i) == node) return i;
+                Expression expression = call.getArguments().get(i);
+                while (expression instanceof EnclosedExpr) {
+                    expression = ((EnclosedExpr) expression).getInner();
```
jlerbsc, line 660: Here argument "expression" is an expression so you can write n.isEnclosedExpr() and use p.asEnclosedExpr() instead of casting the expression.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3859Test.java
```diff
@@ -0,0 +1,59 @@
...
+                "    void foo(Consumer<String> arg) {}\n" +
+                "    void print(Object arg) {}\n" +
+                "    public void bar() {\n" +
+                "        foo(s->print(s));\n" +
+                "    }\n" +
+                "    public void baz() {\n" +
+                "        foo((s->print(s)));\n" +
+                "    }\n" +
+                "}\n";
+        ParserConfiguration configuration = new ParserConfiguration()
```
jlerbsc, line 33: For some time this can be written more simply
> CompilationUnit cu = JavaParserAdapter.of(createParserWithResolver(defaultTypeSolver())).parse(code);

javaparser-core/src/main/java/com/github/javaparser/resolution/Navigator.java
abego: Hi @jlerbsc, you are right, it can be done by recursion. 

However I would prefer to keep it in the loop form, as it is easy to understand and straight forward. Using recursion in a functional language is typically a good idea (and sometimes even necessary). However I would use recursion in Java only when it provides a better solution than the imperative solution, especially as the Java runtime does not provide recursion-related optimisations (like tail call elimination) to make recursion comparable to loops performance-wise. Also taking Java's stack architecture into account using recursion typically requires additional thinking if this recursion may not potentially lead to a stack overflow (not in our case, but think of a loop with 100000 iterations replaced with an equivalent recursive implementation).

If you think this method must get a recursive implementation please explain where you see the benefits. Otherwise I would leave the method as it is

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
abego: This won't work.

Assume I change the implementation to 
```
private static final Predicate<Expression> IS_NOT_ENCLOSED_EXPR = n -> !(n.isEnclosedExpr());
```
As the Predicate is used as a parameter to the new `demandParentNode(Node, Predicate<Node>)` it would require me to change that method to `demandParentNode(Node, Predicate<Expression>)`. However in the implementation of `demandParentNode` we assume the result of a `demandParentNode` call is tested against the predicate , and that result is a `Node`.

Or did I missed something.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/LambdaExprContext.java
abego: Good idea. Will change this to
```
            while (p.isEnclosedExpr()) {
                p = p.asEnclosedExpr().getInner();
            }
```

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
abego: Good idea, agreed.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3859Test.java
abego: Nice, I didn't know that.

Will change accordingly

PR #3886, opened 2023-01-31
Refactor: cleanup/refactor code after fixing #3859

abego:
This commit contains various refactorings following the bug fix for https://github.com/javaparser/javaparser/issues/3859, as discussed in https://github.com/javaparser/javaparser/issues/3859#issuecomment-1402781546 etc.

Brief summary:
- introduce local variables to avoid repeated evaluation of expressions.
- move IS_NOT_ENCLOSED_EXPR to Expression, avoiding code duplication
- add NodeWithArguments#getArgumentPosition (two overloads) and Expression#EXCLUDE_ENCLOSED_EXPR, and use them, avoiding code duplication (in TypeExtractor#getParamPos, LambdaExprContext#pos)
- simplify CompilationUnit parsing (in test code)
- add test code


jlerbsc:

PR #3908, opened 2023-02-19
(feat) #1 Branch coverage for applyAddedDiffElement

flaxxen:
Created a simple structure to keep track of which branches are visited in the class when running all tests. The data structure is kept as a static array in the StaticJavaParser class and prints the result in at the end of all core tests and java-solver tests (not tested since this part of the tests does not build for me locally).


jlerbsc:
What is the purpose of this PR? What is the need? You offer us a PR without explaining the problem you are encountering!

PR #3910, opened 2023-02-20
(feat) #12 branch coverage reduce method

flaxxen:
Added a simple branch coverage tool for the reduce method in the TypeSameAsType class in the symbolsolver project.

Also added Jacoco to the solver project to be able to compare results with another tool.


jlerbsc:
Can you stop making PRs on this subject and close them immediately. THANKS.

PR #3913, opened 2023-02-21
refactoring/#19 plan added to findNodeListName

linus1996:


jlerbsc:
Unfortunately this PR brings nothing to the project. Thank you for not proposing PR without added value and without having taken the time beforehand to explain in detail what you have observed so that we can tell you if your proposal is interesting.

Issue #3936, opened 2023-03-02
LexicalPreservingPrinter prints invalid (JDK15) TextBlock when swapped from an Expression


blacelle:
I want to replace a concatenated String expression by a TextBlock:

```
import com.github.javaparser.StaticJavaParser;
import com.github.javaparser.ast.body.VariableDeclarator;
import com.github.javaparser.ast.expr.TextBlockLiteralExpr;
import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;

public class LexicalPreservingTextBlocks {
	static final String testCase = "package eu.solven.cleanthat.engine.java.refactorer.cases.do_not_format_me;\n"
			+ "\n"
			+ "import java.util.Optional;\n"
			+ "\n"
			+ "public class SomeClass {\n"
			+ "\n"
			+ "	String html = \"\" + \"<html>\\n\"\n"
			+ "			+ \"\\t<head>\\n\"\n"
			+ "			+ \"\\t\\t<meta charset=\\\"utf-8\\\">\\n\"\n"
			+ "			+ \"\\t</head>\\n\"\n"
			+ "			+ \"\\t<body class=\\\"default-view\\\" style=\\\"word-wrap: break-word;\\\">\\n\"\n"
			+ "			+ \"\\t\\t<p>Hello, world</p>\\n\"\n"
			+ "			+ \"\\t</body>\\n\"\n"
			+ "			+ \"</html>\\n\";\n"
			+ "}";
	
	static final String newText = "			<html>\n"
			+ "				<head>\n"
			+ "					<meta charset=\"utf-8\">\n"
			+ "				</head>\n"
			+ "				<body class=\"default-view\" style=\"word-wrap: break-word;\">\n"
			+ "					<p>Hello, world</p>\n"
			+ "				</body>\n"
			+ "			</html>";

	public static void main(String[] args) {
		var unit = StaticJavaParser.parse(testCase);

		unit = LexicalPreservingPrinter.setup(unit);

		var expr = unit.findFirst(VariableDeclarator.class).get();

		expr.setInitializer(new TextBlockLiteralExpr(newText));
		
		System.out.println("OK");
		System.out.println(expr.toString());

		System.out.println("KO");
		System.out.println(LexicalPreservingPrinter.print(expr));
	}
}
```

The prettyPrinter gives valid code:

OK
```
html = """
    <html>
    	<head>
    		<meta charset="utf-8">
    	</head>
[...]

Issue #2677, opened 2020-05-17
`CsmTextBlock` not used anywhere


MysterAitch:
As part of #2676, I've come to realise that textblocks are treated as `CsmString`.

Looking further, I've seen that `CsmTextBlock` isn't referred to at all.

MysterAitch:
Might be simpler than I thought if this is the only place that it needs to be edited:

https://github.com/javaparser/javaparser/blob/cdf31ab586e3a6c09631780441ab8e508d32abe2/javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmElement.java#L63-L65

PR #3937, opened 2023-03-02
Fix TextBlockLiteralExpr in LexicalDifferenceCalculator

blacelle:
Suggested fix for https://github.com/javaparser/javaparser/issues/3936.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue3936Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue3936Test.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2023 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.printer.lexicalpreservation;
+
+import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
+
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.TextBlockLiteralExpr;
+import org.junit.jupiter.api.Test;
+
+public class Issue3936Test extends AbstractLexicalPreservingTest {
+	static final String given = "package eu.solven.cleanthat.engine.java.refactorer.cases.do_not_format_me;\n"
+			+ "\n"
+			+ "import java.util.Optional;\n"
+			+ "\n"
+			+ "public class SomeClass {\n"
+			+ "\n"
+			+ "	String html = \"\" + \"<html>\\n\"\n"
+			+ "			+ \"\\t<head>\\n\"\n"
+			+ "			+ \"\\t\\t<meta charset=\\\"utf-8\\\">\\n\"\n"
+			+ "			+ \"\\t</head>\\n\"\n"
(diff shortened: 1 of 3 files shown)


jlerbsc:
I'm not sure if this is the right solution. I need to take the time to look into this problem.

blacelle:
https://github.com/javaparser/javaparser/issues/3936#issuecomment-1452233358

> We also see a comment:
> 
> > // FIXME: csm should be CsmTextBlock -- See also #2677
> 
> And in `CsmTextBlock`, we see a `\n` after `"""`:
> 
> ```
> @Override
> public void prettyPrint(Node node, SourcePrinter printer) {
>     // Note that values within TextBlocks ALWAYS have the \n line ending, per https://openjdk.java.net/jeps/378#1--Line-terminators
>     printer.print("\"\"\"\n");
>     // TODO: Confirm if we need to force this to use {@code \n} separators
>     printer.print(property.getValueAsStringAttribute(node));
>     printer.print("\"\"\"");
> }
> ```

This made me quite confident some '\n' has been lost in the process.

jlerbsc:
The pretty printer and LPP do not work the same way at all. I have no doubt that it is necessary to add a line break but it is more the solution raises some questions. I need to dig a little deeper into the subject.

jlerbsc:
Since the parser is not very restrictive on text blocks and the definition of element syntax in the LPP does not categorize text block delimiters, your proposal can be accepted. 

However, you should simplify your unit test and avoid referring to your project. Thanks.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue3936Test.java
```diff
@@ -0,0 +1,72 @@
...
+package com.github.javaparser.printer.lexicalpreservation;
+
+import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
+
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.TextBlockLiteralExpr;
+import org.junit.jupiter.api.Test;
+
+public class Issue3936Test extends AbstractLexicalPreservingTest {
+	static final String given = "package eu.solven.cleanthat.engine.java.refactorer.cases.do_not_format_me;\n"
```
jlerbsc: You should simplify your unit test and avoid referring to your project

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalDifferenceCalculator.java
```diff
@@ -290,9 +290,9 @@ private void calculatedSyntaxModelForNode(CsmElement csm, Node node, List<CsmEle
         } else if ((csm instanceof CsmString) && (node instanceof TextBlockLiteralExpr)) {
             // FIXME: csm should be CsmTextBlock -- See also #2677
             if (change instanceof PropertyChange) {
-                elements.add(new CsmToken(GeneratedJavaParserConstants.TEXT_BLOCK_LITERAL, "\"\"\"" + ((PropertyChange) change).getNewValue() + "\"\"\""));
+                elements.add(new CsmToken(GeneratedJavaParserConstants.TEXT_BLOCK_LITERAL, "\"\"\"\n" + ((PropertyChange) change).getNewValue() + "\"\"\""));
             } else {
```
jlerbsc, line 296: To set the most appropriate eof character, you can probably use Node.getLineEndingStyle().toString().

blacelle: This is not a matter of style, it is a matter of Java specification Just like prettyPrinter adds a `\n`.

On a system where EOL is '\r', `Node.getLineEndingStyle().toString()` would break `TextBlockLiteralExpr`

blacelle: This is similar to `CsmTextBlock.prettyPrint(Node, SourcePrinter)` not relying on `printer.println()`

blacelle: https://docs.oracle.com/en/java/javase/14/docs/specs/text-blocks-jls.html

Reading https://www.vojtechruzicka.com/java-text-blocks/ makes me feel I overlooked part of the spec

```
Line terminators are normalized to the ASCII LF character, as follows:

1. An ASCII CR character followed by an ASCII LF character is translated to an ASCII LF character.
2. An ASCII CR character is translated to an ASCII LF character.
```

I'm less comfortable with dynamic EOL in this case, but you're the boss.

Should I do the same in `CsmTextBlock.prettyPrint(Node, SourcePrinter)  `?

blacelle: // Note that values within TextBlocks ALWAYS have the \n line ending, per https://openjdk.java.net/jeps/378#1--Line-terminators
        
(`CsmTextBlock` author also felt '\n' were preferable)

Issue #3983, opened 2023-04-06
why FieldDeclaration in an interface calling isStatic() return false?


XenoAmess:

jlerbsc:
Can you provide us with a unit test that demonstrates your problem?

XenoAmess:
> Can you provide us with a unit test that demonstrates your problem?

@jlerbsc sure, here.
https://github.com/javaparser/javaparser/pull/3984

jlerbsc:
This is right. Thank you for reporting this bug which refers to "Every field declaration in the body of an interface is implicitly public, static, and final."
https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html

PR #3984, opened 2023-04-07
add test for showing interface field shall be static & final

XenoAmess:
https://github.com/javaparser/javaparser/issues/3983


jlerbsc:

Issue #2888, opened 2020-11-07
JEP 360: Sealed Classes (Preview)


MysterAitch:
Multiple examples available at: https://openjdk.java.net/jeps/360

> **Summary**
> Enhance the Java programming language with sealed classes and interfaces. Sealed classes and interfaces restrict which other classes or interfaces may extend or implement them.

> **Description**
> A sealed class or interface can be extended or implemented only by those classes and interfaces permitted to do so.
> 
> A class is sealed by applying the sealed modifier to its declaration. Then, after any extends and implements clauses, the permits clause specifies the classes that are permitted to extend the sealed class. For example, the following declaration of Shape specifies three permitted subclasses:
> 
> ```
> package com.example.geometry;
> 
> public abstract sealed class Shape
>     permits Circle, Rectangle, Square {...}
> ```

> **Java Grammar**
> 
> ```
> NormalClassDeclaration:
>   {ClassModifier} class TypeIdentifier [TypeParameters]
>   [Superclass] [Superinterfaces] [PermittedSubclasses] ClassBody
> 
> ClassModifier:
>   (one of)
>   Annotation public protected private
>   abstract static sealed final non-sealed strictfp
> 
> PermittedSubclasses:
>   permits ClassTypeList
> 
> ClassTypeList:
>   ClassType {, ClassType}
> ```

FeldrinH:
With https://github.com/javaparser/javaparser/pull/3997 merged and released, is there more to do or are sealed classes now fully supported in JavaParser?

jlerbsc:
To my knowledge, sealed classes are supported. Did you find any errors?

FeldrinH:
No, but the codebases that I analyse don't have many sealed classes. I was just wondering if this issue could be closed.

jlerbsc:
"sealed is just a modifier. This issue can be closed.

PR #3997, opened 2023-04-17
add Java 17 sealed/non-sealed classes

kris-scheibe:
This adds sealed/non-sealed classes, which were fully added in Java 17, see #2888


jlerbsc:
Thank you for this suggestion which I will look into very soon.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -148,6 +156,20 @@ public ClassOrInterfaceDeclaration setImplementedTypes(final NodeList<ClassOrInt
         return this;
     }
 
+    @Generated("com.github.javaparser.generator.core.node.PropertyGenerator")
+    public ClassOrInterfaceDeclaration setPermittedTypes(final NodeList<ClassOrInterfaceType> permittedTypes) {
```
jlerbsc, line 160: Can you confirm for me that the generated properties are actually generated?

kris-scheibe: I did run the metamodel and core generator scripts, which added the getters/setters after I added the field. But that didn't modify the `@AllFieldsConstructor`-constructor or the main-constructor. Those I had to update manually.

Also the generator updated pretty much all files, mostly by adding/removing whitespace in imports and such. So I only kept the changes related to the sealed/non-sealed logic and reverted everything else.

jlerbsc: I admit that I don't know this part very well (generator) but what seems curious to me, if I understand your answer correctly, is that the generator generates the following code but does not generate the property "permittedTypes".

    this.permittedTypes.setParentNode(null);

For my information, how did you know you had to run the generators and manipulate the metamodel? Did you consult any documentation?

kris-scheibe: I didn't find any documentation for the generator (only IDE-setup, coding-guidelines, etc.). After adding the `permittedTypes` field I just looked around how to invoke the generator for the accessors and other functions. This lead me to the `javaparser-core-generators` module, the `run-generators` Maven-profile and finally the `run_core_generators.sh` script, which contains the comment
```
# Runs all the code generators.
# If the node structure was changed, run_metamodel_generator.sh first!
```
This is essentially the "documentation" to first run the metamodel-generator and then the code-generator. (Plus a bit of inductive reasoning, as I've seen similar projects and code-generators before.)

So, yes, after adding the `permittedTypes` field, running the generators generated the matching accessors and other utility-functions in `ClassOrInterfaceDeclaration`, updated various visitor- and observer-classes, and the metamodel to each include the new field.
I'm not sure, why the `MainConstructorGenerator` didn't update the "main constructor" properly, as I had to update this one manually.

---

Also the generators changed the formatting of all files, that contained generated code by stripping out empty lines in the import statements, so from:
```java
import com.github.javaparser.resolution.Resolvable;
import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;

import java.util.Optional;

import static com.github.javaparser.utils.Utils.assertNotNull;
```
to
```java
import com.github.javaparser.resolution.Resolvable;
import com.github.javaparser.resolution.declarations.ResolvedReferenceTypeDeclaration;
import java.util.Optional;
import static com.github.javaparser.utils.Utils.assertNotNull;
```

I reverted these changes, as I didn't want to inflate the PR with hundreds of reformatted, but unrelated, java files.

jlerbsc: The only documentation I found on this topic is here. Your suggestions for improving it are welcome. https://javaparser.org/code-generation-and-maven-in-javaparser/

jlerbsc:
If you have any other suggestions, feel free to submit them. For example, there is support for local interfaces/enumerations... (java16) or certain specific features provided by java18 and java19.

jlerbsc:
Although I don't know this part very well it seems to me that this PR is correct. However, it remains to be ensured that the Lexical Preserving Printer and the Symbol Solver also meet expectations. Thank you for your involvement.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
kris-scheibe: I did look on the javaparser website, specifically https://javaparser.org/contributing.html , but this only links back to various github pages.
Honestly, without knowing that the code-generator exists I wouldn't have looked outside the contributing-section.

The contributing-section links to https://github.com/javaparser/javaparser/blob/master/CONTRIBUTING.md, which has links to the github wiki. IMO, the wiki is the correct place to document how to use javaparser and how to contribute. Or at least there should be pages in the wiki that link back to specific javaparser.org blog-articles, so that users/contributers can find the information, without already knowing where to look. 

The blog just isn't a good place for technical documentation, but I also don't know the "official" status of javaparser.org. To me a lot of information and documentation seems to be spread across the javaparser.org and tomassetti.me website/blog and the github wiki and documents, which makes it hard to find specific documentation.

kris-scheibe:
At work we use javaparser as a code-generator, but we only recently transitioned from Java 11 to Java 17 and haven't really started using the new language-features yet. But I can have a look into new language-features, especially as the next LTS-version will be released this fall.

@jlerbsc are you OK with a PR, that only changes the formatting to what the code-generators generate now? This would make future contributions much easier! Ideally running the code-generator on a freshly checked-out project shouldn't change anything, but this isn't the case right now.

jlerbsc:
Yes of course. Personally I have never started it. I know old contributors did this at least once a year.
To be completely honest, I am "the last of the mohicans" the previous historical contributors no longer contribute to the project and I try to maintain it. New contributors would be welcome. Out of curiosity, what company do you work for?

kris-scheibe:
Thank you for keeping the project alive! I know that being a maintainer of an OSS-project can be a thankless job.
I'd be happy to contribute to this project -- at least to the parser (I haven't really worked much with the symbol-solver, but the parser itself seems pretty straight-forward).

I work for a beauty-retailer in Germany. I'm not going to name my employer here, as we don't have an official policy regarding contributions to OSS-projects. In the past, direct contributions have just been the fastest way to get bugs fixed and features added. 
Also I like to contribute back as projects like this one in my free time as it makes my work so much easier!

blacelle:
Hello. I feel this is missing from https://github.com/javaparser/javaparser/blob/master/changelog.md

It impacted us (positively, but unexpectedly) in https://github.com/diffplug/spotless/pull/1725

jlerbsc:
No, the change log is updated automatically when we publish the project. What you see in Next Release is just a template.

blacelle:
No ? Where is this (#3997, or a ref to `sealed` keyword) referred in the changelog then? I believe this is available in JavaParser 3.25.3

jlerbsc:
Yes this PR is merged since 3.25.3 but the changelog does not mention it. It is a mistake. The change log is updated automatically so i don't know why there is an issue.

Auties00:
Is there any reason why com.github.javaparser.ast.Modifer doesn't provide a named constructor for the SEALED and NON_SEALED modifiers? Every other modifier is covered as far as I can tell. I can open a quick PR if it was missed here for whatever reason

jlerbsc:
Thank you for your improvement suggestion.

PR #4049, opened 2023-06-01
Support the switch case block without brace when parent node is SwitchEntry

### LINKED ISSUE #4048: switch case block without brace when parent node is SwitchEntry
when print the final code, the switch have brace like below:

```
            switch (x) {
                case 1: {
                    System.out.println(11);
                }
                break;
                case 2: {
                    System.out.println(22);
                }
                break;
            }
```

should be:
```
            switch (x) {
                case 1:
                    System.out.println(11);
                    break;
                case 2:
                    System.out.println(22);
                    break;
            }
```

shijinkui:
Support the switch case block without brace when parent node is SwitchEntry

Fixes #4048.


jlerbsc:
This fix does not seem correct.

PR #4057, opened 2023-06-07
Fix #4056 isDeclaredInInterface() returns true for fields declared inside enumerations contained in an interface

### LINKED ISSUE #4056: [BUG] isDeclaredInInterface() returns true for fields inside Enums inside Interface
Hi,

The method `#isPublic()` of `FieldDeclaration` checks either for the presence of a `public` modifer, or whether the field is part of an interface (`isDeclaredInInterface()`). The reasoning is that fields inside interfaces are always implicitly `public static final`. Unfortunately, this check is no implemented properly: it searches for the nearest `ClassOrInterfaceDeclaration` in the parent-hierarchy of the node, and returns `true`, if that node is an interface.

In the case where the field is part of an Enum (which is not a `ClassOrInterfaceDeclaration`, but rather a separate `EnumDeclaration`), the Enum is skipped and instead the enclosing interface is found.

Test Case:
```java

import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.hasSize;
import static org.junit.jupiter.api.Assertions.assertFalse;

class JavaParserIsPublicTest {

  @Test
  void testEnumWithPrivateFieldInsideInterface() {
    var source = """
        interface Outer {
          enum Numbers {
            ONE(1),
            TWO(2),
            THREE(3);

            Numbers(int i) {
              this.i = i;
            }

            private final int i;
          }
        }""";
    var cu = StaticJavaParser.parse(source);
    var types = cu.getTypes();
    assertThat(types, hasSize(1)); // OK

    var interfaceMembers = types.get(0).asClassOrInterfaceDeclaration().getMembers();
    assertThat(interfaceMembers, hasSize(1)); // OK

    var enumType = interfaceMembers.get(0).asEnumDeclaration();
    var enumFields = enumType.getFields();
    assertThat(enumFields, hasSize(1)); // OK

    var i = enumFields.get(0);
    assertFalse(i.isPublic()); // FAILS
  }
}
```

It seems this also applies to `isStatic()` and `isFinal()`, as they also use `isDeclaredInInterface`.

Suggested fix: extend `isDeclaredInInterface` to properly also check whether the field is declared in an `EnumDeclaration`, instead of only checking `ClassOrInterfaceDeclaration`.

Tested Version: `3.25.3`

A real-life occurrence of such a code sample (private field inside Enum inside Interface) can be found e.g. in `java.lang.System.Logger.Level#severity` of the JDK (17 Eclipse Adoptium).

Elewyth:
Fixes #4056 .

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/FieldDeclarationTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/FieldDeclarationTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser.ast.body;
 
+import com.github.javaparser.StaticJavaParser;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.Modifier;
 import com.github.javaparser.ast.Modifier.Keyword;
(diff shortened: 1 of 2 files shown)


jlerbsc:

PR #4062, opened 2023-06-11
Fixes missing named constructor in Modifier.java

Auties00:
Related to my comment at https://github.com/javaparser/javaparser/pull/3997#issuecomment-1586218487


jlerbsc:
Can you add test cases?

Auties00:
> Can you add test cases?
Sorry, where should I add them? It's just a method so I'm not sure what I could test

PR #4063, opened 2023-06-11
Fix missing permits clause in DefaultPrettyPrinterVisitor

Auties00:
Let's take this code snippet:
```
var parser = new JavaParser();
parser.getParserConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_17);
var result = parser.parse("sealed interface A permits B {}");
System.out.println(result );
```
To output:
`sealed interface A permits B {}`
Yet, the actual output is:
`sealed interface A {}`
This is because the permits clause doesn't get printed out by the DefaultPrettyPrinterVisitor.
I've fixed this issue(I've only added the printing for classes/interfaces as records and enums can't be sealed)


jlerbsc:
Can do a PR with only appropriate changes.

PR #4078, opened 2023-06-22
test: always use "\r\n" line endings when saving in SourceRootTest

### LINKED ISSUE #4077: After building JavaParser (with tests) on MacOS multiple committed files are modified
When doing a `mvn clean install` of JavaParser on MacOS multiple committed files are modified:
Screenshot 2023-06-22 at 21 59 24

The difference is in the line endings: the files are committed with `"\r\n"` line endings, but after running the tests on MacOS the files have `"\n"` line endings.

Having those files modified is annoying when developing JavaParser code as one always have to rollback those changes or explicitly exclude those files before commiting changes.

_(From what I can tell the files are modified in the `SourceRootTest`, by the test methods `saveInCallback()` and `saveInCallbackParallelized()`.)_

abego:
see https://github.com/javaparser/javaparser/issues/4077

Fixes #4077.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/SourceRootTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/SourceRootTest.java
@@ -25,6 +25,9 @@ import com.github.javaparser.ParseProblemException;
 import com.github.javaparser.ParseResult;
 import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.printer.DefaultPrettyPrinter;
+import com.github.javaparser.printer.configuration.DefaultConfigurationOption;
+import com.github.javaparser.printer.configuration.DefaultPrinterConfiguration;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.mockito.MockedStatic;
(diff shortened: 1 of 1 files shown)


jlerbsc:
I cannot accept this change which will impact all project users.

abego:
> I cannot accept this change which will impact all project users.

Actually, I think this change has no impact on any project *user* (the ones that use JavaParser as a library) as the change is not in the main code, but in the test code.

Anyhow...

As an alternative: would you accept a PR that changes the seven files to use "\n" instead of "\r\n" line separators? This way the files would no longer be marked as modified after running the tests, on all platforms (Unix, MacOS and Windows). This assumes the proper "core.autocrlf" setting is configured  (core.autocrlf=true on Windows machines, core.autocrlf=input on Unix/MacOS). 

The seven files are:

jlerbsc:
> Actually, I think this change has no impact on any project _user_ (the ones that use JavaParser as a library) as the change is not in the main code, but in the test code.

I actually meant the people who compile the project (who are also users).

jlerbsc:
> would you accept a PR that changes the seven files to use "\n" instead of "\r\n" line separators? This way the files would no longer be marked as modified after running the tests, on all platforms (Unix, MacOS and Windows).

It seems safer to me.

abego:
> > would you accept a PR that changes the seven files to use "\n" instead of "\r\n" line separators? This way the files would no longer be marked as modified after running the tests, on all platforms (Unix, MacOS and Windows).
> 
> It seems safer to me.

Fine. 

I will close this PR and create a new one in the next days with the alternative solution.

PR #4083, opened 2023-06-26
fix line separators of selected test files

### LINKED ISSUE #4077: After building JavaParser (with tests) on MacOS multiple committed files are modified

abego:
to resolve https://github.com/javaparser/javaparser/issues/4077

as discussed here: https://github.com/javaparser/javaparser/pull/4078#issuecomment-1606043764


jlerbsc:
It seems correct to me. Thank you for your contribution.

jlerbsc:
After analysis, the cause of this problem comes mainly from the SourceRoot class which uses printing configuration parameters with default values. The default end-of-line character is system-specific (on Windows it's CRLF...). This only raises a problem when this class is asked to parse and save the result, which is done in particular by the SourceRootTest test class. In this case, ideally, one should use the end-of-line character used in the parsed file.

abego:
> After analysis, the cause of this problem comes mainly from the SourceRoot class which uses printing configuration parameters with default values. The default end-of-line character is system-specific (on Windows it's CRLF...). This only raises a problem when this class is asked to parse and save the result, which is done in particular by the SourceRootTest test class. In this case, ideally, one should use the end-of-line character used in the parsed file.

That was also my analysis of the problem and my initial PR  (https://github.com/javaparser/javaparser/pull/4078/files) was exactly your suggestion "one should use the end-of-line character used in the parsed file.", i.e. use "\r\n" as the end-of-line character as that was the one used in the parsed files.

Anyhow, the current solution (store the files with "\n" in Git and let Git change it to the actual system-specific ending on checkout) does not require extra changes in the test code, hence in much simpler, so I would keep it.

jlerbsc:
> Anyhow, the current solution (store the files with "\n" in Git and let Git change it to the actual system-specific ending on checkout) does not require extra changes in the test code, hence in much simpler, so I would keep it.

This solution doesn't work because the test class rewrites the classes after parsing them and it doesn't default to using the system end-of-line character. I noticed this problem after checking out the project and running the tests.

jlerbsc:
> That was also my analysis of the problem and my initial PR (https://github.com/javaparser/javaparser/pull/4078/files) was exactly your suggestion "one should use the end-of-line character used in the parsed file.", i.e. use "\r\n" as the end-of-line character as that was the one used in the parsed files.

Indeed your first solution was correct.

PR #4117, opened 2023-08-14
Improvements and support for annotations

JulianHahn96:
I am using the library to build a gradle plugin, that analyzes whole gradle projects including all its dependencies.

While doing so i noticed a few things that were missing (for example the support to analyse not only annotation classes and default values, but the actual declared annotations and its values). 

Also there were a few bugs / different behaviours between java parser, reflection and javassist implementations.

I tried to fix those and add support for annotations. Hope that this will make it into the official repository, as then i do not need to use my local builds anymore.

For details see the commit messages.


jlerbsc:
Thank you for this PR but there are too many changes in this one. Ideally it should be cut into small changes or corrections that can be verified.

PR #4122, opened 2023-08-20
Replace deprecated command with environment file

70825:
I replace deprecated `set-output` command to `$GITHUB_OUTPUT` environment file to use

reference : https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

### AS-IS
```yaml
echo "::set-output name=sha_long::$SHA_LONG"
```

### TO-BE
```yaml
echo "sha_long=$SHA_LONG" >> $GITHUB_OUTPUT
```

PR #4130, opened 2023-08-27
Fix: apply multiline strings

70825:
I apologize for #4122  😭
Initially, I assumed it was a simple contribution and neglected to run the workflow tests
However, I've now modified the code and verified that the tests are working properly

This was created based on the reference link provided below
reference : https://github.com/kubernetes/minikube/blob/master/.github/workflows/docs.yml#L28-L31


jlerbsc:
Thank you.

PR #4166, opened 2023-10-19
Using JAXP on XMLPrinter implementation

### LINKED ISSUE #2749: Generated XML is invalid
Using XmlPrinter, the code fragment

```
throw new IllegalArgumentException(
                                    "Unrecognized value STRIP_WHITESPACE = '" + value +
                                            "': must be 'all', 'none', or 'ignorable'");
```
produces the ill-formed XML output:

`<statement type='ThrowStmt'><expression type='ObjectCreationExpr'><type type='ClassOrInterfaceType'><name type='SimpleName' identifier='IllegalArgumentException'></name></type><arguments><argument type='BinaryExpr' operator='PLUS'><left type='BinaryExpr' operator='PLUS'><left type='StringLiteralExpr' value='Unrecognized value STRIP_WHITESPACE = ''></left><right type='NameExpr'><name type='SimpleName' identifier='value'></name></right></left><right type='StringLiteralExpr' value='': must be 'all', 'none', or 'ignorable''></right></argument></arguments></expression></statement></statements></entrie></entries></statement><statement type='BreakStmt'></statement></statements></statement>`

which is invalid because of the failure to escape the single quotes.

lcbarcellos:
Fixes #2749

The original implementation "manually" generates XML and does not take care of special cases, when illegal characters are present on attribute values. The proposed solution avoid the "manual" generation of XML, by using JAXP API for outputing XML structure.

Also, instead of generating the complete XML string in memory, user of XMLPrinter has now the option of streaming data directly to an output file, by using public methods of the class.


lcbarcellos:
What does that codecov fail mean? It seems it's not related to the code in the commits, but related to some bug in codecov. Is it possible to rerun that specific check?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/XmlPrinterTest.java
```diff
@@ -25,38 +25,73 @@
...
+            dbf.setCoalescing(true);
+            dbf.setIgnoringElementContentWhitespace(true);
+            dbf.setIgnoringComments(true);
+            db = dbf.newDocumentBuilder();
+        } catch (ParserConfigurationException ex) {
+            throw new RuntimeException(ex);
+        }
+    }
+
+    public Document getDocument(String xml) throws SAXException, IOException {
```
jlerbsc, line 118: Is there no need to close the inputStream?

javaparser-core/src/main/java/com/github/javaparser/printer/XmlPrinter.java
```diff
@@ -42,76 +47,120 @@ public XmlPrinter(boolean outputNodeType) {
...
+        return stringWriterOutput(node, "root").toString();
     }
 
+    // Kept for backward compatibility
     public void output(Node node, String name, int level, StringBuilder builder) {
+        builder.append(stringWriterOutput(node, name).toString());
+    }
+
+    public StringWriter stringWriterOutput(Node node, String name) {
+        StringWriter stringWriter = new StringWriter();
```
jlerbsc, line 87: I imagine that this method allows you to implement data streaming in a file, perhaps you can specify in javadoc that the resource returned by the method must be closed by the user?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/XmlPrinterTest.java
lcbarcellos: As per [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/io/ByteArrayInputStream.html#close--), closing a ByteArrayInputStream has no effect.

jlerbsc:
Thank you for this interesting development. There are some outstanding questions and it seems to me that it would be interesting to add some comments to clarify the intentions of certain added methods. A few additional unit test cases would also be welcome to illustrate the use of the new methods and more particularly the data streaming that you mention.

jlerbsc:
> it's not related to the code

you are right it's not related to the code.

lcbarcellos:
> Thank you for this interesting development. There are some outstanding questions and it seems to me that it would be interesting to add some comments to clarify the intentions of certain added methods. A few additional unit test cases would also be welcome to illustrate the use of the new methods and more particularly the data streaming that you mention.

I've made minor changes and added documentation as well as additional tests. Also, I've made changes in small dedicated commits hoping that will make the revision easier.

jlerbsc:

PR #4177, opened 2023-10-27
Fixes #2625 Add messages to thrown exceptions

### LINKED ISSUE #2625: Add messages to thrown exceptions

oannhpham:
Fixes #2625 

I have added messages to `UnsupportedOperationException` that did not have messages and modified some which I thought were unclear. Please let me know if there are any message that are unclear.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
@@ -222,7 +222,7 @@ public class ConcreteSyntaxModel {
             throw new IllegalStateException(s);
         });
         if (!concreteSyntaxModelByClass.containsKey(nodeClazz)) {
-            throw new UnsupportedOperationException(nodeClazz.getSimpleName());
+            throw new UnsupportedOperationException("Class " + nodeClazz.getSimpleName() + " not found!");
         }
         return concreteSyntaxModelByClass.get(nodeClazz);
     }
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Added.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Added.java
@@ -92,7 +92,7 @@ public class Added implements DifferenceElement {
             if (element instanceof CsmToken) {
             return new TokenTextElement(((CsmToken) element).getTokenType(), ((CsmToken) element).getContent());
         }
-        throw new UnsupportedOperationException(element.getClass().getSimpleName());
+        throw new UnsupportedOperationException("Unsupported element type: " + element.getClass().getSimpleName());
     }
 
     /*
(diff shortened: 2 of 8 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalDifferenceCalculator.java
```diff
@@ -325,7 +325,7 @@ private void calculatedSyntaxModelForNode(CsmElement csm, Node node, List<CsmEle
         } else if (csm instanceof CsmChild) {
             elements.add(csm);
         } else {
-            throw new UnsupportedOperationException(csm.getClass().getSimpleName() + " " + csm);
+            throw new UnsupportedOperationException("Unknown element type: " + csm.getClass().getSimpleName() + " " + csm);
```
jlerbsc: This is a concrete syntax model type

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/Added.java
```diff
@@ -92,7 +92,7 @@ public TextElement toTextElement() {
             if (element instanceof CsmToken) {
             return new TokenTextElement(((CsmToken) element).getTokenType(), ((CsmToken) element).getContent());
         }
-        throw new UnsupportedOperationException(element.getClass().getSimpleName());
+        throw new UnsupportedOperationException("Unsupported element type: " + element.getClass().getSimpleName());
```
jlerbsc, line 95: This is a concrete syntax model type

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -222,7 +222,7 @@ public static CsmElement forClass(Class<? extends Node> nodeClazz) {
             throw new IllegalStateException(s);
         });
         if (!concreteSyntaxModelByClass.containsKey(nodeClazz)) {
-            throw new UnsupportedOperationException(nodeClazz.getSimpleName());
+            throw new UnsupportedOperationException("Class " + nodeClazz.getSimpleName() + " not found!");
```
jlerbsc, line 225: This is a concrete syntax model class

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/NodeText.java
```diff
@@ -137,7 +137,7 @@ public void remove(TextElementMatcher matcher, boolean potentiallyFollowingWhite
                             elements.remove(i);
                         }
                     } else {
-                        throw new UnsupportedOperationException();
+                        throw new UnsupportedOperationException("There is no element to remove!");
```
jlerbsc, line 140: This is a text element

jlerbsc:

Issue #4000, opened 2023-04-19
Improving documentation


jlerbsc:
Suggested by @kris-scheibe 

I did look on the javaparser website, specifically https://javaparser.org/contributing.html , but this only links back to various github pages. Honestly, without knowing that the code-generator exists I wouldn't have looked outside the contributing-section.

The contributing-section links to https://github.com/javaparser/javaparser/blob/master/CONTRIBUTING.md, which has links to the github wiki. IMO, the wiki is the correct place to document how to use javaparser and how to contribute. Or at least there should be pages in the wiki that link back to specific javaparser.org blog-articles, so that users/contributers can find the information, without already knowing where to look.

The blog just isn't a good place for technical documentation, but I also don't know the "official" status of javaparser.org. To me a lot of information and documentation seems to be spread across the javaparser.org and tomassetti.me website/blog and the github wiki and documents, which makes it hard to find specific documentation.

AqilNumber1:
Hi there, I am trying to come up with a solution for this issue :)

jlerbsc:
Thank you for your suggestion but it seems to me that this subject should be taken care of by the JP team.

Issue #3999, opened 2023-04-19
Consider adding a lenient mode


NebelNidas:
Currently, JavaParser is extremely strict and fails immediately when it encounters not 100%-compliant Java code. In many use-cases (like when parsing decompiled code), this behavior is undesirable and should be made optional. This includes rules like `super` constructor calls not being in the first line of a subclass's constructor, or inner classes' names starting with numbers.

jlerbsc:
It is a choice that was certainly made when the project was developed. It seems to me very difficult to modify this choice today. Besides, I don't know all the implications of a less strict parsing. To my knowledge it is indeed a pre-requisite to use JavaParser.

revusky:
Well, the reason that JavaParser cannot be fault-tolerant is that it is built using the legacy JavaCC parser generator, which has zero concept of fault tolerance or error recovery. There is a (vastly) more advanced version of JavaCC, [CongoCC](https://github.com/congo-cc/congo-parser-generator) that can generate fault-tolerant parsers. You can read the key information about how this works [here](https://parsers.org/javacc21/the-promised-land-fault-tolerant-parsing/).

I doubt this is the appropriate venue to be discussing this, however. If anybody wants to ask anything about this topic, I would suggest that they post questions or comments on the [CongoCC discussion forum](https://discuss.congocc.org/). That would be my preference, but I'm hardly picky. I guess you could use the built-in issue tracker or discussions feature that is part of git-hub. (But obviously, I mean the ones that are associated with the [Congo parser generator project](https://github.com/congo-cc/congo-parser-generator/).

AqilNumber1:
Hi, I am trying to work on some sort of solution for this issue

jlerbsc:
This subject is too sensitive to be handled by someone who does not have very good knowledge of the JP project.

AqilNumber1:
> This subject is too sensitive to be handled by someone who does not have very good knowledge of the JP project.

Thanks for the heads up on the difficulty. Although I understand I may not create a proper full solution, I am still interested in exploring this problem to strengthen my knowledge on this project :)

revusky:
> This subject is too sensitive to be handled by someone who does not have very good knowledge of the JP project.

As I said in a previous comment, there is simply no prospect of JavaParser having much in the way of fault-tolerant parsing capabilities. This is because it is built on top of a very old parser generator (JavaCC) that itself has zero concept of fault-tolerant parsing, i.e. error recovery or backtracking. 

[...]

PR #4179, opened 2023-10-28
Lenient mode and docs solution

AqilNumber1:
A partial idea I came up with for the lenient mode issue #3999. I have created a class called LenientParser + LenientToken with the thoughts that it can integrate the parser but at the same time allow the user to tweak settings to make the parser's guidelines less strict.

For issue #4000 , I have tried coming up with a simple solution by making an md file titled "BLOG". The concept was that all articles from both websites can have links into this one document for convenience along with the copied contribution guidelines.

I hope these would spark new ideas for whoever might like to fully finish these issues in the future.


jlerbsc:
I cannot validate this PR because on the one hand there are too many files impacted. I don't understand why there are files that have been deleted. On the other hand, thank you for separating each subject in a PR.

Issue #4067, opened 2023-06-14
COLUMN_ALIGN_PARAMETERS aligns arguments instead of parameters


Auties00:
Hello,
while trying to align the parameters in a record as IntelliJ does, I stumbled upon COLUMN_ALIGN_PARAMETERS.
Unfortunately, that option actually aligns arguments, not parameters.
I could open a PR to modify its behaviour to align parameters(methods and records have parameters), and add a COLUMN_ALIGN_ARGUMENTS option to align arguments(method invocations and class initializations have arguments), but this would be a breaking change.

jlerbsc:
You can make a proposal if you wish.

swathisomayaji:
Hi,

I am Swathi Somayaji, and I am a student at Australian national university. I am interested in working on issue #4067, "COLUMN_ALIGN_PARAMETERS aligns arguments instead of parameters" and i was wondering if this issue could be assigned?

I have been following the javaparser project for some time now, and I am impressed with the work that has been done. I am also interested in learning more about the project and contributing to its development.

I have some experience with Java parsing and type solving, and I am confident that I can fix the issue in a timely and efficient manner. I am also happy to communicate with the maintainers of the project to ensure that my solution meets their requirements.


Thank you for your time and consideration.

Sincerely,
Swathi Somayaji

hubowen7:
Hello,

I am Bowen Hu, and I am also a student at Australian national university. I'm also very interested in issue #4067, and I'm happy to have chosen the same issue as Swathi Somayaji. I would also like to ask if this problem can be assigned to us to solve? I have also studied this project carefully and have some ideas about the solution to this issue. I am very grateful and impressed with the work you have done on this project, it is a very useful project. I also have some learning and practical experience in this area, and I hope to cooperate with developers and other contributors to solve this issue.

Thank you very much for your time and consideration!

Best,
Bowen Hu

jlerbsc:
It's just a naming problem. You should not introduce an API break for this type of issue.
There is also no need for 2 people to work on the same subject. @swathisomayaji since you were the first to wish to work on this subject you can take it in charge.

PR #4182, opened 2023-10-29
An attempt to fix the issue COLUMN_ALIGN_PARAMETERS aligns arguments instead of parameters (#4067)

hubowen7:
In this PR, I set out to add a 'COLUMN_ALIGN_PARAMETERS' option to solve the problem of actually aligning arguments instead of parameters, see issue #4067. At the same time, I also added a few test cases to verify this fix and ensure that the same problem does not occur again in the future. I'm not sure if the modification is correct since this is my first attempt at a pull request. If there are any unreasonable modifications, you can return them to me at any time and make new modifications.


jlerbsc:
How does adding a new position solve the problem?
Furthermore, you have not read my comment "It's just a naming problem. You should not introduce an API break for this type of issue."

jlerbsc:
I am closing this PR because it absolutely does not resolve the issue.

PR #4183, opened 2023-10-29
Fix: issue #2751 new HashSet()" != "new HashSet<>()

### LINKED ISSUE #2751: Generated XML does not distinguish "new HashSet()" from "new HashSet<>()"
The XML serializer does not distinguish an empty NodeList property from a null NodeList property.

In consequence "new HashSet()" and "new HashSet<>()" produce the same output: in one case type.typeArguments is null, in the other case it is empty.

lcbarcellos:
Change XML generator so that an empty type argument list is emmitted as an empty XML element. All other empty lists keep not being emmitted.

Fixes #2751 - Generated XML does not distinguish "new HashSet()" from "new HashSet<>()"


jlerbsc:
This fix seems reasonable to me.

PR #4184, opened 2023-10-31
MethodResolutionLogic.isMoreSpecific throws exception on overloaded method w/varargs

fabgo:
The isMoreSpecific() may throw an ArrayIndexOutOfBoundsException if the method being resolved is overloaded, and one or more method signatures contain a vararg parameter.

In this case, the number of parameters varies, and the existing logic does not exit the loop once a vararg is detected. Instead, it continues, which then causes an ArrayIndexOutOfBoundsException when it tries to retrieve the next method parameter after the vararg.

The fix is to exit the loop and return true as soon as a vararg is detected in the parameter list.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -820,7 +820,7 @@ public class MethodResolutionLogic {
 
             // If B is vararg and A is not, A is more specific
             if (tdB.isArray() && tdB.asArrayType().getComponentType().isAssignableBy(tdA)) {
-                oneMoreSpecificFound = true;
+                return true;
             }
         }
         return oneMoreSpecificFound;


jlerbsc:
Hi, Thank you for this contribution. Can you add unit tests.

fabgo:
There are no existing tests that cover the functionality in this method. I am new to the code base, so no idea where to even begin to add tests. But if you point me in the right direction, e.g. where to add the test, how to run it, etc. I can give it a try.

jlerbsc:
I can create the test case if you provide me with sample code that doesn't work.

fabgo:
OK, let me work on that.

fabgo:
I was not able to distill this down to a test case. I am not quite sure what is the pattern that causes this to fail.

Here is some sample code that is failing though:

```
import org.jooq.DSLContext;
import org.jooq.SQLDialect;
import org.jooq.impl.DSL;

import java.sql.Connection;

import static org.jooq.impl.DSL.field;
import static org.jooq.impl.DSL.table;

public class DatabaseUtil {
    public static void insert(String name, String value) {
        Connection conn = null;
        DSLContext context = DSL.using(conn, SQLDialect.POSTGRES);
        context.insertInto(table("Foo"), field("name"), field("value"))
                .values(name, value)
                .onConflict(field("name"))  // Assuming 'name' is a unique column
                .doUpdate()
                .set(field("value"), value)
                .execute();
    }
}
```

```
Caused by: java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266)
	at java.base/java.util.Objects.checkIndex(Objects.java:359)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at com.github.javaparser.resolution.MethodUsage.getParamType(MethodUsage.java:141)
	at com.github.javaparser.resolution.logic.MethodResolutionLogic.isMoreSpecific(MethodResolutionLogic.java:800)
	at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicableUsage(MethodResolutionLogic.java:834)
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistUtils.solveMethodAsUsage(JavassistUtils.java:85)
	at com.github.javaparser.symbolsolver.javassistmodel.JavassistInterfaceDeclaration.solveMethodAsUsage(JavassistInterfaceDeclaration.java:102)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ContextHelper.solveMethodAsUsage(ContextHelper.java:48)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:181)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:565)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethodAsUsage(MethodCallExprContext.java:122)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solveMethodAsUsage(JavaParserFacade.java:656)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:325)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:65)
	at com.github.javaparser.ast.expr.MethodCallExpr.accept(MethodCallExpr.java:118)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:501)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:349)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:331)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.AbstractJavaParserContext.findTypeDeclarations(AbstractJavaParserContext.java:200)
	... 23 more
```

jlerbsc:
There are too many classes that are not known in your example. Please simplify it so that we can easily reproduce the problem.

jlerbsc:
You did not provide us with a simple test case, perhaps you solved your issue? Without new information I will close this issue/PR.

jlerbsc:
I tried to reproduce this problem with this simple case but I couldn't do it.

    @Test
    public void issue4184() {
        config.setSymbolResolver(new JavaSymbolSolver(new ReflectionTypeSolver(false)));

        		"	interface Bar {\n"
        		+ "		void m(String... params);\n"
        		+ "\n"
        		+ "\n"
        		+ "		void m(String strings);	\n"
        		+ "	}\n"
        		+ "	\n"
        		+ "	class Baz implements Bar {\n"
        		+ "\n"
        		+ "		@Override\n"
        		+ "		public void m(String strings) {\n"
        		+ "			// TODO Auto-generated method stub\n"
        		+ "			\n"
        		+ "		}\n"
        		+ "\n"
        		+ "		@Override\n"
        		+ "		public void m(String... strings) {\n"
        		+ "			// TODO Auto-generated method stub\n"
        		+ "			\n"
        		+ "		}\n"
        		+ "		\n"
        		+ "     void foo() {\n"
        		+ "			m(\"\");\n"
        		+ "		}"
        		+ "	}"
        );

        for (MethodCallExpr mce : cu.findAll(MethodCallExpr.class)) {
				String sign = mce.resolve().getQualifiedSignature();
				System.out.println(sign);
            }
        
	}

PR #4186, opened 2023-11-01
Fixes resolution of method references that are default methods on interfaces

fabgo:
Currently, default methods are treated the same as instance methods. Instead, treat default methods the same as static methods, since they do not have an instance as the first parameter.

Sample code which currently breaks:

public class InterfaceWithDefaultMethod {
    Map<String, String> testMap = new ConcurrentHashMap<>();

    public void invokeDefaultMethod(MyInterface myInterface) {
        Optional<Integer> priority = Optional.of(4);
        priority.map(myInterface::toString).orElse("0");
    }

    public interface MyInterface {
        default String toString(int priority) {
            return Integer.toString(priority);
        }
    }
}

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -405,12 +405,13 @@ public class JavaParserFacade {
         Set<MethodUsage> allMethods = resolvedTypdeDecl.getAllMethods();
 
         if (scope.isTypeExpr()) {
-            // static methods should match all params
-            List<MethodUsage> staticMethodUsages = allMethods.stream()
-                    .filter(it -> it.getDeclaration().isStatic())
+            // default and static methods should match all params
+            List<MethodUsage> nonInstanceMethodUsages = allMethods.stream()
+                    .filter(it -> it.getDeclaration().isStatic() || it.getDeclaration().isDefaultMethod())
                     .collect(Collectors.toList());
 
-            result = MethodResolutionLogic.findMostApplicableUsage(staticMethodUsages, methodReferenceExpr.getIdentifier(), paramTypes, typeSolver);
+            result = MethodResolutionLogic.findMostApplicableUsage(nonInstanceMethodUsages,
+                    methodReferenceExpr.getIdentifier(), paramTypes, typeSolver);
 
             if (!paramTypes.isEmpty()) {
                 // instance methods are called on the first param and should match all other params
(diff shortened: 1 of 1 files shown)


fabgo:
Another fix. I added sample code to this one. If you give me a quick idea of how tests should be added, I can start doing that.

fabgo:
Code I used to test:

```
public class SymbolResolverTest {
        ParserConfiguration config = new ParserConfiguration()
                .setSymbolResolver(new JavaSymbolSolver(new ReflectionTypeSolver()))
                .setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_17);

        CompilationUnit cu = StaticJavaParser.parse(new FileInputStream("InterfaceWithDefaultMethod.java"));
        cu.findAll(MethodCallExpr.class).forEach(MethodCallExpr::resolve);
    }
}
```

jlerbsc:
> Currently, default methods are treated the same as instance methods. Instead, treat default methods the same as static methods, since they do not have an instance as the first parameter.

According to https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-9.4
A default method is an instance method declared in an interface with the default modifier.

fabgo:
OK. This may not be a good fix. The question then is, why don't default methods have an instance parameter?

fabgo:
Turns out this is not limited to default methods. I replaced it with a regular method in a class, and it still fails. 

It looks like this line in toMethodUsage() is the problem:

instanceMethodParamTypes.remove(0); // remove the first one

I removed it, and it works, and all tests pass. 

But I assume this line was added for a reason?

Context:

```
                // instance methods are called on the first param and should match all other params
                List<MethodUsage> instanceMethodUsages = allMethods.stream()
                        .filter(it -> !it.getDeclaration().isStatic())

                List<ResolvedType> instanceMethodParamTypes = new ArrayList<>(paramTypes);
                instanceMethodParamTypes.remove(0); // remove the first one
```

jlerbsc:
> OK. This may not be a good fix. The question then is, why don't default methods have an instance parameter?

I do not understand your question. Can you be more specific or give an example?

jlerbsc:
> Turns out this is not limited to default methods. I replaced it with a regular method in a class, and it still fails.
> 
> It looks like this line in toMethodUsage() is the problem:
> 
> instanceMethodParamTypes.remove(0); // remove the first one
> 
> I removed it, and it works, and all tests pass.
> 
> But I assume this line was added for a reason?
> 
> Context:
> 
> ```
>                 // instance methods are called on the first param and should match all other params
>                 List<MethodUsage> instanceMethodUsages = allMethods.stream()
>                         .filter(it -> !it.getDeclaration().isStatic())
>                         .collect(Collectors.toList());
> 
>                 List<ResolvedType> instanceMethodParamTypes = new ArrayList<>(paramTypes);
>                 instanceMethodParamTypes.remove(0); // remove the first one
> ```

Can you create a new issue for this case and close the PR you proposed.

fabgo:
Filed: https://github.com/javaparser/javaparser/issues/4188

PR #4200, opened 2023-11-10
[pom] Switch from hamcrest-library (deprecated) to hamcrest

hazendaz:
The content of hamcrest-core and hamcrest-library are simply empty with comment to stop using.  See https://github.com/hamcrest/JavaHamcrest/blob/master/hamcrest-library/src/main/java/org/hamcrest/library/deprecated/HamcrestLibraryIsDeprecated.java

This saves having an extra library present in the build that isn't necessary.  No code changes required.


jlerbsc:

PR #4202, opened 2023-11-12
Indent with spaces rather than with tabs

mernst:
The JavaParser codebase mainly uses spaces for indentation.  However, in a few places it inconsistently uses tabs.  This pull request changes tabs to spaces in `.java` files.

### DIFF
--- a/javaparser-core-generators/src/main/java/com/github/javaparser/generator/CompilationUnitGenerator.java
+++ b/javaparser-core-generators/src/main/java/com/github/javaparser/generator/CompilationUnitGenerator.java
@@ -29,18 +29,18 @@ import java.util.List;
 
 public abstract class CompilationUnitGenerator extends Generator {
 
-	protected CompilationUnitGenerator(SourceRoot sourceRoot) {
-		super(sourceRoot);
-	}
+    protected CompilationUnitGenerator(SourceRoot sourceRoot) {
+        super(sourceRoot);
+    }
 
-	@Override
-	public void generate() throws Exception {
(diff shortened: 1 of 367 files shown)


hazendaz:
Hi @mernst, given this is so large of a change and impact it can have on existing pulls, maybe it can be approached in a separate way.

Here are my thoughts on it.
- First run com.github.hazendaz.maven:whitespace-maven-plugin:1.3.0:trim as that one is less intrusive (plugin I maintain) and I saw some trailing spacing in various files I've been in.  I think this is less likely to cause issues with open PRs.
- I would break this up into change per module only so its not so large but do those back to back until done.

Just my food for thought, and I agree tags in any files are bad.  Not very cross platform nice and its getting mixed by contributors is pretty frequent.  If you don't already you may want to make sure to have plugin do this automatically after done so it doesn't happen again. Spotless maven plugin is probably best candidate to cover the files in fast way.

hazendaz:
There are also some tricks, I've seen 'maven' team doing where they make git ignore the commits involved since they add little value and can be seen as a lot of noise.  Not sure that is worth the effort or concern as one can ignore whitespace and this is same line stuff but worth noting that is possible with git.

mernst:
I agree that use of Spotless is better than trying to manually enforce the style.  It can also "ratchet":  enforce formatting only for lines that have changed, so there are no irrelevant changes in the version control history.  I'll leave applying the `trim` plugin and/or Spotless up to you.

johannescoetzee:
This is no longer necessary since https://github.com/javaparser/javaparser/issues/4408 has been implemented. Thanks for the idea to use spotless for this!

PR #4225, opened 2023-11-25
Cleanup poms, use jakarta provided (javax namespace), hamcrest follow up, and switch coveralls plugin - Fixes #4111

### LINKED ISSUE #4111: Move to jakarta.json
For a few annoying reasons, https://mvnrepository.com/artifact/jakarta.json is the replacement for javax.json.

https://blog.ltgt.net/javax-jakarta-mess-and-gradle-solution/
https://blog.ltgt.net/javax-jakarta-mess-even-worse/

I honestly don't know the right move, but making this issue to track.

hazendaz:
Fixes #4111

- Jakarta was the original transition from javax but was still javax namespace.  Its still supported and thus should be used.  Do note if using a dependency bot to update that you will not want subsequent items yet depending on usage as the namespace changes after these revisions.
- Removed already defined versions as flagged by m2e in Eclipse
- Removed invalid setup on site plugin that has no influence on anything as not supported there (was for javadocs but not correct).  This was also flagged by m2e as invalid configuration.
- Cleanup junit 4 usage, it was already set by having junit vintage engine.  Cleaned up transient on hamcrest core.
- Switched coveralls plugin to a hard fork I'm not supporting (I support some half a dozen or more maven plugins including spotbugs).  New version no longer has bind-api concerns.  Added some links on the commit to follow out to the original repo(s) as its been difficult to get releases in years even though changes.  We started using this on MyBatis in last week or so and it works perfect while picking up a lot of newer changes.

### DIFF
--- a/javaparser-core-serialization/pom.xml
+++ b/javaparser-core-serialization/pom.xml
@@ -56,14 +56,14 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
-			<groupId>javax.json</groupId>
-			<artifactId>javax.json-api</artifactId>
-			<version>1.1.4</version>
+			<groupId>jakarta.json</groupId>
+			<artifactId>jakarta.json-api</artifactId>
+			<version>1.1.6</version>
 		</dependency>
 		<dependency>
 			<groupId>org.glassfish</groupId>
-			<artifactId>javax.json</artifactId>
-			<version>1.1.4</version>
+			<artifactId>jakarta.json</artifactId>
+			<version>1.1.6</version>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/javaparser-core-testing-bdd/pom.xml
+++ b/javaparser-core-testing-bdd/pom.xml
@@ -125,7 +125,6 @@
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest</artifactId>
-            <version>2.2</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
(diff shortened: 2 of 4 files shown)


bowbahdoe:
@hazendaz If someone has a newer version of jakarta.json, this will break them.

https://mvnrepository.com/artifact/jakarta.json/jakarta.json-api/2.1.3 exposes jakarta.json.

hazendaz:
@bowbahdoe Using both jakarta namespace and non jakarta namespace is a pretty much no go as it is so I don't think that is accurate.  I've yet to see any project using both with any meaningful success.  It's one or the other.  If this project is not EE centric maybe it should be using one of the half dozen other json libraries instead?  At minimum, I can revert the one out but the test one is fine.

hazendaz:
They could also simply exclude it and use the legacy ones that are not supported.  I've seen that done.

jlerbsc:

bowbahdoe:
@hazendaz Wait, what? That doesn't track.

If someone has the jakarta.json dependency and is using version 2.1.3 in their project, or a dependency of their project is, then that code has imports from `jakarta.json`. This old version has the imports as `javax.json`. 

If we did _nothing_ that would be a better situation, because at least then it wouldn't be interfering with other code/introducing a dependency conflict they have no way of resolving.

@jlerbsc I think this merge was too hasty.

jlerbsc:
> I think this merge was too hasty.

Very much so. I think I rushed.

hazendaz:
@bowbahdoe Lets assume maven, if they are using jakarta namespace, it would simply replace this one which is "true", then they would add javax.json and be done with it.  Its not complicated.  Do this all the time.  The same issue though is present in exact other way around now that this was moved entirely to jakarta namespace.  Now for those lagging behind (a significant amount per springs own research), they will need to do the same overrides.  It changed nothing really other than did the right thing here and just cut entirely to jakarta proper.

bowbahdoe:
Why did you put true in scare-quotes?

hazendaz:
Just calling out that what you noted is in fact true.

PR #4228, opened 2023-11-25
Move mockito to 4.11.0 and handle byte buddy consistently as well as properly define its agent in argLine for surefire

hazendaz:


pom.xml
```diff
@@ -409,6 +411,16 @@
                 <version>4.11.0</version>
                 <scope>test</scope>
             </dependency>
+            <dependency>
+                <groupId>net.bytebuddy</groupId>
```
jlerbsc, line 397: Why are you adding this dependency?

hazendaz:
It's used by assertj and mockito.  If not added it resolves mixed usage of the main byte buddy library and the agent which fails the build.  This ensures it does not regress from expectations.

Not sure if you use eclipse or not but it's m2e resolution easily shows this.  For testing you can look at the open pull requests that already fail.

The agent is added in this way as newer jdks will emit issues with it being auto loaded which will be blocked in the future.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

PR #4229, opened 2023-11-25
[reproducible] Rewire project to handle reproducibility - Fixes #3650

### LINKED ISSUE #3650: unreproducible MAVEN_BUILD_TIMESTAMP
While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that
our `javaparser` 3.24.2 package varied from an embeded `MAVEN_BUILD_TIMESTAMP` value - probably
from https://github.com/javaparser/javaparser/blob/48dfd6d/javaparser-core/src/main/java-templates/com/github/javaparser/JavaParserBuild.java#L12
from https://github.com/javaparser/javaparser/blob/48dfd6d/javaparser-core/pom.xml#L29

There should be some way to create 2 bit-identical builds.

hazendaz:
Fixes #3650

### DIFF
--- a/javaparser-core-serialization/pom.xml
+++ b/javaparser-core-serialization/pom.xml
@@ -25,7 +25,6 @@
 
 	<properties>
 		<java.version>1.8</java.version>
-		<build.timestamp>${maven.build.timestamp}</build.timestamp>
 	</properties>
 
 	<build>
--- a/javaparser-core/pom.xml
+++ b/javaparser-core/pom.xml
@@ -26,7 +26,6 @@
 
     <properties>
         <java.version>1.8</java.version>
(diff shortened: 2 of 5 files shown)


hazendaz:
note dates will be blank inside the jars on this change.  You can either set some reasonable value that is closure.  My sample I pulled from is a super pom I use and its last release date was set as '1699330556', so that would make them show as 11/7/2023.  There is no harm in starting with '1' though and letting the release factor that out but it may be useful to check that to understand better how this works.

jlerbsc:
Thank you for your suggestion but this is not the solution we are going to favor. Firstly, we will simply set the build date during release as recommended in the guide https://maven.apache.org/guides/mini/guide-reproducible-builds.html.

PR #4230, opened 2023-11-25
Stub out support for jdk 19, 20, 21, 22

hazendaz:
This builds off of #4227

Problem today is that we are using impsort-maven-plugin which uses javaparser.  We want to switch our runtimes to hard require java 21 as the runtime (not the build only).  In doing so the missing enums here blocks.  All of these are flagged as not implemented so not to give a false impression they actually solve anything at all on new language usage.  In our case we are not looking at new language yet but also don't want to disable impsort plugin as its import for our users.


hazendaz:
Further why does this matter?  I know what you are thinking, why raise the runtime to java 21 if you are not adding new features.  That is a valid ask, it has more to do with our toolchain jenkins automation that will walk builds up to java 21 runtimes and with fact that jakartaEE is forcing 21 and spring is even now suggesting it.  New coding items that break this tool will certainly happen soon enough but this gives us a window forwards for now until we run into those future issues.

jlerbsc:
This is not desirable because the specifics of these versions are not implemented.

hazendaz:
Hi @jlerbsc, what do you suggest then for those of us that want to target jdk 18 through 21?  Just prior to this I'm pretty sure jdk 18 was added in same way without any updates allowing it to work (was originally in my set too).  Yes new language changes would not work but no one using this will even know that until they are migrated over and find out separately.  The only option we have right now is to just disable usage of this project which I think I'm going to just move forwards with.  Its that or release a temporary fork which I'd like to avoid.

jlerbsc:
> I'm pretty sure jdk 18 was added in same way without any updates allowing it to work

This is not correct, the only feature that is not yet implemented to my knowledge is Pattern Matching for switch which is previewed in this version but which I will work on soon.

Furthermore, the project team has always followed this version deployment strategy. The only difference is that many members are no longer active on the project and there are new features on Java every 6 months.

PR #4237, opened 2023-11-27
Unit tests for class GenericListVisitorAdapter

4everTheOne:
Added unit tests for GenericListVisitorAdapter and added `override` annotation.


jlerbsc:
Thank you for this new contribution which improves JP's test coverage.

PR #4262, opened 2023-12-17
add Java_18 to yieldSupport

Kimmmey:
If you select Java 18, it is not possible to parse switch entries with yield statements.


jlerbsc:

PR #4276, opened 2023-12-26
feat: Add parseArrayInitializerExpr to JavaParser API

iMashtak:
I encountered a problem when did some work with java code generation: there was no API to parse array initializer expression into `Expression`, which is required for `AnnotationMemberDeclaration::setDefaultValue` method

So I simply added that API


jlerbsc:

Issue #2069, opened 2018-01-11
Unable to get types for NameExpr which refer to ClassNames.


fvanham:
I'm unable to get the types for java's common built-in classes, but I cannot figure out where my code is off. Using Symbolsolver 0.6.3 via maven, which pulls in javaparser 3.5.5
```java
public class MathTest {
	public static void main(String[] args) throws FileNotFoundException {
		// set up combined type solver
        CombinedTypeSolver ts = new CombinedTypeSolver();
        ts.add( new ReflectionTypeSolver() );
        ts.add( new JavaParserTypeSolver(new File("./sample") ) );
        // parse to CompilationUnit
        CompilationUnit cu = JavaParser.parse(new File("./sample/Test.java"));
        cu.accept(
            new VoidVisitorAdapter<Void>() {
            	// visit all named expressions
                public void visit( NameExpr n, Void v ) {
                	try
                	{
                		// attempt to get type
                		ResolvedType type = JavaParserFacade.get( ts ).getType( n );
                		System.out.println( n + ":" + type.describe() );
                	}
                	catch ( Exception e )
                	{
                 		e.printStackTrace();
                	}
                	super.visit( n, v );
                }
            },
        null );
    }
}
```
which parses this file content
```java
public class Test {
	public static void main( String[] args )
	{
		System.out.println( Math.round( Math.random() * Integer.parseInt( args[0] ) ) );
	}
}
```
The result is two failed type resolutions on `Math`, one on `System` and one on `Integer`. I'm a bit surprised the typesolver is null in the stacktrace.
```
UnsolvedSymbolException{context='Solving Math', name='Math', typeSolver=null}
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:290)
	at com.github.javaparser.symbolsolver.javaparsermodel.TypeExtractor.visit(TypeExtractor.java:46)
	at com.github.javaparser.ast.expr.NameExpr.accept(NameExpr.java:78)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:388)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:280)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:274)
	at MathTest$1.visit(MathTest.java:33)
[...]

PR #4295, opened 2024-01-19
Fix: issue #3251 Unable to get types for NameExpr which refer to ClassNames

### LINKED ISSUE #3251: Unable to get types for NameExpr which refer to ClassNames
Hello everyone,
Even if the following issue (#2069) seems to be closed, the problem is still here.

```
Math.abs(2);
String.format("%s://%s:%d", "asd", "host", 23);
```

"Math" and "String" cannot be solved.

Any idea?

fishautumn:
Fixes #3251 

on solveSymbol() failure, try solveType(); also change some API methods' return type from ResolvedValueDeclaration to ResolvedDeclaration.


jlerbsc:
You certainly didn't read ftomasseti's comment carefully. Please read his comment more carefully in https://github.com/javaparser/javaparser/issues/2069.
I'm closing this PR because I don't think this is the right way to proceed.

PR #4296, opened 2024-01-22
Fix: issue #3878 resolve MethodReference in ObjectCreationExpr

### LINKED ISSUE #3878: MethodReference calls is not yet fully resolvable
MethodReference call are only resolvable on method call expression.
see MethodReferenceExprContext

fishautumn:
Fixes #3878 resolve MethodReference in ObjectCreationExpr.

mostly copy the MethodCallExpr logic, extract shared code.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodReferenceExprContext.java
```diff
@@ -198,7 +189,40 @@ private List<ResolvedType> inferArgumentTypes() {
...
+
+                resolvedTypes.add(conType);
+            }
+
+            return resolvedTypes;
+        }
+        throw new UnsupportedOperationException();
+    }
+
+    private int pos(NodeWithArguments<?> callExpr, Expression param) {
```
jlerbsc, line 225: You can perhaps improve this method slightly by re-using NodeWithArguments.getArgumentPosition(Expression arg)

PR #4359, opened 2024-03-31
fix: issue 4358 prevent infinite cycles with static imports

### LINKED ISSUE #4358: StackOverflowError with import static in circular dependencies
Hi 👋 I encountered a `StackOverflowError` when trying to use javaparser on a popular open source library.

I managed to reproduce the issue in a minimal example.

> [!NOTE]
> I detected the problem javaparser `3.25.9`, but I was able to reproduce it on older versions.

The code I use for testing:

_Main.java_

``` java
package org.example;

import com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver;
import java.io.File;

public class Main {
    var rts = new ReflectionTypeSolver();
    var jpts = new JavaParserTypeSolver(new File("src/main/java"));
    var cts = new CombinedTypeSolver();
    cts.add(rts);
    cts.add(jpts);
    var pc = new ParserConfiguration()
        .setSymbolResolver(new JavaSymbolSolver(cts));
    StaticJavaParser.setConfiguration(pc);
    var cu = StaticJavaParser.parse(new File("src/main/java/org/example/A.java"));
    cu.findAll(MethodCallExpr.class).forEach(MethodCallExpr::resolve);
  }
}
```

_A.java_

``` java
package org.example;

import static org.example.B.d;

class A {
  void c() {
    d();
  }
}
```

_B.java_

``` java
package org.example;

class B extends A {
  static void d() {
  }
}
```

The issue doesn't happen when I use a fully qualified call (`B.d()`) and don't use `import static`.

Same thing happens with an asterisk import.

> [!NOTE]
> For the curious, the open source library where I discovered this problem is JUnit 5. Class `A` corresponds to [Filter](https://github.com/junit-team/junit5/blob/releases/5.9.x/junit-platform-engine/src/main/java/org/junit/platform/engine/Filter.java), class `B` to [CompositeFilter](https://github.com/junit-team/junit5/blob/releases/5.9.x/junit-platform-engine/src/main/java/org/junit/platform/engine/CompositeFilter.java), method `c` to `composeFilters` and method `d` to `alwaysIncluded`.

kdunee:
Fixes #4358 (infinite loop when `import static` is used between an ancestor and parent class).

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4358Test.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4358Test.java
@@ -1,0 +1,38 @@
+package com.github.javaparser.symbolsolver;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.resolution.declarations.ResolvedMethodDeclaration;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.JavaParserTypeSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
+import java.io.IOException;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+
+public class Issue4358Test extends AbstractSymbolResolutionTest {
+  @Test
+  public void testIssue4358() throws IOException {
+    Path issueResourcesPath = adaptPath("src/test/resources/issue4358");
+    ReflectionTypeSolver rts = new ReflectionTypeSolver();
+    JavaParserTypeSolver jpts = new JavaParserTypeSolver(issueResourcesPath);
+    CombinedTypeSolver cts = new CombinedTypeSolver();
+    cts.add(rts);
+    cts.add(jpts);
+    ParserConfiguration pc = new ParserConfiguration()
+        .setSymbolResolver(new JavaSymbolSolver(cts));
+    StaticJavaParser.setConfiguration(pc);
+    CompilationUnit cu = StaticJavaParser.parse(issueResourcesPath.resolve("foo/A.java"));
+
+    // There's only one method call in this compilation unit
+    ResolvedMethodDeclaration actual = cu.findAll(MethodCallExpr.class).stream()
+        .map(MethodCallExpr::resolve)
+        .findAny().get();
+
+    assertEquals("foo.B.d()", actual.getQualifiedSignature());
+  }
+}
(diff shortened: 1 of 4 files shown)


javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4358Test.java
```diff
@@ -0,0 +1,29 @@
...
+    ReflectionTypeSolver rts = new ReflectionTypeSolver();
+    JavaParserTypeSolver jpts = new JavaParserTypeSolver(issueResourcesPath);
+    CombinedTypeSolver cts = new CombinedTypeSolver();
+    cts.add(rts);
+    cts.add(jpts);
+    ParserConfiguration pc = new ParserConfiguration()
+        .setSymbolResolver(new JavaSymbolSolver(cts));
+    StaticJavaParser.setConfiguration(pc);
+    CompilationUnit cu = StaticJavaParser.parse(issueResourcesPath.resolve("foo/A.java"));
+    cu.findAll(MethodCallExpr.class).forEach(MethodCallExpr::resolve);
```
jlerbsc: Can't we check the result because there's only one method call?

jlerbsc: I'll accept this solution because it doesn't break the existing tests.

kdunee: @jlerbsc  The result is irrelevant, the test verifies that `StackOverflowError` isn't thrown during `resolve`. I considered adding a comment explaining it or explicitly using [assertDoesNotThrow](https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/Assertions.html#assertDoesNotThrow(org.junit.jupiter.api.function.Executable)) to make it more clear.

jlerbsc: Does this mean that the method is not correctly solved?

kdunee: It is correctly solved 😉. I could add the assertions to make it clear that:

- The method is correctly solved.
- `StackOverflowError` isn't thrown.

Should I do that?

jlerbsc: Checking that the method is correctly resolved is sufficient. If we want to find the context, we can refer to the issue.

kdunee: Sounds good to me.
Thanks for reviewing the change. I'll let you know when I implement the assertion, can't do it right now.

jlerbsc: As soon as it's done, I'll validate the PR. Thank you for your contribution.

kdunee: @jlerbsc Done. I also rebased my change onto current master.

Issue #4361, opened 2024-04-02
Question about switch statements and pattern representation


johannescoetzee:
Related to https://github.com/javaparser/javaparser/issues/4343 and https://github.com/javaparser/javaparser/issues/4217

Java 21 extends patterns in 2 ways that affect patterns directly (relevant syntax shown with some alternatives omitted for brevity):

### Patterns in switch statments can now include guard
```
SwitchLabel:
  case CaseConstant { , CaseConstant }
  case CasePattern [ Guard ]
  case null [ , default ]

CasePattern:
  Pattern

Guard:
  when Expression
```

### Record Patterns
```
Pattern:
  TypePattern
  RecordPattern

TypePattern:
  LocalVariableDeclaration

RecordPattern:
  ReferenceType ( [ PatternList ] )

PatternList : 
  Pattern { , Pattern }
```

### Current state

**Patterns:** JavaParser currently only supports `TypePatterns` in `instanceof` expressions and represents these with a `PatternExpr`. 

**Switch labels:** Switch labels are represented as a list of `Expressions` in the corresponding `SwitchEntry`

#### 1. Switch patterns

Since switch labels are currently represented as a single `Expression` per label, I think the best solution in this case is to add a
```java
class SwitchPattern {
	Pattern pattern; // Using Pattern here, since this could also be a record pattern
	Expression guard;
}
```
class to group these. The other option would be to add an `Optional<Expression> guard` to all patterns and then have it be empty
for all non-switch patterns.

Another syntax change to consider is the `case null [ , default ]` case. Currently, we can tell if a `SwitchEntry` is the default
case by checking if the label list is empty, but with the option of having a shared null/default case, this would no longer be 
true. The best solution I can think of for this is to add an `isDefault` field to `SwitchEntry` that will always be set in the 
empty label case (meaning you can use either the empty label list or `isDefault` to determine whether the entry is a default), 
and will also be set in the shared `null/default` case (in which case the empty label list check will return an incorrect result).

The other option I considered is to add some sort of `DefaultExpr` as a placeholder, but I think that causes more trouble than it's worth.

#### 2. Record patterns

This is the trickier change of the two. The ideal case would be to split `PatternExpr` into 2 separate classes: `TypePatternExpr` and 
`RecordPatternExpr`, both implementing a `PatternExpr` interface/base class for generics. This would be a rather large change for 
anyone currently using `PatternExpr`, so an alternative could be to keep `PatternExpr` as the representation for `TypePattern` and to 
[...]

PR #4364, opened 2024-04-03
Add support for `case null, default` in switch and fix concrete syntax model for new switch syntax

johannescoetzee:
Related issue: https://github.com/javaparser/javaparser/issues/4361

Note: This does not address any of the pattern matching-related changes. I decided to start small to build familiarity with the codebase. 

This PR mainly addresses my comment regarding the new `case null, default` possibility for switch labels by introducing an `isDefault` field for the switch entry, setting this where appropriate and using this value  to distinguish cases for PP and LPP.

While testing the LPP, I realized that the new `->` syntax for switch expressions wasn't supported in the `ConcreteSyntaxModel`, so I've updated that to work for what's included here (which excludes the patterns case).

I see Intellij also made some formatting changes, but I use the style config from this project, so the changes should be correct.

For the code generation part, I added the `isDefault` field, ran the code generators, then fixed the constructors by hand.


javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -160,7 +160,7 @@ private static CsmElement typeArguments() {
...
+        concreteSyntaxModelByClass.put(SwitchEntry.class, sequence(comment(), conditional(ObservableProperty.TYPE, IS_STMT_GROUP_ENTRY, conditional(ObservableProperty.LABELS, IS_NOT_EMPTY, sequence(token(GeneratedJavaParserConstants.CASE), space(), list(ObservableProperty.LABELS), token(GeneratedJavaParserConstants.COLON)), sequence(token(GeneratedJavaParserConstants._DEFAULT), token(GeneratedJavaParserConstants.COLON))), conditional(ObservableProperty.LABELS, IS_NOT_EMPTY, conditional(ObservableProperty.DEFAULT, FLAG, sequence(token(GeneratedJavaParserConstants.CASE), space(), list(ObservableProperty.LABELS), comma(), space(), token(GeneratedJavaParserConstants._DEFAULT), space(), token(GeneratedJavaParserConstants.ARROW)), sequence(token(GeneratedJavaParserConstants.CASE), space(), list(ObservableProperty.LABELS), token(GeneratedJavaParserConstants.ARROW))), sequence(token(GeneratedJavaParserConstants._DEFAULT), space(), token(GeneratedJavaParserConstants.ARROW)))), newline(), indent(), list(ObservableProperty.STATEMENTS, newline(), none(), newline()), unindent()));
```
johannescoetzee: The formatting in this file is a mess, so it's easier to see what's going on with sensible formatting:
### Old
```
concreteSyntaxModelByClass.put(
    SwitchEntry.class,
    sequence(
	comment(),
	conditional(
	    ObservableProperty.LABELS,
	    IS_NOT_EMPTY,
	    sequence(
		token(GeneratedJavaParserConstants.CASE),
		space(),
		list(ObservableProperty.LABELS),
		token(GeneratedJavaParserConstants.COLON)
	    ),
	    sequence(
		    token(GeneratedJavaParserConstants._DEFAULT),
		    token(GeneratedJavaParserConstants.COLON)
	    )
	),
	newline(),
	indent(),
	list(ObservableProperty.STATEMENTS, newline(), none(), newline()), unindent()
    )
);
```

### New
```
concreteSyntaxModelByClass.put(
    SwitchEntry.class,
    sequence(
	comment(),
	conditional(
	    ObservableProperty.TYPE,
	    IS_STMT_GROUP_ENTRY,
	    conditional(
		ObservableProperty.LABELS,
		IS_NOT_EMPTY,
		sequence(token(GeneratedJavaParserConstants.CASE), space(), list(ObservableProperty.LABELS), token(GeneratedJavaParserConstants.COLON)),
		sequence(token(GeneratedJavaParserConstants._DEFAULT), token(GeneratedJavaParserConstants.COLON))
	    ),
	    conditional(
		ObservableProperty.LABELS,
		IS_NOT_EMPTY,
		conditional(
		    ObservableProperty.DEFAULT,
		    FLAG,
		    sequence(
			token(GeneratedJavaParserConstants.CASE),
			space(),
			list(ObservableProperty.LABELS),
			comma(),
			space(),
			token(GeneratedJavaParserConstants._DEFAULT),
			space(),
			token(GeneratedJavaParserConstants.ARROW)
		    ),
		    sequence(
			token(GeneratedJavaParserConstants.CASE),
			space(),
			list(ObservableProperty.LABELS),
			token(GeneratedJavaParserConstants.ARROW)
		    )
		),
		sequence(token(GeneratedJavaParserConstants._DEFAULT), space(), token(GeneratedJavaParserConstants.ARROW))
	    )
	),
	newline(),
	indent(),
	list(ObservableProperty.STATEMENTS, newline(), none(), newline()), unindent()
    )
);
```

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
```diff
@@ -93,6 +94,17 @@ boolean evaluate(Node node, ObservableProperty property) {
                 return property.getValueAsBooleanAttribute(node);
             }
         }
+        , IS_STMT_GROUP_ENTRY {
```
johannescoetzee: There may be a more re-usable way to add this, but I couldn't think of anything low-effort enough to justify the effort. Suggestions are very welcome though :)

johannescoetzee:
I see there's a single test failing with
```
Codecov: Failed to write uploader binary: EPERM: operation not permitted, open 'D:\a\_actions\codecov\codecov-action\v4.1.1\dist\codecov.exe'
```
which looks unrelated to this change, but I don' have permission to rerun the action.

javaparser-core/src/main/java/com/github/javaparser/metamodel/SwitchEntryMetaModel.java
```diff
@@ -42,6 +42,8 @@ public class SwitchEntryMetaModel extends NodeMetaModel {
         super(superBaseNodeMetaModel, SwitchEntry.class, "SwitchEntry", "com.github.javaparser.ast.stmt", false, false);
     }
 
+    public PropertyMetaModel isDefaultPropertyMetaModel;
```
jlerbsc, line 44: Did you see this comment "This file, class, and its contents are completely generated based"?

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
```diff
@@ -771,6 +771,8 @@ private static void initializePropertyMetaModels() {
         localRecordDeclarationStmtMetaModel.getDeclaredPropertyMetaModels().add(localRecordDeclarationStmtMetaModel.recordDeclarationPropertyMetaModel);
         returnStmtMetaModel.expressionPropertyMetaModel = new PropertyMetaModel(returnStmtMetaModel, "expression", com.github.javaparser.ast.expr.Expression.class, Optional.of(expressionMetaModel), true, false, false, false);
         returnStmtMetaModel.getDeclaredPropertyMetaModels().add(returnStmtMetaModel.expressionPropertyMetaModel);
+        switchEntryMetaModel.isDefaultPropertyMetaModel = new PropertyMetaModel(switchEntryMetaModel, "isDefault", boolean.class, Optional.empty(), false, false, false, false);
```
jlerbsc, line 774: Did you see this comment "Warning: The content of this class is partially or completely generated - manual edits risk being overwritten"?

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/SwitchEntry.java
```diff
@@ -98,23 +97,24 @@ public enum Type {
...
 
     public SwitchEntry() {
-        this(null, new NodeList<Expression>(), Type.STATEMENT_GROUP, new NodeList<>());
+        this(null, new NodeList<Expression>(), Type.STATEMENT_GROUP, new NodeList<>(), false);
     }
 
     @AllFieldsConstructor
-    public SwitchEntry(final NodeList<Expression> labels, final Type type, final NodeList<Statement> statements) {
-        this(null, labels, type, statements);
+    public SwitchEntry(final NodeList<Expression> labels, final Type type, final NodeList<Statement> statements, final boolean isDefault) {
```
jlerbsc, line 113: I think this is generated by the generators. This is a rather obscure part of JP for me as I've never used it and it's not well documented. As far as I know, the link to the documentation, which can be accessed via the wiki, is https://javaparser.org/code-generation-and-maven-in-javaparser/

javaparser-core/src/main/java/com/github/javaparser/ast/observer/ObservableProperty.java
```diff
@@ -50,6 +49,7 @@ public enum ObservableProperty {
     COMPONENT_TYPE(Type.SINGLE_REFERENCE),
     CONDITION(Type.SINGLE_REFERENCE),
     CONTENT(Type.SINGLE_ATTRIBUTE),
+    DEFAULT(Type.SINGLE_ATTRIBUTE),
```
jlerbsc, line 52: Why have you created this new entry?

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrettyPrinterVisitor.java
```diff
@@ -1243,6 +1243,10 @@ public void visit(final SwitchEntry n, final Void arg) {
                     printer.print(", ");
                 }
             }
+            // `case null, default -> ...` added in JEP 441
+            if (n.getLabels().isNonEmpty() && n.isDefault()) {
+                printer.print(", default");
```
jlerbsc, line 1248: Are you sure that the "default" case must be preceded by a comma?

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1245,6 +1245,10 @@ public void visit(final SwitchEntry n, final Void arg) {
                     printer.print(", ");
                 }
             }
+            // `case null, default -> ...` added in JEP 441
+            if (n.getLabels().isNonEmpty() && n.isDefault()) {
+                printer.print(", default");
```
jlerbsc, line 1250: Are you sure that the "default" case must be preceded by a comma?

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/TokenKindGenerator.java
```diff
@@ -80,7 +80,7 @@ public void generate() {
     }
 
     private void generateValueOfEntry(SwitchStmt valueOfSwitch, String name, IntegerLiteralExpr kind) {
-        final SwitchEntry entry = new SwitchEntry(new NodeList<>(kind), SwitchEntry.Type.STATEMENT_GROUP, new NodeList<>(new ReturnStmt(name)));
+        final SwitchEntry entry = new SwitchEntry(new NodeList<>(kind), SwitchEntry.Type.STATEMENT_GROUP, new NodeList<>(new ReturnStmt(name)), false);
```
jlerbsc, line 83: Can you use the constructor without the 'default' argument to limit the impact on the code?

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
jlerbsc: Can you explain in more detail the purpose of this change (why) and the different cases you want to handle (how)? I understand that it's necessary to modify the CSM but it's still another rather obscure part as it has a big impact on the way the LPP handles differences.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
jlerbsc: Why add this?

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
```diff
@@ -93,6 +94,17 @@ boolean evaluate(Node node, ObservableProperty property) {
...
+            @Override
```
jlerbsc: Can you explain the purpose of this change in more detail?

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java
```diff
@@ -41,6 +42,10 @@ default boolean evaluate(CsmConditional csmConditional, Node node) {
             case IS_PRESENT:
                 return !Utils.valueIsNullOrEmptyStringOrOptional(getValue(csmConditional.getProperty(), node))
                 		&& !isEvaluatedOnDerivedProperty(csmConditional.getProperty());
+	        case IS_STMT_GROUP_ENTRY:
```
jlerbsc: I don't understand this but it must be linked to other changes?

javaparser-core/src/main/javacc/java.jj
```diff
@@ -4625,11 +4624,10 @@ SwitchEntry SwitchEntry():
      |
         "->"
         (
-            expression = Expression()
-            ";"
+            stmt = StatementExpression()
```
jlerbsc, line 4638: I don't understand this change in relation to the purpose of the PR unless it's an improvement on the current version.

jlerbsc:
It's a great job. Thank you very much. Your approach is exactly what I'm looking for. Let's take things one step at a time, because I don't necessarily have the time or the skills to analyse very large PR files. I also prefer it when you first explain what you want to do and what the alternatives are (which you've done perfectly on the other related subjects) because that gives me a slightly clearer idea of the content of the PR.

I'll let you answer my questions before I integrate your work. However, I think that the validation part of the new syntax is missing (validator because it should only be authorised from a certain version of Java). It also seems to me that tests on the LPP are missing.

Just so you know, before integrating this PR, I'm going to publish a new version of JP.

javaparser-core/src/main/java/com/github/javaparser/metamodel/SwitchEntryMetaModel.java
johannescoetzee: This change is generated code

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
johannescoetzee: This change is generated code

javaparser-core/src/main/java/com/github/javaparser/ast/observer/ObservableProperty.java
johannescoetzee: It was generated automatically

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/SwitchEntry.java
johannescoetzee: I added the `isDefault` field and then ran the generators as per the instructions in that page, but this didn't update the 3 constructors in this file. Everything else was correctly generated however.

javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrettyPrinterVisitor.java
johannescoetzee: In this particular case, yes. This only handles the case where `n.getLabels()` is non-empty and the case is the `default` , which would have been illegal code before java 21, but is now allowed in the special case:
```
SwitchLabel:
  case null [, default]
```

From https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-SwitchExpression.

The code as I wrote it doesn't specifically check that `n.getLabels()` only returns the single null expression, but, as I understand it, we assume code parsed by JavaParser compiles, so the check for the null expression is implicit. This could potentially allow users to generate invalid code with manual AST changes, but doing so would require explicit changes to a switch entry, so I feel like in that case it's also not unreasonable to expect the user to not create this rather niche problem.

javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
johannescoetzee: Same comment as above (this is the deprecated printer, but I figured I'd update it here too).

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/TokenKindGenerator.java
johannescoetzee: There isn't a constructor without a `default` argument anymore, but I could add that if you prefer.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
johannescoetzee: This line combines 2 changes: 
* support for switch expression syntax (`case X ->`). This is a fix for an existing bug and would be required even without the other changes in this PR
* support for `case null, default ->` added in Java 21 which is new.

The first point is addressed by the first conditional: 
```
conditional(
    ObservableProperty.TYPE,
    IS_STMT_GROUP_ENTRY,
    ..., // If type == STATEMENT_GROUP, we're in the classic "case X: ..." switch statement
    ..., // Otherwise we're in the new "case X -> ..." switch expression
)
```

`IS_STMT_GROUP_ENTRY` is the new condition I added to do this check. From the [SwitchEntry docs](https://javadoc.io/static/com.github.javaparser/javaparser-core/3.25.9/com/github/javaparser/ast/stmt/SwitchEntry.html), we use `SwitchEntry.type` to distinguish between these cases. The type `STATEMENT_GROUP` corresponds to the classic switch statement, where the other 3 types all correspond to the new switch expression. There may be a simpler way to define the concrete syntax for these alternatives, but I don't understand these components well enough to be confident making big changes. Instead, I decided to stick to what already exists as much as possible.

In the classic switch statement case, nothing has changed, so that's just a copy-paste job from what was already there. In the switch expression case, however, there's my second point above to deal with.

The second conditional handles the regular `default` case (with no labels) vs cases with labels:
```
conditional(
    ObservableProperty.LABELS,
    IS_NOT_EMPTY,
    ..., // If labels is not empty, then we're either in the regular "case X, Y -> ..." case, or the new "case null, default ->" - see below for discussion
    ..., // If labels is empty, we're in the regular "default -> ..." case, so handle as before, but with "->" instead of ":"
)
````

The final conditional is then used to handle the `case null, default -> ...` possibility:
```
conditional(
    ObservableProperty.DEFAULT,
    FLAG,
    ..., // If isDefault is true, expect ", default" after labels
    ..., // Otherwise handle as with switch statement, but "->" instead of ":"
)
```

Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java]: The many nested conditionals are quite complex, but the benefit of checking all these cases explicitly is that there's a fairly direct mapping between these alternatives and what's described in the syntax. It might be possible to simplify and flatten this somewhat, but I don't understand the CSM and various supporting constructs well enough to do this confidently and this will need to be updated further for switch pattern support, so simplifying now might be premature.

File: javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java (hunk 2 above)
Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java]: To answer this and the below comment:

This is the condition I used to distinguish the `case x: ...` and `case x -> ...` options for switch statements and expressions respectively (please see my comment above for a better discussion of how it's used).

To explain what it's doing line-by-line:
```
 if (node instanceof SwitchEntry && property == ObservableProperty.TYPE) {
```
This check only makes sense to do for the `TYPE` property of `SwitchEntry` nodes, so check for that here.

```
SwitchEntry.Type entryType = (SwitchEntry.Type)(property.getRawValue(node));
```
Get the `type` from the `SwitchEntry` here. The `getRawValue` call and the cast are both safe because of the check above. We could also just do something like `((SwitchEntry) node).getType()` here, but doing it via the property is more consistent with how the other conditions work.

```
return entryType == SwitchEntry.Type.STATEMENT_GROUP;
```
This check is effectively the whole purpose of this condition. If the entry type is ` STATEMENT_GROUP`, then the entry is a classic switch statement entry `case X: ...`

Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java]: My comment about a more re-usable way to do this was about this condition being very specifically written for this specific check. It would be much nicer to have a general "does the given property on the given node have this value" sort of check, but couldn't think of a good way to do this. Maybe an anonymous class overriding some `expectedValue` could work, but I didn't want to spend too much time engineering a general solution when there's only one use case at this point.

File: javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java (hunk 10 above)
Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java]: Please see above.

File: javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java (hunk 11 above)
Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java]: This is linked to the `Condition` I added above. I don't understand why this default `evaluate` method is necessary and doesn't just defer to the condition's `evaluate`, but adding handling for the new `Condition` is necessary and the logic here is basically duplicated from above.

File: javaparser-core/src/main/javacc/java.jj (hunk 12 above)
Comment (johannescoetzee) [reply javaparser-core/src/main/javacc/java.jj:4638]: This is related to the CSM fix. I'll try to explain with the example `case X -> foo();`

With how it was written before, we would generate 2 tokens for the body of the entry (ignoring whitespace):
```
TOKEN("foo();")
TOKEN(";")
```
Because the "foo()" call was handled as an expression, it didn't consume the ";", which was then handled separately. The issue (as far as I can tell) was that we were then treating it as an `ExpressionStmt`, which should include the ";" token. This means we were effectively duplicating the ";" token, so when the `Difference` is applied in the LPP, the extra ";" was treated as an unexpected, left-over non-whitespace token which results in a crash.

By treating the body as a statement expression from the start, it takes the ";" into account and avoids this duplicate issue.

File: javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java (hunk 11 above)
Comment (johannescoetzee) [reply javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java]: I also just noticed that the indentation here seems wrong, so I'll fix that :)

File: javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java [hunk 13]
```diff
@@ -56,4 +56,31 @@ void switchEntryCsmHasTrailingUnindent() {
...
+
+    @Test
+    void nullDefaultSwitchEntryPreserved() {
+        String code = "switch (a) { case null, default -> System.out.println(-1); }";
+        Statement stmt = consider(code);
+        NodeList<Statement> statements = stmt.asSwitchStmt().getEntry(0).getStatements();
+        statements.set(0, statements.get(0).clone());
+        assertTransformedToString(code, stmt);
+    }
+
```
johannescoetzee, line 99: These tests check that switch expressions can at least be reproduced correctly (in line with some other tests in the file). If you'd like me to add more involved tests, just say and I'd be happy to do so :)

johannescoetzee:
Thank you for the quick and thorough review! 

Regarding LPP tests, I did add some simple ones (I added a comment to those so that you can see them in the conversation history). If you'd like me to add more involved ones, please let me know and I'll be happy to do so.

I didn't add the validation part to avoid conflicts with https://github.com/javaparser/javaparser/pull/4357, but if this is going to go in first, I'll add that here and then we can deal with conflicts later.

Regarding a new version, that's no problem at all! I don't know what you have in mind for the JavaParser release schedule, but in my opinion releasing all of the non-preview java 21 features at once would be ideal!

jlerbsc:
> I didn't add the validation part to avoid conflicts with #4357, but if this is going to go in first, I'll add that here and then we can deal with conflicts later.

I'm not going to validate the PR #4357 because it contains too many unrelated changes and I'm therefore unable to validate its content. There are also points that need to be reworked. It's a big job but I don't want to risk destabilising the project.

jlerbsc:
> Regarding a new version, that's no problem at all! I don't know what you have in mind for the JavaParser release schedule, but in my opinion releasing all of the non-preview java 21 features at once would be ideal!

I agree with this.

johannescoetzee:
> I'm not going to validate the PR https://github.com/javaparser/javaparser/pull/4357

Good to know. I'll update this PR with the validation tomorrow then.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
jlerbsc: > * support for switch expression syntax (`case X ->`). This is a fix for an existing bug and would be required even without the other changes in this PR

For greater clarity, it would have been preferable to isolate this case in a PR. Maybe for a future fix ;-)

jlerbsc: Can you add your comments on the CSM in the code for other users to understand.

jlerbsc: Can't the FLAG condition solve the problem rather than creating a new, very specific condition (IS_STMT_GROUP_ENTRY)? For example, this condition is used to distinguish between classes and interfaces in the ClassOrInterfaceDeclaration.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
jlerbsc: > It would be much nicer to have a general "does the given property on the given node have this value" sort of check

It seems to me that this is what the FLAG condition does.

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/SwitchEntry.java
jlerbsc: > but this didn't update the 3 constructors in this file

I think this is a logical behaviour as they are certainly not annotated and they have probably been created by providing shortcuts with default values.

javaparser-core/src/main/java/com/github/javaparser/ast/observer/ObservableProperty.java
jlerbsc: This is something I'm discovering with your comments. Thank you.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
johannescoetzee: The FLAG condition only works for boolean properties and just checks whether the relevant property on the given node is `true`. It is (to some degree at least) another special case of the "does the given property on the given node have this value" condition I mentioned below where in this case the specific value is `true`, where for `IS_STMT_GROUP_ENTRY` the value is `STATEMENT_GROUP` (might be wrong on the name).

An alternative to creating this new condition would be to add a `SwitchEntry.isSwitchStmtEntry` or something like that, which we could then check with the FLAG condition. I chose not to do this since that's effectively duplicating information that's already easily available in most cases, but can change this if doing that is preferred.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
johannescoetzee: I think I answered that in my last comment above, but please let me know if anything is still unclear.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
jlerbsc: These tests actually involve the LexicalPreservingPrinter via the consider(..) method.

javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/changes/Change.java
jlerbsc: I don't think this should be necessary if it's possible to use the FLAG condition.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
jlerbsc: What would be the changes in this case? I can only think of creating a method that would certainly check the nature of the type (for example isStmtGroupEntry() or something similar). If this is the case, it seems to me that this is the best method for managing the condition. Even if you have to add an internal attribute, this seems to me to be more in line with the spirit of what has been initiated.

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/TokenKindGenerator.java
jlerbsc: Yes, because it seems to me that there is an update to the code that deals with this point. So it would avoid this change, and it would possibly have an impact on other projects based on Javaparser.

javaparser-core/src/main/javacc/java.jj
```diff
@@ -4597,22 +4598,30 @@ SwitchEntry SwitchEntry():
...
 }
 {
     (
         "case" { begin=token(); }
-        label = ConditionalExpression() { labels = add(labels, label); }
         (
-            ","
-            label = ConditionalExpression()  { labels = add(labels, label); }
-        )*
+            LOOKAHEAD(3)
```
jlerbsc: Would it be possible to organise the configuration of the parser according to the JSL? This hasn't necessarily been the case previously, but I think it would be easier to read. This can be done later if you wish.
https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-14.11.1

johannescoetzee: I had it organised that way before, but ran into an issue where the wrong alternative was chosen. This was at the start of me working on this though, so I was still very much figuring things out. I'll rearrange them again and see if I run into the same issue and if it can be fixed with a lookahead (which I may have placed in the wrong position), because I agree that keeping it consistent with the JLS would be nicer

johannescoetzee:
And just a quick general update: another task came up that took up my time today, hence no update here. I'll be working on this PR again on Tuesday, however.

And thanks again for the quick feedback and open discussion!

jlerbsc:
> And just a quick general update: another task came up that took up my time today, hence no update here. I'll be working on this PR again on Tuesday, however.
> 
> And thanks again for the quick feedback and open discussion!

There's no hurry, I've got other things to do too.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
johannescoetzee: I've updated the PR with this suggestion. I assumed that only fields could be `ObservableProperties`, but then discovered that I can just create a method as you suggested with the `@DerivedProperty` annotation which ensures everything is generated correctly.

javaparser-core/src/main/java/com/github/javaparser/printer/concretesyntaxmodel/CsmConditional.java
johannescoetzee: Resolving this conversation since the latest version replaced this custom Condition with a FLAG

javaparser-core-generators/src/main/java/com/github/javaparser/generator/core/other/TokenKindGenerator.java
johannescoetzee: I added the constructor without the `isDefault` parameter, but left this change in. In my opinion, it's better to be explicit than to have to look at the constructor body to see what's being set, but I can also revert this line if you'd prefer.

javaparser-core/src/main/javacc/java.jj
johannescoetzee: I've fixed this in my latest push as well. It did turn out to be a lookahead issue. Specifically looking ahead a fixed number of tokens worked with the original order in this PR since `case null, default` produced a constant number of tokens, while `ConditionalExpression` did not. I've updated the lookahead to look for `<ignore> "," "default"` instead.

johannescoetzee:
I've updated the PR with changes that address the feedback given and I've mostly mentioned specific changes in the conversation threads where they're relevant.

There are a few extra additions from code generation, however (specifically the `*Visitor` changes). The codegen scripts output is formatted differently from the checked-in source and since the project formatting is inconsistent, I couldn't fix these with automated formatting. After running the generators, I went through the diff manually to separate formatting changes from actual generated code and missed these in the first iteration.

If there's a better way to do this, please let me know! The eventual fix would be to automate formatting so that the project as a whole is consistent, but that's a bigger effort.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java14ValidatorTest.java
```diff
@@ -72,7 +89,7 @@ class RecordDeclarationForbidden {
             void recordDeclaration() {
                 String s = "record X() { }";
                 ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT, provider(s));
-                TestUtils.assertProblems(result, "(line 1,col 1) Record Declarations are not supported. Pay attention that this feature is supported starting from 'JAVA_14' language level. If you need that feature the language level must be configured in the configuration before parsing the source files.");
```
johannescoetzee, line 75: This was Intellij's auto formatting. I could change it back, but the change is more consistent with how other tests do it.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -63,6 +65,17 @@ private Optional<ConfigurationOption> getOption(PrinterConfiguration config, Con
...
+    @BeforeAll
+    public static void setLanguageLevel() {
+        StaticJavaParser.getParserConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.BLEEDING_EDGE);
+    }
+
+    @AfterAll
+    public static void resetLanguageLevel() {
+        StaticJavaParser.getParserConfiguration().setLanguageLevel(storedLanguageLevel);
+    }
+
```
johannescoetzee: There's probably a better way to do this, but I didn't want to change any defaults outside of the scope of the tests

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java21ValidatorTest.java
```diff
@@ -0,0 +1,45 @@
...
+package com.github.javaparser.ast.validator;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseResult;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.ast.expr.Expression;
+import org.junit.jupiter.api.Test;
+
+import static com.github.javaparser.ParseStart.EXPRESSION;
+import static com.github.javaparser.ParserConfiguration.LanguageLevel.JAVA_21_INCOMPLETE;
```
jlerbsc: Can you rename this to JAVA_21.

javaparser-core/src/main/java/com/github/javaparser/ast/stmt/SwitchEntry.java
```diff
@@ -118,6 +127,19 @@ public SwitchEntry(TokenRange tokenRange, NodeList<Expression> labels, Type type
...
 
+    /**
+     * This is required for the ConcreteSyntaxModel, specifically to determine whether this
+     * entry uses the classic switch statement syntax (e.g. `case X: ...`) or the newer
+     * switch expression syntax (`case X -> ...`).
+     *
+     * The entry type is STATEMENT_GROUP in the switch statement case and all other values
+     * are for the various switch expressions.
+     */
+    @DerivedProperty
```
jlerbsc, line 138: What is your understanding of this annotation?

javaparser-core/src/main/javacc/java.jj
```diff
@@ -4603,16 +4603,16 @@ SwitchEntry SwitchEntry():
     (
         "case" { begin=token(); }
         (
-            LOOKAHEAD(3)
-			label = NullLiteral() { labels = add(labels, label); }
-			","
-			"default" { isDefault = true; }
-         |
+			LOOKAHEAD({!(getToken(2).image == "," && getToken(3).image == "default")})
```
jlerbsc: Can you explain this new rule?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java21ValidatorTest.java
johannescoetzee: Done!

javaparser-core/src/main/javacc/java.jj
johannescoetzee: I added an explanation as a comment in code as well, but to repeat here for convenience:

>  Since null is also a conditional expression, a lookahead is necessary to determine
>  whether this is the "null, default" case, or whether the null is a regular
>  conditional expression. The check is negated to maintain the ordering in the JLS.
> 
>  In the "null, default" case, the next 3 tokens will be "null" "," "default". Since
>  "default" can only be preceded by "," in this specific case (as "default" cannot
>  appear in a list elsewhere), it is sufficient to check that "," "default" appear
>  in the correct order. Since the only label that can appear before the "default"
>  is the single-token "null", the "," token must also be the second token.

jlerbsc:
Do you have anything else to add? If not, I'll merge your PR.

johannescoetzee:
If you're happy with the changes here, then I think this PR is done. There are things like the validators, lookahead logic for the different cases, and the CSM that I'll have to change again when adding switch pattern support, but I think it's better to do that when actually adding the relevant features.

jlerbsc:
Thank you for this development.

PR #4371, opened 2024-04-12
Added methods for including children at a given index.

### LINKED ISSUE #4370: Children may be out of order after adding a statement
After adding a statement to a node `n` with with `n.addStatement` method. The children of `n` in `getChildNodes` may be out-of-order with relation to the order of statements.

Consider the following snippet.

```java
   String code =
        """
    class A{
	    void f(){
		    int b = 2;
		    int c = 3;
	    }
    }
    """;
    CompilationUnit cu = StaticJavaParser.parse(code);

    var stmt = cu.findAll(Statement.class).get(1);
    var newStmt = StaticJavaParser.parseStatement("int a = 1;");
    var blockStmt = (BlockStmt) stmt.getParentNode().get();

    System.out.println(cu);
    System.out.println("Children: " + blockStmt.getChildNodes());
    System.out.println("Statements: " + blockStmt.getStatements());

    blockStmt.addStatement(0, newStmt);

    System.out.println(cu);
    System.out.println("Children: " + blockStmt.getChildNodes());
    System.out.println("Statements: " + blockStmt.getStatements());
```
Will result in:
```
    class A {

        void f() {
            int b = 2;
            int c = 3;
        }
    }

    Children: [int b = 2;, int c = 3;]
    Statements: [int b = 2;, int c = 3;]
    class A {

        void f() {
            int a = 1;
            int b = 2;
            int c = 3;
        }
    }

    Children: [int b = 2;, int c = 3;, int a = 1;]
    Statements: [int a = 1;, int b = 2;, int c = 3;]
```
Note that the position of the node `int a = 1;` will differ from `getStatements` and `getChildNodes`, with the children list being with the wrong order.

andrecsilva:
This PR adds some new methods to include children at a given index. It also modifies `addStatements(index, node)` to use those methods.

Fixes #4370.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/stmt/BlockStmtTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/stmt/BlockStmtTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2011, 2013-2024 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


jlerbsc:
Thank you for your suggestion, but this is not the expected behaviour. The children of a node are not necessarily of the same type, as is the case when you consult the list of statements in a block. For example, when you consult the list of children of a class declaration, there may be a name and one or more method declarations. It wouldn't make sense to insert a new method before or after the class name. You can consider that the list of children is not ordered.

Issue #4385, opened 2024-04-22
Record Patterns Implementation Proposal


johannescoetzee:
This is related to https://github.com/javaparser/javaparser/issues/4361, but is a more specific proposal related only to RecordPatterns.

As mentioned in https://github.com/javaparser/javaparser/issues/4217, I would like to add support for Record Patterns introduced in https://openjdk.org/jeps/440 to javaparser. 

### Old Syntax
```
Pattern:
    TypePattern

TypePattern:
    LocalVariableDeclaration
```

### New Syntax
```
Pattern:
    TypePattern
    RecordPattern

TypePattern:
    LocalVariableDeclaration

RecordPattern:
    ReferenceType ( [PatternList] )

PatternList:
    Pattern {, Pattern }
```

## Implementation Proposal

In JP, `PatternExpr` is currently used to represent a `TypePattern`. I propose we split this up into 2 classes: `TypePatternExpr` representing a `TypePattern` and `RecordPatternExpr` representing a  `RecordPattern`. These 2 `XPatternExpr` classes will share a common `Pattern` parent class, so that we end up with something like the following:

```
abstract class Pattern extends Expression {
    // TODO: Do the different classes actually share anything worth sharing in the base class?
}

class TypePatternExpr extends Pattern {
    private NodeList<Modifier> modifiers;

    private SimpleName name;

    private ReferenceType type;
}

class RecordPatternExpr extends Pattern {
    private NodeList<Modifier> modifiers;

    private ReferenceType type;

    private NodeList<Pattern> patterns;
}
```

### Base Class and PatternExpr Considerations
There's a choice to be made about what to use what to do with the existing `PatternExpr`. 

#### Option 1: Use PatternExpr as the new base class

In the examples above, that would mean replacing every instance of `Pattern` with `PatternExpr`.
This approach comes with 2 notable advantages:
1. It's the "cleanest" representation of patterns (in my opinion, at least) and is similar to what JP is already doing with AnnotationExpr.
2.  We don't need to change parameters and fields where `PatternExpr` is expected.

The downside to this option is that we would be changing the behaviour of `PatternExpr` without changing the name. This
could be confusing for users doing JP upgrades who might not understand why `PatternExpr` suddenly doensn't work the 
same anymore.

#### Option 2: Introduce a new class to use as the base class and have PatternExpr extend that as well

In the examples above, this new base class would be `Pattern` and we'd change PatternExpr to be:
```
class PatternExpr extends Pattern {
...
}
```

[...]

PR #4387, opened 2024-04-24
PatternExpr -> TypePatternExpr refactor in preparation for record pattern implementation

johannescoetzee:
See https://github.com/javaparser/javaparser/issues/4385 for details

This is a WIP PR and is still has some TODOs:
* Documentation in comments (I intend to rewrite history and add those with rebasing, so for the final PR each commit will be properly documented)
* A comprehensive review of where `TypePatternExpr` is used
* Stylistic cleanup (e.g. AbstractJavaParserContext copyright comment)
* Possibly some changes to the ConcreteSyntaxModel, although this might only be true once the record pattern PR is in.


javaparser-core/src/main/java/com/github/javaparser/ast/expr/Expression.java
```diff
@@ -881,4 +877,19 @@ public final boolean elidesTypeArguments() {
         NodeWithTypeArguments nwta = (NodeWithTypeArguments) this;
         return scope.elidesTypeArguments() && (!nwta.getTypeArguments().isPresent() || nwta.isUsingDiamondOperator());
     }
+
```
jlerbsc, line 861: This appears to be generated by the TypeCastingGenerator, whereas it seems to have been added manually.

javaparser-core/src/main/java/com/github/javaparser/metamodel/TypePatternExprMetaModel.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
```
jlerbsc: You can delete this line as the class has just been created.

javaparser-core/src/main/java/com/github/javaparser/resolution/Context.java
```diff
@@ -42,21 +40,18 @@
...
+     * Returns the node wrapped in the context
+     */
+    <N extends Node> N getWrappedNode();
 
     /**
      * @return The parent context, if there is one. For example, a method exists within a compilation unit.
      */
     Optional<Context> getParent();
 
-
```
jlerbsc, line 57: Can you adapt the project configuration in your IDE so as not to multiply the code modifications?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/PatternExpr.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
```
jlerbsc: You can delete this line.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/PatternExpr.java
```diff
@@ -21,8 +21,53 @@
...
+import com.github.javaparser.TokenRange;
 
 public abstract class PatternExpr extends Expression {
+
     @AllFieldsConstructor
-    public PatternExpr() { }
+    public PatternExpr() {
+    }
+
+    @Override
```
jlerbsc, line 76: Is it normal that there is no @Generated annotation? What will happen with subsequent generations? Isn't there a risk of inconsistency in the code?

javaparser-core/src/main/java/com/github/javaparser/metamodel/PatternExprMetaModel.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
```
jlerbsc: Idem. But maybe this class has been generated?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/InstanceOfExprTest.java
```diff
@@ -100,7 +101,9 @@ void instanceOf_patternExpression() {
         assertEquals("String", expr.getType().asString());
         assertTrue(expr.getPattern().isPresent());
 
-        TypePatternExpr typePatternExpr = expr.getPattern().get();
+        PatternExpr patternExpr = expr.getPattern().get();
+        assertInstanceOf(TypePatternExpr.class, patternExpr);
+        TypePatternExpr typePatternExpr = (TypePatternExpr) patternExpr;
```
jlerbsc: Can you use the asTypePatternExpr() method to avoid casts?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/InstanceOfExprTest.java
```diff
@@ -140,7 +143,9 @@ void instanceOf_finalPatternExpression() {
...
```
jlerbsc: Idem.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
```diff
@@ -142,7 +142,10 @@ public Optional<SimpleName> getName() {
         if (pattern == null) {
             return Optional.empty();
         }
-        return Optional.of(pattern.getName());
+        if (!(pattern instanceof TypePatternExpr)) {
```
jlerbsc: Can you use the isTypePatternExpr() method to avoid instanceof?

javaparser-core/src/main/java/com/github/javaparser/ast/expr/InstanceOfExpr.java
```diff
@@ -142,7 +142,10 @@ public Optional<SimpleName> getName() {
...
+            return Optional.empty();
+        }
+        return Optional.of(((TypePatternExpr) pattern).getName());
```
jlerbsc: Can you use the asTypePatternExpr() method?

javaparser-core/src/main/java/com/github/javaparser/resolution/Context.java
```diff
@@ -325,10 +326,10 @@ default Optional<TypePatternExpr> patternExprInScope(String name) {
         // FIXME: If there are multiple patterns, throw an error?
         // First check if the pattern is directly declared within this context.
         Node wrappedNode = getWrappedNode();
-        Optional<TypePatternExpr> localResolutionResults = parentContext
+        Optional<PatternExpr> localResolutionResults = parentContext
                 .patternExprsExposedToChild(wrappedNode)
                 .stream()
-                .filter(vd -> vd.getNameAsString().equals(name))
+                .filter(vd -> vd instanceof TypePatternExpr && ((TypePatternExpr) vd).getNameAsString().equals(name))
```
jlerbsc: isTypePatternExpr and asTypePatternExpr

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -1,23 +1,23 @@
...
- *
- * You should have received a copy of both licenses in LICENCE.LGPL and
- * LICENCE.APACHE. Please refer to those files for details.
- *
- * JavaParser is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Lesser General Public License for more details.
- */
+///*
```
jlerbsc: This modification is strange.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -144,16 +144,17 @@ public SymbolReference<? extends ResolvedValueDeclaration> solveSymbolInParentCo
...
         Context parentContext = optionalParentContext.get();
         if(parentContext instanceof BinaryExprContext || parentContext instanceof IfStatementContext) {
-            List<TypePatternExpr> typePatternExprs = parentContext.patternExprsExposedToChild(this.getWrappedNode());
+            List<PatternExpr> typePatternExprs = parentContext.patternExprsExposedToChild(this.getWrappedNode());
 
-            Optional<TypePatternExpr> localResolutionResults = typePatternExprs
+            Optional<PatternExpr> localResolutionResults = typePatternExprs
                     .stream()
-                    .filter(vd -> vd.getNameAsString().equals(name))
+                    .filter(vd -> vd instanceof TypePatternExpr && ((TypePatternExpr) vd).getNameAsString().equals(name))
```
jlerbsc: I'm not going to underline them anymore. I'll let you check cast and instanceof.

jlerbsc:
Very good work. Thank you very much.

Based on your experience, you could create documentation in the wiki for using the generators (when, how, why use them?).
For the moment, to my knowledge, only this documentation is available (https://javaparser.org/code-generation-and-maven-in-javaparser/). It's not enough to ensure the project's long-term future, and it's a part that I'm not familiar with...

javaparser-core/src/main/java/com/github/javaparser/resolution/Context.java
johannescoetzee: I'm using the included `dev-files/JavaParser-idea.xml` for my project configuration, so my IDE formatting should be up to project standards. I suspect this is a change from code generator output. When I run the code generators, it effectively rewrites every source file, even if there isn't anything new being added, resulting in a lot of style changes that aren't relevant to anything being done. I've been cherry-picking files that are actually relevant, so with this being new, it wouldn't have been included before.

The proper fix is to set up automatic formatting and run it as a commit hook, which would avoid any unnecessary style changes, but I'm not sure what the best way is to do that, both in which tools to use and how to minimise impact on open PRs. Until some kind of automatic formatting exists, I think manually removing generated formatting changes from PRs is more effort than it's worth, especially since these would have to be repeated for every new codegen PR.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/PatternExpr.java
johannescoetzee: This is something I've noticed as well. Sometimes the `@Generated` annotation is added and sometimes not. It's also not entirely consistent across runs: I've seen `@Generated` added on a second or later codegen run, but I've never seen it removed. 

As far as I can tell, this annotation doesn't have any effect on the generators themselves, so the risk is that contributors will edit generated code without realising it. This is not an ideal situation, but I don't know what the solution is yet. I'm still learning how the generators work myself, so I'm hoping that adding the `RecordPatternExpr` class later will answer some of the questions I have as well.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
johannescoetzee: I agree. It's probably something I added by accident with some IDE shortcut, but I've fixed it now.

javaparser-core/src/main/java/com/github/javaparser/ast/expr/Expression.java
johannescoetzee: I think the same applies as the comment I left below regarding formatting and project configuration. I think this is generated formatting which just hasn't been included before.

johannescoetzee:
> Based on your experience, you could create documentation in the wiki for using the generators (when, how, why use them?). For the moment, to my knowledge, only this documentation is available (https://javaparser.org/code-generation-and-maven-in-javaparser/). It's not enough to ensure the project's long-term future, and it's a part that I'm not familiar with...

I had the same idea when I started with this, so I kept detailed notes of everything I did for code generation for this PR. Since I'm going to add a new node type in the follow-up PR, I want to use those notes as a reference to confirm that they're correct. Once I've confirmed that, I'll update the wiki page with everything I know about code generation.

johannescoetzee:
@jlerbsc I've updated the PR to address feedback given here (although I left the whitespace changes in with a discussion about why; please let me know if you disagree). I've also included a `PatternExprTest` class which should make codecov bot happy while providing some minimal tests for the auto-generated methods. 

At this point, I no longer consider the PR a WIP since I think it makes more sense to postpone further logic changes to the record patterns pr.

johannescoetzee:
There's something going on with the github runners:

```
Run actions/setup-java@v4
  with:
    distribution: zulu
    java-version: 12
    java-package: jdk
    check-latest: false
    server-id: github
    server-username: GITHUB_ACTOR
    server-password: GITHUB_TOKEN
    overwrite-settings: true
    job-status: success
    token: ***
  env:
    OS: macos-latest
    JDK: 12
Installed distributions
  Trying to resolve the latest version from remote
  Error: Could not find satisfied version for semver 12. 
  Available versions: 22.0.1+8, 22.0.0+36, 21.0.3+9, 21.0.2+13, 21.0.1+12, 20.0.2+9, 19.0.2+7, 18.0.2+1, 17.0.11+9, 17.0.10+7, 17.0.3+7, 16.0.2+7, 15.0.10+5, 13.0.14+5, 11.0.23+9, 11.0.15+10, 8.0.412+8, 8.0.332+9, 22.0.0+36, 21.0.2+13, 21.0.0+35, 20.0.1+9, 20.0.0+36, 19.0.1+10, 19.0.0+36, 18.0.2+9, 18.0.1+10, 18.0.0+37, 17.0.10+7, 17.0.9+8, 17.0.8+1, 17.0.8+7, 17.0.7+7, 17.0.6+10, 17.0.5+8, 17.0.4+1, 17.0.4+8, 17.0.2+8, 17.0.1+12, 17.0.0+35, 16.0.1+9, 16.0.1+9, 16.0.0+36, 15.0.9+5, 15.0.8+4, 15.0.7+4, 15.0.6+5, 15.0.5+3, 15.0.4+5, 15.0.3+3, 15.0.2+7, 15.0.1+9, 15.0.1+9, 13.0.13+5, 13.0.12+4, 13.0.11+4, 13.0.10+5, 13.0.9+3, 13.0.8+5, 13.0.7+5, 13.0.6+5, 13.0.5+1, 13.0.5+1, 13.0.5+1, 11.0.22+7, 11.0.21+9, 11.0.20+1, 11.0.20+8, 11.0.19+7, 11.0.18+10, 11.0.17+8, 11.0.16+1, 11.0.16+8, 11.0.14+1, 11.0.14+9, 11.0.13+8, 11.0.12+7, 11.0.11+9, 11.0.10+9, 11.0.9+1, 11.0.9+1, 11.0.9+1, 8.0.402+6, 8.0.392+8, 8.0.382+5, 8.0.372+7, 8.0.362+9, 8.0.362+8, 8.0.352+8, 8.0.345+1, 8.0.342+7, 8.0.322+6, 8.0.312+7, 8.0.302+8, 8.0.292+10, 8.0.282+8, 8.0.275+1
  ```

johannescoetzee:
It looks like java versions 9, 10, 12, and 14 are no longer installed correctly on the macos runners. I'll have a look and see if I can fix it

johannescoetzee:
I think https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/ is the culprit. The new macos-latest runners are M1 (therefore ARM) based (https://github.com/actions/runner-images#available-images), and there aren't azul JDK ARM releases for the failing versions: https://www.azul.com/downloads/?os=macos&architecture=arm-64-bit&package=jdk&show-old-builds=true#zulu

jlerbsc:
I've opened an exit on the project. We may have to change distrition soon.

jlerbsc:
Do you have anything else to add to this PR?

johannescoetzee:
> Do you have anything else to add to this PR?

If you are satisfied with the changes here, then I don't have anything further to add for now. Everything else would be better suited to the record pattern pr

Issue #4390, opened 2024-04-26
Different nodes have same hashcode


Laughh:
Hello! I'm using JP, and in the following code, I found that variables have the same hashCode in different scopes
```
public void hash(){
    {
        // 1
        Object cnt = new Object();
        cnt.equals(null);
    }
    {
        // 2
        Object cnt = new Object();
        cnt.equals(null);
    }

}
```

Here is the output. The ranges are different, but the hash code is Identical value. And is there any way to distinguish the two VariableDeclarators by hash code?
```
cnt = new Object()
Optional[(line 56,col 20)-(line 56,col 37)]
-628174007
--------------------
cnt = new Object()
Optional[(line 61,col 20)-(line 61,col 37)]
-628174007
--------------------
```

jlerbsc:
Hi, The hashcode calculation for a variable declaration is based on the name, type, comments and initialisation expression attributes.

Laughh:
But why the range are not considered？

jlerbsc:
I don't know exactly. It's a functional equality which is obtained by an in-depth check of all the attributes which describe a node.

jlerbsc:
I'm closing this exit because we've given an answer.

PR #4391, opened 2024-04-27
Add two methods in class VisitorMap for debug purposes

Laughh:
The method `keySet()` and `entrySet()` are still based on hash code and the built-in `equal()` method, and there is no way to get all keys and entries, so I added two address-sensetive methods to get them  which are more convenient for debugging.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -336,7 +336,7 @@
                         <dependency>
                             <groupId>com.puppycrawl.tools</groupId>
                             <artifactId>checkstyle</artifactId>
-                            <version>10.20.0</version>
+                            <version>10.20.1</version>
                         </dependency>
                     </dependencies>
                 </plugin>


jlerbsc:
Thank you for this suggestion, but I'm not sure that it's essential as the VisitorMap class accepts as parameters an implementation of a visitor for calculating the hashcode and another for implementing equals.

Laughh:
> Thank you for this suggestion, but I'm not sure that it's essential as the VisitorMap class accepts as parameters an implementation of a visitor for calculating the hashcode and another for implementing equals.

As mentioned above, the two methods are used for debugging purpose. And the reason I wrote about them was precisely because I was having trouble debugging VisitorMap. And maybe there are two keys in VisitorMap, but when we print it, only one key showed.

Laughh:
Is it useful for debugging purpose？

jlerbsc:
No, I don't think so.

Laughh:
Without the functions，is there any way to obtain all keys by using `keySet()` if two of them have same hash code and equals？Like what I describe in issue #4390

jlerbsc:
There is no way of obtaining this result with the VisitorMap class, as it relies internally on the operation of the HashCodeVisitor class. But you can certainly either use the VisitorSet class or implement your own Visitor class to suit your specific needs or extend the CompilationUnit class and redefine the hashcode and equals methods.

johannescoetzee:

PR #4393, opened 2024-04-27
Fix: Fixes the version currently supported by Javaparser.

### LINKED ISSUE #4392: Meaning of LanguageLevel
I have some Questions about the `LanguageLevel` of the parser.
1. What is difference between `CURRENT` and `BLEEDING_EDGE`? The [descriptions in the Javadoc](https://www.javadoc.io/static/com.github.javaparser/javaparser-core/3.25.10/com/github/javaparser/ParserConfiguration.LanguageLevel.html#CURRENT) are slightly different, but I cannot capture what is the different between "latest Java version" and "newest Java features". (Lately only `BLEEDING_EDGE` [was updated](https://github.com/javaparser/javaparser/commit/c8b5c77775f059e55e7531f34f522ed36fa976f7) by @johannescoetzee to 21, but `CURRENT` is still 16.)
2. The [RAW](https://www.javadoc.io/static/com.github.javaparser/javaparser-core/3.25.10/com/github/javaparser/ParserConfiguration.LanguageLevel.html#RAW) option disable post processing and validation. I understand that validation means that the code is checked to be valid for the specific Java version (e.g., Text Block Literals are not supported by Java versions lower than 15). If I want to parse all Java versions, I do not care about such validation, but what does the "post processing" do?
3. The supported Java versions are also mentioned in the README (1.0 to 18) and in the "About" section of the GitHub Repository (1 to 17). Should these versions only correspond to latest release or the status of the main branch? The range from the README actual correspond to the latest release (3.25.10).

jlerbsc:
Fixes #4392 .

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
@@ -20,6 +20,15 @@
  */
 package com.github.javaparser;
 
+import static com.github.javaparser.ParserConfiguration.LanguageLevel.POPULAR;
+
+import java.nio.charset.Charset;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+
 import com.github.javaparser.UnicodeEscapeProcessingProvider.PositionMapping;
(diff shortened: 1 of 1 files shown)


jlerbsc:
It seems that the azul distribution has removed versions 9, 10, 12 and 14 from their official API version.

Issue #4408, opened 2024-05-02
Spotless formatting implementation proposal


johannescoetzee:
# Background
The lack of automatic formatting in JavaParser makes it quite cumbersome to make any changes that involves code generation. This is because the code generators rewrite the core source files without preserving formatting, resulting in a very large diff with mostly unrelated changes. Since the code style in JP is inconsistent, it's currently impossible to fix this by formatting, so the only solution is to go through the diff and manually add files which contain relevant changes. This is both slow and error-prone (please see https://github.com/javaparser/javaparser/pull/4375 and https://github.com/javaparser/javaparser/pull/4387/ for more discussion about this). 

In https://github.com/javaparser/javaparser/pull/4202, there was the suggestion to use Spotless for automatic code formatting and I've played around with this for a bit and think I found a way to do this while minimizing impact on currently-open PRs (although there aren't that many in any case and most that exist are fairly small). 

# Solution
### Step 1: Add spotless with ratcheting to pom.xml
For this step, the idea is to add the spotless maven plugin configuration to pom.xml with the ratcheting option enabled, but to not run the plugin yet (the ratcheting option here is an option that takes a branch as input (in this case origin/master) and only applies spotless to files that have edits from the target branch). Merge this to master.

### Step 2: Fully format JavaParser
Remove the ratcheting locally, re-run the code generators and reformat the entire project. Add spotless:check to the github workflow (possibly removing checkstyle, but there may be regions where they don't overlap). Merge this to master.

Technically removing the ratcheting here isn't necessary since all the regenerated sources would still be reformatted as well, but if we're going to do a huge reformat, then I think it's worth cleaning up everything. The commit from this should also still have ratcheting enabled since that's much faster than reformatting all files every time.

### Step 3: Ignore the reformat in git blame
It's possible to ignore commits with git blame, so ignore the big format commit to avoid git-blame always pointing at the person who merged that instead of the person who actually wrote the relevant code. Github supports a `.git-blame-ignore-revs` file by default and this can also be set locally with `git config blame.ignoreRevFile` (which works for intellij as it uses cli git, but I havent' been able to find anything conclusive about eclipse support). See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view for more details.

At this point, the reformatting is done!

## Updating open PRs or WIP changes

[...]

PR #4409, opened 2024-05-02
Add spotless plugin configuration

johannescoetzee:
See https://github.com/javaparser/javaparser/issues/4408 for an extended motivation.

Spotless supports a few Java formatters, but I personally like Palantir the most. It generally follows the google style guide for Java, with some nice readability improvements. Some more details can be found at https://github.com/diffplug/spotless/tree/2d21c62cdfb83237ca4a5f48ad3b143a0580c5be/plugin-maven#java and https://github.com/palantir/palantir-java-format


pom.xml
```diff
@@ -340,6 +340,34 @@
...
+                        <formats>
+                            <format>
+                                <includes>
+                                    <include>.gitattributes</include>
+                                    <include>.gitignore</include>
+                                </includes>
+                                <trimTrailingWhitespace/>
+                                <endWithNewline/>
+                                <indent>
+                                    <tabs>true</tabs>
```
jlerbsc: Does this parameter mean that tabs are retained or replaced by spaces?

johannescoetzee: This means tabs are retained, but this section doesn't cover java code (that's handled in the `<java>` section below) but instead just `.gitattributes` and `.gitignore`. 

I think I'll just remove this section for now since the  focus of this change should just be the java source files, which is already a large enough change

PR #4411, opened 2024-05-03
Fixes #4410

### LINKED ISSUE #4410: Tests systematically failing on Windows
To test the project, the readme states to execute the mvn install command with the wrapper:
https://github.com/javaparser/javaparser/blob/master/readme.md?plain=1#L100-L105

On my machine, it fails constantly:
image

My machine: Windows 11 23H2 (build 22631.3447)

I wonder whether it is because there is a space in my user name, and thus in the path used. When I move the project in a path where there is no space, it still fails the same, but the test section still seems to depend on a path related to my user name, even if my project is not there anymore:
image

matthieu-vergne:
settings.localRepository might contains spaces, which splits the arguments once passed to the commande line in Maven. We use single quotes to keep them together as one single argument.

Notice that double quotes did not fix the issue, although many resources on the Web suggest to use double quotes for similar issues.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -148,7 +148,7 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <java.version>1.8</java.version>
         <byte-buddy.version>1.14.14</byte-buddy.version>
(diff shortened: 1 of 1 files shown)


jlerbsc:

PR #4412, opened 2024-05-04
Fixes SYSTEM_EOL warnings

matthieu-vergne:
The first commit fixes the warnings themselves.
The next commit replaces a runtime check by a compile time check to solve an issue discovered while fixing the warnings (see commit description for details).
The last two commits remove the SYSTEM_EOL and EOL deprecated constants, since they are not used anymore.

If the constants should remain there for retrocompatibility, you can ignore the two last commits. I would however suggest to explicitly mention in their Javadoc that they should be removed in the future (preferably by telling the target version for removal, like the next major version or any precise reference).


jlerbsc:
That seems fine to me. Thanks for this PR.

PR #4417, opened 2024-05-07
Fix: <enum>::values() is a static method, was not static

Kimmmey:
Hi everyone,

while working with enums, I noticed that javaparser recognices <enum>::values() as a not static method.

This pull request fixes this bug and adds an additional test.

Best regards
Kim

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java
@@ -387,7 +387,7 @@ public class JavaParserEnumDeclaration extends AbstractTypeDeclaration
 
         @Override
         public boolean isStatic() {
-            return false;
+            return true;
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java
```diff
@@ -387,7 +387,7 @@ public boolean isDefaultMethod() {
 
         @Override
         public boolean isStatic() {
-            return false;
+            return true;
```
jlerbsc, line 390: Can you identify the corresponding section in the JLS?

Kimmmey:
Hi,

the section is jls-8.9.3-100-D.

see: 
Java 8: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.3-100-D
Java 22: https://docs.oracle.com/javase/specs/jls/se22/html/jls-8.html#jls-8.9.3-100-D

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserEnumDeclaration.java
Kimmmey: Hi,

the section is jls-8.9.3-100-D.

see:
Java 8: https://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9.3-100-D
Java 22: https://docs.oracle.com/javase/specs/jls/se22/html/jls-8.html#jls-8.9.3-100-D

jlerbsc:
I can't see anything to indicate that all the methods are static, even though that's what your modification is about. I haven't done a more precise test but I'm not sure that Javaparser supports implicit methods.

jlerbsc:
Your PR seems correct. Sorry, I didn't read it properly.

Issue #4416, opened 2024-05-07
<enum>::values() can not be resolved without the enum name


Kimmmey:
Hi,

I have a problem with resolving the implicit method "values()" of an enum. Resolving is only possible, if the enum is given as the scope.

Example (does not work): 
```java
    @Test
    public void testResolveValuesMethod() {
        String s =
                "public class ClassTest {\n" +
                "    public enum SecurityPolicyScopedTemplatesKeys {\n" +
                "        SUSPICIOUS(\"suspicious\");\n" +
                "        private String displayName;\n" +
                "\n" +
                "        private SecurityPolicyScopedTemplatesKeys(String displayName) {\n" +
                "            this.displayName = displayName;\n" +
                "        }\n" +
                "\n" +
                "        public String getDisplayName() {\n" +
                "            return this.displayName;\n" +
                "        }\n" +
                "    }\n" +
                "\n" +
                "    public SecurityPolicyScopedTemplatesKeys m(int id) {\n" +
                "        return values()[id];\n" +
                "    }\n" +
                "}";
        TypeSolver typeSolver = new ReflectionTypeSolver();
        StaticJavaParser.getParserConfiguration().setSymbolResolver(new JavaSymbolSolver(typeSolver));
        CompilationUnit cu = StaticJavaParser.parse(s);
        MethodCallExpr methodCallExpr = cu.findFirst(MethodCallExpr.class).get();
        ResolvedMethodDeclaration rd = methodCallExpr.resolve();
        assertEquals("values", rd.getName());
        assertEquals("ClassTest.SecurityPolicyScopedTemplatesKeys[]", rd.getReturnType().describe());
    }
```

Example (does work):
```java
    @Test
    public void testResolveValuesMethod() {
        String s =
                "public class ClassTest {\n" +
                "    public enum SecurityPolicyScopedTemplatesKeys {\n" +
                "        SUSPICIOUS(\"suspicious\");\n" +
                "        private String displayName;\n" +
                "\n" +
                "        private SecurityPolicyScopedTemplatesKeys(String displayName) {\n" +
                "            this.displayName = displayName;\n" +
                "        }\n" +
                "\n" +
                "        public String getDisplayName() {\n" +
                "            return this.displayName;\n" +
                "        }\n" +
                "    }\n" +
                "\n" +
                "    public SecurityPolicyScopedTemplatesKeys m(int id) {\n" +
                "        return SecurityPolicyScopedTemplatesKeys.values()[id];\n" +
[...]

PR #4424, opened 2024-05-17
Fix: solveMethodAsUsage() for implicit method <enum>::values()

Kimmmey:
Hi everyone,

this pull request fixes the remaining problem mentioned in #4416 and adds a test. 

Calculating the type of an expression, containing the implicit method `values()`, with `calculateResolvedType()` throws an exception.  

Reason: `calculateResolvedType()` uses the method `solveMethodAsUsage(...)` and this method is not able to resolve the implicit method `values()`.

Best regards 
Kim


jlerbsc:
Thank you for this fix.

PR #4430, opened 2024-05-22
Make resolution of implements and extends types start with the parent…

### LINKED ISSUE #4427: Incorrect resolution of extends and implements types
JavaParserClassDeclaration.toReferenceType is used when trying to resolve class extends and implements lists.
This method attempts to resolve the typename first within the context of the class declaration itself -- which sounds wrong (can the extends list of a class refer to an internally defined type?)

Specifically, here's code that would cause indirect resolution. If we make DerivedClass.BaseClass extend something we can also trigger a stack overflow this way.

BaseClass.java:
```
public abstract class BaseClass {

    protected int x;
}

```

DerivedClass.java:
```
public class DerivedClass extends BaseClass {

    void f() {
        x = x + 1;
    }


    static class BaseClass  {

    }
}

```

Failing testcase:
```
import java.nio.file.Path;

import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;

public class IssueXXXTest extends AbstractSymbolResolutionTest {
  @Test
  public void testIssueXXX() throws IOException {
    Path issueResourcesPath = adaptPath("src/test/resources/issueXXXX");
    ReflectionTypeSolver rts = new ReflectionTypeSolver();
    JavaParserTypeSolver jpts = new JavaParserTypeSolver(issueResourcesPath);
    CombinedTypeSolver cts = new CombinedTypeSolver();
    cts.add(rts);
    cts.add(jpts);
    ParserConfiguration pc = new ParserConfiguration()
        .setSymbolResolver(new JavaSymbolSolver(cts));
    StaticJavaParser.setConfiguration(pc);
    CompilationUnit cu = StaticJavaParser.parse(issueResourcesPath.resolve("DerivedClass.java"));

    // We shouldn't throw a mismatched symbol 
    assertDoesNotThrow(() -> cu.findAll(NameExpr.class).stream()
            .map(NameExpr::resolve)
            .findAny().get());
  }
}

```

eldapiiro:
… context of the class.

Fixes #4427.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -466,6 +466,9 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration
             className = classOrInterfaceType.getScope().get().toString() + "." + className;
         }
 
+        // Since this is used to resolve reference to "extended" and "implemented" types, and since these type references
+        // should not be resolved against member types of the current type, we resolve based on the context containing
+        // the class declaration.
         SymbolReference<ResolvedTypeDeclaration> ref = getContext().getParent()
                 .orElseThrow(() -> new RuntimeException("Parent context unexpectedly empty."))
                 .solveType(className);
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
@@ -465,16 +465,22 @@ public class JavaParserClassDeclaration extends AbstractClassDeclaration
             // look for the qualified name (for example class of type Rectangle2D.Double)
             className = classOrInterfaceType.getScope().get().toString() + "." + className;
         }
-        SymbolReference<ResolvedTypeDeclaration> ref = solveType(className);
+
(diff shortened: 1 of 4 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserClassDeclaration.java
```diff
@@ -465,17 +465,23 @@ private ResolvedReferenceType toReferenceType(ClassOrInterfaceType classOrInterf
             // look for the qualified name (for example class of type Rectangle2D.Double)
             className = classOrInterfaceType.getScope().get().toString() + "." + className;
         }
-        SymbolReference<ResolvedTypeDeclaration> ref = solveType(className);
+
```
jlerbsc, line 468: Please comment on why, at this stage, type resolution should be delegated to the parent context.

eldapiiro: toReferenceType is used to resolve reference to "extends" and "implements" types. These types should not be resolved against e.g types with the same name defined within the class.
As per the test case I included -- 

```
class A extends B { 
static class B ... {} 
}

```
The B in "extends B" should not resolve to the B in "static class B"... Hence we should resolve based on the context containing the class.. Does this make sense?

jlerbsc: I meant adding comments to the code.

eldapiiro: Added, thanks

jlerbsc:

PR #4432, opened 2024-05-23
[JEP 440] Add support for record patterns

johannescoetzee:
Implements https://github.com/javaparser/javaparser/issues/4385. I've already described the general approach I've taken in that issue, so won't repeat that here. 

The diff is huge, but I think all of this is necessary to show that the implementation works. In an attempt to make this more manageable, I used the commits to document all the steps I took while implementing this. This means the commit list is also very long, but this will also be useful as a reference for improving the documentation on adding nodes and code generation in general (which I intend to do soon). 

A lot of the tests added here are just boilerplate tests which either test generated code, or are copies of similar tests for `TypePatternExpr`. I've left comments on the tests that test more interesting functionality.

This PR doesn't include a `ResolvedRecordPatternDeclaration` class. I could implement one similar to `ResolvedTypePatternDeclaration`, but it's not obvious to me how this would be used (and the type pattern declaration one contains a comment suggesting the author of that wasn't too sure either). Any insight on how such a class could be useful would be much appreciated!


javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
```diff
@@ -287,4 +287,37 @@ void testRemoveWithGuard() {
 
         assertFalse(Navigator.findNameExpression(entry, "b").isPresent());
     }
+
+    @Test
```
johannescoetzee, line 291: Simple AST test

javaparser-core-testing/src/test/java/com/github/javaparser/printer/PrettyPrinterTest.java
```diff
@@ -651,4 +651,20 @@ public void testSwitchPatternWithGuard() {
         CompilationUnit cu = parse(code);
         assertEqualsStringIgnoringEol(code, new DefaultPrettyPrinter().print(cu));
     }
+
+    @Test
```
johannescoetzee, line 655: PrettyPrinter test

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
```diff
@@ -114,4 +115,16 @@ void switchPatternWithGuardPreserved() {
         statements.set(0, statements.get(0).clone());
         assertTransformedToString(code, stmt);
     }
+
+    @Test
```
johannescoetzee, line 119: LPP test

javaparser-core/src/main/java/com/github/javaparser/ast/expr/PatternExpr.java
```diff
@@ -138,6 +138,7 @@ public ReferenceType getType() {
     }
 
     @Override
+    @Generated("com.github.javaparser.generator.core.node.ReplaceMethodGenerator")
```
johannescoetzee, line 141: I don't know why these `@Generated` annotations appeared suddenly, but this code was, in fact, generated.

javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java
```diff
@@ -137,11 +136,13 @@ private static CsmElement typeArguments() {
...
+        concreteSyntaxModelByClass.put(RecordPatternExpr.class, sequence(comment(), child(ObservableProperty.TYPE), space(), token(GeneratedJavaParserConstants.LPAREN), list(ObservableProperty.PATTERN_LIST, sequence(comma(), space()), none(), none()), token(GeneratedJavaParserConstants.RPAREN)));
         concreteSyntaxModelByClass.put(SimpleName.class, attribute(ObservableProperty.IDENTIFIER));
         concreteSyntaxModelByClass.put(SingleMemberAnnotationExpr.class, sequence(comment(), token(GeneratedJavaParserConstants.AT), child(ObservableProperty.NAME), token(GeneratedJavaParserConstants.LPAREN), child(ObservableProperty.MEMBER_VALUE), token(GeneratedJavaParserConstants.RPAREN)));
         concreteSyntaxModelByClass.put(StringLiteralExpr.class, sequence(comment(), stringToken(ObservableProperty.VALUE)));
-        concreteSyntaxModelByClass.put(TextBlockLiteralExpr.class, sequence(comment(), textBlockToken(ObservableProperty.VALUE)));
```
johannescoetzee, line 143: I seem to have rearranged some items here to maintain alphabetic order. I can't remember why I chose to do that for this PR, so can undo it if that's preferred.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/SwitchEntryContext.java
```diff
@@ -78,19 +81,6 @@ public SymbolReference<? extends ResolvedValueDeclaration> solveSymbol(String na
...
-                continue;
-            }
-
-            TypePatternExpr typePatternExpr = e.asTypePatternExpr();
-            if (typePatternExpr.getNameAsString().equals(name)) {
-                JavaParserPatternDeclaration decl = JavaParserSymbolDeclaration.patternVar(typePatternExpr, typeSolver);
-                return SymbolReference.solved(decl);
-            }
-        }
-
```
johannescoetzee, line 93: I added this logic while adding support for switch pattern expressions. The possibility of having record patterns as well complicates this, however, so I moved the logic for finding declarations in patterns to `AbstractJavaParserContext` to avoid duplicating that. For that to work, I added the `typePatternExprsExposedToChild` method below which just finds all pattern exprs in the entry label and returns them (since they're all exposed to the children of this entry)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -143,27 +144,29 @@ public SymbolReference<? extends ResolvedValueDeclaration> solveSymbolInParentCo
 
         // First check if there are any pattern expressions available to this node.
         Context parentContext = optionalParentContext.get();
-        if(parentContext instanceof BinaryExprContext || parentContext instanceof IfStatementContext) {
```
johannescoetzee, line 145: The changes in this method are all to keep type solving logic for patterns contained. See a comment about this below in `SwitchEntryContext`

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -287,4 +290,27 @@ public Optional<MethodUsage> solveMethodAsUsage(String name, List<ResolvedType>
     public N getWrappedNode() {
         return wrappedNode;
     }
+
+    public List<TypePatternExpr> typePatternExprsDiscoveredInPattern(PatternExpr patternExpr) {
```
johannescoetzee, line 307: In the symbol solver, we're generally only interested in `TypePatternExprs` since these contain the variable declarations needed. This method just does a breadth-first traversal of the pattern tree to discover all `TypePatternExprs` contained in the nested `RecordPatternExpr` structure.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
```diff
@@ -825,6 +825,57 @@ void localVariablesExposedToChildWithinTryWithResourcesStatement() {
         assertOneVarExposedToChildInContextNamed(stmt, stmt.getTryBlock(), "res1");
     }
 
+    @Nested
+    class RecordPatternExprTests {
```
johannescoetzee, line 829: Tests to ensure declarations in nested patterns are correctly exposed to children

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/InstanceOfTest.java
```diff
@@ -551,6 +560,24 @@ public void test_shouldFail() {
 
     }
 
+    @Nested
```
johannescoetzee, line 563: Test to ensure symbols in record patterns can be solved in instanceof expressions

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/SwitchExprTest.java
```diff
@@ -91,4 +91,20 @@ public void switchPatternInOtherCaseShouldNotResolve() {
 
         assertThrows(UnsolvedSymbolException.class, resolveS);
     }
+
```
johannescoetzee, line 94: Test to ensure symbols in record patterns can be solved in switch expressions

johannescoetzee:
The codecov bot messages suggest I missed some boilerplate tests. I'll add those either tomorrow or early next week

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -287,4 +287,27 @@ public Optional<MethodUsage> solveMethodAsUsage(String name, List<ResolvedType>
     public N getWrappedNode() {
         return wrappedNode;
     }
+
```
jlerbsc, line 292: Comments in the code would be welcome

jlerbsc:
Everything seems fine to me. If you have nothing to add, I will merge your PR. Thank you for your excellent work.

jlerbsc:
Can you resolve the checkstyle violations?

[INFO] There are 2 errors reported by Checkstyle 10.17.0 with dev-files/JavaParser-CheckStyle.xml ruleset.
Error:  src/main/java/com/github/javaparser/resolution/Context.java:[26,8] (imports) UnusedImports: Unused import - com.github.javaparser.ast.expr.PatternExpr.
Error:  src/main/java/com/github/javaparser/ast/expr/RecordPatternExpr.java:[11,8] (imports) UnusedImports: Unused import - java.util.regex.Pattern.

johannescoetzee:
I've updated the pr with a comment explaining the type pattern discovery method, tests that should satisfy some of the codecov issues (but not all) and which fixed checkstyle violations.

@jlerbsc if the remaining codecov violations are acceptable (they're all for generated boilerplate, as far as I can tell, except for the deprecated PrettyPrinterVisitor method which I added for anyone still using that, but which is tested in the new DefaultPrettyPrintVisitor), then I don't have anything further to add to this PR. 

I think fixing https://github.com/javaparser/javaparser/issues/4440 should then be the only remaining task for java 21 support, but please let me know if you can think of anything I missed!

Edit: looks like I added enough tests to get the coverage above average and therefore codecov bot is happy, but I could still add some more if that's preferred :)

PR #4452, opened 2024-06-02
Fixed issue #4450 and added unit tests

### LINKED ISSUE #4450: Endless recursion (-> StackOverflow) with cyclic static references
When analyzing files with mutual static imports, symbol solving decays into an endless loop. In the following example:

`package a;

import static a.RefCycleEnum.*;

public class RefCycleClass {

    public static int value1Value = 1;

    private RefCycleEnum theEnum;

    public void runMe() {
        unknownName.run();
    }
}

-----------------------------------

package a;

import static a.RefCycleClass.*;

public enum RefCycleEnum {
    Value1(value1Value);

    private RefCycleEnum(int value) {
	Underlying = value;
    }
    
    public final int Underlying;
}
`

When trying to solve unknownName, the system will eventually resort to resolve the name within RefCycleEnum (static-imported by the first compilation unit); when resolving unknownName within RefCycleEnum, the system will eventually resort to solving it within RefCycleClass (static-imported), and so on.

Should probably restrict solving process for static imports (e.g only consider public static declarations within the type), and specifically not continue looking in imports within the compilation unit.

eldapiiro:
When looking up a symbol in a type based on a static import, restrict search to members of the type.

Fixes #4450

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedValueDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedValueDeclaration.java
@@ -40,7 +40,9 @@ public interface ResolvedValueDeclaration extends ResolvedDeclaration {
     static boolean IsPublicStaticField(ResolvedValueDeclaration declaration) {
         if(declaration instanceof ResolvedFieldDeclaration) {
             ResolvedFieldDeclaration resolvedFieldDeclaration = (ResolvedFieldDeclaration)declaration;
-            return resolvedFieldDeclaration.isStatic() && resolvedFieldDeclaration.accessSpecifier() == AccessSpecifier.PUBLIC;
+            if(resolvedFieldDeclaration.isStatic() && resolvedFieldDeclaration.accessSpecifier() == AccessSpecifier.PUBLIC) {
+                return true;
+            };
         }
 
         return declaration instanceof ResolvedEnumConstantDeclaration;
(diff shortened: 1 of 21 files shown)


eldapiiro:
@jlerbsc reaching out to you since you looked at my previous PR -- seems like codecov fails running on this one due to a missing token...? Any advice here would be appreciated.

jlerbsc:
I can't validate this PR for several reasons. The first is that I can't reproduce your error with your test case. In addition, your PR is confusing (there are commits that modify other commits) and it is too specific (solvePublicStaticField). I think that if the issue exists, it should be possible to solve it more simply without creating a method responsible for solving fields with certain attributes.

eldapiiro:
> I can't validate this PR for several reasons. The first is that I can't reproduce your error with your test case. In addition, your PR is confusing (there are commits that modify other commits) and it is too specific (solvePublicStaticField). I think that if the issue exists, it should be possible to solve it more simply without creating a method responsible for solving fields with certain attributes.

I'll try to reproduce the error with the latest version of the library. Regarding your comment on the specificity of the solution -- the issue stems from handling static imports. Specifically right now (as of the 25.x) the code looks at symbols that may not be used for resolving static imports... I'm not sure how we can avoid restricting the search for static fields if we want to maintain correctness when resolving static imports. Any thoughts?

eldapiiro:
> > I can't validate this PR for several reasons. The first is that I can't reproduce your error with your test case. In addition, your PR is confusing (there are commits that modify other commits) and it is too specific (solvePublicStaticField). I think that if the issue exists, it should be possible to solve it more simply without creating a method responsible for solving fields with certain attributes.
> 
> I'll try to reproduce the error with the latest version of the library. Regarding your comment on the specificity of the solution -- the issue stems from handling static imports. Specifically right now (as of the 25.x) the code looks at symbols that may not be used for resolving static imports... I'm not sure how we can avoid restricting the search for static fields if we want to maintain correctness when resolving static imports. Any thoughts?

@jlerbsc alas it seems like the issue reproduces with the latest master as well. How can we progress this? Do you want to hop on a zoom and discuss, perhaps?

johannescoetzee:

jlerbsc:
I'm closing this PR because the issue has been resolved in a less specific way. Thank you for your suggestion.

PR #4453, opened 2024-06-04
Add link to the guide to adding nodes in CONTRIBUTING.md

johannescoetzee:
I wrote this as a walkthrough for the changes I made in https://github.com/javaparser/javaparser/pull/4432 and I think this guide would be useful for future contributors who want to add support for new Java features. `CONTRIBUTING.md` already has a link to the wiki, but I think it's worth adding a specific mention of this page since it mentions all of the major `JavaParser` components and is a fairly comprehensive overview of what's required when adding new nodes.


jlerbsc:
Thank you for this guide.

PR #4460, opened 2024-06-06
Fix record declarations nested in annotation declarations

johannescoetzee:
From the grammar
```
AnnotationTypeMemberDeclaration:
    AnnotationTypeElementDeclaration
    ConstantDeclaration
    ClassDeclaration
    InterfaceDeclaration
    ;
ClassDeclaration:
    NormalClassDeclaration
    EnumDeclaration
    RecordDeclaration
```

it is possible to declare a record as a child of an annotation declaration, but this case is currently not handled, leading to parse errors. This PR adds that case to the JavaParser grammar with some tests. I added the lpp test to make sure it works, but no fixes were required there.

(I discovered this issue when parsing some larger projects to see if there are any more switch-related bugs, but this is an older bug that is unrelated to any recent changes)


jlerbsc:
Thank you for your attention.

PR #4461, opened 2024-06-06
Fix crash on SwitchExpr entries if tokens are not stored

johannescoetzee:
When creating SwitchExpr entries with an expression on the RHS, an ExpressionStmt is instantiated with the token range of the child Expression used as the token range for the ExpressionStmt. The problem is that this was not done safely, since `expr.getTokenRange()` can be empty if the `storeTokens` configuration option is disabled and in this case the parser would crash. This PR provides a safe alternative for this case, preventing the crash.

This is an old bug that was most likely introduced with the original `SwitchExpr` support, but I missed it during my refactorings of the code.


jlerbsc:
Thank you for this PR.

PR #4465, opened 2024-06-12
Format code with spotless

johannescoetzee:
The motivation for this PR is discussed in https://github.com/javaparser/javaparser/issues/4408 and this is a follow-up to https://github.com/javaparser/javaparser/pull/4409

The changes in this PR were made by running `./run_core_metamodel_generator.sh && ./run_core_generators.sh && ./mvnw spotless:apply` until the output was stable (that is to say, until running these commands again didn't result in any changes to the source files). I ran these commands with the ratcheting option disabled, so all source files in the repo were reformatted, but this should be unnecessary for regular PRs.

I did have to make a few minor manual changes where line comments `// ...` were used in the middle of multi-line call chains or argument lists since these were somehow being duplicated on the lines above (which is probably a pretty printer bug, but that's out of the scope of this PR). The manual changes I made were either moving these comments to the line directly before the statement they appeared in, or to turn them into block comments `/* ... */`


jlerbsc:
> were used in the middle of multi-line call chains or argument lists

Can you check that these particular cases are not test cases for the LPP or for the comment insertion function.

johannescoetzee:
I double checked, and the ones I changed manually were:
```
javaparser-core/src/main/java/com/github/javaparser/printer/ConcreteSyntaxModel.java

comment(), // FIXME: we should introduce a derived property
child(ObservableProperty.NAME), // list(ObservableProperty.EXTRA_ARRAY_LEVELS),


javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrettyPrinterVisitor.java AND
javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java

final boolean // Either we hit the constant amount limit in the configurations, or...
```

These look unrelated to the LPP or any tests

PR #4477, opened 2024-06-21
Fix: issue 4450 Endless recursion (-> StackOverflow) with cyclic static references

### LINKED ISSUE #4450: Endless recursion (-> StackOverflow) with cyclic static references

jlerbsc:
Fixes #4450 .

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/CompilationUnitContext.java
@@ -21,6 +21,11 @@
 
 package com.github.javaparser.symbolsolver.javaparsermodel.contexts;
 
+import java.util.ArrayList;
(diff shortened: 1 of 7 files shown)


jlerbsc:
Ignoring spotless checks

PR #3606, opened 2022-06-07
Initial implementation for resolving records

MysterAitch:
Work-in-progress

ref #2446

- [ ] JavaParser (`javaparsermodel`) source type solver
- [ ] Jar (`javassist`) type solver - not started
- [ ] Reflection (`reflectionmodel`) type solver - tentative

PR #4489, opened 2024-07-04
Add record resolution [continuation of PR #3606]

johannescoetzee:
This is a continuation of https://github.com/javaparser/javaparser/pull/3606 which makes some changes to the work already done there and also adds javassist support.

I'm opening this as a draft since there is one existing testcase that is failing and there are still more tests required and some changes to be  made.

The following test needs to be fixed:
* [X] The record class should inherit default methods from implemented interfaces (the problem was in the test code, not the implementation)

The following tests are still required:
* [ ] Constructor tests for the javassist model
* [ ] Constructor tests for the reflection model
* [ ] Tests for generic types for record parameters

Currently there is also a lot of duplicated code between the `RecordDeclarationContext`, `JavaParser/Javassist/ReflectionRecordDeclaration` and their`ClassDeclaration*` counterparts. This works, but is not good for maintainability. The two options I can think of for fixing this are:
1. Make the `RecordDeclaration*` classes subclasses of the `ClassDeclaration*` versions and only override methods where the functionality differs. This leads to a strange type hierarchy, however, since `ResolvedRecordDeclaration.isClass` returns false (taken from the earlier draft PR, but I can also change this), so `ResolvedRecordDeclaration.isClass()` is false, but `ResolvedRecordDeclaration instanceof ResolvedClassDeclaration` is true.
2. Extract the common functionality into a common base class (the start of which can already be seen in this PR in the `AbstractClassOrRecordDeclaration` class, but I put this aside while working on functionality since the common functionality occurs over multiple levels of abstraction, some of which are also shared by anonymous classes and such where it wasn't immediately obvious whether they should share this or not.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/logic/AbstractClassOrRecordDeclaration.java
```diff
@@ -0,0 +1,88 @@
...
+import com.github.javaparser.resolution.types.ResolvedReferenceType;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A common ancestor for all ClassDeclarations.
+ *
+ * @author Federico Tomassetti
+ */
+public abstract class AbstractClassOrRecordDeclaration extends AbstractTypeDeclaration
```
johannescoetzee: This is currently unused so will either be updated or removed, depending on which refactor option is chosen

jlerbsc:
It's hard to read because the PR mixes code changes with formatting. But I'm willing to validate it as is because it adds functionality that isn't currently covered.

johannescoetzee:
> It's hard to read because the PR mixes code changes with formatting. But I'm willing to validate it as is because it adds functionality that isn't currently covered.

Sorry about that! This work started as me just playing around with various components to figure out how they work with a lot of back and forth between files as I was putting together the pieces.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/reflectionmodel/ReflectionConstructorDeclaration.java
```diff
@@ -47,8 +46,12 @@ public ReflectionConstructorDeclaration(Constructor<?> constructor, TypeSolver t
     }
 
     @Override
-    public ResolvedClassDeclaration declaringType() {
-        return new ReflectionClassDeclaration(constructor.getDeclaringClass(), typeSolver);
+    public ResolvedReferenceTypeDeclaration declaringType() {
```
johannescoetzee, line 49: An API change is required here since the declaring type for a constructor can be a record as well

johannescoetzee:
> It's hard to read because the PR mixes code changes with formatting.

Some of this was me forgetting to format each commit, so if you're looking commit-by-commit then that's on me. If you're looking at the overall diff, then the formatting changes are due to improperly formatted code being merged to master, which was fixed when I reformatted the code for this PR. https://github.com/javaparser/javaparser/pull/4499 will hopefully fix this once I've rebased.

johannescoetzee:
Once https://github.com/javaparser/javaparser/pull/4499 has been merged, I will update this PR and then I think it will be ready for a proper review. As I mentioned before, the implementation here contains a lot of code duplication which should be cleaned up (although it does continue the pattern followed until now). It's also quite possible that I missed some cases which will still fail to resolve, but the PR as it is, is certainly already an improvement and such edge cases can easily be accounted for later.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistRecordDeclarationTest.java
```diff
@@ -322,6 +330,35 @@ void testNonCanonicalConstructor() {
         assertEquals("java.lang.String", stringConstructor.getParam(0).getType().describe());
     }
 
+    @Test
+    @EnabledForJreRange(min = org.junit.jupiter.api.condition.JRE.JAVA_17)
+    void genericConstructorTest() {
```
jlerbsc, line 335: It seems to me that this is not a Javassist test because there is no reference to jar declaration.

johannescoetzee: The reference to the jar is through the typeSolver. box.GenericBox is imported, but this declaration is included in the Box.jar used to create the type solver.


There's a similar test for the reflection model

jlerbsc:
I'm closing this PR in favour of  https://github.com/javaparser/javaparser/pull/4489

jlerbsc:
Is this PR ready to be merged?

johannescoetzee:
I see codecov bot is complaining that 
```
    default ResolvedReferenceTypeDeclaration getSolvedJavaLangRecord() throws UnsolvedSymbolException {
        return solveType(JAVA_LANG_RECORD);
```
is not covered in the tests. I want to check this quickly, but otherwise I think it should be fine.

jlerbsc:
Thank you for your work.

PR #4498, opened 2024-07-11
Fix project formatting

johannescoetzee:
I just ran `./mvnw spotless:apply` to reformat the project. 

This should clean up the diff for https://github.com/javaparser/javaparser/pull/4489 a bit once I've rebased it. 

Edit: the formatting check is failing here, even though `./mvnw spotless:check` succeeds when I run it locally. I find that very strange


johannescoetzee:
Going to close and re-open to force a re-run

PR #4506, opened 2024-07-22
Allow primitive types for patterns

### LINKED ISSUE #4504: ParseProblemException when parsing record pattern with primitive types
Hi! I'm using 3.26.1, I set language level to ParserConfiguration.LanguageLevel.BLEEDING_EDGE and I'm getting this exception:
```
Exception in thread "main" com.github.javaparser.ParseProblemException: (line 3,col 26) Parse error. Found "(", expected one of  "!=" "%=" "&" "&&" "&=" ")" "*=" "+=" "-=" "->" "/=" "::" "<<=" "=" "==" ">>=" ">>>=" "?" "^" "^=" "|" "|=" "||"
Problem stacktrace :
  com.github.javaparser.GeneratedJavaParser.generateParseException(GeneratedJavaParser.java:14457)
  com.github.javaparser.GeneratedJavaParser.jj_consume_token(GeneratedJavaParser.java:14302)
  com.github.javaparser.GeneratedJavaParser.IfStatement(GeneratedJavaParser.java:6704)
  com.github.javaparser.GeneratedJavaParser.Statement(GeneratedJavaParser.java:5832)
  com.github.javaparser.GeneratedJavaParser.BlockStatement(GeneratedJavaParser.java:6079)
  com.github.javaparser.GeneratedJavaParser.Statements(GeneratedJavaParser.java:2811)
  com.github.javaparser.GeneratedJavaParser.Block(GeneratedJavaParser.java:5955)
  com.github.javaparser.GeneratedJavaParser.MethodDeclaration(GeneratedJavaParser.java:2201)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBodyDeclaration(GeneratedJavaParser.java:1796)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceBody(GeneratedJavaParser.java:1286)
  com.github.javaparser.GeneratedJavaParser.ClassOrInterfaceDeclaration(GeneratedJavaParser.java:538)
  com.github.javaparser.GeneratedJavaParser.CompilationUnit(GeneratedJavaParser.java:156)
  com.github.javaparser.JavaParser.parse(JavaParser.java:125)
  com.github.javaparser.utils.SourceRoot.tryToParse(SourceRoot.java:134)
  com.github.javaparser.utils.SourceRoot.tryToParse(SourceRoot.java:149)
  com.github.javaparser.utils.SourceRoot.parse(SourceRoot.java:265)
```
when trying to parse something like this:
```
class Test {
    void foo(Object o) {
        if (o instanceof RecordType(int i)) {
        }
    }
}
```
While this gets parsed without any issues:
```
class Test {
    void foo(Object o) {
        if (o instanceof RecordType(Int i)) {
        }
    }
}
```
Among other things, that means the example
```
// As of Java 21
static void printSum(Object obj) {
    if (obj instanceof Point(int x, int y)) {
        System.out.println(x+y);
    }
}
```
provided here https://openjdk.org/jeps/440 fails the same way.

From what I could see with my limited knowledge of javaparser, the problem is probably that java.com.github.javaparser.ast.expr.PatternExpr.type is ReferenceType.

johannescoetzee:
Fixes #4504.

The type of a PatternExpr has been represented by a reference type until now, which was correct before the introduction of record patterns (since a top level pattern type must always be a reference type from https://docs.oracle.com/javase/specs/jls/se21/html/jls-14.html#jls-Pattern), but now nested type patterns may have a primitive type as well. The type of a record pattern expr must still be a reference type since there aren't primitive records in java.

This PR updates the type of a pattern to reflect this, which includes making a necessary API change.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
@@ -456,4 +456,28 @@ class SwitchExprTest {
         assertEquals("Foo f", entry.getLabels().get(0).toString());
         assertEquals("f.get();", entry.getStatements().get(0).toString());
     }
+
+    @Test
+    void testRecordPatternWithPrimitiveType() {
+        SwitchExpr switchExpr =
+                parseExpression("switch (foo) { case Foo(int x) -> sink(x); }").asSwitchExpr();
+
+        assertEquals(Node.Parsedness.PARSED, switchExpr.getParsed());
+
+        SwitchEntry entry = switchExpr.getEntry(0);
+        Expression switchLabel = entry.getLabels().get(0);
+
+        assertEquals("Foo(int x)", switchLabel.toString());
+        assertTrue(switchLabel.isRecordPatternExpr());
+
+        RecordPatternExpr recordPattern = switchLabel.asRecordPatternExpr();
+
+        assertEquals("Foo", recordPattern.getType().toString());
+        assertEquals(1, recordPattern.getPatternList().size());
+        assertTrue(recordPattern.getPatternList().get(0).isTypePatternExpr());
+
+        TypePatternExpr innerType = recordPattern.getPatternList().get(0).asTypePatternExpr();
+
+        assertTrue(innerType.getType().isPrimitiveType());
+    }
 }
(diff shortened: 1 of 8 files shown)


jlerbsc:
Thank you for this fix.

PR #4527, opened 2024-08-14
Add missing copyright notice to RecordPatternExpr.java

johannescoetzee:
I missed this in the original PR adding support for record patterns


jlerbsc:
There is a spotless check error!

johannescoetzee:
Hmm, I ran `./mvnw spotless:apply` before creating this PR, but in this case it wasn't enough since the check in PRs also re-run the code generators. I've fixed this now and will follow this PR up with one that adds a `maven format` rule or something that also runs the generators and update documentation to mention that

Issue #4344, opened 2024-03-15
nameExpr.resolve().isField() resolves incorrectly in certain context


YXL-123:
I want to add this to the field reference that omits this in the method, but when there is a variable declaration with the same name in the method, the field reference before the variable declaration will be resolved as a variable instead of a field.
**Main function**
```
    public static void main(String[] args) throws IOException {
        JavaParser javaParser = new JavaParser();
        TypeSolver typeSolver = new CombinedTypeSolver(new ReflectionTypeSolver());
        JavaSymbolSolver symbolSolver = new JavaSymbolSolver(typeSolver);
        javaParser.getParserConfiguration().setSymbolResolver(symbolSolver);
        javaParser.getParserConfiguration().setLanguageLevel(ParserConfiguration.LanguageLevel.JAVA_17);
        ParseResult<CompilationUnit> result = javaParser.parse(Paths.get("This.java"));
        ModifierVisitor visit = new ModifierVisitor(){
            public Visitable visit(MethodDeclaration n, Void arg) {
                List<NameExpr> all = n.findAll(NameExpr.class);
                all.stream().filter(nameExpr -> {
                    try {
                        if (nameExpr.resolve().isField())
                            return true;
                    } catch (UnsolvedSymbolException e){
                        return false;
                    }
                    return false;
                }).forEach(nameExpr -> {
                    Node parent = nameExpr.getParentNode().get();
                    if(parent instanceof FieldAccessExpr && ((FieldAccessExpr) parent).getScope() instanceof ThisExpr) return;
                    FieldAccessExpr fieldAccessExpr = NodeFactory.getFieldAccessExpr(new ThisExpr(), nameExpr.getName());
                    nameExpr.replace(fieldAccessExpr);
                });
                return super.visit(n, arg);
            }
        };
        visit.visit(result.getResult().get(), null);
        System.out.println(result.getResult().get());
    }
```

**This.java**
```
public class This {
    int x = 5;

    void test(){
        System.out.println(x++);  // x here is field
        int x = 0;
        System.out.println(x++);  // x here is the variable
    }
}
```
output:
> public class This {
> 
>     int x = 5;
> 
>     void test() {
>         // x here is field
>         System.out.println(x++);
>         int x = 0;
>         // x here is the variable
>         System.out.println(x++);
>     }
> }

>

jlerbsc:
This is a bug that has already been identified but for which we do not yet have a solution. 

[...]

PR #4539, opened 2024-08-20
Fix scope for variables introduced by pattern expressions

### LINKED ISSUE #4525: Pattern variables not always introduced into the correct scope
https://docs.oracle.com/javase/specs/jls/se21/html/jls-6.html#jls-6.3.1 and onwards in the language specification provides a list of rules for how variables are introduced into various scopes. The JavaParser symbol solver handles many of these correctly, but doesn't implement some of these rules correctly. I've implemented tests for all of the rules in the spec [here](https://github.com/johannescoetzee/javaparser/blob/969555f3c3c41707916b38f079cd30867d11d8be/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/PatternVariableIntroductionTest.java), but include the failing testcases below. I've double checked that these code examples compile, so these variables should all be resolvable.

I'm going to try to fix this, but decided to open this issue for documentation and to discuss ideas.

## Failing tests:
```
@Test
public void ifIntroducesPatternVariable1() {
    CompilationUnit cu = parse("public class Test {\n"
            + "  public void foo(Object o) {\n"
            + "    if (!(o instanceof String s)) {\n"
            + "      return;\n"
            + "    }\n"
            + "    System.out.println(s);\n"
            + "  }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "s").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifIntroducesPatternVariable7() {
    CompilationUnit cu = parse("public class Test {\n"
            + "  public void foo(Object o, boolean b) {\n"
            + "    if (!(o instanceof String s)) {\n"
            + "        return;\n"
            + "    } else if (b) {\n"
            + "        // do nothing\n"
            + "    } \n"
            + "    System.out.println(s);\n"
            + "  }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "s").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifIntroducesPatternVariable8() {
    CompilationUnit cu = parse("public class Test {\n"
            + "  public void foo(Object o, boolean b) {\n"
            + "    if (!(o instanceof String s)) {\n"
            + "        return;\n"
            + "    } else if (b) {\n"
            + "        // do nothing\n"
            + "    } else {}\n"
            + "    System.out.println(s);\n"
            + "  }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "s").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifIntroducesPatternVariable9() {
    CompilationUnit cu = parse("public class Test {\n"
            + "  public void foo(Object o, boolean b) {\n"
            + "    if (!(o instanceof String s)) {\n"
            + "        return;\n"
            + "    } else if (b) {\n"
            + "        // do nothing\n"
            + "    } else {\n"
            + "        return;\n"
            + "    }\n"
            + "    System.out.println(s);\n"
            + "  }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "s").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void conditionalOperator1() {
    CompilationUnit cu = parse("public class Test {\n"
            + "    public void foo(Object o) {\n"
            + "      boolean b = o instanceof String value ? value.length() > 0 : false;\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void conditionalOperator2() {
    CompilationUnit cu = parse("public class Test {\n"
            + "    public void foo(Object o) {\n"
            + "      boolean b = !(o instanceof String value) ? false : value.length() > 5;\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void conditionalOr1() {
    CompilationUnit cu = parse("public class Test {\n"
            + "    public void foo(Object o) {\n"
            + "        if (!(o instanceof String value) || value.length() > 5) { }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void conditionalOr2_1() {
    CompilationUnit cu = parse("public class Test {\n"
            + "    public void foo(Object o, boolean b) {\n"
            + "        if (!(!(o instanceof String value) || b)) {\n"
            + "            System.out.println(value);\n"
            + "        }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void conditionalOr2_2() {
    CompilationUnit cu = parse("public class Test {\n"
            + "    public void foo(Object o, boolean b) {\n"
            + "        if (!(b || !(o instanceof String value))) {\n"
            + "            System.out.println(value);\n"
            + "        }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void doStatements1() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        do {\n"
            + "        } while (!(o instanceof String value));\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void forStatements1() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        for (int i = 0; o instanceof String value; i += value.length()) { }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void forStatements2() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        for (; o instanceof String value; ) {\n"
            + "            System.out.println(value);\n"
            + "        }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void forStatements3() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        for (; !(o instanceof String value); ) {\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifStatements2() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        if (!(o instanceof String value)) {\n"
            + "            return;\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifStatements3() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        if (!(o instanceof String value)) {\n"
            + "            throw new RuntimeException();\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifStatements6() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        if (o instanceof String value) {\n"
            + "        } else {\n"
            + "            return;\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void ifStatements7() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        if (!(o instanceof String value)) {\n"
            + "            return;\n"
            + "        } else {\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");
    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void whileStatements1() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        while (o instanceof String value) {\n"
            + "            System.out.println(value);\n"
            + "        }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void whileStatements2() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        while (!(o instanceof String value)) {\n"
            + "        }\n"
            + "        System.out.println(value);\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

```
@Test
public void whileStatements3() {
    CompilationUnit cu = parse("class Test {\n"
            + "    public void test(Object o) {\n"
            + "        first:\n"
            + "        while (true) {\n"
            + "            while (!(o instanceof String value)) {\n"
            + "                break first;\n"
            + "            }\n"
            + "            System.out.println(value);\n"
            + "        }\n"
            + "    }\n"
            + "}");

    NameExpr name = Navigator.findNameExpression(cu, "value").get();
    assertEquals("java.lang.String", name.resolve().getType().describe());
}
```

johannescoetzee:
Fixes #4525 

I tried to stay as close as possible to the JLS for the changes in this PR since the rules regarding variables introduced by pattern expressions are very specific. For most of the methods I added, I added a comment quoting the rule from the JLS that is implemented by the specific method. If something doesn't make sense, feel free to ask and I'll try to explain (or correct it if I made a mistake) :)

# PatternVariableVisitor
The `PatternVariableVisitor` was added to replace the `*typePatternExprsExposedFromChildren` methods in `Context`. In my opinion this is a cleaner implementation for 2 reasons:
* It avoids having to create contexts for all the children as the tree is traversed downwards looking for patterns
* It keeps the logic for searching for patterns contained in a single file which makes it easier to read through and reason about.

# Context changes
A new `getIntroducedTypePatterns` method was added to `StatementContext` with implementations for `if`, `do`, `while`, and `for` statements. This method is used to find pattern expressions introduced by rules such as
```
A pattern variable is introduced by if (e) S iff (i) it is introduced by e when false and (ii) S cannot complete normally. 
```

When attempting to solve a type in context, one of the steps is now to search for pattern expressions exposed to the child in the parent. I implemented it this way instead of attempting to solve the symbol in the parent context directly since this avoids the problem mentioned in https://github.com/javaparser/javaparser/issues/4344, among others. This adds yet another path through which a symbol can be solved, which is not ideal, but fixing the issues with losing positional information/context history when attempting to solve symbols is outside the scope of this PR (although I'd be happy to take a look at some point since I think I understand the `javaparsermodel` process for solving symbols fairly well by now).

I've also cleaned up implementations for the `*typePatternExprsExposedFromChildren` methods and removed now-unnecessary helper methods.

# NormalCompletionVisitor
Some rules (such as that for if statements mentioned above) require knowledge about whether a statement can complete normally. Similar to the PatternVariableVisitor, I've added a separate visitor for this to keep the logic contained. I've implemented most of the rules described in the JLS, but notably did not implement the rules for loops correctly since this requires knowledge of whether the loop condition is a constant expression that evaluates to true/false. Since this PR was already getting quite big and this only affects an edge-case which I suspect is rather rare, I decided to put it aside for now with a comment of what needs to change. I'll come back to it at some point, but I think that's worth a separate change.

# Test changes
I added tests to test all of the rules for pattern variable scope mentioned in the JLS and also removed some old tests that were testing incorrect behaviour (see the linked issue for a comment describing some of these rules). I verified all of the tests that I've added with javac.

I added some negative tests of the form
```
        @Test
        public void conditionalOr1Negated() {
            CompilationUnit cu = parse("public class Test {\n"
                    + "    public void foo(Object o) {\n"
                    + "        if (o instanceof String value || value.length() > 5) { }\n"
                    + "    }\n"
                    + "}");

            NameExpr name = Navigator.findNameExpression(cu, "value").get();
            assertThrows(UnsolvedSymbolException.class, () -> name.resolve());
        }
```
Even though the code in this test doesn't compile, these tests still confirm that a pattern variable isn't incorrectly introduced into a given scope. As mentioned in the linked issue, one can always add a field `value` which would make this example compile and which would demonstrate how this could cause shadowing issues, but the tests as I wrote them are easier to verify with the compiler and test only how pattern variables are introduced into scope, instead of a combination of how the variables are introduced into scope and resolution of fields, for example.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/PatternVariableVisitor.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/PatternVariableVisitor.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright (C) 2015-2016 Federico Tomassetti
+ * Copyright (C) 2017-2024 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+package com.github.javaparser.symbolsolver.javaparsermodel;
+
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.*;
+import com.github.javaparser.ast.visitor.GenericVisitorWithDefaults;
+import java.util.*;
+
+public class PatternVariableVisitor extends GenericVisitorWithDefaults<PatternVariableResult, Void> {
+
+    private static PatternVariableVisitor patternVariableVisitor = null;
+
+    private PatternVariableVisitor() {}
+
+    public static PatternVariableVisitor getInstance() {
+        if (patternVariableVisitor == null) {
+            patternVariableVisitor = new PatternVariableVisitor();
+        }
+        return patternVariableVisitor;
+    }
+
+    @Override
+    public PatternVariableResult defaultAction(Node node, Void unused) {
+        return new PatternVariableResult();
+    }
+
+    @Override
+    public PatternVariableResult visit(BinaryExpr expression, Void unused) {
+        if (expression.getOperator().equals(BinaryExpr.Operator.AND)) {
+            return getVariablesIntroducedByAnd(expression);
+        }
+        if (expression.getOperator().equals(BinaryExpr.Operator.OR)) {
+            return getVariablesIntroducedByOr(expression);
+        }
+
+        return new PatternVariableResult();
+    }
+
+    /**
+     * The following rules apply to a conditional-and expression a && b:
+     * - A pattern variable is introduced by a && b when true iff either
+     *   (i) it is introduced by a when true or
+     *   (ii) it is introduced by b when true.
+     *
+     * It should be noted that there is no rule for introducing a pattern variable by a && b when false.
+     * This is because it cannot be determined at compile time which operand will evaluate to false.
+     *
+     * https://docs.oracle.com/javase/specs/jls/se22/html/jls-6.html#jls-6.3.1.1
+     */
+    private static PatternVariableResult getVariablesIntroducedByAnd(BinaryExpr expression) {
+        PatternVariableVisitor variableVisitor = PatternVariableVisitor.getInstance();
+        PatternVariableResult introducedByLeft = expression.getLeft().accept(variableVisitor, null);
+        PatternVariableResult introducedByRight = expression.getRight().accept(variableVisitor, null);
+
+        introducedByLeft.addVariablesIntroducedIfTrue(introducedByRight.getVariablesIntroducedIfTrue());
+        introducedByLeft.clearVariablesIntroducedIfFalse();
+
+        return introducedByLeft;
+    }
+
+    /**
+     * The following rules apply to a conditional-or expression a || b:
+     * - A pattern variable is introduced by a || b when false iff either
+     *   (i) it is introduced by a when false or
+     *   (ii) it is introduced by b when false.
+     *
+     * It should be noted that there is no rule for introducing a pattern variable by a || b when true.
+     * This is because it cannot be determined at compile time which operand will evaluate to true.
+     *
+     * https://docs.oracle.com/javase/specs/jls/se22/html/jls-6.html#jls-6.3.1.2
+     */
+    private static PatternVariableResult getVariablesIntroducedByOr(BinaryExpr expression) {
+        PatternVariableVisitor variableVisitor = PatternVariableVisitor.getInstance();
+        PatternVariableResult introducedByLeft = expression.getLeft().accept(variableVisitor, null);
+        PatternVariableResult introducedByRight = expression.getRight().accept(variableVisitor, null);
+
+        introducedByLeft.addVariablesIntroducedIfFalse(introducedByRight.getVariablesIntroducedIfFalse());
+        introducedByLeft.clearVariablesIntroducedIfTrue();
+
+        return introducedByLeft;
+    }
+
+    @Override
+    public PatternVariableResult visit(UnaryExpr expr, Void unused) {
+        if (expr.getOperator().equals(UnaryExpr.Operator.LOGICAL_COMPLEMENT)) {
+            return getVariablesIntroducedByLogicalComplement(expr);
+        }
+
+        return new PatternVariableResult();
+    }
+
+    /**
+     *  The following rules apply to a logical complement expression !a:
+     *  - A pattern variable is introduced by !a when true iff it is introduced by a when false.
+     *  - A pattern variable is introduced by !a when false iff it is introduced by a when true.
+     *
+     *  https://docs.oracle.com/javase/specs/jls/se22/html/jls-6.html#jls-6.3.1.3
+     */
+    private static PatternVariableResult getVariablesIntroducedByLogicalComplement(UnaryExpr unaryExpr) {
+        PatternVariableVisitor variableVisitor = PatternVariableVisitor.getInstance();
+        PatternVariableResult introducedByChild = unaryExpr.getExpression().accept(variableVisitor, null);
+
+        introducedByChild.swapTrueAndFalse();
+
+        return introducedByChild;
+    }
+
+    /**
+     * The following rule applies to an instanceof expression with a pattern operand, a instanceof p:
+     * - A pattern variable is introduced by a instanceof p when true iff the pattern p contains a declaration of the pattern variable.
+     *
+     * https://docs.oracle.com/javase/specs/jls/se22/html/jls-6.html#jls-6.3.1.5
+     */
+    public PatternVariableResult visit(InstanceOfExpr instanceOfExpr, Void unused) {
+        LinkedList<TypePatternExpr> variablesIntroducedIfTrue = new LinkedList<>();
+        LinkedList<TypePatternExpr> variablesIntroducedIfFalse = new LinkedList<>();
+
+        instanceOfExpr.getPattern().ifPresent(patternExpr -> {
+            Queue<PatternExpr> patternQueue = new ArrayDeque<>();
+            patternQueue.add(patternExpr);
+
+            while (!patternQueue.isEmpty()) {
+                PatternExpr toCheck = patternQueue.remove();
+                if (toCheck.isTypePatternExpr()) {
+                    variablesIntroducedIfTrue.add(toCheck.asTypePatternExpr());
+                } else if (toCheck.isRecordPatternExpr()) {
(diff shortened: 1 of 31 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/PatternVariableVisitor.java
```diff
@@ -0,0 +1,167 @@
...
+import com.github.javaparser.ast.visitor.GenericVisitorWithDefaults;
+import java.util.*;
+
+public class PatternVariableVisitor extends GenericVisitorWithDefaults<PatternVariableResult, Void> {
+
+    private static PatternVariableVisitor patternVariableVisitor = null;
+
+    private PatternVariableVisitor() {}
+
+    public static PatternVariableVisitor getInstance() {
```
jlerbsc: Why create a singleton?

jlerbsc:
It's a great job. Let me know when this PR is ready to be integrated. I don't have the knowledge to validate it fully, but if the tests pass and the existing system continues to work, I'm not opposed to integrating it.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/PatternVariableVisitor.java
johannescoetzee: My reasoning was that there's no reason not to do it since it doesn't hold any state that would cause issues and by having it be a singleton, we skip the cost of creating a new instance with every symbol resolution. It's largely a stylistic choice though since I don't think it would have a noticeable performance impact to create new instances when needed. I'd be happy to refactor this if it's inconsistent with the rest of the project

johannescoetzee:
If the singleton comment above isn't a blocker, then I think this should be ready to merge. The various paths symbol resolution can take get quite complex, so I can't say for sure that this won't include any regressions, but I did make sure to check that the existing tests that failed because of the behaviour change (which I then removed) were actually testing the incorrect behaviour. If any issues do come up that look like they may be related to this, please @ me in the issue and I'll take a look :)

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/PatternVariableVisitor.java
jlerbsc: There is no state in this visitor, but the cost of creating the visitor occasionally has no impact on performance. Nor is this a widespread pattern in JP. You can replace this pattern by creating an instance of the visitor.

johannescoetzee: Done!

PR #4540, opened 2024-08-21
Add better instructions for re-formatting the project

johannescoetzee:
The style check I've added has caused some confusion for 2 reasons:
1. What I've explicitly communicated so far is not consistent with what's actually being checked
2. The instructions I added to CONTRIBUTING.md are not sufficient to guarantee that a change will pass the style check.

Specifically, I mentioned in the original spotless PR that the check that is run first runs the code generators, then runs spotless. In later comments and in `CONTRIBUTING.md`, however, I said that running `./mvnw spotless:apply` would fix style errors. This was not correct and I did not consider that the code generators would introduce style changes that wouldn't be "fixed" by spotless. 

There have been a few cases where the generators have made whitespace changes that weren't changed by spotless, however, and the instructions I gave also don't give any guidance in cases where there are legitimate codegen issues (for example if someone implemented a method manually when it should've been generated), so I've updated the instructions in `CONTRIBUTING.md` to be much more clear (and correct) about what's required.

I also mentioned creating a custom maven command  (for example `./mvnw format`) to make this process simpler for contributors, but this would only work on systems with `bash` available and isn't much simpler than running the code generators in any case. I think just having better instructions available is a better solution.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -29,4 +29,51 @@ Our development workflow is based on Pull Request. If you are not familiar with 
 - Pull requests often stay open for at least a few days to give people a chance to review it.
 - A pull request is merged when all comments on it have been resolved.
 - If you create a pull request for an issue, mention the issue in the format #123 to make github link it automatically.
-- Before submitting a pull request, please run `./mvnw spotless:apply` to format your code to avoid any formatting related issues during review.
+- Before creating a commit (or at least before submitting a pull request), please reformat the project with the instructions given below
+  to avoid any formatting-related issues during the review.
+
+### Note on formatting the project:
+
+- If you are developing on a machine with bash installed, execute `./run_core_metamodel_generator.sh && ./run_core_generators.sh`. This
+  will re-run all of the code generators and then re-format the entire project as a final step. This ensures that:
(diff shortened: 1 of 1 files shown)


johannescoetzee:
I hope this addresses any confusion caused, for example in https://github.com/javaparser/javaparser/pull/4538 and https://github.com/javaparser/javaparser/pull/4527 (where I was caught by the same issue)

jlerbsc:
Thank you for these clarifications

ktul:
I just saw that applying spotless to my unit tests inlined lines that were intended for readability. See commit 770f766.

This changed
```java
Node root = parse("package com;"
                + "import com.*;"
                + "import org.*;"
                + "abstract class Foo {}");
```
to
```java
Node root = parse("package com;" + "import com.*;" + "import org.*;" + "abstract class Foo {}");
```
Had I seen the change earlier, I would have changed it to
```
Node root = parse("package com; import com.*; import org.*; abstract class Foo {}");
```

Similar changes happened in commit 5106428.

I couldn't find any options for palantir-java-format that could prevent this kind of inlining. Concatenating two string literals on the same line is also not done automatically.

I suggest mentioning in the instructions to check the changes after applying spotless.

Issue #4512, opened 2024-08-04
javaparser-core.jar is not reproducible due to JavaParserBuild


ebourg:
Hi,

The `JavaParserBuild` class captures several elements from the build environment that prevent the `javaparser-core.jar` file from being [reproducible](https://reproducible-builds.org/).

Since this class doesn't appear to be used, would it be possible to remove it, or at least remove the `MAVEN_*`, `JAVA_*` and `OS_*` fields?

jlerbsc:
The corresponding PR is merged.

PR #4582, opened 2024-10-11
Improve JavaParserBuild reproducibility by removing environment-specific fields

bfsn:
This PR addresses issue #4512.

Changes made:
- Removed MAVEN_*, JAVA_*, and OS_* fields from JavaParserBuild class
- Kept PROJECT_VERSION, PROJECT_NAME, and PROJECT_BUILD_FINAL_NAME fields

Rationale:
These changes improve the reproducibility of the javaparser-core.jar file by removing build environment-specific information. This ensures that the JAR file can be consistently reproduced across different environments.

Testing:
- Compiled the project successfully
- Ran all tests, which passed without any new failures
- Verified the contents of the generated JavaParserBuild.class file

Please review and let me know if any further changes are needed.


jlerbsc:
Thank you for your suggestion, but you don't demonstrate that these modifications allow you to build a reproducible build.

https://maven.apache.org/guides/mini/guide-reproducible-builds.html

bfsn:
> Thank you for your suggestion, but you don't demonstrate that these modifications allow you to build a reproducible build.
> 
> https://maven.apache.org/guides/mini/guide-reproducible-builds.html

Thank you for your feedback. You're right, I will work on building the project and attempting to prove reproducibility in the coming days. I appreciate your guidance on this matter.

jlerbsc:
Given that this modification does not disrupt the build process, we can try to implement it.

Issue #2432, opened 2019-11-26
LexicalPreservingPrinter always indent four spaces 


Implex1v:
Given this code
```java
public class Example1 {
  public void foo() {
    System.out.println("Hello World!");
  }
}
```
when adding an if statement into it, the indent of the if block is four spaces instead of two.
```java
public class Example1 {
  public void foo() {
    System.out.println("Hello World!");
    if (13 > 12) {
        System.out.println("!!");
    }
  }
}
```
Code for reproducing the output:
```java
String content = "public class Example1 {\n" +
        "  public void foo() {\n" +
        "    System.out.println(\"Hello World!\");\n" +
        "  }\n" +
        "}";
JavaParser javaParser = new JavaParser();
ParseResult<CompilationUnit> parseResult = javaParser.parse(content);
CompilationUnit compilationUnit = parseResult.getResult().get();
LexicalPreservingPrinter.setup(compilationUnit);
BlockStmt blockStatement = compilationUnit.findFirst(BlockStmt.class).get();

IfStmt ifStmt = new IfStmt();
ifStmt.setCondition(StaticJavaParser.parseExpression("13 > 12"));
ifStmt.setThenStmt(new BlockStmt().addStatement(StaticJavaParser.parseStatement("System.out.println(\"!!\");")));
blockStatement.addStatement(ifStmt);
System.out.println(LexicalPreservingPrinter.print(compilationUnit));
```

jlerbsc:
The representation of a new node is based on a model of the syntax of each node in the AST (concrete syntax model); this model determines in particular whether the representation in the form of a token should be indented or not. If indentation is to be generated, it is based on a default indentation (consisting of 4 spaces). This is how the LexicalPreservingPrinter works, as it was not designed to manage code indentation more finely. Changing this behaviour does not seem easy.

PR #4583, opened 2024-10-11
Fix: Issue 2432 Incorrect indentation

Universe-E:
## Description
Fix for issue #2432: Incorrect indentation when adding if statement using LexicalPreservingPrinter

This PR addresses the problem of incorrect indentation when adding an if statement to existing code using the LexicalPreservingPrinter. The fix ensures that the added if statement and its content maintain the same indentation style as the surrounding code.

## Changes made
- Modified the LexicalPreservingPrinter class to handle indentation for newly added nodes
- Updated the interpret method to use the existing indentation level when adding new statements

the project can be compiled normally after building:
image

For the expected code block:
```
public class Example1 {
  public void foo() {
    System.out.println("Hello World!");
    if (13 > 12) {
        System.out.println("!!");
    }
  }
}
```

I used the initial code block:
```
public class Example1 {
  public void foo() {
    System.out.println("Hello World!");
  }
}
```
And by adding the if code block through BlockStmt, the expected result is obtained, achieving the test purpose.

```
        ParseResult<CompilationUnit> parseResult = javaParser.parse(initialCode);
        CompilationUnit compilationUnit = parseResult.getResult().get();
        LexicalPreservingPrinter.setup(compilationUnit);

        BlockStmt blockStatement = compilationUnit.findFirst(BlockStmt.class).get();

        IfStmt ifStmt = new IfStmt();
        ifStmt.setCondition(StaticJavaParser.parseExpression("13 > 12"));
        ifStmt.setThenStmt(
                new BlockStmt().addStatement(StaticJavaParser.parseStatement("System.out.println(\"!!\");")));


        blockStatement.addStatement(ifStmt);
```

- Note:  The 4-space indentation in the "if code box" (as well as the "switch", "for" and other code boxes) was originally correct, but the issue proposer thought that the 4-space indentation was incorrect

## How to test
1. Run the Issue2432Test class
2. Verify that the test passes and the indentation of the added if statement matches the existing code style

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
@@ -636,12 +636,6 @@ public class LexicalPreservingPrinter {
                         indentation.add(new TokenTextElement(SPACE, " "));
                     }
                 }
-                else if (indexCurrentElement > 0 && TokenTypes.isEndOfLineToken(IF)
-                        && !(calculatedSyntaxModel.elements.get(indexCurrentElement - 1) instanceof CsmUnindent)) {
-                    for (int i = 0; i < existingIndentationLevel; i++) {
-                        indentation.add(new TokenTextElement(SPACE, " "));
-                    }
-                }
             } else if (element instanceof CsmUnindent) {
                 for (int i = 0; i < Difference.STANDARD_INDENTATION_SIZE && indentation.size() > 0; i++) {
                     indentation.remove(indentation.size() - 1);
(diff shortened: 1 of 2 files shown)


javaparser-core/src/main/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinter.java
```diff
@@ -629,6 +636,12 @@ private static NodeText interpret(Node node, CsmElement csm, NodeText nodeText)
                         indentation.add(new TokenTextElement(SPACE, " "));
                     }
                 }
+                else if (indexCurrentElement > 0 && TokenTypes.isEndOfLineToken(IF)
```
jlerbsc: Thank you for your contribution, but this section of code is too poorly documented for us to understand its intent and context.

Universe-E: OK thanks for your reminding and I will review this part of code soon

Universe-E:
Dear maintainer, I think the problem raised in this issue should be simplified, that is, the 4-space indentation in the "if code box" (as well as the "switch", "for" and other code boxes) was originally correct, but the issue proposer thought that the 4-space indentation was incorrect, so I think I should rollback the changes to LexicalPreservingPrinter.java, but I still keep the test class of Issue2432Test to enrich this "possible" indentation problem test

jlerbsc:
I'm closing this PR because the proposed correction is not quite clear enough.

PR #4587, opened 2024-10-20
Fixed return within void method

### LINKED ISSUE #2962: Double check test case - return within a void method
Follow up of comments (and a change to the test) in #2961 - https://github.com/javaparser/javaparser/pull/2961#discussion_r532243328

This test case has a `return` within a void method:

https://github.com/javaparser/javaparser/blob/master/javaparser-symbol-solver-testing/src/test/resources/LambdaVoid.java.txt#L6-L13

While I don't _think_ it affects the test, it is an example that does not compile - maybe the test should be edited?

https://github.com/javaparser/javaparser/blob/c64823aaaa4e2f7f0f6777f2e171280854955118/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/LambdaResolutionTest.java#L178-L192

Universe-E:
## Description

Fixes #2962 

Solved the problem of an incorrect return statement in a void method in the LambdaVoid.java sample file and updates the corresponding test case in LambdaResolutionTest.java.

## Changes

- Modified the LambdaVoid.java file to remove the incorrect return statement and use the correct lambda syntax
- Updated the typeOfVoidLambda test method in LambdaResolutionTest.java to correctly test the modified sample code
- Fixed line ending issues in test resources files by ensuring consistent LF line endings

## Results
- The project compiles normally after these changes
- The typeOfVoidLambda test passes and correctly identifies the lambda expression type as Consumer<String>
- All test resource files now have consistent line endings (LF)


jlerbsc:
You must use spotless:apply to resolve code formatting issues.

jlerbsc:

PR #4600, opened 2024-10-30
Fixes #4599 making B.class for testing non-empty

### LINKED ISSUE #4599: Empty file in javaparser-core-testing may cause false alarms when checking sources
I noticed, that the class https://github.com/javaparser/javaparser/blob/master/javaparser-core-testing/src/test/resources/com/github/javaparser/storage/B.java is empty. 

This fact might cause false alarms when automated tools are checking whether all sources are not empty.

I think that very simple solution would be to add content like `//empty`, in that way the functionality of `CompilationUnitTest` would stay the same, and the tools checking non-empty files would be satisfied too.

Let me provide a simple PR fixing it.

JiriOndrusek:
Fixes https://github.com/javaparser/javaparser/issues/4599


jlerbsc:
Thank you for your PR.

PR #4608, opened 2024-11-06
Fix issue #4607: don't forget to clone guard when cloning stmt.SwitchEntry

### LINKED ISSUE #4607: javaparser.ast.stmt.SwitchEntry.clone doesn't copy the guard expression
CloneVisitor here https://github.com/javaparser/javaparser/blob/master/javaparser-core/src/main/java/com/github/javaparser/ast/visitor/CloneVisitor.java#L892 clones the guard expression, but it doesn't insert it into the newly created SwitchEntry:
```
    public Visitable visit(final SwitchEntry n, final Object arg) {
        Expression guard = cloneNode(n.getGuard(), arg);
        NodeList<Expression> labels = cloneList(n.getLabels(), arg);
        NodeList<Statement> statements = cloneList(n.getStatements(), arg);
        Comment comment = cloneNode(n.getComment(), arg);
        SwitchEntry r = new SwitchEntry(n.getTokenRange().orElse(null), labels, n.getType(), statements, n.isDefault());
        r.setComment(comment);
        n.getOrphanComments().stream().map(Comment::clone).forEach(r::addOrphanComment);
        copyData(n, r);
        return r;
    }
```

DaniilSuchkov:
Fixes #4607.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
@@ -442,6 +442,7 @@ public final class JavaParserMetaModel {
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.typePropertyMetaModel);
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.statementsPropertyMetaModel);
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.isDefaultPropertyMetaModel);
+        switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.guardPropertyMetaModel);
         switchStmtMetaModel.getConstructorParameters().add(switchStmtMetaModel.selectorPropertyMetaModel);
         switchStmtMetaModel.getConstructorParameters().add(switchStmtMetaModel.entriesPropertyMetaModel);
         synchronizedStmtMetaModel.getConstructorParameters().add(synchronizedStmtMetaModel.expressionPropertyMetaModel);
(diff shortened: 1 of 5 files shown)


jlerbsc:
You must use mvn spotless:apply to format the code you commit.

DaniilSuchkov:
@jlerbsc thank you for pointing that out! Your suggestion made me realize that I need to have another look at CONTRIBUTING.md. There I saw the recommendation to run `./run_core_metamodel_generator.sh && ./run_core_generators.sh`, which I did. And it turned out, my original fix was in the wrong place: I was trying to fix the code which is autogenerated. After realizing that, I found out why it is generated incorrectly and fixed the root cause: the `@AllFieldsConstructor` for SwitchEntry wasn't actually initializing all the fields.

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
```diff
@@ -442,6 +442,7 @@ private static void initializeConstructorParameters() {
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.typePropertyMetaModel);
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.statementsPropertyMetaModel);
         switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.isDefaultPropertyMetaModel);
+        switchEntryMetaModel.getConstructorParameters().add(switchEntryMetaModel.guardPropertyMetaModel);
```
jlerbsc, line 445: Can you explain what you're trying to do?

jlerbsc:
@johannescoetzee Do you agree with these changes?

javaparser-core/src/main/java/com/github/javaparser/metamodel/JavaParserMetaModel.java
jlerbsc, line 445: Did you add this line manually?

johannescoetzee:
This looks good to me! Adding the `guard` parameter to the `@AllFieldsConstructor` and rerunning the code generators as @DaniilSuchkov did is the correct fix

Edit: The fact that the style check passed for the current iteration also means there are no manual changes in the PR that will be overwritten by the code generators

jlerbsc:

DaniilSuchkov:
Thank you for reviewing it so quickly!

PR #4611, opened 2024-11-13
Fix bug in VisitorSet.toString()

Laughh:
Hello, I find a bug in VisitorSet.toString() due to missing a space, and the last useful character will be removed, causing a wrong output.
- The right output: 
image

- The wrong ouput:
image

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/VisitorSetTest.java
@@ -136,8 +136,7 @@ class VisitorSetTest {
 
     @Test
     void visitorSetForType() {
-        Set<Type> set =
-                new VisitorSet<>(new ObjectIdentityHashCodeVisitor(), new ObjectIdentityEqualsVisitor());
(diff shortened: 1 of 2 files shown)


jlerbsc:
Can you add a unit test that demonstrates the problem and its resolution.

Laughh:
Sure, I added a test for VisitorSet with `Type`, and the original output of the test is `[void,Strin]`, which is wrong due to the the Line 157 in `VisitorSet.toString()`,  the first arg `sb.length() - 2` will delete the last useful character, and now I have fixed it by setting the separator as `', '`.

jlerbsc:
I don't think the correction is correct. It would have been better to replace the last comma with a square bracket. This is what the method tries to do when the set contains at least one object. You can propose a new PR that avoids adding elements to the standard output. You can also add tests for cases where there are no elements in the list, or a single element, or more than one element.

Laughh:
> I don't think the correction is correct. It would have been better to replace the last comma with a square bracket. This is what the method tries to do when the set contains at least one object. You can propose a new PR that avoids adding elements to the standard output. You can also add tests for cases where there are no elements in the list, or a single element, or more than one element.

Okay, I will open a new PR to fix it, thank you.

PR #4615, opened 2024-11-18
Fix bug in `VisitorSet.toString()`

Laughh:
Hi, I haved closed previous PR and open a new one, and the case of no element, one element, and multiple elements is considered, and it's also supplemented with tests.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/VisitorSet.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/VisitorSet.java
@@ -149,12 +149,9 @@ public class VisitorSet<N extends Node> implements Set<N> {
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder("[");
-        if (size() == 0) return sb.append("]").toString();
-        for (EqualsHashcodeOverridingFacade facade : innerSet) {
-            sb.append(facade.overridden.toString() + ",");
-        }
-        return sb.replace(sb.length() - 2, sb.length(), "]").toString();
+        return innerSet.stream()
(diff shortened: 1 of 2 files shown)


Laughh:
BTW, I think that the separator `, ` will be more intuitive than single comma, and the thought of this PR is same as the closed one. In this PR,  I use `stream` to make the code more concise.

javaparser-core/src/main/java/com/github/javaparser/utils/VisitorSet.java
```diff
@@ -149,12 +149,9 @@ public <T> T[] toArray(T[] arr) {
...
     public String toString() {
-        StringBuilder sb = new StringBuilder("[");
-        if (size() == 0) return sb.append("]").toString();
-        for (EqualsHashcodeOverridingFacade facade : innerSet) {
-            sb.append(facade.overridden.toString() + ",");
-        }
-        return sb.replace(sb.length() - 2, sb.length(), "]").toString();
+        return innerSet.stream()
+                .map(facade -> facade.overridden.toString())
+                .collect(Collectors.joining(", ", "[", "]"));
```
jlerbsc: I can't accept this PR, even though it undoubtedly improves the code, because you're changing the format of the character string returned by this method, which may have an impact on other users. Please only correct the issue you have identified.

Laughh: Sure, I have solved it and the format is the same as before.

jlerbsc:

Laughh:
> Thank you for this contribution

Thanks for your review.

PR #4634, opened 2024-12-15
Fixed #3674

### LINKED ISSUE #3674: Solving incorrectly a NameExpr Symbol when there is a variable declared after with same name
I'm using the "com.github.javaparser:javaparser-symbol-solver-core:3.24.4" release.

Considering de following class:

```
class Class {

    int attribute;

    private int method() {
        int value = attribute;
        double attribute = 0.0;
        return (int) (attribute + 1);
    }
}
```
When resolving the line:
`int value = attribute;`

Setting the JavaParser with the solver as shown in the following code (Kotlin):

`StaticJavaParser.setConfiguration(ParserConfiguration().setSymbolResolver(JavaSymbolSolver(CombinedTypeSolver())))`
The `NameExpr <attribute>` provides the VariableDeclarationExpr `<double attribute = 0.0>` instead of the `FieldDeclaration <int attribute;>`.

Am I doing anything wrong or is it just a bug?

nzdev:
Solves for the issue #3674 described below where `attribute ` in  `int value = attribute;` should resolve to the field `int attribute;
`
Stores the current statement when resolving via it's parent Block Statement and then limits variable declarations to consider as matches from within the Block Statement to only Variables declared before the stored statement.

Considering the following class:

```
class Class {

    int attribute;

    private int method() {
        int value = attribute;
        double attribute = 0.0;
        return (int) (attribute + 1);
    }
}
```
When resolving the line:
int value = attribute;

Setting the JavaParser with the solver as shown in the following code (Kotlin):

StaticJavaParser.setConfiguration(ParserConfiguration().setSymbolResolver(JavaSymbolSolver(CombinedTypeSolver())))
The NameExpr <attribute> provides the VariableDeclarationExpr <double attribute = 0.0> instead of the FieldDeclaration <int attribute;>.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -171,6 +171,12 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
                     throw new IllegalStateException("Unexpectedly more than one reference in scope");
             }
         }
+        if(parentContext instanceof BlockStmtContext) {
+        	// if parent context is BlockStmtContext, when we call parentContext.solveSymbol(name) we lose the position within the block
+        	// so we pass the current context to the block context
+        	BlockStmtContext context = (BlockStmtContext)parentContext;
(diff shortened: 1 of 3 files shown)


jlerbsc:
Thank you for taking the time to suggest this correction. There is indeed a loss of information about the AST node which makes it impossible to locate in the code, but your solution is not the one we're going to favour. The solution we would like to see would be to modify the method signature by replacing the string with the name expression.

PR #4650, opened 2025-01-12
Reorganize jacoco 

verhasi:
Opening the project in IntelliJ IDEA I was unable to build it. It complained about the generated sources copied into the test modules. I realized that the copy was added as a workaround for the jacoco configuration. This PR gives a solution instead of a workaround. As the project uses CodeCov as the default tool for managing code coverage there is no need to generate the jacoco report, only the measurement result is enough to work. On the other hand, I introduced a new maven module javaparser-jacoco-report in case someone needs the jacoco report. Along this approach there is one enhancement in the jacoco reports aka the source code is also available in the view. Also added the run script to execute it easily.
Some other minor refactors were added to the poms to simplify them.

### DIFF
--- a/javaparser-core-metamodel-generator/pom.xml
+++ b/javaparser-core-metamodel-generator/pom.xml
@@ -28,6 +28,13 @@
                     <skip>true</skip>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skipTests>true</skipTests>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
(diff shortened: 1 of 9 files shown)


jlerbsc:
Please create a separate PR for each subject

verhasi:
@jlerbsc Thank you for your input. To make it clear, you mean separate PR for the following subjects:
- Jacoco reorganization
- pom refactoring
?

jlerbsc:
> On the other hand, I introduced a new maven module javaparser-jacoco-report in case someone needs the jacoco report.

There is no point in creating a new module for this.

jlerbsc:
> Some other minor refactors were added to the poms to simplify them.

Why not, if it really adds value. I don't know what the improvements are because there are no more specific comments on the PR.

jlerbsc:
> I realized that the copy was added as a workaround for the jacoco configuration.

I don't know the background to this choice but perhaps it was put in place to get feedback on test coverage before pushing a PR. I'm not against doing away with it if it really does simplify things and if it's not really being used.

jlerbsc:
As a general rule, I prefer to open an issue that details the problems encountered before proposing a solution. And the proposed solution must detail the motivations and choices made.

verhasi:
Thank you for your support. I close this PR now and going to create the separate issues and PRs respectively.

PR #4658, opened 2025-01-18
#4653 use report-aggregate of jacoco instead of report to use the dep…

### LINKED ISSUE #4653: Unable to build in IntelliJ IDEA
Opening the project in IntelliJ IDEA I was unable to build it. It complained about the generated sources copied into the test modules. I realized that the copy was added as a workaround for the jacoco configuration. I am going to create a PR that gives a solution instead of a workaround. 
I have seen that the project coverage results are split in the CodeCov into core and symbol solver. Is it a real requirement or was introduced as a result of technical limitations? 
The project uses CodeCov as the default tool for managing code coverage. This tool works from the XML format of the jacoco report result. We can also generate the HTML format of the report to see the coverage before pushing the modifications.

verhasi:
- use `report-aggregate` of jacoco instead of `report` to use the dependency modules source codes. 
- Remove the unnecessary resource copy from the poms. 
- Move the jacoco configuration to the plugin management in the parent pom.

Fixes #4653

### DIFF
--- a/javaparser-core-testing-bdd/pom.xml
+++ b/javaparser-core-testing-bdd/pom.xml
@@ -28,60 +28,6 @@
             <plugin>
                 <groupId>org.jacoco</groupId>
                 <artifactId>jacoco-maven-plugin</artifactId>
-                <executions>
-                    <execution>
-                        <id>jacoco-initialize</id>
-                        <goals>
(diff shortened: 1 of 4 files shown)


jlerbsc:
Thank you fir your contribution.

PR #4666, opened 2025-01-26
#4664 remove misleading javadoc

### LINKED ISSUE #4664: Misleading Javadoc in com.github.javaparser.JavaParser.java
The Javadoc states that the parse method `throws ParseProblemException` but this is not true. This exception is thrown by the `JavaParserAdapter` and the `StaticJavaParser`.

```
    /**
     * Parses the Java code contained in code and returns a
     * {@link CompilationUnit} that represents it.
     *
     * @param code Java source code
     * @return CompilationUnit representing the Java source code
     * @throws ParseProblemException if the source code has parser errors
     */
    public ParseResult<CompilationUnit> parse(String code) {
        return parse(COMPILATION_UNIT, provider(code));
    }
```

The same misleading Javadoc can be found in all the overloaded parse methods.

verhasi:
Fixes #4664. It is time to remove the misleading Javadoc after a decade.


jlerbsc:

PR #4679, opened 2025-02-14
Fix added switch entries printed by LPP

### LINKED ISSUE #4646: Adding new `SwitchEntry` to `SwitchExpr` resultis in illeagal code
Somewhat related to #4104 I think (?), was unsure if it was best to add a new issue, or comment on that issue.

if I add a new `SwitchEntry` it generates illegal code, as it places the semicolons in the wrong place. What I am trying to do, is add a 2024 entry to a switch like this:

```
        return switch (year) {
            case 2023 -> new PmV2023();
            default -> throw new IllegalStateException("Cant create for year");
        };

```

with code like this:

```
                            NodeList<SwitchEntry> entries = switchExpr.getEntries();
                            NodeList<Expression> labels = NodeList.nodeList(new IntegerLiteralExpr("2024"));

                            interfaceDeclaration.tryAddImportToParentCompilationUnit(klasseForNyttÅr);
                            ClassOrInterfaceType type = StaticJavaParser.parseClassOrInterfaceType(
                                klasseForNyttÅr.getSimpleName()
                            );
                            NodeList<Statement> expression = NodeList.nodeList(
                                new ExpressionStmt(
                                    new ObjectCreationExpr(
                                        null,
                                        type,
                                        entries
                                            .get(entries.size() - 2)
                                            .getStatements()
                                            .get(0)
                                            .asExpressionStmt()
                                            .getExpression()
                                            .asObjectCreationExpr()
                                            .getArguments()
                                    )
                                )
                            );
                            SwitchEntry newEntry = new SwitchEntry(labels, SwitchEntry.Type.EXPRESSION, expression);
                            entries.add(entries.size() - 1, newEntry);
```

It ends up producing something like this

```
        return switch (year) {
            case 2023 -> new PmV2023()case 2024 ->
                new PmV2024();
            ;
            default -> throw new IllegalStateException("Cant create for year");
        };
```

So it ends up adding it between the previous `SwitchEntry` and the semicolon?

johannescoetzee:
Fixes #4646.

This PR includes 2 changes to fix the token range of switch expression entries. Before, the token range of the `ExpressionStmt` (which is added to the switch entry) was set to the token range of the `Expression` contained in it. This was incorrect, since the token range for the `ExpressionStmt` should include the `;` at the end. This could be fixed in the `SwitchEntry` production/method, but I decided to move it into a separate method is a cleaner solution overall.

After fixing the token range, the LPP output was better, but still formatted strangely:
```
case 2023 -> new Object();
case 2024 ->
    new java.lang.Object();
```
This was because of shared logic between the older switch statement entries with `:` and the newer switch expression entries in the `ConcreteSyntaxModel`. To fix this, I effectively moved a shared `newline()` after the `:/->` to only affect the switch statement entries and added a space after the `->` for switch expressions. I also refactored everything a bit to avoid duplicating the `space(), token(ARROW), space()` at the end of each `->` branch and the `token(COLON), newline()` at the end of each `:` branch.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
@@ -22,13 +22,19 @@
 package com.github.javaparser.printer.lexicalpreservation.transformations.ast.body;
 
 import static com.github.javaparser.StaticJavaParser.parseStatement;
+import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.Range;
 import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.NodeList;
-import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.*;
+import com.github.javaparser.ast.stmt.ExpressionStmt;
 import com.github.javaparser.ast.stmt.Statement;
 import com.github.javaparser.ast.stmt.SwitchEntry;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
 import com.github.javaparser.printer.lexicalpreservation.AbstractLexicalPreservingTest;
 import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;
 import org.junit.jupiter.api.*;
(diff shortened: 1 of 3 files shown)


javaparser-core/src/main/javacc/java.jj
```diff
@@ -4744,6 +4740,36 @@ SwitchEntry SwitchEntry():
...
 
+/**
+* This method provides special handling for the case
+*
+*<pre>{@code
+*     SwitchRule:
+*         SwitchLabel -> Expression ;
+* }</pre}
+*
+* for switch entries in the JLS. The expression in the entry body is
```
jlerbsc, line 4751: Can't this be done in a Validator?

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
```diff
@@ -127,4 +132,57 @@ void switchWithRecordPatternPreserved() {
...
+
+        NodeList<Statement> expressions = NodeList.nodeList(stmt);
+
+        SwitchEntry newEntry = new SwitchEntry(labels, SwitchEntry.Type.EXPRESSION, expressions);
+
+        entries.add(entries.size() - 1, newEntry);
+
+        String output = LexicalPreservingPrinter.print(switchExpr);
+
+        assertEquals(
```
jlerbsc: You can replace assertEquals by TestUtils.assertEqualsStringIgnoringEol

javaparser-core/src/main/javacc/java.jj
johannescoetzee: The problem isn't that the input code is incorrect, but rather that the JavaParser representation for `SwitchEntry` is (slightly) incorrect. JP Assumes that the body of a `SwitchEntry` will be a list of statements which was true before switch expressions were added, but is no longer the case for this rule. 

The workaround for this (I assume) was to wrap the expression in an `ExpressionStmt` node which works well, but is not entirely correct according to the JLS since the `ExpressionStmt` in this specific case can contain any expression, not just those which are legal expression statements according to the JLS, for example below (a lambda is not a valid expression statement as far as I can tell, but the below snippet is still legal Java code):
```
        return switch (o) {
            case String s -> (arg) -> System.out.println(arg + s);
            case null, default -> (arg) -> {};
        };
```

If I understood the situation correctly, then this means that this case does require some special handling, because JP is breaking the rules slightly.

javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/transformations/ast/body/StatementTransformationsTest.java
johannescoetzee: I've updated the PR with this suggestion

jlerbsc:
Once again, thank you for your help.

PR #4704, opened 2025-04-02
Fix for #3710 by cutting off resolution loops involving object creation steps.

### LINKED ISSUE #3710: Inner class instantiation statement cause StackOverflowError when trying to resolve method call to the declaration
Consider the following example:

```java
public class JavaParserStackOverflowErrorTest {

    public static Object test(SomeType value) {
        return value.getFoo().getBar().new SomeAnotherType();
    }

}
```

If we try to parse above source with the JavaParser 3.24.4 (latest available version on Maven Central at the time of writing) and then resolve method call chain `value.getFoo().getBar()` to the declaration with the `MethodCallExpr.resolve()` call we will got `StackOverflowError` instead of expected `UnsolvedSymbolException`.

However, if we drop the `.new SomeAnotherType()` part from the source, then we will got `UnsolvedSymbolException` as expected.

Similar result with `com.github.javaparser.ast.expr.Expression::calculateResolvedType()` invocation on the same method call chain.

Quick dig show that *without* inner class instantiation statement type resolution chain will be just

```text
getType on method call value.getFoo()
getType on name expr value
```

In contast, inner class instantiation statement cause infinite loop:

```text
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
getType on method call value.getFoo()
getType on method call value.getFoo().getBar()
...
```

Here is the reproducer project: https://github.com/Maccimo/JavaParserSOE
Complete source of the above sample for reference: https://github.com/Maccimo/JavaParserSOE/blob/master/src/test/resources/JavaParserStackOverflowErrorTest.java

johanneskloos:
Fixes #3710.
The idea of the fix is that the scope in inner class object creation expressions, e.g., `makeOuter().new Inner()`, can cause resolution loops in just the same way that the scope of a field access or method call can, and should be handled in the same way.

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -117,6 +117,12 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
         Node notMethodNode = parentNode;
         // To avoid loops JP must ensure that the scope of the parent context
         // is not the same as the current node.
+        // For most part, this can be achieved that the scope of the nodes is different,
+        // but in some cases, we may have loops of length > 1. This is the case for expressions
+        // that have something like a "receiver" - field accesses, method calls and the
+        // non-static inner class variant of constructor calls. We handle these by just
+        // skipping all method calls, field accesses, and all constructor calls that have
+        // a receiver (i.e., outer.new Inner()), as identified by hasScope.
         while (notMethodNode instanceof MethodCallExpr
                 || notMethodNode instanceof FieldAccessExpr
                 || (notMethodNode instanceof ObjectCreationExpr && notMethodNode.hasScope())
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -119,6 +119,7 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
         // is not the same as the current node.
         while (notMethodNode instanceof MethodCallExpr
                 || notMethodNode instanceof FieldAccessExpr
+                || (notMethodNode instanceof ObjectCreationExpr && notMethodNode.hasScope())
                 || (notMethodNode != null
                         && notMethodNode.hasScope()
                         && getScope(notMethodNode).equals(wrappedNode))) {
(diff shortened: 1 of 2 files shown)


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
```diff
@@ -119,6 +119,7 @@ public final Optional<Context> getParent() {
         // is not the same as the current node.
         while (notMethodNode instanceof MethodCallExpr
                 || notMethodNode instanceof FieldAccessExpr
+                || (notMethodNode instanceof ObjectCreationExpr && notMethodNode.hasScope())
```
jlerbsc, line 128: I'm not sure that the expression new A().new B() can be solved correctly.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3710Test.java
```diff
@@ -0,0 +1,43 @@
...
+        CompilationUnit compilationUnit = parser.parse(sourceCode).getResult().get();
+        Optional<MethodCallExpr> methodCall = compilationUnit.findFirst(MethodCallExpr.class);
+
+        ResolvedMethodDeclaration resolvedMethod = methodCall.get().resolve();
+
+        Assertions.assertEquals(
+                "com.github.javaparser.symbolsolver.Outer.make()", resolvedMethod.getQualifiedSignature());
+    }
+}
+
```
jlerbsc, line 40: Please include these classes in the source code to be parsed.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
jlerbsc: Please add a comment explaining simply and clearly why this has been added

johanneskloos: I have added a comment, and a unit test for the case you mentioned.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3710Test.java
johanneskloos: done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
johanneskloos: In fact, I added two unit tests :)

jlerbsc:

PR #4709, opened 2025-04-03
Implement MethodResolutionCapability in JavassistRecordDeclaration

johanneskloos:
Add an implementation of `MethodResolutionCapability` to `JavassistRecordDeclaration`, directly based on that from
`JavasassistClassDeclaration`. Comes with a unit test.


jlerbsc:
That looks fine to me. Thank you for your contribution.

PR #4717, opened 2025-04-09
Generalize DefaultPrettyPrinterVisitor

johanneskloos:
Make the code in `DefaultPrettyPrinterVisitor` a bit easier to extend.

- Add Javadoc to the various helper methods.
- Make most of the helper methods `protected` so that they can be used from overriding classes.
- Add `preVisit`/`postVisit` callbacks that can be used to gather information when nodes are about to be printed, or have just been printed.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrettyPrinterVisitor.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/DefaultPrettyPrinterVisitor.java
@@ -1846,7 +1846,7 @@ public class DefaultPrettyPrinterVisitor implements VoidVisitor<Void> {
         printer.print("???;");
     }
 
-    private void printImports(NodeList<ImportDeclaration> imports, Void arg) {
+    protected void printImports(NodeList<ImportDeclaration> imports, Void arg) {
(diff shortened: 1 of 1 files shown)


jlerbsc:
Thank you for this PR. It would be easier to merge the changes if you avoided mixing several subjects in a single PR.

In fact, the preVisit/postVisit callbacks features has never been requested by the user community. So why should we include this type of evolution in the default implementation. You can integrate this using the composition pattern in your own visitor.

johanneskloos:
> Thank you for this PR. It would be easier to merge the changes if you avoided mixing several subjects in a single PR.

Will do, creating two PRs for the first two commits then.

> 
> In fact, the preVisit/postVisit callbacks features has never been requested by the user community. So why should we include this type of evolution in the default implementation. You can integrate this using the composition pattern in your own visitor.

Hm, I'll try and see if I can easily implement this in my own code.

johanneskloos:
Dropped the first commit (now in PR #4718) and the third commit.

jlerbsc:
Please create a new PR to integrate new modifications. This PR cannot be integrated in this state.

PR #4721, opened 2025-04-10
Fix range for cast expression with lambda child

### LINKED ISSUE #4488: Lexical parsing on a lambda with type fails to replace
Hi,

I have encountered the following which I think is a bug:

When using LexicalPreservation, doing changes inside the following lambda does not succeed:

```java
(Function<String, String>) s -> SomeFunction.withAMethodHere("test").build()
```

However, if I remove the `(Function<String, String>)` part, it behaves as expected.

Here is a test case showing what I mean:
```java
	@Test
	void cannotChangeMethodNameInLambda() {
		ParserConfiguration parserConfiguration = new ParserConfiguration();
		parserConfiguration.setLexicalPreservationEnabled(true);
		StaticJavaParser.setConfiguration(parserConfiguration);

		CompilationUnit cu = StaticJavaParser.parse("""
			class Test {
				private Map<String, String> dummyMap = new HashMap<>();
			
				public String dummyFunction(String name) {
					return dummyMap.computeIfAbsent(name,
						(Function<String, String>) s -> SomeFunction.withAMethodHere("test").build());
				}
			}
			""");

		cu.accept(new ModifierVisitor<>() {
			@Override
			public Visitable visit(MethodCallExpr mc, Object arg) {
				if (mc.getNameAsString().equals("withAMethodHere")) {
					return mc.setName("replacedMethodHere");
				}
				return super.visit(mc, arg);
			}
		}, null);

		assertEquals("""
		class Test {
			private Map<String, String> dummyMap = new HashMap<>();
		
			public String dummyFunction(String name) {
				return dummyMap.computeIfAbsent(name,
					(Function<String, String>) s -> SomeFunction.replacedMethodHere("test").build());
			}
		}
		""", LexicalPreservingPrinter.print(cu));
	}
```

The above test fails (the method name is not replaced). However, as I stated above, if I remove the `(Function<String, String>)` part, the test will pass as expected.

This is in the latest version of the lib (`3.26.1` as of now).

johannescoetzee:
Fixes #4488.

The cause of the issue is that lambda expressions are created in 2 steps (seemingly to avoid ambiguity issues). First, only a node for the lambda parameters is created (which could be a `NameExpr` or a `LambdaExpr` with only the params set) and then later in parsing the full `LambdaExpr` is created or the body is added to the `LambdaExpr`, depending on the context. The problem is that, for cast expressions, the range of the cast expression wasn't correctly adjusted.

In the recursive `generateLambda` calls, `propagateRangeGrowthOnRight` is called to update the range of the lambda expression and all ancestor nodes, but this is called before the lambda expression `inner` is added as a child to the `CastExpr` node, so the range of the cast was never updated. This PR fixes that by explicitly calling `propagateRangeGrowthOnRight` after the lambda is added to the cast.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue4488Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue4488Test.java
@@ -1,0 +1,49 @@
+package com.github.javaparser.printer.lexicalpreservation;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.visitor.ModifierVisitor;
+import com.github.javaparser.ast.visitor.Visitable;
+import org.junit.jupiter.api.Test;
+
+public class Issue4488Test {
+    @Test
+    void cannotChangeMethodNameInLambda() {
+        ParserConfiguration parserConfiguration = new ParserConfiguration();
+        parserConfiguration.setLexicalPreservationEnabled(true);
+        StaticJavaParser.setConfiguration(parserConfiguration);
+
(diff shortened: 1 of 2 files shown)


jlerbsc:
Perfect.

PR #4729, opened 2025-04-12
Make some helper methods protected in DefaultPrettyPrinterVisitor

johanneskloos:
This PR has been spun off from the closed #4717.

Making these helper methods protected makes it much easier to override visitor methods in customer pretty-printer visitors.


jlerbsc:
I have no objection to integrate these changes, even if the methods concerned are not intended to be overloaded.

PR #4752, opened 2025-05-12
Fix resolution of lambdas used as varargs

johannescoetzee:
This PR fixes an issue I encountered when attempting to resolve lambdas that were used as varargs, for example the second argument in
```
class Test {
    void acceptConsumers(Consumer<String>... consumers) {}
    void test(Consumer<String> first) {
        acceptConsumers(first, s -> System.out.println(s));
    }
}
```

At first this resulted in an `IndexOutOfBounds` crash when trying to get the type of the parameter with index 1. The fix for this was to make `MethodResolutionLogic.getMethodsExplicitAndVariadicParameterType` public and to use that to get the type (since this accounts for varargs already) and to add a `getMethodUsageExplicitAndVariadicParameterType` to do the same for `MethodUsage`. 

This fixed the `IndexOutOfBounds` crash, but resulted in a new issue where `acceptConsumers` couldn't be resolved. This is because, when getting the expected parameter type from the parent (in this case the `acceptConsumers` call), the returned type is an array of `Consumer` due to the varargs. It's not possible for a lambda to define an array type, however, so the resolution fails. This broke the rest of the lambda resolution logic, so to fix it the type is now unwrapped and the component type is used instead.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -741,14 +741,24 @@ public class MethodResolutionLogic {
         return true;
     }
 
-    private static ResolvedType getMethodsExplicitAndVariadicParameterType(
-            ResolvedMethodLikeDeclaration method, int i) {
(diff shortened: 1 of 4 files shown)


jlerbsc:
That sounds fine to me. Thank you for your PR.

PR #4754, opened 2025-05-15
Fix IndexOutOfBoundsException resulting from empty varargs call as method usage

johannescoetzee:
This PR fixes a crash due to `java.lang.IndexOutOfBoundsException: Index -1 out of bounds for length 0` when attempting to resolve a call to a method with only a single variadic parameter while providing no arguments, for example the `foo` call in:
```
public class Test {
  static String foo(String... values) { return null; }
  static void id(String value) { return null; }

  void test() {
    id(foo());
  }
}
```

This crash only occurs when attempting to resolve it via the `MethodUsage` code path, which is why the `foo` call is used as an argument to the `id` call. For the unit test, including all the code in the test code string didn't trigger this path either, so I added a test jar to be able to test this properly, along with the source and class file used to create the jar.

The fix is to check whether the argument list is empty (and return if it is) before attempting to match/replace type parameters using the actual argument types.


javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/MethodCallExprContext.java
```diff
@@ -305,16 +305,21 @@ private MethodUsage resolveMethodTypeParameters(MethodUsage methodUsage, List<Re
...
-                if (!expectedType.isAssignableBy(actualType)) {
-                    throw new UnsupportedOperationException(String.format(
-                            "Unable to resolve the type typeParametersValues in a MethodUsage. Expected type: %s, Actual type: %s. Method Declaration: %s. MethodUsage: %s",
-                            expectedType, actualType, methodUsage.getDeclaration(), methodUsage));
+                // ResolvedType actualType = new ResolvedArrayType(actualParamTypes.get(actualParamTypes.size() - 1));
+                // It is possible that the method is called with no arguments, in which case we can't get any further
+                // type information about the parameters
+                if (actualParamTypes.isEmpty()) {
+                    return methodUsage;
+                } else {
```
jlerbsc: Is the else statement mandatory?

johannescoetzee: It is not. I've updated the PR to remove it.

jlerbsc:
Thank you for this PR.

PR #4757, opened 2025-05-16
Use lambda parameter counts and block bodies for improved resolution

johannescoetzee:
This PR improves lambda-related method resolution by using the lambda parameter count and body for disambiguation in two cases described below. I implemented this by adding the required information to the `LambdaArgumentTypePlaceholder` class, but made sure that, if this information is not provided (such as for `MethodReferenceExpr`, type resolution behaviour will be the same as before this PR.

## Different parameter counts
```

class Test {
    void foo(Consumer<String> consumer) {}
    void foo(Runnable r) {}

    void test() {
        foo(input -> {});
    }
}
```
`Consumer.accept` has one parameter while `Runnable.run` has zero, so since the lambda has one parameter, we know this must resolve to `foo(Consumer<String>)`

## Disambiguation by return type
```
import java.util.function.Function;

class Test {
    void foo(Consumer<String> consumer) {}
    void foo(Function<Integer, String> func) {}

    void test() {
        foo(input -> {});
    }
}
```
In this example, the body of the lambda is a block statement which does not contain any return statements. This means that the lambda can only define a method with a `void` return type, so it must define `Consumer<String>` (the same would be true for `foo(input -> { return; })`.

If it were `foo(input -> { return ""; })` instead, then the reverse logic would apply. Since the return type of the lambda is not `void`, it cannot define `Consumer` and must therefore define `Function`.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -24,6 +24,7 @@ import com.github.javaparser.resolution.MethodAmbiguityException;
 import com.github.javaparser.resolution.MethodUsage;
 import com.github.javaparser.resolution.TypeSolver;
 import com.github.javaparser.resolution.declarations.*;
+import com.github.javaparser.resolution.model.LambdaArgumentTypePlaceholder;
 import com.github.javaparser.resolution.model.SymbolReference;
 import com.github.javaparser.resolution.model.typesystem.ReferenceTypeImpl;
 import com.github.javaparser.resolution.types.*;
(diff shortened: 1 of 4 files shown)


johannescoetzee:
I don't understand why the codecov/patch check is failing. The tests are covered, so the logic it's complaining about must be too, otherwise the tests don't pass

johannescoetzee:
This PR might introduce a new crash. I've changed it to a draft while I look into this, so this shouldn't be merged until I've either fixed the new crash or it turns out to be something else.

johannescoetzee:
I've found the source of the new crash and added the `lambdaUsedAsPolymorphicArgument` to test it. The issue was that by the time resolution got to the `isConflictingLambdaType` method, type variables have not been substituted yet, so `LambdaInterfaceLogic.getFunctionalMethod` crashed with an `UnsupportedOperationException` when attempting to find the functional interface for the type variable. I've fixed this by simply not rejecting the candidate in this case (which would be the behaviour before this PR), but this does still result in a `MethodAmbiguityException` in some cases. I've added a disabled test ` lambdaUsedAsOverloadedPolymorphicArgument1` to illustrate this. In my opinion, handling this case as well would be better left for a separate PR as it would require a deeper dive into the whole resolution process to avoid doing something that would result in duplicate solving tasks.

johannescoetzee:
I found another deeper issue that comes up when trying to find the corresponding functional interface, so I'm going to close this PR until I've fixed this one and am confident that there aren't any more

PR #4789, opened 2025-07-04
Fix javadoc of Name class

bannmann:
I could not find the mentioned `name(...)` method in this class. After a while, I found `StaticJavaParser.parseName()` and I guess that is the one that should be linked here.


jlerbsc:
It seems to me that this refers to an old method that no longer exists, but I don't think it's necessary to introduce a reference to the StaticJavaParser class. I'd opt for the Name class constructor instead.

bannmann:
> It seems to me that this refers to an old method that no longer exists, but I don't think it's necessary to introduce a reference to the StaticJavaParser class. I'd opt for the Name class constructor instead.

I was looking for a constructor that accepts one fully qualified name string and did not find one. So I was actually glad that the comment hinted at the existence of a helper method, even though I still had to find it.

So my suggestion is to either add that link or to move `StaticJavaParser.parseName()` to the `Name` class (and having the old method call the new one).

PR #4790, opened 2025-07-04
Bugfix: setIdentifier() etc could not reliably detect unchanged arguments

bannmann:
These methods seemed to assume that in order to detect an unchanged argument (and to skip property change notification), identity comparison is appropriate. But this [would only work in rare circumstances](https://stackoverflow.com/a/32926523/7641). See also [SonarQube rule 4973](https://rules.sonarsource.com/java/type/Bug/RSPEC-4973).


jlerbsc:
This is generated code that cannot be modified in this way.

bannmann:
I see, thanks. Will rework this.

jlerbsc:
I have created issue https://github.com/javaparser/javaparser/issues/4791 to reference this bug.

Issue #4805, opened 2025-08-04
XmlPrinter: Generated data throws exception: Attribute 'type' redefined.


sgqy:
Version: `com.github.javaparser:javaparser-symbol-solver-core:3.27.0`

Convert java source to xml:

```java
XmlPrinter printer = new XmlPrinter(true);
return printer.output(compileUnit);
```

Get xml with this part:

```xml
<type type="PrimitiveType" type="BOOLEAN">
```

Use other tool to parse, eg, Python lxml:

```py
lxml.etree.fromstring(ast_xml)
```

Raise exception:

```
XMLSyntaxError: Attribute type redefined, ...
```

Patch #4806 changes `type` to `_type`, with a _private_ meaning.

jlerbsc:
JP is not intended to provide an XML representation that is compatible with another library.

sgqy:
@jlerbsc 

Sorry for my XY problem description.

The reason is not _another library_. It is the _standard_.

https://stackoverflow.com/a/7058404

PR #4806, opened 2025-08-06
XmlPrinter: fix duplicate attribute name in generated xml

sgqy:
Patch of #4805

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/printer/XmlPrinter.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/printer/XmlPrinter.java
@@ -182,7 +182,7 @@ public class XmlPrinter {
         xmlWriter.writeStartElement(name);
         // Output node type attribute
         if (outputNodeType) {
-            xmlWriter.writeAttribute("type", metaModel.getTypeName());
+            xmlWriter.writeAttribute("_type", metaModel.getTypeName());
         }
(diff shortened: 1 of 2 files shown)


jlerbsc:
JP is not intended to provide an XML representation that is compatible with another library.

javaparser-core/src/main/java/com/github/javaparser/printer/XmlPrinter.java
```diff
@@ -182,7 +182,7 @@ public void outputNode(Node node, String name, XMLStreamWriter xmlWriter) throws
         xmlWriter.writeStartElement(name);
         // Output node type attribute
         if (outputNodeType) {
-            xmlWriter.writeAttribute("type", metaModel.getTypeName());
+            xmlWriter.writeAttribute("_type", metaModel.getTypeName());
```
jlerbsc: Can you name this attribute “nodeType”?

sgqy: Yes. I've renamed this key name.

jlerbsc:
You must format the code with the command: mvn spotless:apply

sgqy:
@jlerbsc thanks. the command worked with modification.

jlerbsc:
Thank you for this PR.

Issue #4781, opened 2025-06-25
`LexicalPreservingPrinter` does not print new `PrimitiveType`s


abextm:
`LexicalPreservingPrinter` can't print `PrimitiveType`s that it didn't also parse. 

`getOrCreateNodeText` eventually calls itself, so nothing ever fills in the `NodeText` that is allocated.
```java
import com.github.javaparser.JavaParser;
import com.github.javaparser.ParserConfiguration;
import com.github.javaparser.ast.type.PrimitiveType;
import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;

class Scratch
{
	public static void main(String[] args)
	{
		var config = new ParserConfiguration();
		config.setLexicalPreservationEnabled(true);

		var cu = new JavaParser(config).parse("class Foo {\n" +
				"}")
			.getResult()
			.get();

		LexicalPreservingPrinter.setup(cu);

		cu.getType(0).addField(PrimitiveType.intType(), "foo");

		System.out.println(LexicalPreservingPrinter.print(cu));
	}
}
```

Output:
```java
class Foo {

 foo;
}
```

Expected:

```java
class Foo {

int foo;
}
```

Version:
`3.27.0`

jlerbsc:
Thank you for reporting this bug.

jlerbsc:
There is a bug but for the moment you can simply remove this line which is useless in your case.

    config.setLexicalPreservationEnabled(true);

jlerbsc:
This is a fairly complex bug to correct that was introduced with the latest version of JP, which includes the option of declaring the lexical printer as the default printer. It seems to be specific to adding dynamic code (versus parsed code). In this particular case the default printing mechanism is not initialised on the new node and this gives rise to the error you have seen.

VKS1572:
import com.github.javaparser.JavaParser;
import com.github.javaparser.ParserConfiguration;
import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
import com.github.javaparser.ast.type.PrimitiveType;
import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;

public class Scratch {
    public static void main(String[] args) {
        var config = new ParserConfiguration();
        config.setLexicalPreservationEnabled(true);

        var parser = new JavaParser(config);

        var cu = parser.parse(
                "class Foo {\n" +
                "}"
        ).getResult().orElseThrow();

        // Enable lexical preservation
[...]

PR #4817, opened 2025-08-18
fix: LexicalPreservingPrinter can't print PrimitiveType

seokjun7410:
fixes. #4781 

## Issue Analysis
When adding primitive type fields via LexicalPreservingPrinter, keywords like `int`, `boolean` get omitted:

```java
// Expected: class Foo { int foo; }
// Actual:   class Foo { foo; }    // 'int' missing
```

## Root Cause
Dynamically created `PrimitiveType` nodes lack proper NodeText initialization. The existing `prettyPrintingTextNode()` relied on `node.toString()` which returns empty strings for uninitialized nodes.

## My Solution: CSM-Based Architecture Approach

**Removed Special Case Handling**: Eliminated the hardcoded PrimitiveType switch statement in `prettyPrintingTextNode()` (38 lines removed).

**Enhanced `getOrCreateNodeText()`**: Added proper handling for dynamically created PrimitiveType nodes:
```java
if (!node.hasRange() && node instanceof PrimitiveType) {
    interpret(node, ConcreteSyntaxModel.forClass(node.getClass()), nodeText);
} else {
    prettyPrintingTextNode(node, nodeText);
}
```

**Made Primitive enum implement Stringable**: For consistency with JavaParser's design patterns.

nstead of special-casing PrimitiveType, I leveraged JavaParser's existing CSM (ConcreteSyntaxModel) infrastructure which already knows how to render primitive types correctly.

## Question
Does this CSM-based approach align well with JavaParser's lexical preservation architecture? I chose this over the simpler hardcoded string approach because it maintains consistency with how other AST nodes are handled.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/LexicalPreservingPrinterTest.java
@@ -36,6 +36,7 @@ import com.github.javaparser.ast.body.*;
 import com.github.javaparser.ast.comments.LineComment;
 import com.github.javaparser.ast.expr.*;
 import com.github.javaparser.ast.stmt.*;
+import com.github.javaparser.ast.type.PrimitiveType;
 import com.github.javaparser.ast.type.Type;
 import com.github.javaparser.ast.type.UnionType;
 import com.github.javaparser.ast.type.VoidType;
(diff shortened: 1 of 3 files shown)


jlerbsc:
Your PR contains two separate changes. One is an interesting improvement to the existing code, and the other tries to fix a bug. 

It would be more practical to separate the two types of changes, as I don't think the bug fix is very relevant. 

My observation is that when propagating the change in the list of members (here the fields) of the class declaration, we try to modify the parent node of the list of fields (more precisely in the Node.setParentNode(...) method). According to my debugger, this is when the data field of the node to be added, which is supposed to contain the text form of the field declaration, is modified. Before the assignment ‘parentNode = newParentNode;’, my debugger indicates that the data field (of the node to be added) is null, and after the assignment, the data field is initialised, which subsequently disrupts lexical preservation. I don't know if this is a bug in the debugger or if it's a side effect related to changing the parent node of the node to be added. Perhaps you could confirm this analysis in your work environment.

seokjun7410:
Thank you for the insightful review. I'll debug the `setParentNode()` issue first and separate the changes into two PRs as you suggested.

PR #4855, opened 2025-09-26
Fix issue 4846

### LINKED ISSUE #4846: JavaParserTypeAdapter throw doesn't account for record declarations
`JavaParserTypeAdapter.solveType` contains code to wrap TypeDeclarations in their equivalent JavaParser type, but is missing a check for records, causing it to throw UnsupportedOperationException when it encounters one.

PiTheGuy:
Fixes #4846.

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4846Test.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4846Test.java
@@ -1,17 +1,20 @@
 package com.github.javaparser.symbolsolver;
 
-import com.github.javaparser.*;
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParserConfiguration;
 import com.github.javaparser.ast.CompilationUnit;
 import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.body.TypeDeclaration;
-import com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade;
 import com.github.javaparser.symbolsolver.resolution.AbstractResolutionTest;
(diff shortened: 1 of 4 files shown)


javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4846Test.java
```diff
@@ -0,0 +1,42 @@
...
+
+public class Issue4846Test extends AbstractResolutionTest {
+    public static final Path SRC_DIR = adaptPath("src/test/resources/issue4846");
+
+    private JavaParser javaParser;
+
+    @BeforeEach
+    void setup() {
+        // clear internal caches
+        JavaParserFacade.clearInstances();
```
jlerbsc: This is not necessary in your test.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue4846Test.java
```diff
@@ -0,0 +1,42 @@
...
+
+        javaParser = new JavaParser(configuration);
+    }
+
+    @Test
+    void test() throws IOException {
+        CompilationUnit cu = javaParser.parse(SRC_DIR.resolve("foo").resolve("Main.java")).getResult().get();
+        TypeDeclaration<?> typeDec = cu.getType(0);
+        MethodDeclaration methodDec = typeDec.getMethodsByName("foo").get(0);
+        methodDec.toDescriptor();
```
jlerbsc: There is no verification.

jlerbsc:
You still need to use spotless to format your changes (mvn sptoless:apply).

jlerbsc:

PR #4867, opened 2025-10-09
Add support for match-all patterns

### LINKED ISSUE #4713: Anonymous Variables and Patterns
https://openjdk.org/jeps/456

johannescoetzee:
Resolves https://github.com/javaparser/javaparser/issues/4713

This PR mainly introduces support for match-all patterns, but does contain a small change to the grammar involving identifiers which was required to disambiguate match-all pattern parsing.

# Major changes

## MatchAllPatternExpr added
The main point of this  PR is to add support for `MatchAllPatternExprs`. This includes a grammar change in `java.jj`, a new node type, and support in the `ConcreteSyntaxModel`. I initially went back-and-forth on the naming, but settled for `MatchAllPatternExpr` in the end since this is consistent with the JLS and is a good description of what it actually is. The other option was `UnnamedPatternExpr`, as it is often referred to in text, but in my opinion this could easily be confused with an unnamed pattern variable which is actually contained in a TypePatternExpr, e.g. `String _`.

While the actual `MatchAllPatternExpr` class  and syntax are very simple, fairly large changes to the pattern type hierarchy are required. Previously, the assumption was made that (1) all patterns can be used as top-level patterns in instanceof expressions and switch entry labels, and (2) all pattern expressions have a type. While this was true for `TypePatternExpr` and `RecordPatternExpr`, this no longer holds, leading to the next major change.

## Pattern type hierarchy changed
To account for the fact that not all patterns can be used as top-level patterns and not all patterns are typed, a new abstract base class `ComponentPatternExpr` is introduced. Following the structure in the updated Java syntax, a `ComponentPatternExpr` can be either a `MatchAllPatternExpr`, or a `PatternExpr`. `PatternExpr` remains unchanged from before and can still be a `TypePatternExpr` or a `RecordPatternExpr`.

For reference, the updated syntax is:
```
Pattern:
  TypePattern
  RecordPattern

TypePattern:
  LocalVariableDeclaration

RecordPattern:
  ReferenceType ( [ComponentPatternList] )

ComponentPatternList:
  ComponentPattern {, ComponentPattern }

ComponentPattern:
  Pattern
  MatchAllPattern

MatchAllPattern:
  _
```

## Java 22 validator added
I've added a Java 22 validator class with 2 new validators: a check to ensure that match-all patterns aren't used as top-level patterns and a check to ensure that unnamed variables are only used where allowed by JEP456. The logic is quite particular, so I've added tests for all of the cases mentioned in JEP456 that should serve as decent documentation.

# Other notes

## Grammar change for identifiers
I had to make a change to the `IDENTIFIER` token to get `MatchAllPatternExpr` to parse correctly. The `IDENTIFIER` token now excludes the identifier `_`. It must either be an underscore followed by a non-zero number of characters, or a non-underscore character followed by  zero or more other characters. In order to continue parsing Java <= 8 code where `_` is an allowed identifier, I added `"_"` as an explicit case in the `Identifier` production, similar to other keywords. This disambiguates parsing for `MatchAllPatternExpr` without affecting identifier parsing.

## Note on testing
I've added tests for everything I could think of that should be affected by this change. This includes the `IfStatmentContextTests` and `SwitchEntryContextTests`. Some of the tests added in these suites aren't for examples that should be affected by the changes here, but there was a gap in testing so I thought it good to add them. If you can think of anything that I missed, please let me know and I'll add tests for that.

## Note on commits
The early commits are a bit of a mess since I didn't settle on the current naming scheme/pattern hierarchy until I'd done a significant amount of work. I initially decided to use `TypedPatternExpr` as the common ancestor for `TypePatternExpr` and `RecordPatternExpr`, and `PatternExpr` as the common ancestor for `TypedPatternExpr` and `MatchAllPatternExpr`. I later changed my mind, since `TypedPatternExpr` and `TypePatternExpr` were too close together and none of this matched the JLS. At this point it was easier to rename the classes than to redo everything, so apologies for the messy back-and-forth in the commits.

### DIFF
--- a/javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
+++ b/javaparser-core-metamodel-generator/src/main/java/com/github/javaparser/generator/metamodel/MetaModelGenerator.java
@@ -31,7 +31,6 @@ import com.github.javaparser.ast.NodeList;
 import com.github.javaparser.ast.body.ClassOrInterfaceDeclaration;
 import com.github.javaparser.ast.body.CompactConstructorDeclaration;
 import com.github.javaparser.ast.body.MethodDeclaration;
-import com.github.javaparser.ast.expr.MatchAllPatternExpr;
 import com.github.javaparser.ast.stmt.Statement;
 import com.github.javaparser.generator.AbstractGenerator;
 import com.github.javaparser.printer.DefaultPrettyPrinter;
@@ -138,7 +137,7 @@ public class MetaModelGenerator extends AbstractGenerator {
             add(com.github.javaparser.ast.expr.NormalAnnotationExpr.class);
             add(com.github.javaparser.ast.expr.NullLiteralExpr.class);
             add(com.github.javaparser.ast.expr.ObjectCreationExpr.class);
-            add(com.github.javaparser.ast.expr.PatternExpr.class);
+            add(com.github.javaparser.ast.expr.ComponentPatternExpr.class);
             add(com.github.javaparser.ast.expr.TypedPatternExpr.class);
             add(com.github.javaparser.ast.expr.RecordPatternExpr.class);
             add(com.github.javaparser.ast.expr.SingleMemberAnnotationExpr.class);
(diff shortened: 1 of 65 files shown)


johannescoetzee:
It looks like the codecov misses are just in the generated boilerplate code. If this is an issue, let me know and I'll add tests targeting this specifically.

jlerbsc:
It's excellent work. It is indeed difficult to keep track of the changes, but overall it seems correct to me.

PR #4872, opened 2025-10-15
Fix : Make SourceRoot.saveAll respect new root for absolute paths and add tests #4795

### LINKED ISSUE #4795: Potential defect in SourceRoot.saveAll(Path)...
Using JavaParser version 3.26.4 along with JDK 21.

The ```SourceRoot``` has a method ```saveAll(Path root)```. The purpose of this method is to save the source files to new path provided as argument. However, this has an issue, which I think could be a defect.

The above method in turn calls ```saveAll(Path root, Charset encoding)``` method. This method has a statement:
```
final Path path = root.resolve(cu.getKey());
```

When the ```CompilationUnit``` has a relative path "src/main/java/some/package/SourceFile.java" and I provide the new path as "/my-modified-project", the root.resolve() would get the path as "/my-modified-project/src/main/java/some/package/SourceFile.java". In this case, the ```saveAll(Path)``` method works as intended. It keeps the source file in the original location untouched and writes the modified source file in the new path that I specified.

When the ```CompilationUnit``` has an absolute path already, irrespective of the path specified for ```saveAll(Path)```, the modified file gets written to the original location. Please refer to [Java Path reference](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#resolve-java.nio.file.Path-). This may have an impact on the intended behavior of ```saveAll(Path)``` method.

Joyce-5:
This pull request resolves #4795
 by fixing inconsistent behavior in SourceRoot.saveAll() when handling absolute paths.
Previously, when a CompilationUnit had an absolute storage path, the method ignored the new root and saved the file to its original location.
This update ensures all files, whether relative or absolute  are saved under the specified new root consistently.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/SourceRootPathResolutionTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/SourceRootPathResolutionTest.java
@@ -1,0 +1,56 @@
+package com.github.javaparser.utils;
+
+import com.github.javaparser.StaticJavaParser;
+import com.github.javaparser.ast.CompilationUnit;
+import org.junit.jupiter.api.Test;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+class SourceRootPathResolutionTest {
+
+    @Test
+    void relativeKey_resolvesUnderNewRoot() {
+        Path newRoot = Paths.get("/out");
+        Path oldRoot = Paths.get("/proj");
+        Path key = Paths.get("src/main/java/p/A.java");
+
+        Path result = SourceRoot.resolveSavePath(newRoot, oldRoot, key, Optional.empty());
+        assertEquals(Paths.get("/out/src/main/java/p/A.java"), result);
+    }
+
+    @Test
+    void absoluteUnderOldRoot_isRelativizedThenReRooted() {
+        Path newRoot = Paths.get("/out");
+        Path oldRoot = Paths.get("/proj");
+        Path key = Paths.get("/proj/src/main/java/p/A.java");
+
+        Path result = SourceRoot.resolveSavePath(newRoot, oldRoot, key, Optional.empty());
+        assertEquals(Paths.get("/out/src/main/java/p/A.java"), result);
+    }
+
+    @Test
+    void absoluteOutsideOldRoot_usesPackageAndPrimaryType() {
+        Path newRoot = Paths.get("/out");
+        Path oldRoot = Paths.get("/proj");
+        Path key = Paths.get("/somewhere/else/X.java");
+
+        CompilationUnit cu = StaticJavaParser.parse("package p.q; public class A {}");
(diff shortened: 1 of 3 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -477,18 +483,104 @@ private SourceRoot save(CompilationUnit cu, Path path, Charset encoding) {
...
+     *
+     * @param newRoot the destination root directory
+     * @param oldRoot the original source root
+     * @param key     the cached path key (may be relative or absolute)
+     * @param unit    the compilation unit if available
+     * @return the resolved and normalized target path under newRoot
      */
+
+
+    static Path resolveSavePath(Path newRoot, Path oldRoot, Path key, Optional<CompilationUnit> unit) {
```
jlerbsc: Why is this method static?

jlerbsc: "Optional" should not be used for parameters

jlerbsc: Rename the method to resolvePath

jlerbsc:
@Joyce-5 thank you for your proposal but I cannot validate this PR because there are too many changes in how the method works and too many inaccuracies. I suggest we redo the PR by breaking down and ordering the commits and taking into account the various comments.
First, isolate the path resolution method without making any changes to it and set up unit tests that will allow you to test your changes (1st commit).
Secondly, introduce changes into the resolution method that enable the implementation of the resolve method specification https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html and adapt the unit test cases (2nd commit).
There are no other cases to implement.

javaparser-core-testing/src/test/java/com/github/javaparser/utils/SourceRootPathResolutionTest.java
```diff
@@ -1,56 +1,46 @@
...
-    void relativeKey_resolvesUnderNewRoot() {
-        Path newRoot = Paths.get("/out");
-        Path oldRoot = Paths.get("/proj");
-        Path key = Paths.get("src/main/java/p/A.java");
+    void relative_is_resolved_under_newRoot_currentBehaviour() throws IOException {
 
-        Path result = SourceRoot.resolveSavePath(newRoot, oldRoot, key, Optional.empty());
-        assertEquals(Paths.get("/out/src/main/java/p/A.java"), result);
-    }
+        Path srcRoot = Files.createTempDirectory("jp-src-");
```
jlerbsc: Once the path resolution method has been isolated, there is no longer any need to create test directories.

Joyce-5:
@jlerbsc I’ve restructured the commits as suggested:
- First commit isolates the path resolution logic and adds tests.
- Second commit applies behavioral fixes and removes `Optional` usage.

Please let me know if any further refinements are needed. Thank you again!

jlerbsc:
Please make a new PR so that your changes are readable.

Joyce-5:
ok, I will do it now

PR #4874, opened 2025-10-16
Include source file path in failed ParseResult when parsing via SourceRoot #4786

### LINKED ISSUE #4786: Failed ParseResult needs reference to the source file
JavaParser version: 3.26.2
JDK version: 21

Context: 

I obtained the ```ParseResults``` using ```SourceRoot.tryToParse()``` on a source root. For the unsuccessful parse, the ```ParseResult.getResult()``` is null, which is fair. However, I get only the list of ```Problem``` having error message and stack trace, but not the reference to the file that failed parsing. It would be helpful to include at least the FQN to the class name or the path to the source file. If it is possible keep reference to unparsed file as broken, it would be wonderful, so that I can get the contents as text continue the processing.

JIN-RUI-LIU:
Fixes #4786.
Problem: Failed parses from SourceRoot.tryToParse() don’t indicate which file failed (ParseResult has Problems but no file path).
Change: ParseResult now carries an internal sourcePath set by SourceRoot; on failures, ParseResult.toString() includes “for <absolute-path>”. Successful parses still set CompilationUnit storage as before.
Tests: Added issues/Issue4786Test to reproduce the “file content is only ‘\u’” failure and assert diagnostics include the absolute file path.
Compatibility: Backward compatible.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ParseResult.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParseResult.java
@@ -22,6 +22,7 @@ package com.github.javaparser;
 
 import com.github.javaparser.ast.comments.CommentsCollection;
 import com.github.javaparser.utils.LineSeparator;
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Consumer;
@@ -36,6 +37,17 @@ public class ParseResult<T> {
     private final List<Problem> problems;
 
     private final CommentsCollection commentsCollection;
+
+    /**
+     * Optional source path associated with this parse operation. This may be set by higher-level utilities
+     * (e.g., {@code SourceRoot}) to allow callers to correlate parse problems with the originating file path
+     * even when no {@code CompilationUnit} is produced.
+     * <p>
+     * Contract:
+     * - If present, it's expected to be an absolute {@link Path} to the source input used for this parse.
+     * - This value should be set immediately after parsing is performed and treated as effectively immutable thereafter.
+     */
+    private Path sourcePath;
 
     /**
      * General constructor.
@@ -66,6 +78,19 @@ public class ParseResult<T> {
     }
 
     /**
+     * Associates a source path with this parse result. Returns {@code this} for chaining.
+     * <p>
+     * Notes:
+     * - Intended for use by infrastructure (e.g., {@code SourceRoot}) to publish the originating file path.
+     * - Should be called immediately after parsing and not modified afterwards to avoid surprises in concurrent contexts.
+     * - The provided path is expected to be absolute.
+     */
+    public ParseResult<T> setSourcePath(Path sourcePath) {
+        this.sourcePath = sourcePath;
+        return this;
+    }
+
+    /**
      * @return the list of encountered parsing problems. Empty when no problems were encountered.
      */
     public List<Problem> getProblems() {
@@ -93,12 +118,23 @@ public class ParseResult<T> {
         return Optional.ofNullable(result);
     }
 
+    /**
+     * @return the absolute path to the source file this parse result originates from, or empty if no file parsing.
+     */
+    public Optional<Path> getSourcePath() {
+        return Optional.ofNullable(this.sourcePath);
+    }
+
     @Override
     public String toString() {
         if (isSuccessful()) {
             return "Parsing successful";
         }
-        StringBuilder message = new StringBuilder("Parsing failed:").append(LineSeparator.SYSTEM);
+        StringBuilder message = new StringBuilder("Parsing failed");
+        if (sourcePath != null) {
+            message.append(" for ").append(sourcePath);
+        }
+        message.append(":").append(LineSeparator.SYSTEM);
         for (Problem problem : problems) {
             message.append(problem.toString()).append(LineSeparator.SYSTEM);
         }
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
@@ -131,7 +131,8 @@ public class SourceRoot {
         final Path path = root.resolve(relativePath);
         Log.trace("Parsing %s", () -> path);
         final ParseResult<CompilationUnit> result = new JavaParser(configuration)
-                .parse(COMPILATION_UNIT, provider(path, configuration.getCharacterEncoding()));
+                .parse(COMPILATION_UNIT, provider(path, configuration.getCharacterEncoding()))
+                .setSourcePath(path);
         result.getResult().ifPresent(cu -> cu.setStorage(path, configuration.getCharacterEncoding()));
         cache.put(relativePath, result);
         return result;
@@ -277,7 +278,8 @@ public class SourceRoot {
         Path localPath = root.relativize(absolutePath);
         Log.trace("Parsing %s", () -> localPath);
         ParseResult<CompilationUnit> result = new JavaParser(configuration)
-                .parse(COMPILATION_UNIT, provider(absolutePath, configuration.getCharacterEncoding()));
+                .parse(COMPILATION_UNIT, provider(absolutePath, configuration.getCharacterEncoding()))
+                .setSourcePath(absolutePath);
         result.getResult().ifPresent(cu -> cu.setStorage(absolutePath, configuration.getCharacterEncoding()));
         switch (callback.process(localPath, absolutePath, result)) {
             case SAVE:
--- a/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
@@ -20,8 +20,6 @@
  */
 package com.github.javaparser.utils;
 
-import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
-import static com.github.javaparser.Providers.provider;
 import static com.github.javaparser.utils.CodeGenerationUtils.*;
 import static com.github.javaparser.utils.Utils.assertNotNull;
 import static java.nio.file.FileVisitResult.*;
@@ -130,10 +128,7 @@ public class SourceRoot {
         }
         final Path path = root.resolve(relativePath);
         Log.trace("Parsing %s", () -> path);
-        final ParseResult<CompilationUnit> result = new JavaParser(configuration)
-                .parse(COMPILATION_UNIT, provider(path, configuration.getCharacterEncoding()))
-                .setSourcePath(path);
-        result.getResult().ifPresent(cu -> cu.setStorage(path, configuration.getCharacterEncoding()));
+        final ParseResult<CompilationUnit> result = new JavaParser(configuration).parse(path);
         cache.put(relativePath, result);
         return result;
     }
@@ -277,10 +272,7 @@ public class SourceRoot {
             throws IOException {
         Path localPath = root.relativize(absolutePath);
         Log.trace("Parsing %s", () -> localPath);
-        ParseResult<CompilationUnit> result = new JavaParser(configuration)
-                .parse(COMPILATION_UNIT, provider(absolutePath, configuration.getCharacterEncoding()))
-                .setSourcePath(absolutePath);
-        result.getResult().ifPresent(cu -> cu.setStorage(absolutePath, configuration.getCharacterEncoding()));
+        ParseResult<CompilationUnit> result = new JavaParser(configuration).parse(absolutePath);
         switch (callback.process(localPath, absolutePath, result)) {
             case SAVE:
                 result.getResult().ifPresent(cu -> save(cu, absolutePath));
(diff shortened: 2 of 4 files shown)


javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -445,7 +448,8 @@ public SourceRoot add(CompilationUnit compilationUnit) {
             final Path path = compilationUnit.getStorage().get().getPath();
             Log.trace("Adding new file %s", () -> path);
             final ParseResult<CompilationUnit> parseResult =
-                    new ParseResult<>(compilationUnit, new ArrayList<>(), null);
+                    new ParseResult<>(compilationUnit, new ArrayList<>(), null)
+                            .setSourcePath(path);
```
jlerbsc: Why are you adding the source file path when there is no file parsing?

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -430,7 +432,8 @@ public SourceRoot add(String startPackage, String filename, CompilationUnit comp
         assertNotNull(compilationUnit);
         Log.trace("Adding new file %s.%s", () -> startPackage, () -> filename);
         final Path path = fileInPackageRelativePath(startPackage, filename);
-        final ParseResult<CompilationUnit> parseResult = new ParseResult<>(compilationUnit, new ArrayList<>(), null);
+        final ParseResult<CompilationUnit> parseResult = new ParseResult<>(compilationUnit, new ArrayList<>(), null)
+                .setSourcePath(root.resolve(path));
```
jlerbsc: Why are you adding the source file path when there is no file parsing?

javaparser-core/src/main/java/com/github/javaparser/ParseResult.java
```diff
@@ -65,6 +79,19 @@ public void ifSuccessful(Consumer<T> consumer) {
...
 
+    /**
+     * Associates a source path with this parse result. Returns {@code this} for chaining.
+     * <p>
+     * Notes:
+     * - Intended for use by infrastructure (e.g., {@code SourceRoot}) to publish the originating file path.
+     * - Should be called immediately after parsing and not modified afterwards to avoid surprises in concurrent contexts.
+     * - The provided path is expected to be absolute.
+     */
+    public ParseResult<T> setSourcePath(Path sourcePath) {
```
jlerbsc, line 88: Curiously, there is no method for retrieving the path to the source file.

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
```diff
@@ -0,0 +1,70 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Regression for issue 4786: when parsing via SourceRoot and a file fails to parse,
+ * diagnostics should indicate which file failed.
+ */
+public class Issue4786Test {
```
jlerbsc: These tests should be implemented in the SourceRootTest class.

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    void failedParse_shouldIncludeFilePath_whenFileContainsBackslashUOnly() throws IOException {
+        // Arrange: create a temporary source root and a file whose content triggers a scanner failure
+        Path tempDir = Files.createTempDirectory("jpv-issue4786");
```
jlerbsc: There is no need to create directories or new files because everything you need is in the ‘resources’ directory.

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
```diff
@@ -0,0 +1,70 @@
...
+        Path srcRoot = tempDir.resolve("src");
+        Files.createDirectories(srcRoot);
+
+        Path broken = srcRoot.resolve("OnlyUnicode.java");
+        // Write exactly two bytes: backslash and 'u'
+        Files.write(broken, new byte[] { (byte) '\\', (byte) 'u' });
```
jlerbsc: ??

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
```diff
@@ -0,0 +1,70 @@
...
+        // Act
+        List<ParseResult<CompilationUnit>> results = sourceRoot.tryToParse("");
+
+        // Assert
+        assertEquals(1, results.size(), "Expected one parse result");
+        ParseResult<CompilationUnit> result = results.get(0);
+        assertFalse(result.isSuccessful(), "Parsing should fail for this content");
+        assertFalse(result.getProblems().isEmpty(), "Problems should be present on failure");
+
+        String diag = result.toString();
```
jlerbsc: Useless

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
```diff
@@ -0,0 +1,70 @@
...
+        String expectedPath = broken.toAbsolutePath().toString();
+        assertTrue(diag.contains(expectedPath),
```
jlerbsc: The check must be based on a getSourcePath method that does not currently exist.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
```diff
@@ -432,8 +432,7 @@ public SourceRoot add(String startPackage, String filename, CompilationUnit comp
         assertNotNull(compilationUnit);
         Log.trace("Adding new file %s.%s", () -> startPackage, () -> filename);
         final Path path = fileInPackageRelativePath(startPackage, filename);
-        final ParseResult<CompilationUnit> parseResult = new ParseResult<>(compilationUnit, new ArrayList<>(), null)
-                .setSourcePath(root.resolve(path));
+        final ParseResult<CompilationUnit> parseResult = new ParseResult<>(compilationUnit, new ArrayList<>(), null);
```
jlerbsc, line 427: Creating a new commit to undo a previous commit is not the best solution for keeping a clean history. It makes your work unreadable. It would have been simpler to use Git's ‘Interactive Rebase’ feature.

jlerbsc:
I am waiting for you to make the other changes before committing your work.

javaparser-core/src/main/java/com/github/javaparser/utils/SourceRoot.java
JIN-RUI-LIU: OK，I‘ll try to figure it out

javaparser-core-testing/src/test/java/com/github/javaparser/issues/Issue4786Test.java
JIN-RUI-LIU: In my first attempt I added a test resource whose content was a single backslash-u ("\u"). The goal was to deterministically trigger a scanner failure so that ParseResult would be unsuccessful and result() would be empty (as issue #4786 ). This let me verify that getSourcePath is present on failed parses.
Based on your feedback, I removed the custom “\u” fixture and now use an existing Java file under the utils test resources to trigger the failure. I also moved the test into SourceRootTest as requested and simplified the assertions to only check ParseResult.getSourcePath() (no dependency on toString).

JIN-RUI-LIU:
> I am waiting for you to make the other changes before committing your work.

Thank you very much for your guidance!
I’ve updated all the changes you suggested.
Could you please let me know if there’s anything else that needs improvement?

jlerbsc:
Can you squash your commits to make it more readable? Thank you.

JIN-RUI-LIU:
> Can you squash your commits to make it more readable? Thank you.

Thanks for the reminder! I’ve already squashed all the commits into one commit.

jlerbsc:
Now what happens if parsing fails? The issue mainly concerned cases where parsing failed, because in the other case the path is already available via cu.getStorage().getPath().

JIN-RUI-LIU:
> Now what happens if parsing fails? The issue mainly concerned cases where parsing failed, because in the other case the path is already available via cu.getStorage().getPath().

In my first attempt I added a test resource whose content was a single backslash-u ("\u"). The goal was to deterministically trigger a scanner failure so that ParseResult would be unsuccessful and result() would be empty (as issue https://github.com/javaparser/javaparser/issues/4786 ). This let me verify that getSourcePath is present on failed parses.
The figure below shows the case when parsing fails. We can see that the path
image

Based on your feedback, I removed the custom “\u” fixture and now use an existing Java file under the utils test resources to trigger the failure.

JIN-RUI-LIU:
I added sourcePathToStringTest to cover ParseResult.toString() with and without sourcePath, improving branch coverage and guarding against regressions. And I ran Spotless locally.
image
image

jlerbsc:
On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.

JIN-RUI-LIU:
> On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.

Thanks for the suggestion. To confirm: 
Do you mean that we should still generate a result even if parsing fails? (In this case, the result is null, and I would create a new one.) I'm not sure if that could have any side effects. If that’s what you intend, please let me know — I’ll create a new branch to handle it.

image

JIN-RUI-LIU:
> On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.

I thought about it again — or do you mean setting the sourcePath here, so that only the files that failed to parse would have their sourcePath set? Please let me know if that’s what you mean, and I’ll give it a try.
image

jlerbsc:
> > On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.
> 
> Thanks for the suggestion. To confirm: Do you mean that we should still generate a result even if parsing fails? (In this case, the result is null, and I would create a new one.) I'm not sure if that could have any side effects. If that’s what you intend, please let me know — I’ll create a new branch to handle it.
> 
> image

In practice, ParseResult is instantiated even in the event of a parsing error.

jlerbsc:
> > On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.
> 
> I thought about it again — or do you mean setting the sourcePath here, so that only the files that failed to parse would have their sourcePath set? Please let me know if that’s what you mean, and I’ll give it a try. image

If the parsing runs correctly, the compilation unit already knows the path of the parsed file (storage). However, in the event of an error, there is no information because the compilation unit is not instantiated. So in this case, the path should be stored in the ParseResult that is returned.
Furthermore, JavaParser declares parsing methods for which the provider is instantiated. It is therefore preferable to use these methods in the SourceRoot class.

JIN-RUI-LIU:
> > > On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.
> > 
> > 
> > I thought about it again — or do you mean setting the sourcePath here, so that only the files that failed to parse would have their sourcePath set? Please let me know if that’s what you mean, and I’ll give it a try. image
> 
> If the parsing runs correctly, the compilation unit already knows the path of the parsed file (storage). However, in the event of an error, there is no information because the compilation unit is not instantiated. So in this case, the path should be stored in the ParseResult that is returned. Furthermore, JavaParser declares parsing methods for which the provider is instantiated. It is therefore preferable to use these methods in the SourceRoot class.

Thank you for your guidance; I’ll give it a try.

jlerbsc:
Why did you close this PR? Just make sure to squash your commits and isolate improvements to other classes in a specific commit.

JIN-RUI-LIU:
> On second thought, I wonder if it wouldn't be better to set the source path in the JavaParser class at the point where methods take a file or path as a parameter.

Sorry to bother you again, but I’m a bit confused about this.
You suggested setting the source path in JavaParser.
However, SourceRoot.tryToParse() internally uses the Provider entry point:
image
in JavaParser.
Since the Provider doesn’t expose the file path, it seems there’s no way to retrieve it inside JavaParser.parse().
Please correct me if I’m missing something — thank you!

JIN-RUI-LIU:
Here is my attempt to move setSourcePath into JavaParser. Since SourceRoot.tryToParse() calls the overload that also instantiates the provider, we cannot retrieve the path at the JavaParser layer. That means we would need to switch the parse() overload used inside SourceRoot.tryToParse(), but this does not seem reliable.

Here is SourceRoot.tryToParse() . I’m a bit worried that replacing the .parse() call with another method might cause some unintended side effects.


image


Here is what I changed to setSourcePath in JavaParser, but this method is Deprecated.


image
image

jlerbsc:
The method that defines character encoding is deprecated because the input character encoding must be a configuration option so you can use the method that does not define the encoding.

jlerbsc:

JIN-RUI-LIU:
> Thank you for your contribution.

Thank you for your guidance. I’ve learned a lot from this task.

Issue #4869, opened 2025-10-09
JEP 512: Compact Source Files and Instance Main Methods


johannescoetzee:
https://openjdk.org/jeps/512 introduces new syntax for compact source files with instance main methods. Details can be found in the JEP, but to illustrate using the examples from the JEP, the program
```
public class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello, World!");
    }
}
```
can be simplified to 
```
void main() {
    System.out.println("Hello, World!");
}
```
in Java 25.

In my opinion, the best way to implement this in JavaParser would be to still create a `ClassOrInterfaceDeclaration` for this CU, with a field specifying that this is a compact class. This would limit the scope of this change to the grammar and concrete syntax model.

For this approach, the following changes need to be made:
- the grammar must be updated to support the new syntax
- an `isCompact` field must be added to `ClassOrInterfaceDeclaration`
- the CSM must be updated to support the new syntax
- compact classes implicitly import the `java.base` module, so this must be taken into account in the symbol resolver/type solvers

johannescoetzee:
Apologies if this is a duplicate. I searched for similar issues and didn't find anything.

I plan to work on this next, so would appreciate any input on the idea proposed.

jlerbsc:
What bothers me about this approach is that we are dealing with Java 25 features before those of Java 23 and 24. I admit that I haven't taken the time to check whether these versions contain features to be implemented in JP.

johannescoetzee:
I remember checking this before starting work on the unnamed patterns support and I didn't see any non-preview features in Java 23 or 24 that would need explicit changes in JavaParser. I agree that it doesn't make sense to start on Java 25 features before everything up to and including Java 24 is supported, so I'll go through the release notes again tomorrow to double check

jlerbsc:
There is this one https://github.com/javaparser/javaparser/issues/4660

johannescoetzee:
Ah, I did miss that one. I can start there then :)

jlerbsc:
> Ah, I did miss that one. I can start there then :)

Thank you!

jlerbsc:
Done.

PR #4875, opened 2025-10-16
[JEP 467] Add support for MarkdownComments

### LINKED ISSUE #4660: Consider supporting the document comments in the form of the markdown included in Java23 (JEP 467)?
Those start with///

Image

johannescoetzee:
Fixes https://github.com/javaparser/javaparser/issues/4660.

This PR adds support for Markdown comments as described in JEP 467. I ran into a few issues while implementing this, so the end result is a compromise between the ideal design and what is doable in a reasonable amount of time and without rewriting all of the comment handling (and requiring users to do the same).
 
## MarkdownComment node
As it is, this PR adds the `MarkdownComment` node, which is distinct from `JavadocComment`. I considered merging these under a common class, but in my opinion the benefits to doing so is far outweighed by the complexity of the change required. The `content` field of the `MarkdownComment` contains the raw text, including the `///` and leading spaces after the first line, which is somewhat consistent with how block comments are handled. Keeping this information is necessary for the pretty printer and LPP. 

It also contains a method `getMarkdownContent` which strips the leading whitespace, `///` and indents the remaining content consistent with the description provided in JEP 467. I suspect that this will be more useful to users than using the raw content.

## Parser changes
I initially tried updating the grammar to parse markdown comments as a single token, but ran into some difficulties doing this. From what I can see it looks like it should be possible, but would require manually manipulating the tokenizer input stream and I wasn't confident enough that this would work out to spend too much time on it. Instead, the `MarkdownComment` consists of a range of tokens starting with the first `SINGLE_LINE_COMMENT`, including all leading and trailing whitespace for the body and ending with the last `SINGLE_LINE_COMMENT`. The handling of all this is done in the `CommonTokenAction` method defined in `java.jj` which is added to the `GeneratedJavaParserTokenManager`. 

## Pretty printer and LPP
Support in the pretty printer is mostly a copy-paste of block comments are handled with only minor tweaks to account for the lack of `/**` and `*/`. Support for the LPP is more complicated since the assumption was made that a comment will always consist of a single token. I've updated the code handling adding, removing, and replacing comments to account for the possibility of having multiple tokens corresponding to the comment.

## Note on the implementation
The implementation and general commit history for this is a bit messy again. The comment handling logic was surprisingly tricky to extend to handle this case, so I tried to implement things in the least disruptive way. As I progressed, and especially when getting to the printers, I ran into issues with my initial implementation (especially the content field) and had to backtrack and change things a bit. Sorry about that.


jlerbsc:
I tried several markdown examples provided by the specification https://openjdk.org/jeps/467 without causing any parsing errors. Ultimately, why should we make changes in JP since it seems to work with the current version?

johannescoetzee:
@jlerbsc I assumed you wanted this based on your comment in https://github.com/javaparser/javaparser/issues/4869. This PR treats markdown comments as a single comment instead of a series of line comments, which makes it easier for users to use them. With how line comments are handled, only the last line comment would be associated with the commented method. Any preceding comments would be orphans, so correctly handling those currently involves handling ranges manually.

In that sense, I still think the change here is beneficial, but if simply parsing the comments without errors or crashing is sufficient, then this PR isn't needed and can be closed without issue

jlerbsc:
@johannescoetzee I agree with you that markdown comments are multi-line comments similar to javadoc comments. What I don't understand is that grouping these elements into a common class or using a class hierarchy makes change more complex. Could you elaborate on this point?

@ftomassetti , what is your opinion on integrating Markdown comments into the existing comment architecture? Do you agree with the suggestion made in this PR? Personally, I don't know JP well enough to have a strong opinion on this subject, but it seems to me that this PR introduces additional complexity into comment management.

johannescoetzee:
@jlerbsc I've thought about your question/concern some more and I've changed my mind. I don't think this is really a problem anymore.

I had the idea that grouping them into a common class while I was still using the first representation I tried for this (seen in the earlier commits). In that representation, the `content` field contained what is now returned by the `getMarkdownContent` method. In that version, the usual reconstruction of comments `comment.getHeader() + comment.getContent() + comment.getFooter()` no longer made sense, and in my opinion changing that anywhere javadoc comments were handled (including in user code) was not worth the benefit of uniting those comments. 

In the current version proposed by this PR, this is no longer an issue. I would therefore propose doing the following:

1. Rename `JavadocComment` to `TraditionalJavadocComment`, following the naming convention from https://docs.oracle.com/javase/specs/jls/se25/html/jls-3.html#jls-3.7 (where it's called `TraditionalComment`, but keeping the `Javadoc` in the name is useful in JavaParser to distinguish it from `BlockComment`
2. Create an abstract `JavadocComment` class that is extended by both `TraditionalJavadocComment` and `MarkdownComment`
3. Update the `Javadoc` and `JavadocParser` classes to handle `MarkdownComments`, which should now be a fairly small change. 

Please let me know if you agree with this approach. Once I hear back from you, I'll update the PR with what we've discussed.

johannescoetzee:
I've pushed changes starting with the [Rename JavadocComment to TraditionalJavadocComment](https://github.com/javaparser/javaparser/pull/4875/commits/c0f488a4a15f8edff33b5507007eef21cf3dabc7) commit that implement the shared ancestor for `TraditionalJavadocComment` and `MarkdownComment` that also demonstrate the code changes required to handle this case. Overall the changes on the user side will be fairly minor. The majority of the manual changes I had to make were for javaparser internal components. It's possible I missed something, but I've covered the things I found and I would, of course, fix any issues related to this that come up.

johannescoetzee:
It looks like the test failures are slow tests that aren't run with `mvn test` by default. I'm updating expectations for those now

Edit: Done

jlerbsc:
@johannescoetzee Thank you for the valuable work you have done. For now, I will wait for @ftomassetti 's opinion to shed further light on how to proceed.

johannescoetzee:
@jlerbsc Sounds good. Since this isn't required to parse Java 23 code, would you be open to accepting Java 25 contributions before this is finalised? I'm on vacation next week, but I've made good progress with handling module imports, so will likely have that ready soon after I'm back.

jlerbsc:
Hi @johannescoetzee , I am going to ask you to modify your PR to simplify it and isolate the integration of Markdown comments from the current comment management. Would it be possible for you to separate the comment class hierarchy into a separate PR (this would involve preparing the code to support Markdown comments) and then create a new PR to manage Markdown comments?

jlerbsc:
Do you think that a solution consisting, initially, of parsing Markdown comments as inline comments and then creating and transforming these comments into Markdown in post-processing would be easier to maintain than a solution consisting of parsing Markdown comments directly?

johannescoetzee:
> Do you think that a solution consisting, initially, of parsing Markdown comments as inline comments and then creating and transforming these comments into Markdown in post-processing would be easier to maintain than a solution consisting of parsing Markdown comments directly?

I did consider this approach as well, but thought it would end up being more complex. The main problem is that all but the last line comment before a method end up as orphan comments, so, to correctly reconstruct the markdown comments, it would necessary to connect these line comments by looking at token ranges and the tokens between them. While separating this logic from the parser would bring some benefit, I think it would be harder to reason about than the approach implemented in this PR where the token stream is being processed directly. It would also only work if tokens are saved (as far as I understand), which is a significant downside.

jlerbsc:
> > Do you think that a solution consisting, initially, of parsing Markdown comments as inline comments and then creating and transforming these comments into Markdown in post-processing would be easier to maintain than a solution consisting of parsing Markdown comments directly?
> 
> I did consider this approach as well, but thought it would end up being more complex. The main problem is that all but the last line comment before a method end up as orphan comments, so, to correctly reconstruct the markdown comments, it would necessary to connect these line comments by looking at token ranges and the tokens between them. While separating this logic from the parser would bring some benefit, I think it would be harder to reason about than the approach implemented in this PR where the token stream is being processed directly. It would also only work if tokens are saved (as far as I understand), which is a significant downside.

Could you add this analysis, along with the other approaches you considered, to the description of the PR, in order to document in the JP project the approach that was chosen and those that were considered? Thank you.

jlerbsc:
Can you close this PR if it is no longer relevant?

johannescoetzee:
Closed in favour of https://github.com/javaparser/javaparser/pull/4885 and https://github.com/javaparser/javaparser/pull/4899

PR #4902, opened 2025-11-20
Implement JEP 512: Compact Source Files and Instance Main Methods

### LINKED ISSUE #4869: JEP 512: Compact Source Files and Instance Main Methods

rpx99:
## Summary
Implements JEP 512, allowing Java source files to contain top-level methods and fields without an explicit class declaration.

## Changes
- Added minimal Java 25 language level support (`JAVA_25` enum, validator, post-processor)
- Added `isCompact` field to `ClassOrInterfaceDeclaration` to mark implicit classes
- Override `getNameAsString()` to return empty string for compact classes
- Extended JavaCC grammar to recognize top-level methods and fields
- Added `CompactClassMember` production for parsing top-level declarations
- Parser wraps top-level members in implicit `ClassOrInterfaceDeclaration`
- Added comprehensive tests for compact class features (7 test cases)

## Implementation Details
- Compact classes use placeholder name `"$CompactClass"` internally
- `getNameAsString()` returns `""` for compact classes to indicate unnamed/implicit class
- Regular type declarations are tried first, `CompactClassMember` as fallback
- Compact class is added at beginning of types list for proper ordering

## Examples
```java
// Simple main method (no explicit class needed)
void main() {
    System.out.println("Hello, World!");
}

// With fields and multiple methods
String greeting = "Hello";

void main() {
    helper();
}

void helper() {
    System.out.println(greeting);
}
```
Notes

    JEP 511 (Module Import Declarations) and JEP 513 (Flexible Constructor Bodies) are not yet implemented
    This PR focuses solely on JEP 512

Related Issues

Closes #4869
Testing

All existing tests pass + 7 new tests for compact class functionality:

    Simple main method
    Main with parameters
    Compact class with fields
    Multiple methods
    Public modifiers
    Mixed compact and regular classes
    Regular classes still work


rpx99:
Note: This is the initial implementation focusing on parsing compact classes. 
Still TODO for complete JEP 512 support:

1. Instance main method validation
2. Flexible main method signatures (int return type, optional params)
3. Compact class restrictions (no extends/implements)
4. Enhanced Java25Validator with JEP 512 rules

Happy to implement these in follow-up commits if you'd like! 
Let me know your preference.

rpx99:
Updated JEP 512 implementation with complete validation:
- Added CompactClassValidator with all JEP 512 rules
- Flexible main method signature validation (void/int, static/instance)
- 9 additional tests for edge cases and validation
- Total 16 tests, all passing ✅

rpx99:
@jlerbsc Passes now.

jlerbsc:
Please squash your commits.

johannescoetzee:
A minor note regarding type resolution:  Compact source files have an implicit `java.base` module import, so `List`, for example, should resolve to `java.util.List` without providing any specific imports. I'm not sure if it's better to handle this case in this PR or in https://github.com/javaparser/javaparser/pull/4905, but either way could work

rpx99:
Let me run the format checks again for this one

rpx99:
Is this build failure due to my code?

jlerbsc:
> Is this build failure due to my code?

No.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/CompactClassTest.java
```diff
@@ -0,0 +1,294 @@
...
+        ClassOrInterfaceDeclaration implicitClass = cu.getClassByName("").get();
+        MethodDeclaration mainMethod = implicitClass.getMethods().get(0);
+
+        assertFalse(mainMethod.isStatic(), "Main method can be instance in JEP 512");
+    }
+
+    @Test
+    void mainMethodWithIntReturnType() {
+        // JEP 512 allows main to return int
+        String code = "int main() { return 0; }";
```
johannescoetzee: This is not true. Trying to compile this code gives the following error
```
Test.java:1: error: compact source file does not have main method in the form of void main() or void main(String[] args)
int main() { return 0; }
```

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/CompactClassTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+    // ==================== Validation Tests: Invalid Main Signatures ====================
+
+    @Test
+    void mainMethodWithInvalidReturnType() {
+        // Main must return void or int
+        String code = "String main() { return \"invalid\"; }";
+        ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT, provider(code));
+
+        assertProblems(result, "(line 1,col 1) Main method must have return type 'void' or 'int', found: 'String'.");
```
johannescoetzee: Related to the above, the problem reporting is wrong. Main methods may only have a 'void' return type.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/CompactClassTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        assertNoProblems(result);
+        assertTrue(result.isSuccessful());
+
+        CompilationUnit cu = result.getResult().get();
+        ClassOrInterfaceDeclaration implicitClass = cu.getClassByName("").get();
+        MethodDeclaration mainMethod = implicitClass.getMethods().get(0);
+
+        assertFalse(mainMethod.isStatic(), "Main method can be instance in JEP 512");
+    }
```
johannescoetzee: This test is very similar to the `simpleMainMethod` above. I think this can be removed if you move the `assertFalse(mainMethod.isStatic(), "Main method can be instance in JEP 512");` check to that method.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -69,6 +69,14 @@ public class ClassOrInterfaceDeclaration extends TypeDeclaration<ClassOrInterfac
 
     private NodeList<ClassOrInterfaceType> permittedTypes;
 
+    /**
+     * Indicates whether this is a compact class (JEP 512 - implicit class from top-level members).
+     * Compact classes are created when parsing top-level methods and fields without an explicit class wrapper.
+     *
+     * @since 3.27.2
+     */
+    private boolean isCompact;
```
johannescoetzee, line 78: Please add a parameter corresponding to this field to the `@AllFieldsConstructor` annotated constructor since this is used for code generation and it's currently not possible to create an instance of a compact ClassOrInterfaceDeclaration without a separate `setCompact` call.

To avoid requiring users to modify existing code, it may be necessary to add another constructor with the same parameter list as the current `AllFieldsConstructor`. I'd like input from the project maintainer @jlerbsc about this, however, since adding another constructor for each added field could lead to a lot of bloat over time and make it more difficult for users to figure out which constructor should be used. Requiring users to potentially update the majority of their constructor invocations to support a field that is only relevant in a small minority of cases feels very disruptive, however.

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/chunks/CompactClassValidator.java
```diff
@@ -0,0 +1,131 @@
...
+     * JEP 512 allows flexible main method signatures:
+     * - Can be static or instance
+     * - Can be public or package-private
+     * - Can return void or int
+     * - Can have String[] parameter or no parameters
+     */
+    private void validateMainMethods(ClassOrInterfaceDeclaration node, ProblemReporter reporter) {
+        node.getMethodsByName("main").forEach(method -> {
+            validateMainMethodSignature(method, reporter);
+        });
```
johannescoetzee, line 89: It is not correct to validate every method called `main` this way. JEP 512 only requires that a main method with a valid "main" signature exists in a compact class, not that every main method has such a signature. I'm not sure whether this is something we even need a validator for, however. As far as I understand, the point of the validators is not to re-implement the java compiler's error checking, but rather to report when language features not matching the selected version are used. Methods called `main` with any signature are already allowed in general. The signature of the main method only matters when trying to determine the entry-point for the program, which isn't something JavaParser attempts to do in general. We also assume that input code compiles, so checking the signature of `main` is redundant.

## Example 1 does not compile
```
void main(String[] args, int x) { }
```
## Example 2 does compile
```
void main(String[] args, int x) { }

void main(String[] args) { }
```
## Example 3 does compile
```
class Foo {
  // This isn't a valid entry point into the application, but is a valid method regardless
  void main(String[] args, int x) { }
}
```

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/Java25Validator.java
```diff
@@ -0,0 +1,57 @@
...
+     * - Compact classes are implicitly final
+     * - Main methods have valid signatures (instance or static, void or int return)
+     */
+    final Validator compactClassValidator =
+            new SingleNodeTypeValidator<>(ClassOrInterfaceDeclaration.class, new CompactClassValidator());
+
+    public Java25Validator() {
+        super();
+        // JEP 512: Compact Source Files and Instance Main Methods
+        add(compactClassValidator);
```
johannescoetzee: I do not think a complex validator is needed here (see my comment below for more of a discussion about why). Instead, I think a Java1Validator should be added to validate that no compact classes are created and then that validator should be removed here since that feature is available for Java versions >= 25. Checking details like whether the compact class extends other classes is effectively just checking for JavaParser bugs and should rather be covered by unit tests. Checking whether a valid main method exists is the responsibility of the java compiler and JavaParser assumes any input code compiles, so checking that is redundant.

javaparser-core/src/main/javacc/java.jj
```diff
@@ -1202,22 +1204,63 @@ public CompilationUnit CompilationUnit():
...
+                );
+                compactClass.setCompact(true);
+
+                // Add compact class at the beginning of types list
+                NodeList<TypeDeclaration<?>> newTypes = emptyNodeList();
+                newTypes = add(newTypes, compactClass);
+                for (TypeDeclaration<?> type : types) {
+                    newTypes = add(newTypes, type);
+                }
+                types = newTypes;
```
johannescoetzee: This isn't correct. With the way compact classes are handled here
```
void main() {}

class Foo {}
```
will effectively be represented as 
```
class UnnamedClass {
  void main() {}
}

class Foo {}
```
when it should be represented as
```
class UnnamedClass {
  void main() {}

  class Foo {}
}
```
The explanation for this is in https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.1.8
> The body of the class contains every class member declared in the compact compilation unit (these are declarations of fields ([§8.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.3)), methods ([§8.4](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.4)), member classes ([§8.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.5)), and member interfaces ([§9.1.1.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-9.html#jls-9.1.1.3))). It is not possible for a compact compilation unit to contain declarations of instance initializers ([§8.6](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.6)), static initializers ([§8.7](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.7)), or constructors ([§8.8](https://docs.oracle.com/javase/specs/jls/se25/html/jls-8.html#jls-8.8)).

where

> A member class is a class whose declaration is directly enclosed in the body of another class or interface declaration

javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/CompactClassTest.java
```diff
@@ -0,0 +1,294 @@
...
+    @Test
+    void mixedCompactAndRegularClass() {
+        String code = "void main() { new Helper().help(); }\n" + "class Helper { void help() { } }";
+        ParseResult<CompilationUnit> result = javaParser.parse(COMPILATION_UNIT, provider(code));
+
+        assertNoProblems(result);
+        assertTrue(result.isSuccessful());
+
+        CompilationUnit cu = result.getResult().get();
+        assertEquals(2, cu.getTypes().size(), "Should have compact class and regular class");
```
johannescoetzee: `Helper` should be a nested class in the compact class. See my comment on the `java.jj` changes for a more detailed explanation.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -364,6 +372,49 @@ public void ifClassOrInterfaceDeclaration(Consumer<ClassOrInterfaceDeclaration>
...
+     * Returns the name of this class as a string.
+     * For compact classes (JEP 512), this returns an empty string to indicate the implicit/unnamed class.
+     *
+     * @return the class name, or empty string for compact classes
+     * @since 3.27.2
+     */
+    @Override
+    public String getNameAsString() {
+        if (isCompact) {
+            return "";
```
johannescoetzee: I can see 2 issues here:
1. I'm not sure if the empty string is a good name to use. It makes the naming of nested classes awkward. I think that keeping a placeholder name would be better for consistent naming (but this is subjective so please share your thoughts on this @jlerbsc)
2. Overriding the name here is very brittle. `compactClass.getName().toString()` still returns `$CompactClass`, for example. I assume the pretty printer and various other mechanisms to get a string representation for the class would have the same issue.

johannescoetzee (changes requested):
There are a few issues that need to be fixed before this can be merged. PrettyPrinter and LexicalPreservingPrinter tests are also missing and compact classes would almost certainly require special handling in both of these. The required steps to get this mergeable (in my opinion, but @jlerbsc would have the final say) would be:
1. Fix the issues I've pointed out so far
2. Go through the unit tests and confirm that all of the code examples are valid compiling code (unless there is a reason to test something that doesn't compile, in which case you should please add a motivation explaining why it is necessary or useful)
3. Go through the assumptions made in this PR (what the unit tests are testing, how the output AST is structured, etc.) and make sure all of those assumptions are backed by the JEP or Java language specification.
4. Go through https://github.com/javaparser/javaparser/wiki/A-Detailed-Guide-to-Adding-New-Nodes-and-Fields and ensure all steps have been followed (this includes support for the printers that I mentioned above)

johannescoetzee:
@rpx99 AI code generation can be a powerful tool, but please check the output thoroughly before opening a PR. A few of the errors were easy to catch just by running the test examples through the Java compiler and I do not think the burden of doing so should lie on the reviewers.

Adding feature support to JavaParser is a challenging task since there are so many connected components to consider, but if you are interested in finishing this implementation, please feel free to ask if you get stuck on something. I'm always happy to share my knowledge with anyone who is interested :)

javaparser-core/src/main/java/com/github/javaparser/ast/validator/postprocessors/Java25PostProcessor.java
```diff
@@ -16,12 +16,13 @@
  * JavaParser is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU Lesser General Public License for more details.
+ * GNU Lesser General Public Package for more details.
```
johannescoetzee: This should not be changed. Please review the generated code before pushing it.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
jlerbsc: > To avoid requiring users to modify existing code, it may be necessary to add another constructor with the same parameter list as the current `AllFieldsConstructor`. I'd like input from the project maintainer @jlerbsc about this, however, since adding another constructor for each added field could lead to a lot of bloat over time and make it more difficult for users to figure out which constructor should be used. Requiring users to potentially update the majority of their constructor invocations to support a field that is only relevant in a small minority of cases feels very disruptive, however.

As you pointed out, I think we need to add a new constructor that will be referenced by all the others. I don't think there will be many more changes to this class in the future, and this will have less impact on code using JavaParser.

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/chunks/CompactClassValidator.java
jlerbsc: > We also assume that input code compiles

That is indeed a requirement.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
jlerbsc: Does the JLS specify anything on this subject? Perhaps a test on a compact class containing an included class could improve our understanding of what needs to be done.

jlerbsc:
My observation is that this is too a complex subject for someone who is unfamiliar with the project. AI cannot compensate for this knowledge. You have worked hard, but we cannot spend our time checking and commenting on all your iterations. Perhaps it would have been more reasonable to start by fixing simpler issue.
@johannescoetzee, I suggest you continue this work in another PR, while of course retaining a reference to the work that has been done by @rpx99 . Thank you for your contribution.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
johannescoetzee: JEP 512 says the following:
> A compact source file declares a class implicitly, so the class does not have a name that can be used in source code. The Java compiler generates a class name when compiling a compact source file, but that name is implementation-specific and should not be relied upon in any source code — not even source code in the compact source file itself.

OpenJDK defaults to using the source filename for the class name which is a sensible choice consistent with Java convention, but that information won't always be available. Perhaps an approach where we use the filename if it is available, otherwise fall back to some (perhaps configurable) default naming scheme could work.

rpx99:
@johannescoetzee Please take over. Thanks.

jlerbsc:
I am closing this PR. Thank you for your involvement. This experience shows us that AI-generated PRs cannot be integrated as they are and that an experienced eye is needed to adapt the AI's proposal. For your next PR, I suggest you choose much less complex topics to familiarise yourself with how JavaParser works.

rpx99:
Yes, it was very interesting - i did not read any spec at all or knew what to do. So it is huge to see what Claude was capable to code. Think GPT 5.2 think or so released yesterday could have solved it properly!

jlerbsc:
What you say confirms my initial impression. I think that in future I will reject all PRs generated by AI, because not only are the results unreliable, but contributors are tempted not to invest themselves in understanding the project, and maintainers spend much more time validating PRs. We mustn't forget that we devote time to this project that we could otherwise spend on our hobbies or with our families...

rpx99:
You can of course do that and i am sorry it did not work out perfectly. But, this won't stop the future. AI will do this soon like a REAL expert.

PR #4903, opened 2025-11-21
Implement JEP 513: Flexible Constructor Bodies

rpx99:
## Summary

This PR implements **JEP 513: Flexible Constructor Bodies** for Java 25, which allows statements to appear before explicit constructor invocations (`super()` or `this()` calls).

### Grammar
- Modified `ConstructorDeclaration` production to allow statements in any order
- Added `ExplicitConstructorInvocation` as a valid `BlockStatement` 
- Constructor invocations can now appear anywhere in the constructor body, not just at the beginning

### Validation
- Created `FlexibleConstructorValidator` to enforce JEP 513 rules:
  - Statements before `super()`/`this()` (prologue) cannot reference the current instance (`this`)
  - Only one `super()`/`this()` call is allowed per constructor
- Added `Java25Validator` with the flexible constructor validator
- Added `Java25PostProcessor`

### Configuration
- Added `JAVA_25` language level to `ParserConfiguration`
- Updated `BLEEDING_EDGE` constant to `JAVA_25`
- Added `JAVA_25` to yield support array

### Tests
Created `FlexibleConstructorTest` with 10 comprehensive test cases:

**Valid cases:**
- ✅ Validation before `super()` call
- ✅ Field initialization before `super()` 
- ✅ Multiple statements before `super()`
- ✅ Statements before `this()` call
- ✅ Statements after `super()` call
- ✅ Constructor without explicit invocation

**Invalid cases:**
- ❌ `this` reference in prologue
- ❌ `this` method call in prologue
- ❌ Multiple explicit constructor calls
- ❌ `this` in field initializer before super

All tests pass successfully.

## Related
- Implements: https://openjdk.org/jeps/513
- Related issue: https://github.com/javaparser/javaparser/issues/4869


jlerbsc:
Please update your version before submitting a PR and finalise the PR https://github.com/javaparser/javaparser/pull/4902 before resubmitting it.

jlerbsc:
Thank you for your work. Could you please remove everything related to JEP 512 from the PR?

javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
```diff
@@ -186,7 +186,11 @@ public enum LanguageLevel {
         /**
          * Java 22
          */
-        JAVA_22(new Java22Validator(), new Java22PostProcessor());
+        JAVA_22(new Java22Validator(), new Java22PostProcessor()),
```
jlerbsc, line 189: Could you update the code from the latest version of JP?

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/chunks/FlexibleConstructorValidator.java
```diff
@@ -0,0 +1,116 @@
...
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.ast.validator.language_level_validations.chunks;
```
jlerbsc, line 21: Could you remove the ‘chunk’ package and integrate the content of this class into the validator (as is done for the other validators)?

jlerbsc:
@johannescoetzee Have you had a look at this PR and do you agree?

johannescoetzee:
> @johannescoetzee Have you had a look at this PR and do you agree?

I haven't had a look yet, but will do so now

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/chunks/FlexibleConstructorValidator.java
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Validates JEP 513: Flexible Constructor Bodies.
+ *
+ * Rules:
+ * 1. Statements before super()/this() (prologue) cannot reference 'this'
+ * 2. Only one super()/this() call allowed per constructor
+ *
+ * @since 3.27.0
+ */
+public class FlexibleConstructorValidator implements TypedValidator<ConstructorDeclaration> {
```
johannescoetzee, line 41: My concern with this validator is similar to my comment on github.com/javaparser/javaparser/pull/4902. These checks are effectively re-running checks that would've been done by the compiler and checking that we report problems for code that isn't valid in ANY Java version. 

The validators should check that, when using a language level < 25, code that wouldn't have been valid for those versions but now parses correctly for java >= 25 is correctly reported as not being valid in that version. This means that a java 1 validator must be added to check this which is removed in java 25

javaparser-core/src/main/javacc/java.jj
```diff
@@ -2547,17 +2610,12 @@ ConstructorDeclaration ConstructorDeclaration(ModifierHolder modifier):
         ("," throwType = AnnotatedReferenceType() { throws_ = add(throws_, throwType); })*
     ]
     "{" { blockBegin=token(); }
-    [
-        LOOKAHEAD(ExplicitConstructorInvocation())
-        exConsInv = ExplicitConstructorInvocation()
-    ]
+    // JEP 513: Flexible Constructor Bodies
+    // ExplicitConstructorInvocation can now appear anywhere in the body, not just at the beginning
```
johannescoetzee, line 2614: Without the context of the lines that have been removed, this comment does not make sense. I would remove it since the same is already explained below.

javaparser-core/src/main/java/com/github/javaparser/ast/validator/language_level_validations/chunks/FlexibleConstructorValidator.java
johannescoetzee: > Could you remove the ‘chunk’ package and integrate the content of this class into the validator (as is done for the other validators)?

I agree with this idea. If the code for a validator method is too complex to reasonably include in the validator class itself, then I think the validation logic should be reconsidered. It may make sense in some cases, but most validator should just check whether a flag is set or a simple condition (in this case whether any statements appear before the constructor invocation)

johannescoetzee (changes requested):
At a quick glance the changes to java.jj look good, but I didn't go through the tests in detail to check whether they compile and match expectations. I have a few requests. @rpx99 could you please:
-  confirm whether you've checked that all the tests samples that should compile do, and that the expectations are consistent with the JEP. 
- check which of the tests are just repeating checks that are already performed by the compiler and remove those.
- add some AST tests as well. The validator tests here test that we don't encounter problems, but not that the AST output is correct.
- update the validator according to my comment 

It is difficult to review this PR while changes from JEP 512 are included (I prefer to look at the full diff to get a holistic idea of the change instead of per-commit) so I just did a quick read-through. I will do a more thorough review once the JEP 512 changes have been removed an my initial comments have been adressed.

rpx99:
@johannescoetzee Please take over. Thanks.

PR #4905, opened 2025-11-24
[JEP 511] Module Import Declarations

### LINKED ISSUE #4848: JEP 511 : Module Import Declarations
Java 25 introduces the ability to import required packages into Java source code with a single import module statement in the Java source file. This simplifies dependency management and improves code readability by reducing the number of lines required to import multiple packages from the same module.

johannescoetzee:
Fixes https://github.com/javaparser/javaparser/issues/4848

This PR adds parser support for module import statements, as well as support for resolving types from modules.

## Changes to ImportDeclaration/CompilationUnit

Constructors and an `addImport` method with an option to specify whether this is a module import or not (indicated by the value of the new `isModule` field in ImportDeclaration) were added. The old versions of these constructors and this method without the `isModule` parameter were kept to avoid breaking user code.

## JarTypeSolver

Javassist does not provide much support for handling modules, so I ended up parsing the module attribute manually. This is a bit cumbersome, but does work and I hope the comments explain what is happening sufficiently well.

## ClassLoaderTypeSolver

I couldn't figure out a way to do this without using Java 9+ types and methods, so I used a lot of reflection for the implementation to maintain compatibility with Java 8 (although this will always fail if JavaParser is run with Java 8). I also couldn't figure out a good way to get to a list of modules loaded by a classloader (the mechanisms I could find that looked promising were all JDK-internal or blocked for security reasons). The solution as it is presented here requires users to provide a list of module layers that the typesolver will use to find modules. This is done automatically for the ReflectionTypeSolver.


johannescoetzee:
I realised while reviewing https://github.com/javaparser/javaparser/pull/4902 that I forgot the validator and printer support for this. I'm closing this for now and will re-open it once that's done.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
```diff
@@ -39,9 +41,60 @@ public class ClassLoaderTypeSolver implements TypeSolver {
...
+        addModuleLayers(moduleLayers);
+    }
+
+    public void addModuleLayers(Iterable<Object> moduleLayers) {
+        for (Object moduleLayer : moduleLayers) {
+            if (!moduleLayer.getClass().getCanonicalName().equals("java.lang.ModuleLayer")) {
+                continue;
+            }
+            try {
+                /* This code is equivalent to the snippet below, but is done with reflection to maintain compatibility
```
jlerbsc, line 68: Can you isolate this code in a class that could be called, for example, ModuleLayerHelper?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
```diff
@@ -68,17 +121,24 @@ protected boolean filterName(String name) {
     @Override
     public SymbolReference<ResolvedReferenceTypeDeclaration> tryToSolveTypeInModule(String qualifiedModuleName, String simpleTypeName) {
```
jlerbsc: Is it really necessary to create a new method to handle this particular case?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ReflectionTypeSolver.java
```diff
@@ -77,6 +79,22 @@ public class ReflectionTypeSolver extends ClassLoaderTypeSolver {
     public ReflectionTypeSolver(Predicate<String> classFilter) {
         super(ReflectionTypeSolver.class.getClassLoader());
         this.classFilter = classFilter;
+
+        HashSet<Object> moduleLayers = new HashSet<>();
+        try {
+            /* This code is equivalent to the snippet below, but is done with reflection to maintain compatibility with
```
jlerbsc, line 85: Probably to be declared in a ‘ModuleLayerHelper’ class.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue1364Test.java
```diff
@@ -69,6 +69,11 @@ public SymbolReference<ResolvedReferenceTypeDeclaration> tryToSolveType(String n
 
                 return SymbolReference.unsolved();
             }
+
```
jlerbsc, line 72: To make the PRs easier to read, would it be possible to isolate the JP test cases in a single commit, and also isolate the code generated by JP, which is specific to printers, the features of the symbol resolver, and the tests which are specific to symbol resolvers?

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
```diff
@@ -193,6 +200,59 @@ private void addPathToJar(String pathToJarOrClassFileHierarchy) throws IOExcepti
         }
     }
 
+    private void registerModuleInfo() {
```
jlerbsc, line 203: This code could be externalised into a helper class until the Javassist project offers this functionality.
Perhaps we could either report an issue or make a contribution to the project.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
```diff
@@ -79,11 +77,16 @@ public void addModuleLayers(Iterable<Object> moduleLayers) {
                  *   ...
                  * }
                  */
-                Set<Object> modulesSet = (Set<Object>) moduleLayer.getClass().getMethod("modules").invoke(moduleLayer);
+                Set<Object> modulesSet = (Set<Object>)
```
jlerbsc, line 80: To outsource.

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javaparsermodel/declarations/JavaParserVariableDeclarationTest.java
```diff
@@ -91,21 +95,49 @@ void test3631() {
...
 
         List<VariableDeclarator> variables = cu.findAll(VariableDeclarator.class);
 
         ResolvedValueDeclaration rvd = variables.get(0).resolve();
 
         assertEquals("java.util.List<java.lang.String>", rvd.getType().describe());
     }
+
+    @Test
+    void testJavaModuleImportFromSource() {
```
jlerbsc, line 118: @EnabledForJreRange ?

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/javassistmodel/JavassistModuleTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
+
+class JavassistModuleTest extends AbstractSymbolResolutionTest {
+
+    private TypeSolver typeSolver;
+
+    @BeforeEach
+    void setup() throws IOException {
+        Path pathToJar = adaptPath("src/test/resources/modules/moduletest/com-github-javaparser-testmodule.jar");
```
jlerbsc, line 46: Can you delete the directory "moduletest" from the path?

jlerbsc:
Thank you for this PR. I have a few comments and I have noted that you have other points to consider.

johannescoetzee:
I opened https://github.com/javaparser/javaparser/pull/4910 with the changes requested here. I could not reopen this PR since I force pushed changes

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
johannescoetzee: Done

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
johannescoetzee: A new method may not be the only solution, but we do need to keep track of the fact that we want to solve the type in a module, as well as the name of the module. This is not information that is currently tracked with the regular `solveType` methods. Since we know exactly when solving a type in a module is needed (that only being when handling module imports), I think a separate method is a much cleaner approach than trying to modify the already complex `solveType` structure/API to record the information we need, while also greatly reducing the risk of introducing new bugs into existing resolution logic.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/JarTypeSolver.java
johannescoetzee: I think creating an issue is a good idea just to get an idea of whether this is something they would be interested in. The implementation here is very specific to what we need, so a more general approach would require some thought and could lead to some questions about attribute handling in javassist in general, which may be beyond the scope of what they want to do. I'll collect my thoughts on the issue and ask them about it

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/resolution/typesolvers/ClassLoaderTypeSolver.java
jlerbsc: > A new method may not be the only solution, but we do need to keep track of the fact that we want to solve the type in a module, as well as the name of the module. This is not information that is currently tracked with the regular `solveType` methods. Since we know exactly when solving a type in a module is needed (that only being when handling module imports), I think a separate method is a much cleaner approach than trying to modify the already complex `solveType` structure/API to record the information we need, while also greatly reducing the risk of introducing new bugs into existing resolution logic.

I do understand the concerns raised. I was thinking of scanning all modules for the class declaration, but that wouldn't be very efficient.

PR #4910, opened 2025-12-02
[JEP 511] Module Import Declarations

### LINKED ISSUE #4848: JEP 511 : Module Import Declarations

johannescoetzee:
I tried to re-open https://github.com/javaparser/javaparser/pull/4905, but since I force-pushed changes to restructure the commits as requested this option was no longer available. 

Fixes https://github.com/javaparser/javaparser/issues/4848

This PR adds parser support for module import statements, as well as support for resolving types from modules.

## Changes to ImportDeclaration/CompilationUnit

Constructors and an `addImport` method with an option to specify whether this is a module import or not (indicated by the value of the new `isModule` field in ImportDeclaration) were added. The old versions of these constructors and this method without the `isModule` parameter were kept to avoid breaking user code.

## JarTypeSolver

Javassist does not provide much support for handling modules, so I ended up parsing the module attribute manually. This is a bit cumbersome, but does work and I hope the comments explain what is happening sufficiently well.

## ClassLoaderTypeSolver

I couldn't figure out a way to do this without using Java 9+ types and methods, so I used a lot of reflection for the implementation to maintain compatibility with Java 8 (although this will always fail if JavaParser is run with Java 8). I also couldn't figure out a good way to get to a list of modules loaded by a classloader (the mechanisms I could find that looked promising were all JDK-internal or blocked for security reasons). The solution as it is presented here requires users to provide a list of module layers that the typesolver will use to find modules. This is done automatically for the ReflectionTypeSolver.


javaparser-core/src/main/java/com/github/javaparser/printer/PrettyPrintVisitor.java
```diff
@@ -1792,6 +1792,9 @@ public void visit(final ImportDeclaration n, final Void arg) {
         if (n.isStatic()) {
             printer.print("static ");
         }
+        if (n.isModule()) {
+            printer.print("static ");
```
jlerbsc: ?

jlerbsc: Is the PR finished? If so, as soon as you have fixed this small issue, I will validate this PR. Thank you for your contribution.

johannescoetzee: Oops, that was a copy-paste issue. I would've thought that the tests I added would catch this, but apparently this visitor isn't being tested correctly. Tomorrow morning I'll update the test to ensure this is actually covered and fix this typo, but after that this PR is finished if you're satisfied with the rest

jlerbsc: > but after that this PR is finished if you're satisfied with the rest

The solution of creating a new method for resolving types does not satisfy me intellectually, but I have no other solution to propose that would be as effective.

johannescoetzee: I've fixed the static/module issue. I saw this is in the deprecated PrettyPrintVisitor and it looks like this visitor either isn't being tested at all, or only in few specific testcases. Most testcases use the DefaultPrettyPrinter instead. If we want to keep the PrettyPrintVisitor, ideally we'd refactor it so that it just uses the DefaultPrettyPrinter methods, removing the need to maintain two largely identical implementations. I haven't looked into the configuration options enough to get an idea of how hard porting the configuring aspect will be, but the printing methods themselves should be simple. Regardless, I think refactoring the printer and fixing the testing situation is probably best handled separately.

jlerbsc: It may be time to do some cleaning up and remove this part of the code. Could you create an issue to address this topic? However, I would prefer to postpone resolving this new issue until later, as I do not want to include it in the next version of JavaParser to avoid too much impact on users' code.

jlerbsc:
Thank you for this contribution. It is very readable and understandable.

PR #4919, opened 2025-12-10
[JEP 513] Add support for flexible constructor bodies

### LINKED ISSUE #4849: JEP 513 : Flexible Constructor Bodies
In previous LTS versions, it was not possible to write instructions before calling the class's parent constructor (super() method). In some cases, this could result in unnecessary calls, particularly when parameter validations were required before initialising the object. The language now allows statements to be written before the call to super(), which offers more flexibility in managing object initialisation.

johannescoetzee:
Fixes https://github.com/javaparser/javaparser/issues/4849
Replaces https://github.com/javaparser/javaparser/pull/4903

This PR removes the restriction that explicit constructor invocations must appear as the first statement in a constructor body. See https://openjdk.org/jeps/513 for more details regarding the motivations behind this change.

The grammar change was inspired by https://github.com/javaparser/javaparser/pull/4903 opened by @rpx99, although I did add my own comment explaining why handling it this way should be fine, even if it isn't 100% correct according to the JLS (we handle compiling code correctly, but also parse some non-compiling code without reporting errors; if users choose to parse non-compiling code, then they will need to do their own validation according to their needs).

I've also included a simplified validator compared to https://github.com/javaparser/javaparser/pull/4903 with validator tests and AST tests.

Finally, I've also included pretty printer and LPP tests, but didn't need to update either of those to support this.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/ConstructorDeclarationTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/ConstructorDeclarationTest.java
@@ -21,8 +21,18 @@
 
 package com.github.javaparser.ast.body;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
+import static com.github.javaparser.Providers.provider;
+import static com.github.javaparser.utils.TestUtils.assertNoProblems;
+import static org.junit.jupiter.api.Assertions.*;
 
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParseResult;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.stmt.Statement;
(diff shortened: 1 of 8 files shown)


johannescoetzee:
@jlerbsc I would like to discuss some things regarding the project with you (how we could best collaborate, etc.) and sent you a message on the JavaParser Gitter to that effect. Since that doesn't seem to be a platform either of us uses regularly, feel free to contact me via email (should be visible in my commit information) or suggest a suitable platform if you're interested :)

jlerbsc:
Does this seem correct to me? Can I merge this PR?

jlerbsc:
> @jlerbsc I would like to discuss some things regarding the project with you (how we could best collaborate, etc.) and sent you a message on the JavaParser Gitter to that effect. Since that doesn't seem to be a platform either of us uses regularly, feel free to contact me via email (should be visible in my commit information) or suggest a suitable platform if you're interested :)

I have replied to your message on Gitter. I hope you find my answers helpful.

johannescoetzee:
> Does this seem correct to me? Can I merge this PR?

Yes, this can be merged if you're satisfied with it

jlerbsc:

PR #4930, opened 2025-12-18
Add UnaryExpr, BinaryExpr, and some record/enum tests to improve overall test coverage

johannescoetzee:
While working on https://github.com/javaparser/javaparser/pull/4929 and the upcoming compact class PR I added the following tests. These are mostly sanity-checks that already passed and aren't strictly related to either of those changes, but are still worth having for the sake of overall test coverage to potentially catch future regressions. I think it makes more sense to add as a separate PR instead of including this in either of the other PRs, distracting from the main purpose of those.

These tests are mostly AI-generated, but I did verify them myself and fixed a few issues manually.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/RecordDeclarationTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/body/RecordDeclarationTest.java
@@ -714,6 +714,64 @@ public class RecordDeclarationTest {
         });
     }
 
+    /**
+     * "module" became a keyword in Java 9, but can still be used as an identifier
+     * in certain contexts. This test verifies the AST for a record named "module"
+     * that also uses "module" as a type and in object creation.
+     */
+    @Test
+    void recordWithModuleAsName() {
+        String s = "record module(String s) {\n"
+                + "  void foo() {\n"
+                + "    module m = new module(\"hello\");\n"
+                + "  }\n"
+                + "}\n";
+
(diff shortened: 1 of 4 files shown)


johannescoetzee:
Check failed due to a connection issue:
```
Error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond. (api.github.com:443)
```

johannescoetzee:
@jlerbsc what are your thoughts on adding tests like these? Should I just keep adding tests for simple cases like these when there's a gap in our test coverage, even if the tests themselves are very basic?

jlerbsc:
> @jlerbsc what are your thoughts on adding tests like these? Should I just keep adding tests for simple cases like these when there's a gap in our test coverage, even if the tests themselves are very basic?

If you think these tests add value, then you can write them, but don't go overboard, as it can become time-consuming. For me, the most important thing is to fix bugs to make Javaparser more reliable.

johannescoetzee:
> > @jlerbsc what are your thoughts on adding tests like these? Should I just keep adding tests for simple cases like these when there's a gap in our test coverage, even if the tests themselves are very basic?
> 
> If you think these tests add value, then you can write them, but don't go overboard, as it can become time-consuming. For me, the most important thing is to fix bugs to make Javaparser more reliable.

I wouldn't go out of my way to write tests, but I wrote these as part of figuring out what works and what doesn't when investigating the grammar issue

PR #4940, opened 2025-12-24
[JEP 512] Add support for compact source files

### LINKED ISSUE #4869: JEP 512: Compact Source Files and Instance Main Methods

johannescoetzee:
Resolves https://github.com/javaparser/javaparser/issues/4869 and replaces https://github.com/javaparser/javaparser/pull/4902 by @rpx99. I adapted the logic introduced to `java.jj` in that PR to fix an issue with the ordering of the members of the compact class.

This PR adds support for compact source files/classes introduced in Java 25. It does so by explicitly creating the `ClassOrInterfaceDeclaration` node for the compact class with a new `isCompact` field set to true. 

Adding support for the LPP turned out to be quite complicated. Support for adding/removing members from a compact class was fairly straightforward, but I ran into lots of issues when adding support for converting a non-compact class a compact class and vice-versa. The implementation here still doesn't handle whitespace 100% correctly, but it's close to correct and I've already spent about as much time as I reasonably can debugging it. In my opinion this niche case should not be a blocker, especially considering the LPP does support "normal" operations on these classes.


jlerbsc:
@johannescoetzee The PR will be merged as soon as the errors have been corrected. A new version of JavaParser will then be released.

 JavaParserVariableDeclarationTest.javaBaseTypeFromImplicitCompactClassImport:171 » UnsolvedSymbol Unsolved symbol : List

johannescoetzee:
I missed this testcase since I was running all the tests with Java 8, apparently. It's all fixed now

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
```diff
@@ -196,6 +197,25 @@ public ClassOrInterfaceDeclaration(
         customInitialization();
     }
 
+    /**
```
jlerbsc, line 200: I didn't really understand what the point of it was.

javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/JavaParserTypeDeclarationAdapter.java
```diff
@@ -81,6 +83,17 @@ public SymbolReference<ResolvedTypeDeclaration> solveType(String name, List<Reso
             return SymbolReference.solved(JavaParserFacade.get(typeSolver).getTypeDeclaration(wrappedNode));
         }
 
+        if (this.wrappedNode.isClassOrInterfaceDeclaration()
```
jlerbsc, line 86: Isn't this redundant with what is already done in the ClassOrInterfaceDeclarationContext class?

johannescoetzee: What I did in the ClassOrInterfaceDeclarationContext class was actually incorrect. That logic was triggered for solving symbols as values (for example resolving the identifier `listMember` to the relevant member), so it was actually just attempting to resolve names like `java.util.listMember`. I've removed this logic from `ClassOrInterfaceDeclarationContext` and added a test to demonstrate that resolving such symbols works.

javaparser-core/src/main/java/com/github/javaparser/ast/body/ClassOrInterfaceDeclaration.java
johannescoetzee: I've updated the comment to simplify it somewhat. Please let me know if it's still unclear

jlerbsc:
Have you finished this PR?

jlerbsc:
I will have to be away for two or three days, but as soon as I am back, I will create a new version of JP.

jlerbsc:
Thank you for this important new PR

PR #4950, opened 2026-01-12
Fix: issue 4949 Parsing issue in switch with yield with Java 25 parser configuration

### LINKED ISSUE #4949: Parsing issue in switch with yield with Java 25 parser configuration
Hi,

We basically noticed the same issue as mentioned in #4549. A class using "yield" can not be parsed by JavaParser, although it is valid Java 25 code.

```java
final ParserConfiguration configuration = new ParserConfiguration( );
// configuration.setLanguageLevel( ParserConfiguration.LanguageLevel.JAVA_21 );
final JavaParser javaParser = new JavaParser( configuration );
final ParseResult<CompilationUnit> parseResult = javaParser.parse( """
		class A {
		
			public static String doSomething( int value ) {
				return switch ( value ) {
					case 0 -> "Zero";
					case 1 -> {
						System.out.println( value );
						yield "One";
					}
					default -> "N/A";
				};
			}
		
		}
		""" );
System.out.println( parseResult.isSuccessful( ) );
```

The above code prints "false" for language level JAVA_25, but "true" for JAVA_21. This should not be the case. The code remains valid in Java 25 and should thus be parsable. 

Thank you and best regards

  Nils

jlerbsc:
Fixes #4949 .

Support for the yield statement must be added for each version of Java that supports it.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/ParserConfiguration.java
@@ -215,12 +215,12 @@ public class ParserConfiguration {
         /**
          * The latest Java version that is available.
          */
-        public static LanguageLevel CURRENT = JAVA_18;
+        public static LanguageLevel CURRENT = JAVA_25;
 
         /**
          * The newest Java features supported.
          */
-        public static LanguageLevel BLEEDING_EDGE = JAVA_24;
+        public static LanguageLevel BLEEDING_EDGE = JAVA_25;
 
         final Validator validator;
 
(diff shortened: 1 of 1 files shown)


martinaldrin:
Thanks for a quick fix.  when can we expect a new release?

jlerbsc:
Unfortunately, I will not be releasing a new version with just one bug fix. At best, I can plan for a new version next month.

martinaldrin:
but isn't that the purpose of the minor version, e.g. 3.28.1 to allow patches on previous versions?

jlerbsc:
Please remember that this work is done in our spare time. It is time that we are not spending with our families, children, and friends, and we are not paid for the time spent.

martinaldrin:
Hi, I fully understand that. I did not expect that a release is time consuming. I thought it was kind of automatic.

jlerbsc:
You can also build the snapshot version yourself by cloning the repository and using the maven install command.

PR #4968, opened 2026-02-03
Implement switch expr type extraction

mrjameshamilton:
I came across an `UnsupportedOperationException` like the following when experimenting with parsing switch expressions:

```
java.lang.UnsupportedOperationException: com.github.javaparser.ast.expr.SwitchExpr
	at com.github.javaparser.symbolsolver.javaparsermodel.DefaultVisitorAdapter.visit(DefaultVisitorAdapter.java:530)
	at com.github.javaparser.symbolsolver.javaparsermodel.DefaultVisitorAdapter.visit(DefaultVisitorAdapter.java:37)
	at com.github.javaparser.ast.expr.SwitchExpr.accept(SwitchExpr.java:90)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getTypeConcrete(JavaParserFacade.java:670)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:424)
	at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.getType(JavaParserFacade.java:389)
	at com.github.javaparser.symbolsolver.JavaSymbolSolver.calculateType(JavaSymbolSolver.java:443)
```

The cause was that the TypeExtractor did not implement the type extraction for switch expressions. For example, I had a case like this which threw the exception because the type of the method call argument (the result of the switch expression) could not be computed:

```
    methodCall(
        switch ("a") {
          case "a" -> 3;
          default -> 0;
        });
```

The PR implements:

* Type extraction for switch expressions based on [JLS 15.28.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.28.1)
<quote>
If the result expressions all have the same type (which may be the null type), then that is the type of the switch expression.

Otherwise, if the type of each result expression is boolean or Boolean, then an unboxing conversion ([§5.1.8](https://docs.oracle.com/javase/specs/jls/se21/html/jls-5.html#jls-5.1.8)) is applied to each result expression of type Boolean, and the switch expression has type boolean.

Otherwise, if the type of each result expression is convertible to a numeric type ( [§5.1.8](https://docs.oracle.com/javase/specs/jls/se21/html/jls-5.html#jls-5.1.8)), then the type of the switch expression is the result of general numeric promotion ([§5.6](https://docs.oracle.com/javase/specs/jls/se21/html/jls-5.html#jls-5.6)) applied to the result expressions.

Otherwise, boxing conversion ([§5.1.7](https://docs.oracle.com/javase/specs/jls/se21/html/jls-5.html#jls-5.1.7) ) is applied to each result expression that has a primitive type, after which the type of the switch expression is the result of applying capture conversion ([§5.1.10](https://docs.oracle.com/javase/specs/jls/se21/html/jls-5.html#jls-5.1.10)) to the least upper bound ([§4.10.4](https://docs.oracle.com/javase/specs/jls/se21/html/jls-4.html#jls-4.10.4)) of the types of the result expressions.
</quote>

* Validation for switch expressions with no result expressions based on [15.28.1](https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.28.1)

<quote>
It is a compile-time error if a switch expression has no result expressions.
</quote>

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/SwitchExprTest.java
@@ -78,12 +78,15 @@ class SwitchExprTest {
 
     @Test
     void aThrowStatement() {
-        SwitchExpr switchExpr = parseExpression(
-                        "switch (k) {\n" + "        case 1 -> throw new Exception(\"one\");\n" + "    }")
+        SwitchExpr switchExpr = parseExpression("switch (k) {\n"
+                        + "        case 1 -> throw new Exception(\"one\");\n"
+                        + "        case 2 -> 42;\n"
+                        + "    }")
                 .findFirst(SwitchExpr.class)
                 .get();
 
         assertEquals(THROWS_STATEMENT, switchExpr.getEntry(0).getType());
+        assertEquals(EXPRESSION, switchExpr.getEntry(1).getType());
     }
 
     @Test
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java14ValidatorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/Java14ValidatorTest.java
@@ -23,6 +23,7 @@ package com.github.javaparser.ast.validator;
 
 import static com.github.javaparser.ParseStart.COMPILATION_UNIT;
 import static com.github.javaparser.ParseStart.EXPRESSION;
+import static com.github.javaparser.ParseStart.STATEMENT;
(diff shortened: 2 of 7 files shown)


jlerbsc:
This work seems absolutely excellent to me. What do you think @johannescoetzee ?
Perhaps you could rearrange some commits to make them more readable.

johannescoetzee:
Looks good to me! Thanks for the contribution @mrjameshamilton 

I would say the commits are fine if we squash and merge, since everything will be flattened anyways

jlerbsc:
Thank you for your work.

Issue #4975, opened 2026-02-24
Fix BlockStmtContext.typePatternExprsExposedToChild performance issue


johannescoetzee:
I found this while debugging a performance issue when using `javasrc2cpg` in https://github.com/joernio/joern (which uses JavaParser for AST creation and type resolution) to analyze https://github.com/kekingcn/kkFileView. This took much longer than expected to finish and according to the attached flamegraph created with async-profiler, the culprit seems to be `BlockStmtContext.typePatternExprsExposedToChild`. 

This project has some methods with very, very long bodies and finding the child node index with
```
        int position = wrappedNode.getStatements().indexOf(child);
```
is very slow. It is also necessary to start at the correct position to avoid finding patterns exposed by statements after the child in question.

A possible fix for this and similar situations would be to add an `order` property to each Statement, storing the index of that node in its parent block or `Statement` node list (for switch entries where they need not be wrapped in a block statement). This property can be set once at AST creation (and modified later if nodes are inserted/removed) and can be used for a constant time lookup in this case. Alternatively this can be set for all AST nodes for consistency, although it's less useful for expressions. A proper implementation of this fix will likely be fairly high-effort and could introduce performance costs elsewhere for use-cases where nodes are added/removed frequently.

Another option is to add support for this to `NodeList` specifically with a more sophisticated `indexOf` operation. This would still need an `index` field added to `Node`, but the management of this field can be done in `NodeList` which is significantly easier to optimise.

johannescoetzee:
https://github.com/javaparser/javaparser/pull/4976 is a proposal for fixing the issue described above. 

There is a second issue causing the bad performance that is much more difficult to solve. In `BlockStmtContext.typePatternExprsExposedToChild`, once the index of the child we've traversed up from is found, the solver steps back over all the preceding nodes in the block to find type patterns introduced by them. This is also very slow in large blocks. 

The problem with solving this is that a new `BlockStmtContext` is created for each type solver query, so any kind of preprocessing that may be useful would just be discarded. The best solution I can think of here is to also add a `BlockStmtContext` field to the `BlockStatement` node. This would mean that the context for a given block is only created once and can be re-used for future queries. This would make it possible to, for example, keep a list of which indices introduce which type patterns. This list would then be created once upon context creation and will be much faster to traverse than repeating the same full traversal on each type solver query

PR #4976, opened 2026-02-25
Add first optimization for solving types in a block context

johannescoetzee:
This is related to https://github.com/javaparser/javaparser/issues/4975 and fixes the first issue causing slowdowns in `BlockStmtContext`. 

This PR adds an optimized `indexOf` method to `NodeList` which relies on a new `nodeListIndex` field in `Node` which is used to look up the node at the given index in the node list directly. If that node equals the given node, then the index is found and a linear search is not necessary. 

This solution to the problem has some downsides. Since it relies on the `nodeListIndex` field in the `Node`, adding a node to multiple node lists or using `fastIndexOf` on sublists of a list is likely to cause incorrect behavior and there is not a good mechanism to handle these cases. This means that `fastIndexOf` should not be used as a general purpose method, but in most "regular" use-cases (in particular where it's used in `BlockStmtContext`), it is easy to avoid these issues as regular list mutations (adding, removing, and replacing elements) are handled. 

Another issue with this approach is that `Node` now has this `nodeListIndex` field which doesn't make sense when viewed only from the context of the `Node` class. It needs the context of the completely separate `NodeList` class to be understood. The alternative I mentioned in the linked issue of adding an `order` field or something to AST nodes solves the issue of not making sense, but will likely run into many of the other issues mentioned.

I also considered other ways of implementing this that don't require this `nodeListIndex` field in `Node`. The most obvious idea is to keep some sort of hashmap of `Node -> indices at which equal nodes exist`, but this has a memory cost and introduces a much more difficult problem to solve, namely that of figuring out which of the indices at which the equal node exists is actually the correct one.

A possible change to this implementation that makes it even less generally useful but perhaps less confusing would be to change the logic to:
```
int fastIndexOf(Object o) {
  if (indicesInvalidated) resetIndices();
  if (o instanceof Node) {
    int index = ((Node) o).getNodeListIndex();
    // Specifically check reference equality
    if (nodeList.get(index) == o) {
      return index;
    }
  }
  // Only return if exact match is found
  return -1;
}
```

This would mean the method can only be used by providing reference to an item in the list, not an equal copy.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/NodeListTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/NodeListTest.java
@@ -29,6 +29,7 @@ import com.github.javaparser.ast.observer.AstObserver;
 import com.github.javaparser.ast.observer.ObservableProperty;
 import com.github.javaparser.printer.lexicalpreservation.AbstractLexicalPreservingTest;
 import com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter;
+import java.lang.reflect.Field;
 import java.util.*;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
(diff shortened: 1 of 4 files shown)


johannescoetzee:
@jlerbsc What do you think of this change? I'm not quite happy with it, but it's the best I can think of that doesn't require an enormous change (e.g. adding an index field to all AST nodes which needs to be set and maintained correctly everywhere)

PR #4982, opened 2026-03-06
Fix circular static import causing stackoverflow

ellamongo:
# Fix circular static import resolution causing StackOverflowError

## Summary

Resolves unbounded recursion and `StackOverflowError` when the symbol solver resolves symbols in a compilation unit that has **cyclic static imports** (e.g. type A statically imports from B, B from C, and C or an intermediate type leads back to the same compilation unit context).

## Problem

When `CompilationUnitContext#solveSymbol(String)` looks up a name, it iterates over static imports and calls `solveSymbolInType` on each imported type. If those types live in the same compilation unit (or in types that eventually delegate to the same CU via their parent context), the flow can re-enter `CompilationUnitContext` from the type’s context. That creates a cycle (e.g. `WorkfrontTestConstants` → … → `DocumentTestConstants` → … → back to the same CU), causing unbounded recursion and a `StackOverflowError`.

## Solution

- Introduced a **thread-local “resolving member in type only”** flag in `CompilationUnitContext`.
- When we are already resolving a symbol **as a member of a type** (inside `solveSymbolInType`), any re-entry into this compilation unit context (via the type’s parent) **does not** iterate static imports; it returns `unsolved` immediately, breaking the cycle.
- All existing static-import resolution that calls `solveSymbolInType` is wrapped in a scope that sets this flag for the duration of the call.

## Reason this works (Why we don't resolve the second CUC entry)

Java does not use the CU's static imports when we're already resolving a name as a member of a type (no transitive import). Re-applying them on re-entry would repeat the same lookup and create a cycle (e.g. ZooTestConstants → ElephantTestConstants → CUC → …). So when the flag is set we return unsolved and do not iterate static imports—that breaks the cycle and matches Java semantics.

---

### When the flag is set (three entry points)

The flag is set only when CUC calls `solveSymbolInType` from one of these three paths:

| # | Entry point | Zoo example |
|---|-------------|-------------|
| 1 | **Qualified reference** (lines 97–108) | Resolving `"Lion.OBJCODE"` in ElephantBuilder's CU: we resolve type `Lion`, then call `solveSymbolInType(Lion, "OBJCODE")` inside `runWithMemberInTypeOnlyScope`. |
| 2 | **Static import asterisk** (lines 112–120) | Resolving simple name `"Lion"` in ElephantBuilder's CU with `import static ...ZooTestConstants.*`: we call `solveSymbolInType(ZooTestConstants, "Lion")` inside `runWithMemberInTypeOnlyScope`. |
| 3 | **Static import single** (lines 126–132) | Resolving a name that matches a single static import (e.g. `import static junit4.zoo.constants.ZooTestConstants.ID`): we call `solveSymbolInType(ZooTestConstants, "ID")` inside `runWithMemberInTypeOnlyScope`. |

In each case we are "inside a member-of-type resolution" for the duration of that call.

#### Example 1: Entry point 1 — Qualified reference

**Example:** In ElephantBuilder we resolve **`Lion.OBJCODE`** (a qualified name).

- CUC enters `solveSymbol("Lion.OBJCODE")`, flag false.
- Name contains `"."` → we take typeName = `"Lion"`, memberName = `"OBJCODE"`.
- We call `solveType("Lion")`, then `runWithMemberInTypeOnlyScope(() -> solveSymbolInType(Lion, "OBJCODE"))`. **Flag is set** for the duration of `solveSymbolInType(Lion, "OBJCODE")`.
- Lion is in zoo-sdk-stub; its context might delegate to a parent. If that parent is a CUC and it calls `solveSymbol(...)`, we get a **second entry** with the flag still true → guard returns unsolved. In this case, if OBJCODE is not found in Lion we do not resolve further in a CUC, because Java does not support transitive import—for a qualified reference like `Lion.OBJCODE`, resolution is only within the type `Lion`; we do not fall back to the CU's static imports. So returning unsolved on re-entry is correct.

#### Example 2: Entry point 2 — Static import asterisk

**Example:** In ElephantBuilder we resolve the simple name **`"Lion"`** with `import static ...ZooTestConstants.*`.

- CUC enters `solveSymbol("Lion")`, flag false.
- No dot → we go to the static-import loop.
- For `import static ...ZooTestConstants.*` we call `runWithMemberInTypeOnlyScope(() -> solveSymbolInType(ZooTestConstants, "Lion"))`. **Flag is set** for the duration of `solveSymbolInType(ZooTestConstants, "Lion")`.
- ZooTestConstants doesn't have `Lion`; its context's parent is the CUC of the CU that contains ZooTestConstants. That CUC's `solveSymbol("Lion")` is invoked → **second entry** with flag true → guard returns unsolved. If Lion is not in ZooTestConstants, we do not need to resolve further in the CUC, because Java does not allow transitive import—when resolving a name as a member of a statically imported type, we only look within that type; we do not fall back to the CU's static imports again in that nested context. Letting the CUC iterate static imports on re-entry would re-enter the same logic (ZooTestConstants, ElephantTestConstants, …) and cause the cycle, so returning unsolved is correct.

#### Example 3: Entry point 3 — Static import single

**Example:** Suppose the CU has **`import static junit4.zoo.constants.ZooTestConstants.ID`** and we resolve the simple name **`"ID"`**.

- CUC enters `solveSymbol("ID")`, flag false.
- No dot → we go to the static-import loop.
- For that single static import we see memberName `"ID"` equals the requested name, so we call `runWithMemberInTypeOnlyScope(() -> solveSymbolInType(ZooTestConstants, "ID"))`. **Flag is set** for the duration of `solveSymbolInType(ZooTestConstants, "ID")`.
- If ZooTestConstants' context delegates to its parent CUC (e.g. to resolve something needed for `ID`), that CUC's `solveSymbol(...)` is invoked → **second entry** with flag true → guard returns unsolved. If ID is not in ZooTestConstants, we do not need to resolve further in the CUC, because Java does not allow transitive import—when resolving a name as a member of a statically imported type (single or asterisk), we only look within that type; we do not fall back to the CU's static imports again in that nested context. Re-applying the CU's static imports on re-entry would create a cycle, so returning unsolved is correct.

## Testing

- **New test:** `resolveMethodCallsInDocumentBuilderWithCyclicStaticImportsWithoutStackOverflow` in `CompilationUnitContextResolutionTest`.

## Failing case (regression test)

A separate PR ([#2 — Add test expecting StackOverflowError for cyclic static imports](https://github.com/ellamongo/javaparser/pull/2)) adds a **regression test** that demonstrates the bug on `master` (unpatched code). That test, `resolveSymbolInElephantBuilderWithCyclicStaticImportsThrowsStackOverflow`, uses the same zoo fixture and asserts that `CompilationUnitContext#solveSymbol("Lion")` **throws** `StackOverflowError` when run without this fix. With this fix applied, the same scenario completes without throwing (and is covered by `resolveMethodCallsInDocumentBuilderWithCyclicStaticImportsWithoutStackOverflow`). The two tests together document the before/after behavior.


johannescoetzee:
At first glance this looks good to me. I want to run a few more tests next week, but I don't expect that this will cause any issues. 

Please refer to https://github.com/javaparser/javaparser/blob/master/CONTRIBUTING.md for instructions on fixing the code formatting

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/contexts/CompilationUnitContextResolutionTest.java
```diff
@@ -268,4 +283,42 @@ void solveMethodStaticallyImportedWithoutAsterisk() throws ParseException, IOExc
...
+        typeSolver.add(new JavaParserTypeSolver(testJava));
+        typeSolver.add(new JavaParserTypeSolver(mainJava));
+        typeSolver.add(new JavaParserTypeSolver(zooSdkStub));
+
+        CompilationUnit cu = parseSampleWithStandardExtension(
+                "static_import_cycle_fixture/app/src/test/java/junit4/zoo/builders/ElephantBuilder",
+                typeSolver);
+
+        try {
+            new CompilationUnitContext(cu, typeSolver).solveSymbol("Lion");
```
jlerbsc, line 318: It's cleaner to use the JavaParserFacade.
JavaParserFacade.get(typeSolver).solve(new NameExpr("Lion"))

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/contexts/CompilationUnitContextResolutionTest.java
```diff
@@ -268,4 +283,42 @@ void solveMethodStaticallyImportedWithoutAsterisk() throws ParseException, IOExc
...
+     * own static imports. Without the "member in type only" fix, re-entry into CUC via the type's
+     * parent context causes unbounded recursion and StackOverflowError.
+     * <p>
+     * This test triggers that path directly with {@code CUC.solveSymbol("Lion")}. With the
+     * fix, resolution succeeds (Lion from zoo-sdk-stub) or returns unsolved; without it, the test fails with StackOverflowError.
+     */
+    @Test
+    void resolveMethodCallsInDocumentBuilderWithCyclicStaticImportsWithoutStackOverflow() throws IOException, URISyntaxException {
+        Path fixtureRoot = fixtureRootForStaticImportCycle();
+        Path testJava = fixtureRoot.resolve("src/test/java");
```
jlerbsc, line 304: Could you simplify your test case?

jlerbsc:
Thanks for the PR — the fix itself looks solid. However, the test fixture seems more complex than it needs to be. The actual cycle here is just ZooTestConstants ↔ ElephantTestConstants (mutual static imports), but reproducing it requires 7 fixture files across a multi-module structure with 4 TypeSolver instances.

Since the bug is purely about cyclic static imports causing unbounded recursion, a minimal fixture with 3 files in a single source root would be easier to understand and maintain:

// cycle/A.java
package cycle;
import static cycle.B.*;
public class A {
    public static final String X = "x";
}

// cycle/B.java
package cycle;
import static cycle.A.*;
public class B {
    public static final String Y = "y";
}

// cycle/User.java
package cycle;
import static cycle.A.*;
public class User {
    String ref = A.X;
}
@Test
void cyclicStaticImportsDoNotCauseStackOverflow() {
    Path root = adaptPath("src/test/resources/cycle_static_import");
    CombinedTypeSolver typeSolver = new CombinedTypeSolver(
        new ReflectionTypeSolver(),
        new JavaParserTypeSolver(root));

    CompilationUnit cu = parseSampleWithStandardExtension(
        "cycle_static_import/User", typeSolver);

    NameExpr nameExpr = cu.findFirst(NameExpr.class,
        n -> n.getNameAsString().equals("A")).get();
    assertDoesNotThrow(() -> JavaParserFacade.get(typeSolver).solve(nameExpr));
}
This makes the cycle immediately visible in the test, removes the multi-module setup that doesn't contribute to exercising the fix, and keeps a single TypeSolver. The "zoo" domain modeling is realistic but it obscures what the test is actually verifying.

Would you be open to simplifying the fixture along these lines?

ellamongo:
> Thanks for the PR — the fix itself looks solid. However, the test fixture seems more complex than it needs to be. The actual cycle here is just ZooTestConstants ↔ ElephantTestConstants (mutual static imports), but reproducing it requires 7 fixture files across a multi-module structure with 4 TypeSolver instances.
> 
> Since the bug is purely about cyclic static imports causing unbounded recursion, a minimal fixture with 3 files in a single source root would be easier to understand and maintain:
> 
> // cycle/A.java package cycle; import static cycle.B.*; public class A { public static final String X = "x"; }
> 
> // cycle/B.java package cycle; import static cycle.A.*; public class B { public static final String Y = "y"; }
> 
> // cycle/User.java package cycle; import static cycle.A.*; public class User { String ref = A.X; } @test void cyclicStaticImportsDoNotCauseStackOverflow() { Path root = adaptPath("src/test/resources/cycle_static_import"); CombinedTypeSolver typeSolver = new CombinedTypeSolver( new ReflectionTypeSolver(), new JavaParserTypeSolver(root));
> 
> ```
> CompilationUnit cu = parseSampleWithStandardExtension(
>     "cycle_static_import/User", typeSolver);
> 
> NameExpr nameExpr = cu.findFirst(NameExpr.class,
>     n -> n.getNameAsString().equals("A")).get();
> assertDoesNotThrow(() -> JavaParserFacade.get(typeSolver).solve(nameExpr));
> ```
> 
> } This makes the cycle immediately visible in the test, removes the multi-module setup that doesn't contribute to exercising the fix, and keeps a single TypeSolver. The "zoo" domain modeling is realistic but it obscures what the test is actually verifying.
> 
> Would you be open to simplifying the fixture along these lines?

Hey, thanks for the reply, I was in the middle of updating didn't expect a review this fast, apologise for that. 

I just pushed a refactor of the test files, which kind of simplified the fixture along those lines — the multi-module app/ + zoo-sdk-stub/ structure is gone, replaced with a flat directory and a single JavaParserTypeSolver. The zoo naming is also stripped down significantly.

We did need to keep a 3-node import structure rather than the 2-node A ↔ B cycle you suggested, as a simple mutual cycle is already fully guarded by the existing resolvedStaticImport ThreadLocal in master (added for issues #4450/#2720).

The overflow requires a dead-end import to appear first in one of the cyclic classes. When that dead-end returns unsolved, it calls resolvedStaticImport.remove(), flushing the entire guard set mid-search. The subsequent cyclic import then runs with an empty guard list, allowing unbounded re-entry. In our fixture, ZooQueryConstants plays that dead-end role — it's the first import in ZooTestConstants, and being an interface with no static imports of its own, it always returns unsolved and clears the list before ElephantTestConstants.* is tried.

Let me know how you think about it, thanks!

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/contexts/CompilationUnitContextResolutionTest.java
ellamongo: Addressed in 1486e7fb053205042b944ec6704c079ffef3d3f2

javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/javaparser/contexts/CompilationUnitContextResolutionTest.java
ellamongo: addressed in 8557a370821fb3bd270e0b4847ab2579c6b03653

jlerbsc:
Thanks for the fix and for simplifying the test fixture — the diagnosis is right: the old resolvedStaticImport history was cleared by nested calls, which made the guard ineffective for cycles of 3+ nodes. The ThreadLocal scope approach is cleaner. A few points before merging, in priority order:

Must fix

The guard is broader than its stated intent. The early return unsolved() in solveSymbol short-circuits the whole method — including the absolute-reference block (Type.member) — and applies to any CompilationUnitContext on the thread, not just the one that started the resolution. During solveSymbolInType(A, name), the class context of A legitimately delegates to A's own CU (Java allows A's internals to use A's own imports); only the transitive import lookup should be cut. Please move the check so it only skips the static-import loop, or explain why cutting the absolute-reference path is safe.
The fixture isn't legal Java. Static imports from the default package are rejected by javac. Please add a package declaration (e.g. package cycle;) to the 5 fixture files so the scenario represents compilable code.
Should fix

Restore the previous flag value instead of forcing false in the finally (boolean prev = get(); ... set(prev)), or call remove() — safer for future call paths and avoids leaving a ThreadLocal entry per pooled thread.
Strengthen the test assertion: besides assertDoesNotThrow, resolve Lion.OBJCODE and assert the resolved type. Also rename the local typeSolver (it shadows the field) and consider using adaptPath(...) directly like the other tests instead of the classloader fallback.
Nice to have

Make the fixture self-explanatory: structural names (e.g. EntryPoint, CycleA, CycleB) or a short comment describing the import cycle, and drop the non-structural bits (builder pattern, requireNonNull, @SuppressWarnings, unused constants). Please verify any simplification still reproduces the StackOverflow with the fix reverted.
Finally, could you squash the 4 commits into a single one (or two: fix + test)? 
That will make the history easier to follow when merging.

ellamongo:
> Thanks for the fix and for simplifying the test fixture — the diagnosis is right: the old resolvedStaticImport history was cleared by nested calls, which made the guard ineffective for cycles of 3+ nodes. The ThreadLocal scope approach is cleaner. A few points before merging, in priority order:
> 
> Must fix
> 
> The guard is broader than its stated intent. The early return unsolved() in solveSymbol short-circuits the whole method — including the absolute-reference block (Type.member) — and applies to any CompilationUnitContext on the thread, not just the one that started the resolution. During solveSymbolInType(A, name), the class context of A legitimately delegates to A's own CU (Java allows A's internals to use A's own imports); only the transitive import lookup should be cut. Please move the check so it only skips the static-import loop, or explain why cutting the absolute-reference path is safe. The fixture isn't legal Java. Static imports from the default package are rejected by javac. Please add a package declaration (e.g. package cycle;) to the 5 fixture files so the scenario represents compilable code. Should fix
> 
> Restore the previous flag value instead of forcing false in the finally (boolean prev = get(); ... set(prev)), or call remove() — safer for future call paths and avoids leaving a ThreadLocal entry per pooled thread. Strengthen the test assertion: besides assertDoesNotThrow, resolve Lion.OBJCODE and assert the resolved type. Also rename the local typeSolver (it shadows the field) and consider using adaptPath(...) directly like the other tests instead of the classloader fallback. Nice to have
> 
> Make the fixture self-explanatory: structural names (e.g. EntryPoint, CycleA, CycleB) or a short comment describing the import cycle, and drop the non-structural bits (builder pattern, requireNonNull, @SuppressWarnings, unused constants). Please verify any simplification still reproduces the StackOverflow with the fix reverted. Finally, could you squash the 4 commits into a single one (or two: fix + test)? That will make the history easier to follow when merging.

Squashed into two commits (fix + test). All points addressed — two notes on the shape of the fix:

1. The guard. 

Moved after the absolute-reference block, so Type.member is no longer cut.

It tracks the name being resolved as a member of a type; while that name is in flight, no CU applies its static imports to it. Each recursion step was a CU re-applying its imports to a name that had arrived as a member lookup — a transitive import Java doesn't perform — so this removes the edge the cycle is built from rather than detecting the loop after the fact, as the old history did.

Because the key is the name, "A's internals use A's own imports" still works: those are different names. When name itself reaches A's CU it's because A has no member name — the outside query falling through into A's imports, not A's internals using them.

On the ThreadLocal suggestion: there's no boolean flag left to restore. It's now a set of in-flight names, each frame removing only the entry it added, and the ThreadLocal is remove()d once the set drains.

2. A follow-up I'd suggest. 

This is a dynamic guard, not a structural property of the resolver.

The JLS keeps two things apart that the resolver conflates. The members of a type are only those declared in it plus those inherited (JLS 8.2) — a static import never adds a member to the type it imports into. A simple name, by contrast, is resolved against the enclosing scope, which does include that file's static imports (JLS 6.5.6.1). So a member lookup must stop at the type; only simple-name resolution may continue outward into the compilation unit.

That's exactly what goes wrong here: `solveSymbolInType` is a member lookup, but it reaches the type's Context, and `ClassOrInterfaceDeclarationContext`:71 falls through to the parent unconditionally — silently turning the §8.2 question into the §6.5.6.1 one, and letting the query escape into that file's imports. The method side already fixed this: `solveMethod` resolves static imports with staticOnly = true, and `JavaParserTypeDeclarationAdapter`:317 skips the fall-through when it's set (added for #4358). Value lookup never got the equivalent signal.

The right fix is to give it one. I have a prototype that does this and deletes the `ThreadLocal` entirely, with both suites identical to this PR — but it touches two public interfaces in javaparser-core, so I've kept it out. Happy to open it as a follow-up issue with the prototype, if you agree with the direction.

jlerbsc:
Re-reviewed the rebased version — this addresses everything from the previous round, and I verified it locally rather than by reading:

Full symbol-solver suite on your branch: 2091 tests, 0 failures.
Reverting CompilationUnitContext to master while keeping your test makes exactly one test fail, yours, with StackOverflowError — the other 14 in the class still pass. The fixture isolates the regression precisely.
spotless:check clean.

The design holds up. The name-keyed guard can't break field inheritance, because ClassOrInterfaceDeclarationContext.solveSymbol resolves inherited fields through hasVisibleField on the declaration's ancestor chain, never through the ancestors' compilation-unit contexts. The ThreadLocal bookkeeping is leak-free (an added == false frame always has an ancestor still holding the entry), and solveSymbolAsValue inherits the default that delegates to solveSymbol, so it's covered too.

One thing I'd like fixed before merge, both about the description rather than the code:

- The solveMethod claim in the commit message is inaccurate. Method resolution isn't immune "this way" — its immunity is structural: CompilationUnitContext.solveMethod passes staticOnly = true, and JavaParserTypeDeclarationAdapter.solveMethod only delegates back to the parent context when !staticOnly (with a comment citing #4358). No thread-local involved. The real asymmetry is that the value side has no staticOnly equivalent — which is exactly the structural follow-up you proposed. Worth saying it that way.

Nice to have: a test that asserts the non-transitivity itself. Right now nothing locks that semantic in — only the absence of the overflow plus one positive type assertion.

Nits: the comment cites JLS 7.5.4, but the guard also covers single static imports (7.5.3); CycleB.CYCLE_B_ID is unused and not needed for the repro; and findFirst(FieldAccessExpr.class) could take a predicate like the NameExpr lookup above it.

ellamongo:
> Re-reviewed the rebased version — this addresses everything from the previous round, and I verified it locally rather than by reading:
> 
> Full symbol-solver suite on your branch: 2091 tests, 0 failures. Reverting CompilationUnitContext to master while keeping your test makes exactly one test fail, yours, with StackOverflowError — the other 14 in the class still pass. The fixture isolates the regression precisely. spotless:check clean.
> 
> The design holds up. The name-keyed guard can't break field inheritance, because ClassOrInterfaceDeclarationContext.solveSymbol resolves inherited fields through hasVisibleField on the declaration's ancestor chain, never through the ancestors' compilation-unit contexts. The ThreadLocal bookkeeping is leak-free (an added == false frame always has an ancestor still holding the entry), and solveSymbolAsValue inherits the default that delegates to solveSymbol, so it's covered too.
> 
> One thing I'd like fixed before merge, both about the description rather than the code:
> 
> * The solveMethod claim in the commit message is inaccurate. Method resolution isn't immune "this way" — its immunity is structural: CompilationUnitContext.solveMethod passes staticOnly = true, and JavaParserTypeDeclarationAdapter.solveMethod only delegates back to the parent context when !staticOnly (with a comment citing [StackOverflowError with import static in circular dependencies #4358](https://github.com/javaparser/javaparser/issues/4358)). No thread-local involved. The real asymmetry is that the value side has no staticOnly equivalent — which is exactly the structural follow-up you proposed. Worth saying it that way.
> 
> Nice to have: a test that asserts the non-transitivity itself. Right now nothing locks that semantic in — only the absence of the overflow plus one positive type assertion.
> 
> Nits: the comment cites JLS 7.5.4, but the guard also covers single static imports (7.5.3); CycleB.CYCLE_B_ID is unused and not needed for the repro; and findFirst(FieldAccessExpr.class) could take a predicate like the NameExpr lookup above it.

Thanks for the reply  — all four points are addressed, pushed as the same two commits.

**The solveMethod claim.** I removed it from the commit message rather than rewording it: the commit doesn't touch method resolution, so the comparison doesn't belong in it. But your framing is the correct one and I want it on record here — method resolution's immunity is structural, not a guard: CompilationUnitContext.solveMethod passes staticOnly = true, and JavaParserTypeDeclarationAdapter.solveMethod only delegates to the parent context when !staticOnly (citing #4358).

**Non-transitivity test.** Added, using the existing fixture: SolvableConstants.ID is reachable from EntryPoint only through CycleA's own static import, so it must not resolve, with a control asserting it does resolve from CycleA itself. Asserted through CompilationUnitContext so the fixture stays compilable.

**Comment Nits** Comment now cites 7.5.3 and 7.5.4; CYCLE_B_ID dropped (re-verified the repro still overflows without it); findFirst(FieldAccessExpr.class) takes an OBJCODE predicate.

PR #4986, opened 2026-03-08
Respect visibility in MethodResolution(fix #4969)

### LINKED ISSUE #4969: Method Resolution: Repescting Visibility in the Call Context
When attempting to resolve the call `a.m(1)` in the example below, JP correctly identifies the two possible candidates (`m(int)` and `m(long)`) and correctly treats `m(int)` as the most specific, which ought to be selected.

However, in this case, `m(long)` must be selected, because `m(int)` is not visible to `B`.

```java
final class A {
  public  byte m(long i) { return 2; }

  private int m(int i) { return 1; }
}

final class B {
  static int callM() {
     A a = new A();
     return a.m(1);
  }
}
```

I did take a cursory look at the code and it seems that the information of where the call originates is lost during resolution. Is this a JP bug or am I simply using the wrong method resolution method(s)?

wadoon:
Fixes issue #4969. 

The resolution of methods now respects their visibility. For example, calling `m(5)` should resolve to `m(long)` if the sibling method `m(int)` is not visible from the caller context. 

This patch extends the `solveMethod` interface with an additional caller context (`ResolvedReferenceTypeDeclaration`). The final visibility resolution is done in `MethodResolutionLogic#findMostApplicable`).

Side edit: `<maven.compiler.release>8</maven.compiler.release>` makes it compile under modern JDK.

Original patch by @unp1

### DIFF
--- a/javaparser-core-testing/src/test/resources/com/github/javaparser/bulk_test_results/javaparser_test_results_javaparser_symbol_solver_testing_src_test_java.txt
+++ b/javaparser-core-testing/src/test/resources/com/github/javaparser/bulk_test_results/javaparser_test_results_javaparser_symbol_solver_testing_src_test_java.txt
@@ -1,1 +1,0 @@
-0 problems in 0 files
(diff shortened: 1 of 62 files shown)


jlerbsc:
Please squash your commits to make your PR easier to read.

wadoon:
The two commits are squashed now.

jlerbsc:
Thank you for your suggestion, but I don't think it's the right solution, even though I haven't looked into the issue in great detail. Furthermore, this solution has a major impact on the community's use of Javaparser, which is something we don't want.

wadoon:
Taking back the discussion to the issue #4969

PR #5011, opened 2026-05-07
build: add OSGi support to all deployable modules (#1477)

### LINKED ISSUE #1477: Improving OSGI support
We have pretty basic osgi support right now. I'm pretty sure it's broken right now because we keep forgetting to add new package to the configuration. We could automatically generate those packages, and maybe set a real version number too.

ethan-godden:
Closes #1477.

Adds OSGi metadata to all published modules automatically by wiring bnd-maven-plugin into the parent POM. Each module's jar now exposes Bundle-SymbolicName, Bundle-Version, and appropriate Export-Package / Import-Package headers, so the artifacts can be dropped into Felix/Equinox without repackaging.
Verification

`mvn clean install` succeeds; generated MANIFEST.MF inspected for each module.
Loaded javaparser-core jar into a OSGi project to confirm the bundle resolves.
No changes to non-OSGi Maven/Gradle consumers — same coordinates, same classes.

Future work (not in this PR)
P2 repository / Eclipse RCP (Equinox) support would require either consumers building from source or hosting an Eclipse update site — happy to open a separate issue if there's interest.


ethangodden:
One thing that I could add is support for P2 repositories (i.e. the Eclipse platform), but this would either require user to compile this project from source or deploying to a Eclipse update site.

jlerbsc:
It seems to me that this press release is lacking in explanation. For example, could you explain why you are removing the generator...

ethan-godden:
@jlerbsc 
> It seems to me that this press release is lacking in explanation. For example, could you explain why you are removing the generator...

- The purpose of `BndGenerator` was to generate the bnd.bnd file: a file used by the BND plugin to generate the OSGi MANIFEST.MF 
- The bnd.bnd file is not needed anymore because its now defined concisely in the [parent pom.xml](https://github.com/ethangodden/javaparser/blob/f50d7c558fc0927374ae30908f13769f2eb47f77/pom.xml#L206-L223) file.
```
<configuration>
    <!--
        Bundle-SymbolicName is derived from groupId + artifactId. OSGi permits
        dashes here, so this preserves names like
        `com.github.javaparser.javaparser-core` for OSGi consumers.
    
        Automatic-Module-Name uses ${jpms.moduleName} because JPMS module names
        disallow dashes — each module sets this property to a dot-separated,
        dash-free identifier.
    -->
    <bnd><![CDATA[
        Bundle-SymbolicName: ${project.groupId}.${project.artifactId}
        Automatic-Module-Name: ${jpms.moduleName}
        -exportcontents: ${osgi.export}
        -nodefaultversion: true
        Bundle-DocURL: https://javaparser.org
    ]]></bnd>
</configuration>
```
- Previously, the BND plugin required packages to be declared explicitly in the `-exportcontents:` section, but it now allows for wildcards, which is defined as a property [here](https://github.com/ethangodden/javaparser/blob/f50d7c558fc0927374ae30908f13769f2eb47f77/pom.xml#L155-L161). OSGi wildcards are sometimes are frowned upon because you are possibly exporting implementation classes users do not need. However, `BndGenerator` does not seem to filter out any packages, so this is functionally equivalent to what it was before. 
- In the generated MANIFEST.MF, the packages are declared explicitly based on the pattern passed
- Even when configuration of explicit exports are needed, we could just define them via the `osgi.export` maven property with syntax similar to how the MANIFEST.MF is defined above, which I believe is simpler that the `BndGenerator` approach

jlerbsc:
Why are you including the test projects in this PR? What added value do these projects bring to OSGi?

ethan-godden:
> Why are you including the test projects in this PR? What added value do these projects bring to OSGi?

@jlerbsc 

There is no added value. I included them because it reduces the added build complexity of this PR with a negligible performance hit to the build. If a new module is added to this project, the only OSGi configuration needed would be to define `jpms.moduleName` without dashes. We could even define a default `jpms.moduleName` as `${project.groupId}.${project.artifactId}`  in the parent to avoid this. 

I have commited a fix (ced4bfe2ba600e50ae170f76076c7cb25d60c4b3) regarding this. I had to add configurations per deployable module. Any new module would also need this.

jlerbsc:
Could you please squash your commits to make the PR easier to read? Thank you.

ethan-godden:
> Could you please squash your commits to make the PR easier to read? Thank you.

@jlerbsc Done

jlerbsc:
"Future work (not in this PR)
P2 repository / Eclipse RCP (Equinox) support would require either consumers building from source or hosting an Eclipse update site — happy to open a separate issue if there's interest."

There is no need to submit this pull request, as JP is not intended to provide features for other projects. However, you could create a separate project specifically designed to provide these user-friendly features.

jlerbsc:
Thank you for this PR.

ethan-godden:
@jlerbsc I noticed that the changes made in this PR seem to have been reverted in [this commit](https://github.com/javaparser/javaparser/commit/5a43c63f9b3344c0bfe67ecaeca039dbcad04966). Was this intentional?

jlerbsc:
Hi @ethan-godden  — while reviewing recent history I found that PR [#5011](https://github.com/javaparser/javaparser/pull/5011) (OSGi support, [#1477](https://github.com/javaparser/javaparser/issues/1477)) was accidentally reverted five days after it was merged: PR [#5019](https://github.com/javaparser/javaparser/pull/5019) was based on a stale master and its commit unintentionally restored the old build files (BndGenerator, bnd.bnd, and the previous pom configuration). As a result, the OSGi improvements never shipped in 3.28.2, even though [#1477](https://github.com/javaparser/javaparser/issues/1477) was closed as completed.

I’m going to draft a PR that will restore the original commit as-is (clean cherry-pick of b4e443b, original authorship preserved). I verified the build and confirmed the generated MANIFEST.MF contains the expected Bundle-SymbolicName, Bundle-Version, Export-Package, and Automatic-Module-Name headers. Apologies to @ethan-godden for the accidental revert — the fix will be included in the next release.

PR #5017, opened 2026-05-18
fix(grammar): allow text block ending in escaped backslash before closing """

### LINKED ISSUE #4894: Text block ending in double backslash is parsed incorrectly
**JavaParser-Core Version**: 3.27.1
**Parser Configuration**: language level `JAVA_15` (or higher)

JavaParser fails to parse valid Java code when a text block (triple-quoted string) ends with a double-backslash immediately before the closing `"""`.

## Input

Affected source sample:
```java
public class MinimalTextBlockBug {

    // This text block ends with a pair of backslashes.
    // JavaParser fails to recognize the end of the block and encounters an unexpected EOF.
    public String testCase1() {
        return """
            arbitrary text
            \\""";
    }

}
```

## Actual Behavior

Output from parsing:
```text
Lexical error at line 12, column 0.  Encountered: <EOF> after : ""
```

## Expected Behavior

The file should parse successfully. This is valid Java code that compiles with Java 15+.
The text block contains two backslash characters which form a literal backslash in the string.

## Related Observation

Adding another character after the double backslash fixes the parse error and the printed parse result correctly shows the double blackslash and the character after it.

It seems that this bug is only triggered when backslashes are the very last characters before the closing `"""`.

ethan-godden:
Fixes #4894.

## Summary

A text block whose content ends in an escaped backslash (`\\`) immediately before
the closing `"""` failed to parse, e.g.:

```java
String s = """
    arbitrary text
    \\""";
```

produced `Lexical error ... Encountered: <EOF> after : ""`. `javac` accepts this
form per JLS §3.10.6. 

## Root cause

In `javaparser-core/src/main/javacc/java.jj`, the `TEXT_BLOCK_CONTENT` rule only
treated `\"` as an atomic escape:

```
<IN_TEXT_BLOCK> MORE : { <TEXT_BLOCK_CONTENT: ( "\\" "\"" | ~[] ) > }
```

Given input `\\"""`, JavaCC's longest-match consumed the first `\` as a 1-char
`~[]` match, then matched the second `\` together with the first `"` of the
closing delimiter as a 2-char `\"` escape — leaving only `""`, which can never
satisfy the 3-quote `TEXT_BLOCK_LITERAL` token. The lexer ran off the end of
input.

## Fix

Recognise *any* `\X` (backslash + any character) as a single atomic chunk:

```
<IN_TEXT_BLOCK> MORE : { <TEXT_BLOCK_CONTENT: ( "\\" ~[] | ~[] ) > }
```

So `\\` is now consumed as one unit and the closing `"""` stays intact. `\"`
continues to work (still a 2-char match, still wins over the 1-char fallback),
and bare `"` inside the block is unaffected because the 3-char
`TEXT_BLOCK_LITERAL` token outranks `~[]` at the closer.

## Tests

Two regression tests added in
`javaparser-core-testing/.../ast/expr/TextBlockLiteralExprTest.java`:

- `textBlockEndingInDoubleBackslashAdjacentToCloserParses` — canonical failing
  form from the issue (`\\` immediately before `"""`).
- `textBlockEndingInDoubleBackslashOnSeparateLineParses` — the `\\` on its own
  line variant, asserting both raw value and `translateEscapes()` output.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
@@ -21,6 +21,7 @@
 
 package com.github.javaparser.ast.expr;
 
+import static com.github.javaparser.utils.TestParser.parseExpression;
 import static com.github.javaparser.utils.TestParser.parseStatement;
 import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.toList;
@@ -196,5 +197,30 @@ class TextBlockLiteralExprTest {
                 .findFirst(TextBlockLiteralExpr.class)
                 .get();
         assertEquals("\nHello\n" + "World", textBlock.translateEscapes());
+    }
+
+    /**
+     * Regression test for <a href="https://github.com/javaparser/javaparser/issues/4894">issue #4894</a>.
+     * <p>
+     * A text block whose last content characters before the closing {@code """} are an
+     * escaped backslash ({@code \\}) on a separate line must parse successfully.
+     */
+    @Test
+    void textBlockEndingInDoubleBackslashOnSeparateLineParses() {
+        TextBlockLiteralExpr textBlock = parseExpression("\"\"\"\n" + "        foo\\\\\n" + "        \"\"\"");
+        assertEquals("foo\\\n", textBlock.translateEscapes());
+    }
+
+    /**
+     * Regression test for <a href="https://github.com/javaparser/javaparser/issues/4894">issue #4894</a>.
+     * <p>
+     * Canonical failing form: an escaped backslash ({@code \\}) sits immediately adjacent
+     * to the closing {@code """} (no intervening newline)
+     */
+    @Test
+    void textBlockEndingInDoubleBackslashAdjacentToCloserParses() {
+        TextBlockLiteralExpr textBlock =
+                parseExpression("\"\"\"\n" + "        arbitrary text\n" + "        \\\\\"\"\"");
+        assertEquals("arbitrary text\n\\", textBlock.translateEscapes());
     }
 }
--- a/javaparser-core/src/main/javacc/java.jj
+++ b/javaparser-core/src/main/javacc/java.jj
@@ -840,8 +840,13 @@ TOKEN :
 MORE: { <ENTER_TEXT_BLOCK: "\"\"\"" >: IN_TEXT_BLOCK }
 <IN_TEXT_BLOCK> TOKEN : { <TEXT_BLOCK_LITERAL: "\"\"\"" >: DEFAULT }
 // Need to recognize an escaped " to prevent it from parsing as "backslash doublequote" -
-// which might match that doublequote with following doublequotes.
-<IN_TEXT_BLOCK> MORE :{ <TEXT_BLOCK_CONTENT:  ( "\\" "\"" | ~[] ) > }
+// which might match that doublequote with following doublequotes. The ~[] is needed so the
+// backslash escape ("\\" followed by any character) atomically so
+// the escape's trailing character (most importantly, the " in \")
+// cannot be re-used as part of the closing """ delimiter. The first alternative
+// wins by JavaCC's longest-match rule whenever the current char is a backslash;
+// the second alternative is the fallback for ordinary text-block characters.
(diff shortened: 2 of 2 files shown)


jlerbsc:
Is this an AI-generated PR?

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
```diff
@@ -197,4 +197,43 @@ void whiteSpaceLineShorterThanMiniumCommonPrefix() {
...
+
+    /**
+     * Regression test for <a href="https://github.com/javaparser/javaparser/issues/4894">issue #4894</a>.
+     * <p>
+     * A text block whose last content characters before the closing {@code """} are an
+     * escaped backslash ({@code \\}) on a separate line must parse successfully.
+     */
+    @Test
+    void textBlockEndingInDoubleBackslashOnSeparateLineParses() {
+        TextBlockLiteralExpr textBlock = parseStatement("String s = \"\"\"\n" + "        foo\\\\\n" + "        \"\"\";")
```
jlerbsc: The test can be simplified by parsing an expression directly.

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
```diff
@@ -197,4 +197,43 @@ void whiteSpaceLineShorterThanMiniumCommonPrefix() {
...
+     * Canonical failing form: an escaped backslash ({@code \\}) sits immediately adjacent
+     * to the closing {@code """} (no intervening newline). Before the fix, the lexer rule
+     * for {@code TEXT_BLOCK_CONTENT} only recognised {@code \"} as an atomic escape; it
+     * therefore consumed the second backslash together with the first {@code "} of the
+     * closing delimiter as a single {@code \"} chunk, leaving only {@code ""} and failing
+     * with {@code Lexical error ... Encountered: <EOF> after : ""}.
+     */
+    @Test
+    void textBlockEndingInDoubleBackslashAdjacentToCloserParses() {
+        TextBlockLiteralExpr textBlock = parseStatement(
```
jlerbsc: The test can be simplified by parsing an expression directly.

ethan-godden:
> Is this an AI-generated PR?

@jlerbsc 

It is partially. I had it check my work and make improvements, mostly the comments,  tests, and PR markdown so I do not miss anything

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
ethan-godden: Should be fixed in c7652d8c048f6b45627b258dc471add9783be03e

javaparser-core-testing/src/test/java/com/github/javaparser/ast/expr/TextBlockLiteralExprTest.java
ethan-godden: Should be fixed in c7652d8c048f6b45627b258dc471add9783be03e

javaparser-core/src/main/javacc/java.jj
```diff
@@ -840,8 +840,13 @@ TOKEN :
...
 // Need to recognize an escaped " to prevent it from parsing as "backslash doublequote" -
-// which might match that doublequote with following doublequotes.
-<IN_TEXT_BLOCK> MORE :{ <TEXT_BLOCK_CONTENT:  ( "\\" "\"" | ~[] ) > }
+// which might match that doublequote with following doublequotes. The ~[] is needed so the
+// backslash escape ("\\" followed by any character) atomically so
+// the escape's trailing character (most importantly, the " in \")
+// cannot be re-used as part of the closing """ delimiter. The first alternative
+// wins by JavaCC's longest-match rule whenever the current char is a backslash;
+// the second alternative is the fallback for ordinary text-block characters.
+<IN_TEXT_BLOCK> MORE :{ <TEXT_BLOCK_CONTENT:  ( "\\" ~[] | ~[] ) > }
```
ethan-godden, line 849: One thing to note is we could make this more explicit by listing all possible escape sequences instead of use a `~[]` to represent all possible second characters in a escape sequence. I think what we have is more simple and less risky. The counter argument to that is listing all escape sequences would be closer to the actual Java spec.

ethan-godden:
Sorry for the spam and poor quality tests

jlerbsc:
Thank you for this PR.

PR #5018, opened 2026-05-19
Upgrade Java source/target from 8 to 17

GG-Feng:
## Summary

This PR upgrades the project from Java 8 to Java 17, bringing the codebase up to modern Java standards.

## Changes

- Updated source/target from 8 to 17 in pom.xml
- Applied instanceof pattern matching
- Added @Serial annotation to serialVersionUID fields
- Replaced new Integer/Long/Double with valueOf
- Converted String.format to .formatted
- Converted multiline strings to text blocks
- Migrated Paths.get to Path.of

## Statistics

334 files changed, 7774 insertions, 6165 deletions

## Verification

- All 2047 tests pass with 0 failures

### DIFF
--- a/.github/workflows/formatting_check.yml
+++ b/.github/workflows/formatting_check.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Set up JDK 11
         uses: actions/setup-java@v5
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'zulu'
       - name: Validate Project Checkstyle
         run: ./mvnw -B checkstyle:check
@@ -53,7 +53,7 @@ jobs:
       - name: Set up JDK 11
         uses: actions/setup-java@v5
         with:
-          java-version: '11'
+          java-version: '17'
           distribution: 'zulu'
       - name: Generate code and format
         run: |
--- a/.github/workflows/prepare_release_changelog.yml
+++ b/.github/workflows/prepare_release_changelog.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions/setup-java@v5
         with:
           distribution: zulu
-          java-version: 11
+          java-version: 17
 
       # Set environment variables
       - name: Generate Changelog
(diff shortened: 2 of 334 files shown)


jlerbsc:
Thank you for this contribution — the effort is clearly substantial and the direction is worth discussing. However, before we can evaluate whether to merge anything here, there is a more fundamental question to settle: should JavaParser drop Java 8 support at all?

This is not a trivial decision. JavaParser is widely used as a library, and a significant part of its user base still runs on Java 8–16. Raising the baseline to Java 17 is a one-way door: users who cannot migrate would have to freeze on the last Java 8-compatible release. We would need strong motivation — concrete benefits that cannot be achieved while maintaining Java 8 compatibility — to justify that cost.

A few questions that should be answered before this conversation goes further:

Has there been any prior discussion or consensus among maintainers about dropping Java 8?
What are the concrete, user-facing benefits that Java 17 unlocks and that cannot be backported?
Is the community aware this is being considered? A change of this scope typically warrants an issue or discussion thread first.
If the decision is made to proceed, the PR as submitted would also need significant rework regardless:

334 files across infrastructure, syntax modernization, and API changes should be split into separate PRs.
ReflectionTypeSolver silently changes behaviour for users analysing Java 8 source on a Java 17 JVM — this needs documentation.
The CI matrix would need to drop JDKs below 17.
NodeList.getFirst(), getLast(), addFirst(), and addLast() have return types incompatible with java.util.List defaults added in Java 21 — a follow-up issue should track that.
We appreciate the initiative, but we would like to resolve the strategic question first. Could you point to any prior discussion that motivated this PR?

GG-Feng:
First of all, I sincerely apologize for being too hasty with this PR. That was my mistake.
I don’t intend or request JavaParser to drop Java 8 support at all. My original thought came from seeing many ongoing issues about JDK version compatibility in the repository, so I assumed you might be considering version upgrades.
I also understand this PR cannot and should not be directly merged into the main branch right now. What I’d like to do instead is:
1： Open a dedicated issue to formally discuss this JDK version topic
2： Create a separate branch to maintain these changes independently
3： Keep this branch continuously updated long‑term, providing more JDK version options for JavaParser users

Regarding Java version discussions, Java 17 is just one possible option. If needed, I can also work on supporting more JDK versions.
This way it can help expand JavaParser’s usability and benefit more community members, while keeping the main repository’s Java 8 compatibility fully intact.
I totally agree we should first resolve the strategic questions you raised before any further technical discussion, and I’m happy to follow your guidance throughout this process.

jlerbsc:
For now, I'm closing this PR.

A discussion on this topic is open here https://github.com/javaparser/javaparser/issues/5020

PR #5053, opened 2026-06-27
Add comprehensive unit tests for EqualsVisitor

verhasi:
### Summary

Add 394 unit tests achieving near-100% coverage of `EqualsVisitor.java`. This provides a safety net for a planned refactoring where `EqualsVisitor` will extend `NoCommentEqualsVisitor`.

### Changes

14 new test files + enhancements to existing test files in `javaparser-core-testing/src/test/java/com/github/javaparser/ast/visitor/equals/`:

- `EqualsVisitorMethodTest` — MethodDeclaration, Parameter, InitializerDeclaration
- `EqualsVisitorTypeTest` — ClassOrInterfaceType, PrimitiveType, ArrayType, ArrayCreationLevel, IntersectionType, UnionType, VoidType, WildcardType, UnknownType, VarType
- `EqualsVisitorExpressionTest` — ArrayAccessExpr, ArrayCreationExpr, ArrayInitializerExpr, AssignExpr, BinaryExpr, CastExpr, ClassExpr, ConditionalExpr, EnclosedExpr, FieldAccessExpr, InstanceOfExpr
- `EqualsVisitorLiteralTest` — all literal expression types + TextBlockLiteralExpr
- `EqualsVisitorCallAndAccessTest` — MethodCallExpr, NameExpr, ObjectCreationExpr, ThisExpr, SuperExpr, UnaryExpr, VariableDeclarationExpr
- `EqualsVisitorAnnotationExprTest` — MarkerAnnotationExpr, SingleMemberAnnotationExpr, NormalAnnotationExpr, MemberValuePair
- `EqualsVisitorStatementTest` — ExplicitConstructorInvocationStmt, LocalClassDeclarationStmt, LocalRecordDeclarationStmt, AssertStmt, BlockStmt, LabeledStmt, EmptyStmt, ExpressionStmt
- `EqualsVisitorControlFlowTest` — SwitchStmt, SwitchEntry, BreakStmt, ReturnStmt, IfStmt, WhileStmt, ContinueStmt, DoStmt, ForEachStmt, ForStmt, ThrowStmt, SynchronizedStmt, TryStmt, CatchClause, YieldStmt
- `EqualsVisitorLambdaAndRefTest` — LambdaExpr, MethodReferenceExpr, TypeExpr, SwitchExpr
- `EqualsVisitorModuleTest` — ModuleDeclaration, all module directives
- `EqualsVisitorRecordTest` — RecordDeclaration, CompactConstructorDeclaration
- `EqualsVisitorPatternTest` — TypePatternExpr, RecordPatternExpr, MatchAllPatternExpr
- `EqualsVisitorImportTest` — ImportDeclaration
- `EqualsVisitorMiscTest` — ReceiverParameter, Modifier, NodeList, UnparsableStmt, MarkdownComment

Also added missing tests to pre-existing files: ConstructorTest (thrownExceptions, typeParameters, annotations), ClassOrInterfaceTest (isCompact).

### Coverage results

- **Methods**: 113/113 (100%)
- **Lines**: 587/600 (97.8%)
- **Branches**: 627/748 (83.8%)

### Unreachable code identified

The remaining uncovered branches are structurally unreachable through the public API:

1. **Null guards in private helpers** — `nodesEquals(List, List)` null checks and `objEquals` null check. These inputs are never null (`getOrphanComments()` always returns non-null list; `getOperator()`, `isVarArgs()`, `getIdentifier()` etc. never return null).

2. **Comment check inside each visit method** — Every `visit()` method ends with `if (!nodeEquals(n.getComment(), n2.getComment())) return false;`. This branch can never return false because `commonNodeEquality()` already checks comments *before* `accept()` is called. If comments differ, `nodeEquals` returns false at `commonNodeEquality` and never enters `visit()`.

These findings support a future refactoring to:
- Remove redundant comment checks from each visit method (they're dead code)
- Extract `EqualsVisitor` to extend `NoCommentEqualsVisitor` since the only difference is the comment comparison in `commonNodeEquality`

### Testing

All 394 tests pass. No changes to production code.


verhasi:
Dear @jlerbsc ,

Do you expect from me anything on this PR or you are just busy/on vacation?

br,
István

jlerbsc:
Hi @verhasi, thanks a lot for this substantial effort

On the "planned refactoring": the PR description presents EqualsVisitor extending NoCommentEqualsVisitor as a planned refactoring, but I'm not aware of such a plan on the project side. Note that both classes are generated (by EqualsVisitorGenerator and NoCommentEqualsVisitorGenerator), so any inheritance change would have to happen in the generators, not in the generated files. Could you open an issue to discuss that refactoring first? It would be better to agree on the target design before merging a large test suite whose main purpose is to support it.

In addition, below is a list of comments on the content of the PR.

1. The parallel NoComment* hierarchy and the equalsNodes indirection. I don't see much added value in re-running all ~370 structural tests against NoCommentEqualsVisitor: both visitors are generated from the same metamodel and differ only in comment handling. The 23 NoComment* subclasses (each ~38 lines of identical boilerplate overriding equalsNodes and commentsAffectEquality) exist only to enable this, and the equalsNodes indirection makes the main tests harder to read for little benefit. I'd suggest keeping the EqualsVisitor*Test classes calling EqualsVisitor.equals(...) directly, and testing NoCommentEqualsVisitor only for its delta — comment insensitivity — with a handful of targeted tests like your NoCommentEqualsVisitorCommentTest. That would remove ~900 lines and 22 files with no real coverage loss.

2. Reflection on the private SINGLETON field. Several test classes use getDeclaredField("SINGLETON") + setAccessible(true) to call visit(...) methods directly and reach branches that, as your own description notes, are unreachable through the public API. This is fragile (it depends on a private field name of generated code) and counterproductive for the refactoring: these tests will break exactly when the refactoring removes those dead branches — which is what it should be free to do. I'd rather leave unreachable branches uncovered (or remove them as part of the refactoring) and test only through the public equals(...) methods.

3. The assumeTrue(commentsAffectEquality()) pattern (101 occurrences). In the NoComment* subclasses these tests all report as skipped, which adds noise to test reports. A stronger formulation would assert the expected result as a function of the flag, e.g. assertThat(result, is(!commentsAffectEquality())) — the same test would then verify that EqualsVisitor returns false and NoCommentEqualsVisitor returns true, which is precisely the behavioral contract between the two classes. (Minor: the assumption is also evaluated after parseAndClone(...), doing the parsing work before deciding to skip.)

4. Global parser state in the base class. EqualsVisitorTest.parse() mutates StaticJavaParser configuration, which can leak into other tests running in the same JVM and create order dependence. Please use a local instance instead: new JavaParser(configuration).parse(...). Also, cloneNode can simply be node.clone().

5. The MarkdownComment test pins what looks like a bug. visit_markdownComment_differentContent_false asserts that NoCommentEqualsVisitor compares MarkdownComment content while ignoring the content of the three other comment types. That inconsistency is real in the generated code, but it's almost certainly an oversight in NoCommentEqualsVisitorGenerator (MarkdownComment was added later). I'd prefer to fix the generator rather than lock this behavior in with a test — could you open a separate issue for it?

6. Style consistency. A few things diverge from the codebase conventions: Hamcrest assertThat(result, is(true)) instead of the JUnit assertTrue/assertFalse used elsewhere; many fully-qualified class names inline instead of imports; unnecessary (Visitable) casts; and the abstract base named EqualsVisitorTest looks like a concrete test class (AbstractEqualsVisitorTest would be clearer). The new equals/nocommentequals packages also fragment the visitor tests — note there's already a NoCommentEqualsVisitorTest in com.github.javaparser.ast.visitor.

verhasi:
Hi @jlerbsc , thanks a lot for this substantial effort and the incredibly thorough review! This is exceptionally helpful feedback.

Apologies for the confusion regarding the phrasing! By "planned refactoring," I meant my own intended goal for this contribution, not an official project roadmap item.

Regarding the test suite: I designed these strictly as black-box behavioral tests. Because EqualsVisitor and NoCommentEqualsVisitor are public APIs, the test suite asserts what they do, not how they are implemented under the hood. The standalone value of these tests is actually already proven: while discussing them, they immediately exposed an active regression where `NoCommentEqualsVisitor` erroneously evaluates `MarkdownComment` content. I have isolated that generator bug into a separate tracking issue here: #5056. Even if we decide never to touch the generator architecture, this test suite provides crucial regression protection for the current implementation.

That said, I completely agree we should align on the target design first! I have opened a separate issue to pitch and discuss the generator inheritance proposal here: #5055, so we can agree on the direction before changing any generated files.

To address your excellent feedback on the content of the PR and streamline the test suite, here is my plan to enhance the test suite:

1. **Eliminate Parallel Hierarchy & assumeTrue noise**: I completely see your point regarding the 101 skipped test warnings in the test report—that definitely adds unnecessary noise.
However, because `EqualsVisitor` and `NoCommentEqualsVisitor` represent two entirely distinct public API contracts, I want to ensure we treat this strictly as a black-box behavioral test suite. Relying on the implementation detail that they share a meta-model generator leaves us exposed if the generator itself ever introduces a regression unique to one variant.
To honor the black-box principle while entirely eliminating the skipped test noise, I will refactore the suite to use a stronger dynamic assertion formulation. Instead of skipping the comment-related tests via assumeTrue in the NoComment subclasses, the tests will actively assert that NoCommentEqualsVisitor successfully ignores the comment discrepancies (returning true), while EqualsVisitor catches them (returning false).
This keeps full contract coverage for both public APIs while delivering a 100% clean, warning-free test report.
2. **Remove Reflection**: I will entirely remove the reflection hacks targeting the private SINGLETON field. I agree that we should only test the public equals() contract and leave unreachable generated branches alone for now.
3. **Fix Global State**: I will migrate away from StaticJavaParser to local new JavaParser(configuration) instances to prevent test pollution, and use standard node.clone().
4. **Remove MarkdownComment Pinning**: I will remove the test pinning the `MarkdownComment` bug from this PR so we don't accidentally lock in the buggy behavior, since we are tracking the generator fix separately in  #5056.
5. **Codebase Style alignment**: I will rename the base class to `AbstractEqualsVisitorTest`, consolidate the packages to match the existing layout under com.github.javaparser.ast.visitor, switch from Hamcrest to standard JUnit assertions (`assertTrue`/`assertFalse`), fix the inline FQCNs, and remove unnecessary casts.

I will ping you here as soon as the updated, streamlined test suite is ready for another look. Let's move the architecture discussion over to #5055, and hopefully, we can get this test suite merged to protect the current codebase!

jlerbsc:
Please rebase rather than merge when you update — the branch history still carries ~90 commits duplicating already-merged work.

verhasi:
Apologies for the cluttered history earlier! I have rebased the branch against upstream/master to clean it up. The ~90 duplicate commits have been dropped, leaving only my 2 unique commits with the new unit tests.

verhasi:
Hi @jlerbsc!

Thanks for the excellent architectural feedback again! I have completely overhauled the test suite to treat both visitors strictly as distinct public API contracts under a pure black-box testing strategy, while entirely removing the duplicate boilerplate and report noise.

Here is a summary of the updates now live in the branch:

- **Eliminated Parallel Hierarchy & Fragmentation**: Completely removed the 22 boilerplate subclasses, saving ~900 lines of redundant code. To prevent package fragmentation, all refactored files have been moved directly up into the parent com.github.javaparser.ast.visitor package.
- **Unified Parameterized Strategy**: Both public contracts are now verified using a clean JUnit 5 @ParameterizedTest runner driven by a Strategy enum.
- **Silenced Test Noise & Fixed Naming**: Replaced the 101 assumeTrue skips with functional, active assertions. The suite now actively verifies that NoCommentEqualsVisitor ignores comment changes while EqualsVisitor flags them. Additionally, I cleaned up all 101 method names to drop misleading _false suffixes.
- **Purged Reflection Hacks**: Stripped out every instance of getDeclaredField("SINGLETON") and setAccessible(true). The entire suite now communicates exclusively through the public, static equals(Node, Node) API contracts, ensuring it is entirely refactor-safe.
- **Isolated Parser State**: Fixed the global state leak by discarding mutations to StaticJavaParser in favor of local, isolated JavaParser instances per test execution.

Ready for your review!

jlerbsc:
Thanks for the rework — the parameterized `Strategy` approach is a clear improvement. I checked the mutations against `EqualsVisitor`: the enumeration is complete today, all snippets parse at `BLEEDING_EDGE`, and I found no wrong assertion. A few things to fix before I merge.

**Blocking**

1. `EqualsVisitorsMiscTest` (lines 174-204) — the `MarkdownComment` × `NO_COMMENT` tests pass vacuously. `NoCommentEqualsVisitor` never traverses `getComment()`/orphan comments from a `CompilationUnit` root, so the buggy `visit(MarkdownComment)` (#5056, the only comment visit that compares content) is unreachable from all ~390 tests. Neither the bug nor its fix would be detected. Please add a direct-node test — `NoCommentEqualsVisitor.equals(new MarkdownComment("a"), new MarkdownComment("b"))` should be `true` — marked `@Disabled` with a reference to #5056.

2. `EqualsVisitorsMiscTest#equals_differentMarkdownCommentComment` (line 196) — the body removes the comment from its parent, which fails on the generic `commonNodeEquality` check; the comment-on-a-comment line in `EqualsVisitor.visit(MarkdownComment)` is never reached. It also duplicates `equals_differentMarkdownCommentContent`.

3. `AbstractEqualsVisitorsTest#parse` (line 36) — `getResult().get()` without `isSuccessful()`. On a recovered tree both visitors return `true` for `UnparsableStmt`, so every `*_same_true` test would pass vacuously if a snippet ever stops parsing. Please use the existing `TestParser.parseCompilationUnit()` (`BLEEDING_EDGE`, parser cached per level, `fail()` with the parse problems) instead of the local helper — note that master has meanwhile migrated the sibling visitor tests, including `NoCommentEqualsVisitorTest`, to `StaticJavaParser.newParserAdapter()`.

4. Vacuous/misnamed tests, one line each:
   - `EqualsVisitorsMethodTest#equals_sameParameter_true` (141) is byte-identical to `equals_sameMethod_true` (44) — drop it.
   - `EqualsVisitorsRecordTest#equals_sameCompactConstructor_true` (131) is byte-identical to `equals_sameRecord_true` (42) — drop it.
   - `EqualsVisitorsMiscTest#equals_nodeListDirectly_true` (210) cannot reach `visit(NodeList)`: list children go through the private `nodesEquals(NodeList, NodeList)`, and `equals(Node, Node)` never calls `accept` on a `NodeList`. Drop it or the misleading section comment.
   - `EqualsVisitorsFieldTest#equals_differentMember_false` (45) clears the modifiers — rename to `equals_differentModifiers_false`.

**Worth doing now rather than later**

The comment-mutation body is copy-pasted ~66 times and the `parseAndClone` → `findFirst(...).get()` → mutate → `assertFalse` shape ~200 times. Two or three helpers in `AbstractEqualsVisitorsTest` — e.g. `assertNotEqualAfterMutation(code, type, mutation, strategy)`, `assertCommentChangeHandled(code, type, strategy)`, `<T extends Node> T rightNode(Class<T>)` — would cut each test to one line and keep the comment-equality contract in one place. This is mechanical, and much harder to get done once 4,600 lines are on master.

**Nits**

- `NoCommentEqualsVisitorTest` is left in place and now overlaps `EqualsVisitorsCommentTest`; fold its three cases in or delete the file, so there is a single home for the no-comment contract.
- `Strategy.commentMatters` is only ever read through the negation `expectedResultOnDifferentComments()` (102 call sites) — storing the expected value directly would read better. `cloneNode()` is a one-line pass-through to `clone()`.

**Follow-up, not for this PR**

`EqualsVisitor` is generated from the metamodel, but the test enumeration is manual with nothing keeping the two in sync — a new node type or property will get no test and nothing will flag it. `EqualsVisitorsCommentTest.commentTypesAndStrategies()` already shows the drift (hard-coded `LineComment`/`BlockComment`/`JavadocComment`, `MarkdownComment` patched in elsewhere). Driving the matrix from `JavaParserMetaModel.getNodeMetaModels()`, or at least a sentinel test comparing covered types against the metamodel, would fix this for good. Happy to open a separate issue.

verhasi:
Addressed all review feedback. Rebased on latest master and force pushed (`9955e2f04`).

**Blocking items — all fixed:**
1. Added `@Disabled` direct-node MarkdownComment test with reference to #5056 (includes WARNING to check for duplication with #5057 before enabling)
2. Removed vacuous `equals_differentMarkdownCommentComment` (duplicated and unreachable)
3. Replaced local `parse()` with `StaticJavaParser.newParserAdapter(BLEEDING_EDGE)` — no more `getResult().get()` without success check, no global state mutation
4. Dropped duplicate tests (`equals_sameParameter_true`, `equals_sameCompactConstructor_true`, `equals_nodeListDirectly_true`); renamed `equals_differentMember_false` → `equals_differentModifiers_false`

**Worth doing now — done:**
- Extracted `assertCommentChangeHandled(code, nodeExtractor, strategy)` — reduced ~86 comment-mutation tests to one-liners
- Extracted `assertNotEqualAfterMutation(code, nodeExtractor, mutation, strategy)` — reduced ~205 mutation tests to one-liners
- 2 remaining use complex multi-node mutations (conditional logic, modifying both left and right) — left explicit

**Nits — done:**
- Deleted `NoCommentEqualsVisitorTest` (overlapped with `EqualsVisitorsCommentTest`)
- Simplified `Strategy` enum: stores `expectedResultOnDifferentComments` directly (no double-negation)
- Removed `cloneNode()` pass-through

All 781 tests pass, spotless-clean.

verhasi:
@jlerbsc do you expect from me anything more before review/merge?

PR #5060, opened 2026-07-10
Remove redundant Spotless configuration (trimTrailingWhitespace, endWithNewline)

verhasi:
## Summary

Remove `trimTrailingWhitespace` and `endWithNewline` from the Spotless Maven plugin configuration.

## Rationale

`palantirJavaFormat` already handles both trailing whitespace trimming and final newline insertion. These two explicit steps are no-ops and can be safely removed.

## Verification

Tested by adding a probe file with trailing whitespace and no final newline, then running `spotless:apply` with only `palantirJavaFormat` — both issues are corrected by the formatter alone.

## Changes

- Removed `<trimTrailingWhitespace/>` from Spotless configuration
- Removed `<endWithNewline/>` from Spotless configuration

Net diff: 2 lines deleted.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -449,8 +449,6 @@
                             <!-- google-java-format, but better: see https://github.com/palantir/palantir-java-format -->
                             <palantirJavaFormat/>
                         </java>
-                        <trimTrailingWhitespace/>
-      					<endWithNewline/>
                     </configuration>
                 </plugin>
             </plugins>


jlerbsc:
Thanks for splitting this out from #5057 — the scope separation is appreciated, and the placement itself is right: format steps belong inside a format block, not at the top level of <configuration>.

That said, I tried to reproduce the motivation locally (Maven 3.8.6, Spotless 2.46.1, JDK 17) and ran into two issues.

1. I can't reproduce the build warnings. Running mvn spotless:check -pl javaparser-core before and after the change gives identical output — BUILD SUCCESS, 570 files clean, and no warning about the misplaced elements in either case. Maven appears to ignore them silently. Could you share the exact warning you saw, along with your Maven and JDK versions?

2. The two steps look redundant. I added a probe file with trailing whitespace and no final newline, removed both steps entirely, and ran spotless:apply with palantirJavaFormat alone. Palantir stripped the trailing whitespace and added the final newline on its own. So once moved inside <java>, trimTrailingWhitespace and endWithNewline are effectively no-ops.

Given that, deleting the two lines may be the better change — same diff size, but with a verifiable justification. Happy to go either way.

Two smaller points:

The branch has 5 commits for a 2-line net diff, and 4 of them cancel out (the Spotless revert/re-apply pair, plus the NoComment rollback). A squash into a single commit would make this much easier to review.
The description's stated benefit ("eliminates build warnings") should be updated to match whatever we settle on.

Unrelated, but worth a separate issue: the comment on the Spotless version says "This is the last version with Java 8 support", but 2.46.1 is compiled for Java 11 (class file version 55) and fails outright on JDK 8 with UnsupportedClassVersionError. That comment looks stale.

verhasi:
Thanks for the thorough investigation!

You're right — it wasn't a Maven build warning. IntelliJ IDEA's plugin configuration validation shows the elements in red with a mouse-over suggestion to move them inside `<java>`. Maven itself silently ignores the placement.

Given your finding that `palantirJavaFormat` already handles both `trimTrailingWhitespace` and `endWithNewline` on its own, I agree the cleanest solution is to simply delete the two redundant lines. I'll squash and update the description accordingly.

Regarding the stale Java 8 comment on the Spotless version — happy to open a separate issue for that.

Will push the updated commit shortly.

jlerbsc:
Thanks for the quick turnaround — this addresses everything.

The IntelliJ clarification is the useful piece: that explains why I couldn't reproduce anything on the Maven side. Worth remembering that IDEA's XML validation and Maven's actual behaviour diverge here, since Maven silently ignores misplaced steps rather than warning about them.

I verified the resulting configuration locally (Maven 3.8.6, Spotless 2.46.1, JDK 17). With palantirJavaFormat alone, spotless:check keeps all 570 files in javaparser-core clean, and a probe file with trailing whitespace and no final newline gets both fixed by spotless:apply. The removed steps were indeed no-ops.

The squash and the revised description make this much easier to review — single commit, +0 / -2, and a claim that matches what the code actually does.

PR #5062, opened 2026-07-11
Fix lexical error for Unicode escapes with multiple u characters (#5012)

### LINKED ISSUE #5012: Lexical error when parsing Unicode escapes with multiple 'u' characters
JavaParser fails to parse valid Java source code containing Unicode escape sequences with multiple u characters (e.g., `private String s = "\uuuu1234"`). According to the Java Language Specification, this is a valid syntax.

[JLS-3.3 Unicode Escapes](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html) defines the syntax for Unicode escapes:

> UnicodeMarker:
u {u}

[JLS-2.4](https://docs.oracle.com/javase/specs/jls/se21/html/jls-2.html) defines the grammar notation:

> The syntax {x} on the right-hand side of a production denotes zero or more occurrences of x.
For example, the syntactic production:
ArgumentList:
Argument {, Argument}
states that an ArgumentList consists of an Argument, followed by zero or more occurrences of a comma and an Argument. The result is that an ArgumentList may contain any positive number of arguments.

Therefore, `\u` followed by {u} (zero or more additional 'u's) is a standard-compliant sequence.

- Minimal Reproducible Example:
```java
public class UnicodeTest {

    private String s1 = "t\u1234";
    private String s2 = "t\uuuu1234";

        ParserConfiguration configuration = new ParserConfiguration();
        configuration.setLanguageLevel(ParserConfiguration.LanguageLevel.BLEEDING_EDGE);
        JavaParser parser = new JavaParser(configuration);
        File file = new File(currentFilePath());

        ParseResult<CompilationUnit> result = parser.parse(file);
        System.out.println(result.isSuccessful());
        System.err.println(result.getProblems());
    }

    static String currentFilePath() {
        // change it to your actual path
        return "src/test/java/" +
                UnicodeTest.class.getPackage().getName().replace(".", "/") + "/" +
                UnicodeTest.class.getSimpleName() + ".java";
    }
}
```

kamilkrzywanski:
## Summary

Fixes #5012.

According to [JLS §3.3](https://docs.oracle.com/javase/specs/jls/se21/html/jls-3.html#jls-3.3), a Unicode escape is:

```text
UnicodeEscape:
    \ UnicodeMarker HexDigit HexDigit HexDigit HexDigit
UnicodeMarker:
    u {u}
```

so sequences such as `\uuuu1234` are valid Java. With the default configuration (`preprocessUnicodeEscapes = false`), JavaParser rejected them with a lexical error because the lexer grammar only accepted a single `u`.

### Changes
- Allow zero or more additional `u` characters in `#UNICODE_ESCAPE` (`"\\u" ("u")* …`)
- Reuse `<UNICODE_ESCAPE>` from `STRING_LITERAL` instead of a duplicated single-`u` pattern (covers strings, chars, and identifiers)
- Add `Issue5012Test` regression coverage for multi-`u` escapes with and without Unicode preprocessing

### Note
With `setPreprocessUnicodeEscapes(true)`, this already worked via `UnicodeEscapeProcessingProvider`. The bug was only in the default lexer path.

## Test plan

- [x] `Issue5012Test` (string/char/identifier multi-`u` escapes, full CU example, preprocess on/off)
- [x] Related suites: `UnicodeEscapeProcessingProviderTest`, `StringLiteralExprTest`, `CharLiteralExprTest`, `SimpleNameTest`, `PositionMappingTest`
- [x] Manual reproduction of the issue sample now parses successfully without preprocessing
- [x] `./mvnw spotless:apply` run per CONTRIBUTING.md


jlerbsc:

PR #5063, opened 2026-07-11
Fix method reference resolution after Class::cast in stream pipelines (#4989)

### LINKED ISSUE #4989: bug
Image
When I called the `methodCallExpr.resolve().getQualifiedSignature();` method to parse the code in the image above, I found that it could not parse the method `this::convertToolUseBlockToToolCall`. The error returned was `UnsolvedSymbolException{context='contentBlocks.stream().filter(ToolUseBlock.class::isInstance).map(ToolUseBlock.class::cast).map(this::convertToolUseBlockToToolCall).collect(Collectors.toList())', name='contentBlocks.stream().filter(ToolUseBlock.class::isInstance).map(ToolUseBlock.class::cast).map(this::convertToolUseBlockToToolCall)', cause='java.lang.UnsupportedOperationException: Unable to resolve method reference this::convertToolUseBlockToToolCall with param types [WildcardUsage{type=EXTENDS, boundedType=TypeVariable {ReflectionTypeParameter{typeVariable=R}}}]. Candidates found: 1 method(s) named 'convertToolUseBlockToToolCall' in type ReferenceType{io.agentscope.core.chat.completions.builder.ChatCompletionsResponseBuilder, typeParametersMap=TypeParametersMap{nameToValue={}}}'} 
at com.github.javaparser.symbolsolver.javaparsermodel.contexts.AbstractJavaParserContext.findTypeDeclarations(AbstractJavaParserContext.java:212) 
at com.github.javaparser.symbolsolver.javaparsermodel.contexts.MethodCallExprContext.solveMethod(MethodCallExprContext.java:152) 
at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:326) 
at com.github.javaparser.symbolsolver.javaparsermodel.JavaParserFacade.solve(JavaParserFacade.java:134) 
at com.github.javaparser.symbolsolver.JavaSymbolSolver.resolveDeclaration(JavaSymbolSolver.java:190) 
at com.github.javaparser.ast.expr.MethodCallExpr.resolve(MethodCallExpr.java:332) 
at com.agent.preprocessing.MethodExtractor.lambda$buildMethodJson$12(MethodExtractor.java:236) 
at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)’

kamilkrzywanski:
## Summary

- Fixes symbol solver failure when resolving stream pipelines that use `X.class::cast` followed by instance method references such as `this::convert` (issue #4989).
- Root cause: `toMethodUsage` resolved methods from the raw type declaration, so `Class.cast` returned type variable `T` instead of the concrete type argument (e.g. `Sub` from `Class<Sub>`). Poly inference therefore left `map(Sub.class::cast)` typed as `Stream<Base>` instead of `Stream<Sub>`.
- Fix: substitute type arguments from the method-reference scope into candidate methods before overload resolution / return-type inference.

## Test plan

- [x] Added `Issue4989Test` covering:
  - `map(Sub.class::cast)` refines stream element type to `Stream<Sub>`
  - `this::convert` resolves after `class::isInstance` / `class::cast`
  - `methodCallExpr.resolve().getQualifiedSignature()` on the full pipeline from the issue
  - baseline `this::method` without cast still works
- [x] `./mvnw -pl javaparser-symbol-solver-testing -Dtest=Issue4989Test,MethodReferenceResolutionTest test`
- [x] Manual variants (lambda after cast, cast without filter, unbound `Sub::getText`) resolve correctly

Fixes #4989

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -635,11 +635,9 @@ public class JavaParserFacade {
      * return type of {@code cast} is still the type variable {@code T}; this method replaces it
      * with {@code Sub}.
      */
-    private static MethodUsage applyTypeParametersFromScope(
-            MethodUsage methodUsage, ResolvedReferenceType scopeType) {
+    private static MethodUsage applyTypeParametersFromScope(MethodUsage methodUsage, ResolvedReferenceType scopeType) {
         MethodUsage result = methodUsage;
-        for (Pair<ResolvedTypeParameterDeclaration, ResolvedType> typeParamEntry :
-                scopeType.getTypeParametersMap()) {
+        for (Pair<ResolvedTypeParameterDeclaration, ResolvedType> typeParamEntry : scopeType.getTypeParametersMap()) {
             result = result.replaceTypeParameter(typeParamEntry.a, typeParamEntry.b);
         }
         return result;
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/JavaParserFacade.java
@@ -51,6 +51,7 @@ import com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParse
 import com.github.javaparser.symbolsolver.resolution.SymbolSolver;
 import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
 import com.github.javaparser.utils.Log;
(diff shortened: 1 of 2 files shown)


jlerbsc:
Thanks a lot for digging into #4989 and for the thorough tests — the diagnosis is spot on.

That said, I won't be merging this as-is. The fix substitutes type arguments after the fact in toMethodUsage, whereas the real gap is that ResolvedReferenceType doesn't apply its type arguments to methods the way it already does for fields (getFieldType → useThisTypeParametersOnTheGivenType). Fixing it there rather than in toMethodUsage has two concrete advantages: it reuses the existing substitution primitive (which correctly resolves type variables through ancestors and handles bounded wildcards), instead of a parallel getTypeParametersMap() + replaceTypeParameter path that only covers the direct type and misses inherited generic methods.

So the right place is the type layer, symmetric with fields, which also lets us close the long-standing // TODO replace variables in getDeclaredMethods(). I'll take care of that fix myself shortly.

Thanks again — the reproducer and tests are genuinely useful and I'll keep them.

Issue #4439, opened 2024-05-28
A StackOverflowError occurred when parsing a deep string concatenation


capjiang:
~~~
    @Test
    fun testStackOverFlow() {
        val javaParser = JavaParser()
        val parseResult = javaParser.parse(Paths.get("DeepStringConcat.java")) // error
        if (!parseResult.result.isPresent) {
            return
        }
		...
    }
~~~

I'm using the kotlin language, but that shouldn't affect this routine! The file is https://github.com/openjdk/jdk/blob/master/test/langtools/tools/javac/DeepStringConcat.java

~~~
java.lang.StackOverflowError
	at java.util.ArrayList.addAll(ArrayList.java:583)
	at com.github.javaparser.metamodel.BaseNodeMetaModel.getAllPropertyMetaModels(BaseNodeMetaModel.java:119)
	at com.github.javaparser.ast.validator.language_level_validations.chunks.CommonValidators.lambda$new$7(CommonValidators.java:60)
	at com.github.javaparser.ast.validator.TreeVisitorValidator.accept(TreeVisitorValidator.java:38)
	at com.github.javaparser.ast.validator.TreeVisitorValidator.accept(TreeVisitorValidator.java:40)
	at com.github.javaparser.ast.validator.TreeVisitorValidator.accept(TreeVisitorValidator.java:40)
	at com.github.javaparser.ast.validator.TreeVisitorValidator.accept(TreeVisitorValidator.java:40)
	at com.github.javaparser.ast.validator.TreeVisitorValidator.accept(TreeVisitorValidator.java:40)
	Repeat this.
~~~

jlerbsc:
This is a borderline case linked to the verification of AST nodes. Here the field declaration is initialised with a very large concatenation of character strings and the recursive traversal of the AST raises this problem.

ethan-godden:
I am working on this

jlerbsc:
Closed by this PR https://github.com/javaparser/javaparser/pull/5064.

PR #5064, opened 2026-07-14
fix(#4439): convert TreeVisitorValidator to be iterative

ethan-godden:
This addresses #4439

The original issue in #4439 involves `TreeVisitorValidator`, which was recursive. I have converted it to be iterative by using `Node::walk(Consumer)`. This should be equivalent in functionality while avoiding large calls stacks.

#4439 also notes duplicate issues linked regarding `VoidVisitorAdapter`. I am not sure there is a better solution than just increasing the amount of memory. 

- One idea would be to use a "trampoline" strategy by returning the next nodes to visit rather than using recursive calls. We could do this by creating a `GenericVisitorWithDefaults` that, by default, returns the child nodes and then wrapping this with a `Deque` that keeps track of the node to visit. The problem with this approach is when a user overrides any of these methods by calling `super.visit`.  The execution order would change when compared with the recursive implementation. 
- We could create a new seperate implementation or even a new interface, but I believe this would be confusing 
- We could use reflection or bytecode injection, but I believe this overcomplicates things for a fundamental class

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/TreeVisitorValidatorTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/validator/TreeVisitorValidatorTest.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2026 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+package com.github.javaparser.ast.validator;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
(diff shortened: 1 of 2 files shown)


jlerbsc:
Thanks for the PR — the core change is correct and minimal. I checked Node.walk(Consumer): it delegates to PreOrderIterator, which uses a heap-allocated stack and pushes children in reverse order, so the visit order is strictly identical to the previous recursive pre-order traversal. A request before merging:

Could you add a test asserting the traversal order?

The atomic counter in deeplyNestedTreeIsVisitedWithoutStackOverflow would pass with any traversal order (post-order, BFS, …), yet preserving pre-order — parent visited before its children — is precisely the property most at risk in this refactoring, and validators may depend on it. A small tree where the test records the sequence of visited nodes and compares it against the expected pre-order would lock this in.

ethan-godden:
@jlerbsc 

> Thanks for the PR — the core change is correct and minimal. I checked Node.walk(Consumer): it delegates to PreOrderIterator, which uses a heap-allocated stack and pushes children in reverse order, so the visit order is strictly identical to the previous recursive pre-order traversal. A request before merging:
> 
> Could you add a test asserting the traversal order?
> 
> The atomic counter in deeplyNestedTreeIsVisitedWithoutStackOverflow would pass with any traversal order (post-order, BFS, …), yet preserving pre-order — parent visited before its children — is precisely the property most at risk in this refactoring, and validators may depend on it. A small tree where the test records the sequence of visited nodes and compares it against the expected pre-order would lock this in.

I pushed a new test verifying pre-order traversal.

jlerbsc:
Thanks for adding the pre-order traversal test — it looks good and locks in exactly the property I was concerned about.

One last thing before this can be merged: the Spotless check is failing on the new test file (it's the only red check, everything else is green). Most likely cause is the blank lines separating the import groups, which the project's formatting rules don't allow. Running:

mvn spotless:apply
and pushing the result should fix it.

ethan-godden:
> Thanks for adding the pre-order traversal test — it looks good and locks in exactly the property I was concerned about.
> 
> One last thing before this can be merged: the Spotless check is failing on the new test file (it's the only red check, everything else is green). Most likely cause is the blank lines separating the import groups, which the project's formatting rules don't allow. Running:
> 
> mvn spotless:apply and pushing the result should fix it.

@jlerbsc It should be good now

jlerbsc:
Thank you for this PR.

PR #5066, opened 2026-07-16
(fix: #5065) Fix method resolution when a parameter is a primitive array

### LINKED ISSUE #5065: Method Resolution fails for Parameters of Primitive Array Types
Encountered in the [KeY project](https://github.com/KeYProject/key/pull/3933).

When we have a class like `Arrays`:
```java
class Arrays {
  public static int[] copyOf(int[] original, int newLength) { ,,, }
  public static float[] copyOf(float[] original, int newLength) { ... }
}
```
JavaParser fails to resolve the following call:
```java
int[] arr = {1};
Arrays.copyOf(arr, 1);
```
with the following stack trace (line numbers may be different as we use a fork; the source of the error is unchanged on upstream):
```
com.github.javaparser.resolution.MethodAmbiguityException: Ambiguous method call: cannot find a most applicable method: JavaParserMethodDeclaration{wrappedNode=public static double[] copyOf(double[] original, int newLength);, typeSolver=de.uka.ilkd.key.java.loader.JavaParserFactory$DynamicTypeSolver@787b2e03}, JavaParserMethodDeclaration{wrappedNode=public static float[] copyOf(float[] original, int newLength);, typeSolver=de.uka.ilkd.key.java.loader.JavaParserFactory$DynamicTypeSolver@787b2e03}
        at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:1123)
        at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:993)
        at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveMethod(JavaParserTypeDeclarationAdapter.java:319)
        at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveMethod(ClassOrInterfaceDeclarationContext.java:111)
        at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.solveMethod(JavaParserClassDeclaration.java:339)
        at com.github.javaparser.resolution.logic.MethodResolutionLogic.solveMethodInType(MethodResolutionLogic.java:1364)
        at com.github.javaparser.resolution.logic.MethodResolutionLogic.solveMethodInType(MethodResolutionLogic.java:1352)
```

The problem seems to be `isBoxingCompatibleWithTypeSolver` in `MethodResolutionLogic`. When called with the types `float[]` and `int[]` it calls itself recursively with `float` and `int` and then return `true`, because `float.isAssignableBy(int)` is true.

Instead, `equals` should be used in `isBoxingCompatibleWithTypeSolver` as in `ResolvedArrayType` when the component types are primitive.

Drodt:
Fixes #5065.

The comment in `isBoxingCompatibleWithTypeSolver` was incorrect, as was the logic: When we compare `float[]` and `int[]` and then consider their component types, we should use `equals` as in `ResolvedArrayType#isAssignableBy`.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/TokenTypes.java
@@ -248,7 +248,7 @@ public class TokenTypes {
             case RSIGNEDSHIFT:
             case GT:
                 return JavaToken.Category.OPERATOR;
-                // The following are tokens that are only used internally by the lexer
+            // The following are tokens that are only used internally by the lexer
             case ENTER_JAVADOC_COMMENT:
             case ENTER_MULTILINE_COMMENT:
             case COMMENT_CONTENT:
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -49,8 +49,7 @@ public class MethodResolutionLogic {
             // TODO if there are no variadic values we should default to the bound of the formal type
             res.add(variadicType);
         } else {
-            ResolvedType componentType = findCommonType(variadicValues);
-            res.add(convertToVariadicParameter(componentType));
+            res.add(convertToVariadicParameter(variadicType));
         }
         return res;
     }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -657,11 +657,11 @@ public class MethodResolutionLogic {
                 return false;
             }
         }
-        // Unboxing (primitive expected, reference type provided)
+        // Both are primitive types
         if (expectedType.isPrimitive() && actualType.isPrimitive()) {
             ResolvedPrimitiveType expectedPrimitive = expectedType.asPrimitive();
             ResolvedPrimitiveType actualPrimitive = actualType.asPrimitive();
-            return expectedPrimitive.isAssignableBy(actualPrimitive);
+            return expectedPrimitive.equals(actualPrimitive);
         }
         // Both are reference types but one might be a constrained/wildcard type
         // This can happen after type variable substitution
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -151,7 +151,7 @@ public class MethodResolutionLogic {
                 // array.
                 // Confirm all of these grouped "trailing" arguments have the required type -- if not, this is not a
                 // valid type. (Maybe this is also done later..?)
-                for (int variadicArgumentIndex = countOfMethodParametersDeclared;
+                for (int variadicArgumentIndex = getLastParameterIndex(countOfMethodParametersDeclared);
                         variadicArgumentIndex < countOfNeedleArgumentsPassed;
                         variadicArgumentIndex++) {
                     ResolvedType currentArgumentType = needleArgumentTypes.get(variadicArgumentIndex);
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionLogicTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodsResolutionLogicTest.java
@@ -360,7 +360,7 @@ class MethodsResolutionLogicTest extends AbstractResolutionTest {
                 + "     public static int[] copyOf(int[] original, int newLength) { return original; }"
                 + "     public static float[] copyOf(float[] original, int newLength) { return original; }"
                 + "     public static void test() {"
-                + "       copyOf(new int[]{1});"
+                + "       copyOf(new int[]{1}, 1);"
                 + "     }"
                 + "   }";
         StaticJavaParser.getParserConfiguration().setSymbolResolver(new JavaSymbolSolver(new ReflectionTypeSolver()));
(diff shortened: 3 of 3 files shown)


jlerbsc:
Thanks for the fix! The change is correct: primitive array types are invariant per the JLS (int[] is not assignable to float[]), and using equals on the component types makes isBoxingCompatibleWithTypeSolver consistent with ResolvedArrayType#isAssignableBy. I also verified there's no regression on primitive widening for plain arguments (int → long), since both call sites only reach this fallback after isAssignableBy has already failed on the full types.

One request: could you add a test, e.g. Issue5065Test in javaparser-symbol-solver-testing, covering the repro from the issue (overloads copyOf(int[], int) / copyOf(float[], int) resolved with an int[] argument) and ideally a widening non-regression case (m(long) called with an int)?

LGTM otherwise.

jlerbsc:
Thanks for the fix — the change itself is correct: primitive array types are invariant per the JLS, and using equals on the component types is consistent with ResolvedArrayType#isAssignableBy.

However, this change breaks MethodsResolutionLogicTest.testPrimitiveWideningVarargs (print(double... values) called with print(1, 2, 3)), and the failure reveals a second, pre-existing bug that the old lenient check was silently compensating for.

What happens: when the trailing varargs arguments are grouped into an array, groupVariadicParamValues derives the component type from findCommonType, which is a // TODO implement this decently stub that just returns the type of the first argument. So the synthetic grouped argument becomes int[] instead of double[]. The main loop then compares expected double[] against actual int[], regular assignability fails (array invariance), and it falls back to isBoxingCompatibleWithTypeSolver — where the old isAssignableBy widened int to double and accepted the match. With equals, the match is now rejected, even though per JLS 15.12.4.2 widening applies element-wise in a variadic invocation. At that point in the code the synthetic group is indistinguishable from a genuine int[] argument (the [#5065](https://github.com/javaparser/javaparser/issues/5065) case), so the boxing fallback can't tell the two apart.

Proposed fix: rather than re-loosening the primitive check, I'd suggest fixing the grouping so both bugs are addressed together:

In the excess-arguments pre-check loop in isApplicable, start at lastMethodParameterIndex instead of countOfMethodParametersDeclared — currently the first grouped argument is never validated element-wise (this off-by-one is exactly the gap the old widening was papering over).
Once every grouped element has been validated against the declared variadic component type, group them into an array of the declared component type (methodDeclaration.getLastParam().getType()) instead of convertToVariadicParameter(findCommonType(...)). The grouped argument for print(1, 2, 3) then becomes double[] and matches trivially, while a genuine int[] argument against a non-variadic float[] parameter ([#5065](https://github.com/javaparser/javaparser/issues/5065)) is still correctly rejected.
With that, your equals change stays as-is and both testPrimitiveWideningVarargs and the [#5065](https://github.com/javaparser/javaparser/issues/5065) scenario pass. It would also be great to add a test for the issue's repro (copyOf(int[], int) / copyOf(float[], int) resolved with an int[] argument), e.g. an Issue5065Test in javaparser-symbol-solver-testing.

Drodt:
I added the tests. But why are all the CI tests failing? I also can't run the maven tests on my machine (maven complains about wrong return types) so I can't really test any changes to method resolution.

jlerbsc:
Please see my previous comment to understand why the tests are failing.

Drodt:
It seems to be fixed now. The method resolution tests all run now.

jlerbsc:
<p>Thanks for iterating on this. I checked out the branch and ran comparative probes against <code>master</code>. <strong>The <code>equals</code> fix is correct, but the varargs grouping change introduces regressions.</strong></p>
<p><strong>What works.</strong> The core fix is genuinely right, and I confirmed it against the real repro: <code>int[] a = {1}; java.util.Arrays.copyOf(a, 1);</code> throws <code>MethodAmbiguityException</code> on <code>master</code> and resolves correctly on this branch. Same for a lone <code>void f(float[] a)</code> overload called with <code>int[]</code>.</p>
<p><strong>Blocking issue.</strong> The rewrite of <code>groupVariadicParamValues</code> makes varargs applicability effectively unconditional for single-argument calls. These are all correctly rejected on <code>master</code> and <strong>wrongly accepted</strong> on this branch:</p>

call | master | this PR
-- | -- | --
print(double... v) with print("hello") | rejected ✅ | accepted ❌
print(String... v) with print(1) | rejected ✅ | accepted ❌
print(int... v) with print(1.5) | rejected ✅ | accepted ❌


<p>Root cause: <code>convertToVariadicParameter</code> is idempotent on array types, and <code>variadicType</code> is always already an array at all three call sites, so both branches of the <code>if/else</code> now do the same thing — the grouped argument type no longer depends on the actual arguments at all. The index fix only guards the <code>countOfNeedleArgumentsPassed &gt; countOfMethodParametersDeclared</code> branch; the equal-arity path (a single variadic argument, by far the most common case) never validates element-wise, and now the actual type is overwritten before the final comparison can catch it.</p>
<p><strong>Two more points:</strong></p>
<ul>
<li><code>testIssue5065</code> <strong>passes on <code>master</code> without the fix</strong> — it doesn't reproduce the bug. With <code>int[]</code> declared first, most-specific-method selection masks the over-permissive <code>isApplicable</code>. Please use a repro that actually fails, e.g. <code>java.util.Arrays.copyOf(a, 1)</code> with <code>int[] a</code>, or a single <code>void f(float[] a)</code> overload called with <code>int[]</code>.</li>
<li>The full <code>javaparser-symbol-solver-testing</code> suite is green (2078 tests, 0 failures) despite the regressions above, so CI isn't a safety net here — there's no negative test for a single incompatible variadic argument.</li>
</ul>

<p><strong>Suggestion:</strong> keep the <code>equals</code> change (it alone fixes #5065), drop the <code>groupVariadicParamValues</code> rewrite, and handle variadic widening narrowly — allow component-wise compatibility only for the last parameter when it is variadic. That cleanly separates the two cases, since the offending <code>float[] original</code> parameter in #5065 isn't variadic. Please add negative tests for the three cases above.</p>

<p>Minor: The <code>TokenTypes.java</code> hunk looks like an unrelated spotless artifact worth dropping from this PR.</p></body></html>

jlerbsc:
You need to make some changes so that we can merge your PR.

PR #5075, opened 2026-07-24
fix: bind inherited functional-method type variables in method references (fixes #4936)

### LINKED ISSUE #4936: Unsolved symbol in Stream.of(new BigDecimal(0L)).reduce(BigDecimal::add).orElse(null)
code1 throws a UnsolvedSymbolException
```java


public class Issue10086Test {

	@Test()
	void test() {
		String code = "import java.math.BigDecimal;\n"
			+ "import java.util.stream.Stream;\n"
			+ "public class Test{\n"
			+ "\t\t\t\tpublic void test(){\n"
			+ "\t\t\t\t\tStream.of(new BigDecimal(0L)).reduce(BigDecimal::add);\n"
			+ "\t\t\t\t}\n"
			+ "\t\t\t}";

		combinedTypeSolver.add(new ReflectionTypeSolver());
		JavaSymbolSolver symbolSolver = new JavaSymbolSolver(combinedTypeSolver);
		ParserConfiguration parserConfiguration = new ParserConfiguration()
			.setSymbolResolver(symbolSolver);
		parserConfiguration.setStoreTokens(false);
		JavaParser javaParser = new JavaParser(parserConfiguration);


		CompilationUnit cu = javaParser.parse(code).getResult().get();
		for (MethodCallExpr methodCallExpr : cu.findAll(MethodCallExpr.class)) {
			methodCallExpr.resolve();
		}

		String code1 = "import java.math.BigDecimal;\n"
			+ "import java.util.stream.Stream;\n"
			+ "public class Test{\n"
			+ "\t\t\t\tpublic void test(){\n"
			+ "\t\t\t\t\tStream.of(new BigDecimal(0L)).reduce(BigDecimal::add).orElse(null);\n"
			+ "\t\t\t\t}\n"
			+ "\t\t\t}";

		CompilationUnit cu1 = javaParser.parse(code1).getResult().get();
		for (MethodCallExpr methodCallExpr : cu1.findAll(MethodCallExpr.class)) {
			methodCallExpr.resolve();
		}
	}
}
```

ethan-godden:
Fixes #4936

### Issue

Resolving `Stream.of(new BigDecimal(0L)).reduce(BigDecimal::add).orElse(null)` threw
`UnsolvedSymbolException`. `BinaryOperator<T>` inherits its functional method `apply(T, U)` from
`BiFunction<T, U, R>`, so the method's formal types are expressed with the *ancestor's* type
parameters. The substitution in `TypeExtractor.visit(MethodReferenceExpr)` used the interface's
direct type-parameter map with name-based matching: `BiFunction.T` was bound only through its
accidental name collision with `BinaryOperator.T`, and `BiFunction.U` leak
`BigDecimal::add` was then looked up with parameter types `[BigDecimal, U]` and failed.

### Fix

Replace the name-based `replaceTypeParameter` loop with the ancestor-aware
`useThisTypeParametersOnTheGivenType()`, which resolves each type variable
hierarchy. Ancestor variables such as `BiFunction.U` are bound by the interface's parametrization
instead of by name coincidence.

### Tests

Each test was verified to fail before the fix, covers a distinct scenario,
resolved method signatures (e.g. that `reduce` resolves to the single-argument
`Stream.reduce(BinaryOperator<T>)` overload) rather than just that resolut

- `issue4979_inheritedFunctionalMethodTypeParametersDoNotLeak` — the repor
  instance reference against `BinaryOperator<BigDecimal>` (leaked `[BigDecimal, U]` pre-fix)
- `…_noOwnTypeParameters` — `interface DecimalOp extends BinaryOperator<Bi
  type parameters of its own, so name matching had nothing to bind (leaked `[T, U]`)
- `…_transitiveInheritance` — two-level generic chain with non-colliding p
  transitive ancestor substitution (leaked `[T, U]`)
- `…_staticMethodReference` — static reference `Integer::sum`, pinning the
  branch (leaked `[Integer, U]`)
- `…_wildcardTypeArgument` — `Op<? super BigDecimal>` target, pinning subs
  unwrapping (leaked `[T, U]`)

### DIFF
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/TypeExtractor.java
@@ -53,6 +53,7 @@ import com.github.javaparser.resolution.model.typesystem.ReferenceTypeImpl;
 import com.github.javaparser.resolution.promotion.ConditionalExprHandler;
 import com.github.javaparser.resolution.types.ResolvedArrayType;
 import com.github.javaparser.resolution.types.ResolvedPrimitiveType;
+import com.github.javaparser.resolution.types.ResolvedReferenceType;
 import com.github.javaparser.resolution.types.ResolvedType;
 import com.github.javaparser.resolution.types.ResolvedVoidType;
 import com.github.javaparser.symbolsolver.resolution.SymbolSolver;
@@ -60,7 +61,6 @@ import com.github.javaparser.symbolsolver.resolution.promotion.ConditionalExprRe
 import com.github.javaparser.symbolsolver.resolution.typeinference.LeastUpperBoundLogic;
 import com.github.javaparser.symbolsolver.resolution.typeinference.TypeHelper;
 import com.github.javaparser.utils.Log;
-import com.github.javaparser.utils.Pair;
 import com.google.common.collect.ImmutableList;
 import java.util.List;
 import java.util.Objects;
@@ -843,10 +843,23 @@ public class TypeExtractor extends DefaultVisitorAdapter {
                 if (functionalMethodOpt.isPresent()) {
                     MethodUsage functionalMethod = functionalMethodOpt.get();
 
-                    for (Pair<ResolvedTypeParameterDeclaration, ResolvedType> typeParamDecl :
-                            result.asReferenceType().getTypeParametersMap()) {
-                        functionalMethod = functionalMethod.replaceTypeParameter(typeParamDecl.a, typeParamDecl.b);
+                    // Bind the type variables that appear in the functional method's signature using
+                    // the type arguments of the functional interface. The functional method is often
+                    // inherited (e.g. BinaryOperator<T> inherits apply from BiFunction<T, T, T>), so its
+                    // formal types are expressed with the *ancestor's* type variables. Resolving them via
+                    // useThisTypeParametersOnTheGivenType() walks the type hierarchy, so ancestor type
+                    // variables such as BiFunction's U are bound instead of leaking out unresolved (which
+                    // previously happened because a direct, name-based substitution only matched the
+                    // interface's own type parameters). See issue #4936.
+                    ResolvedReferenceType functionalInterfaceType = result.asReferenceType();
+                    for (int i = 0; i < functionalMethod.getNoParams(); i++) {
+                        functionalMethod = functionalMethod.replaceParamType(
+                                i,
+                                functionalInterfaceType.useThisTypeParametersOnTheGivenType(
+                                        functionalMethod.getParamType(i)));
                     }
+                    functionalMethod = functionalMethod.replaceReturnType(
+                            functionalInterfaceType.useThisTypeParametersOnTheGivenType(functionalMethod.returnType()));
 
                     // replace wildcards
                     for (int i = 0; i < functionalMethod.getNoParams(); i++) {
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodReferenceResolutionTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/MethodReferenceResolutionTest.java
@@ -573,6 +573,169 @@ class MethodReferenceResolutionTest extends AbstractResolutionTest {
         assertEquals(0, errorCount, "Expected zero UnsolvedSymbolException s");
     }
 
+    /**
+     * BinaryOperator inherits its functional method apply(T, U) from BiFunction, so resolving
+     * BigDecimal::add must bind the ancestor's type parameters instead of leaking them unresolved.
+     * See <a href="https://github.com/javaparser/javaparser/issues/4936">issue #4936</a>.
+     */
+    @Test
+    public void issue4936_inheritedFunctionalMethodTypeParametersDoNotLeak() {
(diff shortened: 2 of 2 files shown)


jlerbsc:
The root-cause analysis is correct and I verified it against the code: the previous loop over getTypeParametersMap() combined with the name-based matching in ResolvedReferenceType.replaceTypeVariables() (tp.asTypeVariable().describe().equals(tpToReplace.getName())) only bound BinaryOperator.T by accidental name collision, leaving the ancestor's BiFunction.U/R unbound → the leak. Switching to useThisTypeParametersOnTheGivenType() is the right, hierarchy-aware fix and is consistent with how MethodCallExprContext/LambdaExprContext already do this. 👍

Two things before merge:

1. Issue number inconsistency (blocking). The PR title targets #4936, but every test method is named issue4979_... and all Javadoc/code comments reference #4979. Please align everything on a single number (or state the relationship explicitly if #4979 is a duplicate). As-is it breaks traceability.

2. Add a non-regression test for the case that already worked (reduce(...) without .orElse(...)), to lock the nominal path alongside the new failing case.

Nice test breadth otherwise (no-own-type-params, transitive inheritance, static method reference, wildcard) — the fix itself is well targeted and I'd be happy to see it merged once the issue-number references are cleaned up.

ethan-godden:
> The root-cause analysis is correct and I verified it against the code: the previous loop over getTypeParametersMap() combined with the name-based matching in ResolvedReferenceType.replaceTypeVariables() (tp.asTypeVariable().describe().equals(tpToReplace.getName())) only bound BinaryOperator.T by accidental name collision, leaving the ancestor's BiFunction.U/R unbound → the leak. Switching to useThisTypeParametersOnTheGivenType() is the right, hierarchy-aware fix and is consistent with how MethodCallExprContext/LambdaExprContext already do this. 👍
> 
> The root-cause analysis is correct and I verified it against the code: the previous loop over getTypeParametersMap() combined with the name-based matching in ResolvedReferenceType.replaceTypeVariables() (tp.asTypeVariable().describe().equals(tpToReplace.getName())) only bound BinaryOperator.T by accidental name collision, leaving the ancestor's BiFunction.U/R unbound → the leak. Switching to useThisTypeParametersOnTheGivenType() is the right, hierarchy-aware fix and is consistent with how MethodCallExprContext/LambdaExprContext already do this. 👍
> 
> Two things before merge:
> 
> 1. Issue number inconsistency (blocking). The PR title targets [Unsolved symbol in Stream.of(new BigDecimal(0L)).reduce(BigDecimal::add).orElse(null) #4936](https://github.com/javaparser/javaparser/issues/4936), but every test method is named issue4979_... and all Javadoc/code comments reference [fix(deps): update byte-buddy.version to v1.18.7-jdk5 #4979](https://github.com/javaparser/javaparser/pull/4979). Please align everything on a single number (or state the relationship explicitly if [fix(deps): update byte-buddy.version to v1.18.7-jdk5 #4979](https://github.com/javaparser/javaparser/pull/4979) is a duplicate). As-is it breaks traceability.
> 2. Add a non-regression test for the case that already worked (reduce(...) without .orElse(...)), to lock the nominal path alongside the new failing case.
> 
> Nice test breadth otherwise (no-own-type-params, transitive inheritance, static method reference, wildcard) — the fix itself is well targeted and I'd be happy to see it merged once the issue-number references are cleaned up.

@jlerbsc I have added the test and fixed the 4979 typos

jlerbsc:
Both earlier points are addressed cleanly — issue references are now consistently #4936, the non-regression test (_withoutChainedCall) is in, and the history is squashed to a single commit. Thanks. The fix itself is unchanged and correct.

One remaining nit, on commit-message quality:

The commit subject fix(#4936): resolve inherited functional method type variables in method reference resolution doesn't match this repo's convention and won't auto-close the issue. Compare with recent history, e.g. build: restore OSGi support accidentally reverted by #5019 (fixes #1477) — the issue reference goes in a (fixes #NNNN) suffix (or a Fixes #NNNN footer), not in the scope position. As written, merging this will not close [#4936](https://github.com/jlerbsc/javaparser/issues/4936).

Could you amend + force-push (still one commit) to:

use the repo style and add auto-close, e.g. fix: bind inherited functional-method type variables in method references (fixes #4936);
shorten the subject to ≤72 chars;
add a short body capturing the root cause and fix so git log is self-contained, e.g.:
The previous binding looped over the interface's own type-parameter map
and substituted by variable *name*, so an inherited functional method
(BinaryOperator apply() inherited from BiFunction<T,U,R>) left the
ancestor variables U/R unbound and leaking as UnsolvedSymbolException.
Bind through the type hierarchy with useThisTypeParametersOnTheGivenType()
instead, which resolves ancestor type variables regardless of name.
Once the message is tidied up, this LGTM.

ethan-godden:
@jlerbsc I have modified the commit message

jlerbsc:
Thanks for the thorough work on this, and apologies for the late feedback.

After re-testing, [#4936](https://github.com/jlerbsc/javaparser/issues/4936) is already fixed on master by [#5080](https://github.com/javaparser/javaparser/pull/5080), which filled the // TODO replace variables in ReferenceTypeImpl.getDeclaredMethods() so that a parameterized type now substitutes its type arguments into each method usage's parameter, return and exception types (via useThisTypeParametersOnTheGivenType). 

The change in TypeExtractor re-applies that same substitution on the consumer side, which is exactly the kind of manual re-substitution [#5080](https://github.com/javaparser/javaparser/pull/5080) was meant to remove — so it's now redundant (and narrower, since it doesn't cover exception types). For that reason I'm going to close this PR.

Thanks again for contributing.

PR #5076, opened 2026-07-25
fix: describe lambda parameter types without a spurious '? super' prefix

### LINKED ISSUE #5040: Type Resolution Fails for Lambda Expressions When Used as Method Parameters
📌 Description
When using a lambda expression as a method parameter (e.g., inside Map.forEach()), JavaParser fails to correctly resolve the type of lambda parameters (like v in (key, v) -> ...) even when the enclosing context clearly defines the generic type (e.g., Map<String, Value>).

Specifically:

The lambda parameter v is expected to be resolved as Value, but JavaParser returns UnknownType or fails to infer its type.
The method CombinedTypeSolver.tryToSolveType(...) is not triggered during lambda parameter analysis.
This prevents accurate semantic analysis of expressions like new Value<String>(v) — where the type of v should be inferred as String.
This issue breaks type inference in functional contexts, especially for:

Map<K, V>.forEach((k, v) -> ...)
Stream<T>.forEach(...) with lambda parameters
Generic functional interfaces (Function<T, R>, BiConsumer<T, U>)
✅ Steps to Reproduce
```java
public class Test {
    public void example() {
        Map<String, String> request = ...;
        Map<String, Value> result = new HashMap<>();

        request.forEach((key, v) -> {
            result.put(key, new Value<String>(v)); // ← v should be inferred as String
        });
    }
}
```
Expected Behavior:

The lambda parameter v should be correctly resolved as type **String**.
CombinedTypeSolver.tryToSolveType(...) should be invoked during parameter analysis.
Expression new Value<String>(v) should allow proper type checking (e.g., confirm v is compatible with Value).
Actual Behavior:

v is resolved as UnknownType or ``.
No call to tryToSolveType() occurs during lambda parameter resolution.
Subsequent type checks fail or produce incorrect results.
❗️Note: The generic context (Map<String, Value>) is clear, but JavaParser does not use it to infer the lambda parameter types.

arimu1:
### Summary

Fixes #5040

`ResolvedLambdaConstraintType.describe()` hard-coded a `"? super "` prefix:

```java
public String describe() {
    return "? super " + bound.describe();
}
```

So a lambda parameter inferred as `String` was reported as `"? super java.lang.String"`. **No wildcard is involved.** When the functional interface method parameter is a wildcard, `LambdaExprContext` already unwraps it and stores the *bounded type* as the constraint's bound:

```java
if (argType.isWildcard()) {
    conType = ResolvedLambdaConstraintType.bound(argType.asWildcard().getBoundedType());
}
```

For the issue's example, `v` already resolves to a constraint whose bound is `java.lang.String` — the prefix described a wildcard that had already been stripped. This reports the bound itself.

This matches @johannescoetzee's conclusion in [the issue](https://github.com/javaparser/javaparser/issues/5040#issuecomment-3011452403) that `? super java.lang.String` is not correct and the answer should be exactly `java.lang.String`.

**Correction (per @jlerbsc, 2026-07-25):** the *reason* is not that `String` is final. Finality collapses `? extends String`, since no proper subtype exists — it does nothing to `? super String`, which still admits `CharSequence`, `Serializable` and `Object`. The parameter type is `String` because of JLS §15.27.1 / §9.9 / §9.8: an implicitly typed lambda parameter takes its type from the function type of the target type's **non-wildcard parameterization**, in which a lower-bounded argument `? super L` maps directly to `L`. The inferred type is therefore never a wildcard, and `? extends …` would be equally wrong. The behaviour holds for non-final bounds too — see the new test below.

### On the earlier revision of this PR

The first version of this PR was wrong and I've replaced it entirely. For the record, it:

- unwrapped constraint types in `JavaParserParameterDeclaration.getType()` and `JavaSymbolSolver`, which regressed `Issue2065Test` (`Math.max` inside `Stream.reduce` became unresolvable);
- relaxed `ReferenceTypeImpl.isAssignableBy` to treat `? super T` like `? extends T`, which is unsound — `Foo.isAssignableBy(? super Foo)` must be `false` because `? super Foo` may be `Object`. That broke `ReferenceTypeTest.testIsAssignableByGenerics` and `WildcardUsageTest.testIsAssignableBySimple`, which encode correct Java semantics;
- carried a test that asserted `"? super java.lang.String"` — the value the issue calls incorrect — and therefore **passed on an unmodified `master`** without testing anything.

None of those changes are in this revision; inference behaviour is untouched.

### Assertions updated

Seven assertions pinned the old rendering and now expect the bound. All are lambda parameters resolved through `ResolvedLambdaConstraintType`; no assertion on a real `ResolvedWildcard` changed (e.g. `WildcardUsageTest` lines 189/192 still expect `? super …`).

- `Issue3859Test` ×2
- `ContextTest.resolveReferenceToLambdaParamBase`, `ContextTest.resolveReferenceToLambdaParam`
- `LambdaResolutionTest.lambdaParameterTypeIsResolvedToStringNotTypeVariable`
- `LambdaExprContextResolutionTest.solveParameterOfLambdaInMethodCallExpr`, `…InStaticGenericMethodWithOuterContextInference`

### Test Plan

`lambdaParameterOfWildcardBoundedFunctionalInterfaceIsDescribedWithoutTheWildcard` uses the issue's snippet and asserts on both the parameter declarations (`resolve().describeType()`, the API the issue quotes) and the use site (`calculateResolvedType().describe()`). Verified it **fails on `master`** (`expected: <java.lang.String> but was: <? super java.lang.String>`) and passes with the fix.

`lambdaParameterOfWildcardBoundedFunctionalInterfaceKeepsANonFinalBound` (added at @jlerbsc's suggestion) pins the behaviour to the §9.8 rule rather than to finality, using `Consumer<? super Number>` — `Number` is not final, and the parameter still resolves to exactly `java.lang.Number`. Also verified to fail without the `describe()` change:

```
AssertionFailedError: expected: <java.lang.Number> but was: <? super java.lang.Number>
```

`LambdaResolutionTest` — 29 tests, 0 failures, 0 errors (1 pre-existing skip).

Full suites green on JDK 17:

- `javaparser-core-testing` — 2832 tests, 0 failures, 0 errors
- `javaparser-symbol-solver-testing` — 2078 tests, 0 failures, 0 errors
- `spotless:check` clean

### Note

@johannescoetzee, you mentioned you'd take a look at this one — happy to close this if you have a fix in progress, or to adjust the approach. `describe()` is the narrowest place to correct the rendering, but if you'd rather the constraint were unwrapped at the resolution boundary instead, that path needs the `Issue2065Test` interaction sorted out first.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedLambdaConstraintType.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/types/ResolvedLambdaConstraintType.java
@@ -28,9 +28,15 @@ public class ResolvedLambdaConstraintType implements ResolvedType {
         this.bound = bound;
     }
 
+    /**
+     * The bound is the type the lambda parameter was inferred to, so that is what callers are
+     * shown. Rendering it as {@code "? super " + bound} used to suggest a lower-bounded wildcard,
+     * but no wildcard is involved: when the functional interface method parameter is one,
+     * {@code LambdaExprContext} already stores its bounded type here.
+     */
     @Override
     public String describe() {
-        return "? super " + bound.describe();
+        return bound.describe();
     }
 
     public ResolvedType getBound() {
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3859Test.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/Issue3859Test.java
@@ -62,12 +62,10 @@ public class Issue3859Test extends AbstractResolutionTest {
 
         List<NameExpr> exprs = cu.findAll(NameExpr.class);
         assertEquals(2, exprs.size());
-        assertEquals(
-                "? super java.lang.String", exprs.get(0).calculateResolvedType().describe());
+        assertEquals("java.lang.String", exprs.get(0).calculateResolvedType().describe());
         // Before the fix the following statement failed with an
         // `UnsupportedOperationException` because an extra `(...)` around
         // an argument wasn't handled.
-        assertEquals(
-                "? super java.lang.String", exprs.get(1).calculateResolvedType().describe());
+        assertEquals("java.lang.String", exprs.get(1).calculateResolvedType().describe());
     }
 }
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/ContextTest.java
@@ -320,7 +320,7 @@ class ContextTest extends AbstractSymbolResolutionTest {
         JavaParserFacade javaParserFacade = JavaParserFacade.get(typeSolver);
         ResolvedType ref = javaParserFacade.getType(refToT);
 
-        assertEquals("? super java.lang.String", ref.describe());
+        assertEquals("java.lang.String", ref.describe());
     }
 
     @Test
@@ -564,7 +564,7 @@ class ContextTest extends AbstractSymbolResolutionTest {
         TypeSolver typeSolver = new CombinedTypeSolver(new ReflectionTypeSolver(), new JarTypeSolver(pathToJar));
         ResolvedType typeOfT = JavaParserFacade.get(typeSolver).getType(referenceToT);
 
-        assertEquals("? super com.github.javaparser.ast.body.TypeDeclaration", typeOfT.describe());
+        assertEquals("com.github.javaparser.ast.body.TypeDeclaration", typeOfT.describe());
     }
 
     @Test
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/LambdaResolutionTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/resolution/LambdaResolutionTest.java
@@ -30,6 +30,8 @@ import com.github.javaparser.ast.body.MethodDeclaration;
 import com.github.javaparser.ast.expr.Expression;
 import com.github.javaparser.ast.expr.LambdaExpr;
 import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
 import com.github.javaparser.ast.stmt.ReturnStmt;
 import com.github.javaparser.resolution.Navigator;
 import com.github.javaparser.resolution.types.ResolvedType;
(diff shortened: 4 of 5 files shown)


jlerbsc:
The fix is correct in direction: for an implicit lambda parameter, the inferred type is never a wildcard. Per JLS §15.27.1 / §9.9 / §9.8, the parameter type comes from the function type of the target's non-wildcard parameterization, where a lower-bounded argument ? super L maps directly to L. So dropping the ? super prefix in describe() and returning the bare bound (java.lang.String) matches the spec — and returning ? extends … would be equally wrong.

Two notes worth capturing:

Justification. If the rationale is "since String is final, it can be refined to exactly String", that reasoning is unsound for this case. Finality collapses ? extends String (no proper subtype exists), but not ? super String, which still admits CharSequence, Serializable, Object. The parameter is String because of the §9.8 rule, not because String is final — e.g. Consumer<? super Number> yields exactly Number, and Number isn't final.

Not purely cosmetic. describe() also feeds resolution logic (JavaParserTypeAdapter.solveType(type.describe()), type.describe().equals(qualifiedName) in the reflection/javassist adapters). The old "? super java.lang.String" string could break those lookups when a lambda constraint reached them, so this also removes a latent hazard. isConstraint() still returns true, so callers that need to distinguish a constraint use the API, not the string — no regression there.

Suggest adding a test covering a non-final bound (e.g. Consumer<? super Number>) alongside the String case, to lock the behavior to the §9.8 semantics rather than to finality.

arimu1:
@jlerbsc thank you — both points taken, and both are now reflected in the PR.

**1. The justification was wrong.** You are right that finality is the wrong lever. It collapses `? extends String` (no proper subtype exists) but leaves `? super String` wide open — `CharSequence`, `Serializable`, `Object` all satisfy it. I had carried that reasoning over from the issue thread without checking it. The PR description now cites JLS §15.27.1 / §9.9 / §9.8 instead: the implicitly typed parameter takes its type from the function type of the target's **non-wildcard parameterization**, where `? super L` maps directly to `L`. That is why the result is never a wildcard, and why `? extends …` would be equally wrong.

**2. Added the non-final-bound test.** `lambdaParameterOfWildcardBoundedFunctionalInterfaceKeepsANonFinalBound` in `LambdaResolutionTest` uses exactly your example:

```java
class Test {
    public void example(Consumer<? super Number> consumer) {
        consumer.accept(1);
    }
    public void call() {
        example(n -> n.intValue());
    }
}
```

`Number` is not final, and the parameter still resolves to exactly `java.lang.Number` — so the test is pinned to §9.8 and not to finality. I confirmed it is a real regression test by reverting `describe()` to `"? super " + bound.describe()` and rebuilding `javaparser-core`:

```
AssertionFailedError: expected: <java.lang.Number> but was: <? super java.lang.Number>
```

With the fix restored: `LambdaResolutionTest` — 29 tests, 0 failures, 0 errors (1 pre-existing skip), `spotless:apply` no-op.

**On your "not purely cosmetic" note** — agreed, and that is the more important half. `describe()` feeds `JavaParserTypeAdapter.solveType(type.describe())` and the `type.describe().equals(qualifiedName)` comparisons in the reflection/javassist adapters, so the old `"? super java.lang.String"` string could silently miss a lookup whenever a lambda constraint reached them. `isConstraint()` still returns `true`, so callers that genuinely need to distinguish a constraint use the API rather than the rendered string — no information is lost.

Pushed as `395265fc1`.

jlerbsc:

PR #5086, opened 2026-07-31
Make PrimitiveType keyword generation locale-independent

### LINKED ISSUE #3018: Code generation is locale-sensitive and thus produces illegal Java code
There are locales, where lower-case / upper-case mappings are not what you expect.
If you for example have a turkish locale (`tr`), the lower-case version of `I` is not `i` but `ı` (no dot).
In version 3.6.8 this for example caused `public` to be `publıc` and thus non-valid Java code due to `final String codeRepresentation = this.name().toLowerCase();` in the `Modifier` enum.
This changed in 3.9.0 when `Modifier` was made a node and the code representations hard-coded.

But there are other such bugs. For anything non-human-readable like this, `toLowerCase` and `toUpperCase` should always be used with an explicit locale, just as files should always be read with an explicit encoding.

Another example is `this.codeRepresentation = name().toLowerCase();` in `Primitive` which still fails to work properly in the current 3.18.0.

This small snippet for example demonstrates the problem:
```java
Locale.setDefault(Locale.forLanguageTag("tr"));
CompilationUnit cu = new CompilationUnit();
cu.addInterface("Foo").addMethod("foo").addParameter(PrimitiveType.intType(), "foo");
new PrettyPrinter().print(cu);
```

You can even reproduce (at least some?) with your own tests if you run them with turkish locale.
If you add
```xml
<plugin>
    <artifactId>maven-surefire-plugin</artifactId>
    <configuration>
        <argLine>-Duser.language=tr</argLine>
    </configuration>
</plugin>
```
to your `javaparser-core-testing/pom.xml` build plugins, the tests start to fail due to this problem.

Basically you should at least review all `toLowerCase` and `toUpperCase` calls in your code as to whether they should be better called with an explicit locale instead of `Locale.getDefault()` which is what the parameterless version does.

tarann26:
## Problem

Code generation is locale-sensitive. `PrimitiveType` lowercases the primitive enum names with the default-locale `String.toLowerCase()`, so on a JVM whose default locale folds `I`/`i` irregularly (Turkish, `tr`), `INT` becomes `ınt` (dotless-ı) and JavaParser emits illegal Java:

```java
Locale.setDefault(Locale.forLanguageTag("tr"));
StaticJavaParser.parse("class A { void f(int x){} }").toString();
// -> "void f(ınt x)"   (illegal)
```

`PrimitiveType.byTypeName("int")` likewise fails to resolve under the same locale.

## Fix

Pin both `toLowerCase()` calls in `PrimitiveType` to `Locale.ROOT`, so keyword generation is locale-independent (as suggested in the issue). One import, two one-line edits — no API change.

## Test

Adds `PrimitiveTypeTest#issue3018`, which sets the default locale to Turkish, confirms a parsed `int` parameter still prints as `int` (not `ınt`) and that `byTypeName("int")` resolves, and restores the original default locale in a `finally` block so no other test is affected.

Part of #3018.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/ast/type/PrimitiveTypeTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/ast/type/PrimitiveTypeTest.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2026 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.ast.type;
+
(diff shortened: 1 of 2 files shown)


jlerbsc:
Thank you for this PR.

PR #5087, opened 2026-08-05
perf: Optimize ancestor deduplication in ResolvedReferenceTypeDeclaration (O(N^2) to O(N))

lucasrmendonca7:
I did some profiling and noticed that `depthFirstFunc` and `breadthFirstFunc` in `ResolvedReferenceTypeDeclaration` can get quite slow on large inheritance graphs.

Currently, the code uses `ancestors.contains(...)` inside the traversal loop to prevent duplicates. Since `ancestors` is an `ArrayList`, this results in an $O(N^2)$ time complexity.

To fix this, I added a local `HashSet` (`seen`) to handle the deduplication in $O(1)$ average time. We still use the `ArrayList` to collect the results, ensuring that the insertion order and the original return contracts are fully preserved.

I ran JMH benchmarks (5 warm-up iterations, 30 measurement iterations) on a graph with N = 10,000 elements, to evaluate the impact. Here are the results:

**Execution Time**

| Graph Density | Original (`ArrayList`) | Proposed (`List` + `Set`) | Speedup |
| :--- | :--- | :--- | :--- |
| **Low Duplication** | ~ 86.8 ms/op | ~ 0.42 ms/op | 204x |
| **Medium Duplication** | ~ 22.0 ms/op | ~ 0.24 ms/op | 91x |
| **High Duplication** | ~ 4.9 ms/op | ~ 0.22 ms/op | 21x |

**Memory Allocation (`gc.alloc.rate.norm`)**

| Graph Density | Original (`ArrayList`) | Proposed (`List` + `Set`) | Difference |
| :--- | :--- | :--- | :--- |
| **Low Duplication** | ~ 110.5 KB/op | ~ 436.7 KB/op | + 326.2 KB |
| **Medium Duplication** | ~ 32.9 KB/op | ~ 126.9 KB/op | + 94.0 KB |
| **High Duplication** | ~ 6.6 KB/op | ~ 30.3 KB/op | + 23.7 KB |

Full benchmark output (all densities × input sizes, both time and memory) is available here: https://gist.github.com/lucasrmendonca7/b460378dda3b0ef19ca62ae278229336

Trading ~326 KB of short-lived local memory for a 200x CPU speedup seems like a very reasonable trade-off here. The local `HashSet` is garbage collected right after the method returns.

All tests passed.

grafico_comparativo

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -154,9 +154,9 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
             seen.addAll(queuedAncestors);
             while (!queuedAncestors.isEmpty()) {
                 ResolvedReferenceType queuedAncestor = queuedAncestors.removeFirst();
-                queuedAncestor.getTypeDeclaration().ifPresent(rtd -> new LinkedHashSet<>(
-                        queuedAncestor.getDirectAncestors())
-                        .forEach(ancestor -> {
+                queuedAncestor
+                        .getTypeDeclaration()
+                        .ifPresent(rtd -> new LinkedHashSet<>(queuedAncestor.getDirectAncestors()).forEach(ancestor -> {
                             queuedAncestors.add(ancestor);
                             if (seen.add(ancestor)) {
                                 ancestors.add(ancestor);
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/declarations/ResolvedReferenceTypeDeclaration.java
@@ -124,17 +124,21 @@ public interface ResolvedReferenceTypeDeclaration extends ResolvedTypeDeclaratio
      * In the example above, this method returns B,C,E,D
      */
     Function<ResolvedReferenceTypeDeclaration, List<ResolvedReferenceType>> depthFirstFunc = (rrtd) -> {
-        Set<ResolvedReferenceType> ancestors = new LinkedHashSet<>();
+        List<ResolvedReferenceType> ancestors = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


lucasrmendonca7:
Hi @jlerbsc, hope this is okay to ask. Since you have a lot of context on this part of the codebase, I wanted to reach out. If you have a chance to take a look, I'd highly appreciate your review.

jlerbsc:
Thanks for looking into this — the intuition (contains() in a traversal loop is quadratic) is reasonable, but I can't reproduce your conclusion.

Could you push the benchmark code? It isn't part of the PR, so the numbers in the description can't be re-run.

And please measure on a hierarchy whose types have different bodies. A synthetic 10,000-node graph isn't what the symbol solver actually walks — real ancestor lists are tens of types, with different members. That matters: contains() goes through EqualsVisitor, which compares extendedTypes/implementedTypes first and short-circuits almost immediately on differing types, whereas seen.add() goes through HashCodeVisitor, a full recursive hash of the declaration's AST subtree, unmemoized and with no short-circuit. The change trades N cheap early-exiting comparisons for one unconditional deep traversal per element.

Two things about the PR itself:

Please rebase onto a topic branch with one or two commits whose messages describe this change — feat: update data structure doesn't.

Keep the existing comments (// We want to avoid infinite recursion..., // init direct ancestors, etc.). They're still accurate and unrelated to the optimization. The .stream().forEach(...) → .forEach(...) cleanup is worth keeping.

skrcode:
I ran JAIPilot Cloud against this exact PR head. It found that breadth-first traversal still queued an ancestor before checking seen, so shared ancestors were re-expanded and a cyclic mocked graph could refill the queue indefinitely.

A one-line reorder makes seen.add gate both result insertion and queue insertion. Diamond traversal order remains unchanged, a shared ancestor is expanded once instead of twice, and a mocked X to Y to X fixture moves from timeout to about 20 ms. The full reactor passed 2,094 tests with zero failures or errors, and Spotless passed.

PR directly onto this source branch: https://github.com/lucasrmendonca7/javaparser/pull/1  
Cloud-generated draft and full evidence: https://github.com/skrcode/javaparser/pull/3

Real JVM hierarchies reject inheritance cycles, so the cycle fixture proves the traversal function itself; the duplicate-expansion reduction applies directly to ordinary acyclic diamond graphs.

PR #5088, opened 2026-08-08
fix: prevent trailing line comment from being duplicated across sibling nodes

### LINKED ISSUE #4960: A bit of a weird comment parsing
I wrote a simple code to demonstrate it:
```java
package org.example;

public class Test {
    static void main() {
        CompilationUnit parsed = StaticJavaParser.parse(
                """
                class TestClass {
                    public void a(int a //something
                        ) { }
                }
                """
        );
        System.out.printf("Parsed: %s%n", parsed);

        MethodDeclaration method = parsed.getType(0).asClassOrInterfaceDeclaration().getMethods().getFirst();
        System.out.printf("Method type comment: %s%n", method.getType().getComment());
        System.out.printf("Method name comment: %s%n", method.getName().getComment());
    }
}
```

Expected:
```
Parsed: class TestClass {

    public void a(//something
    int a) {
    }
}

Method type comment: Optional.empty
Method name comment: Optional.empty
```

Got:
```
Parsed: class TestClass {

    public //something
    void //something
    a(//something
    int a) {
    }
}

Method type comment: Optional[//something
]
Method name comment: Optional[//something
]
```
Tried different ways to write comments and got that the bug is triggered only when the comment is after the method name, triggered for return type only when it's on the same line as the comment and triggered for the method name only when it's on the same line as the comment as well. The same error also occurs for comments like this:
```java
class TestClass {
    public void a(int a) //something
    { }
}
```

Not really sure, maybe it's intentional? It's just weird to see the same comment suddenly appear multiple times. Checked on 3.28.0 and on master

fudianchn:
## What

Fixes a bug where a trailing line comment (e.g. at the end of a parameter line) was duplicated across sibling nodes that merely shared the same line, instead of being attributed only to the node it trails.

Fixes #4960.

## Why

Given:

```java
class TestClass {
    public void a(int a //something
        ) { }
}
```

The comment `//something` was attached to **both** the return type (`void`) and the method name (`a`), because both nodes end on the same line as the comment. The parsed output incorrectly showed:

```java
    public //something
    void //something
    a(//something
    int a) {
    }
```

The comment should belong only to the parameter `int a` that it trails, not to unrelated preceding siblings.

## Root cause

`CommentsInserter.attributeLineCommentsOnSameLine` iterated over **every** child whose `end.line` matched the comment's `begin.line`, calling `attributeLineCommentToNodeOrChild` for each. The first match would set the comment on a node, but the loop continued and the recursive helper would set the *same* comment object on additional sibling nodes — duplicating it in the output.

## How

`attributeLineCommentsOnSameLine` now selects only the child whose end position is **nearest** to the comment (the one the comment actually trails) and delegates to it exactly once. This prevents the comment from being attributed to unrelated siblings that merely share the line.

## Testing

`CommentsInserterTest` has three regressions for this issue, and each asserts the actual invariant rather than only that some node lost the comment:

- Global uniqueness: a `commentAttachmentSites` helper walks the whole CU and records every node whose `getComment()` is present, **without deduplicating by `Comment` identity** (the bug is one comment instance attached to several nodes, so identity-dedup would hide exactly that).
- Positive attribution: `assertSame(parameter, attachmentSites.get(0))` pins the comment to the parameter it trails.
- Back-pointer consistency: `assertSame(node, node.getComment().get().getCommentedNode().get())`.
- Both reproducers from the issue (`int a //something` and `int a) //something`), plus a two-parameter case.
- Full-rendering output assertions (`assertEqualsStringIgnoringEol`, per the file's convention).

A separate `Issue4960Test` under `lexicalpreservation` verifies `LexicalPreservingPrinter` reproduces the original source verbatim for a trailing comment after the last parameter, after `)`, after a statement and after an enum constant.

Mutation check against master `2ec0219`: with the production code on master but these new tests, the three `CommentsInserterTest` regressions fail on attachment-site counts `3 vs 1`, `3 vs 1`, and `4 vs 2`. With this branch's production change, the same tests pass with counts `1`, `1`, and `2`.

Verified locally on JDK 17: `CommentsInserterTest` and `Issue4960Test` focused runs pass (13 tests); `spotless:check` and `checkstyle:check` pass; the metamodel and core generators produce no diff; the full `clean test --activate-profiles AlsoSlowTests` run passes with **5128 tests, 0 failures, 0 errors, 37 skipped**.

## Verification of the original issue

Before the fix, parsing the issue's reproducer:
```
Method type comment: Optional[//something]
Method name comment: Optional[//something]
```

After the fix:
```
Method type comment: Optional.empty
Method name comment: Optional.empty
```

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/CommentsInserterTest.java
@@ -24,12 +24,22 @@ package com.github.javaparser;
 import static com.github.javaparser.utils.TestUtils.assertEqualToTextResourceNoEol;
 import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.body.MethodDeclaration;
+import com.github.javaparser.ast.body.Parameter;
+import com.github.javaparser.ast.comments.Comment;
 import com.github.javaparser.ast.comments.CommentsCollection;
 import com.github.javaparser.utils.LineSeparator;
 import com.github.javaparser.utils.TestParser;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
 import org.junit.jupiter.api.Test;
 
 class CommentsInserterTest {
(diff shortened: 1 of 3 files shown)


jlerbsc:
The diagnosis is right and the deduplication works, but the PR ships with a CI-breaking formatting violation and a test that proves almost nothing.

Blocking

1. Spotless. mvn spotless:check fails on the two line-wraps introduced in CommentsInserter.  mvn spotless:apply.

2. Test coverage — the real problem. The PR adds two negative assertions:

    assertFalse(method.getType().getComment().isPresent());
    assertFalse(method.getName().getComment().isPresent());

That test passes if the comment vanishes entirely — the worst possible regression for a comment-attribution change is not covered. It also says nothing about where the comment went, nor whether it's still attached exactly once.

Required before merge:

- Uniqueness, globally. Walk the whole CU, collect every node with a comment, assert the count is 1. That is the actual bug, and nothing in the PR asserts it.

- Positive attribution. Assert the Parameter is the node that received it — assertSame, not just "the others don't have it".

- Back-pointer consistency. comment.getCommentedNode() must be the same node. On master this invariant is broken (three nodes hold the comment, getCommentedNode() returns only VoidType) — it should be locked so it can't silently break again.

- Both reproducers from the issue. The PR only covers void a(int a //something; the issue also reports void a(int a) //something before the brace. Untested.
- Output assertions, per the file's own convention. CommentsInserterTest compares full renderings with assertEqualsStringIgnoringEol (issue200…, issue234…). Two assertFalse calls are out of style and out of strength.
- Round-trip via LexicalPreservingPrinter. This is where the "comment moves" concern belongs. Rather than introducing any new concept, the PR must demonstrate with tests that LPP still reproduces the original source for a line comment after the last parameter, after ), after a statement, and after an enum constant. If LPP output changes, that's a regression the current test suite would not catch, and it must be surfaced as an explicit assertion — not left to chance.
- No comment loss in the sibling cases. The old code fell back to earlier siblings when attributeLineCommentToNodeOrChild returned false; the new code gives up. I couldn't construct a losing case, but the behavior changed and nothing tests it. At minimum: a case where the last node on the line already carries a comment.

Non-blocking, should be fixed

- The code comment claims "the child whose end is nearest to the comment", but the implementation selects the max by begin position via PositionUtils.areInOrder. Same result (siblings are disjoint), unstated reasoning.
- Missing guard childEnd.isBefore(commentBegin) — free, and makes the intent self-evident.
- areInOrder(a, b) hardcodes ignoringAnnotations = false, while the rest of insertComments passes configuration.isIgnoreAnnotationsWhenAttributingComments().

fudianchn:
Updated in `8c7953d`. Addressing every point.

**Blocking**

1. Spotless. Ran `mvn spotless:apply`; `spotless:check` now passes.

2. Test coverage. The two `assertFalse` calls are replaced with tests that assert the actual invariant.

- Uniqueness, globally. `attachedComments(cu)` walks the whole CU via `findAll(Node.class)` and collects every attached comment; both reproducer tests assert the count is 1.
- Positive attribution. `assertSame(attached.get(0), parameter.getComment().get())` pins the comment to the parameter it trails.
- Back-pointer consistency. `assertSame(parameter, comment.getCommentedNode().get())` (and the same check on the holder in the second reproducer).
- Both reproducers. `issue4960CommentAfterParameterIsAttributedOnceToThatParameter` covers `int a //something`, `issue4960CommentAfterClosingParenIsNotDuplicated` covers `int a) //something`.
- Output assertions. Both use `assertEqualsStringIgnoringEol` on `cu.toString()`, matching `issue200` / `issue234`.
- Round-trip via LexicalPreservingPrinter. New `Issue4960Test` (extends `AbstractLexicalPreservingTest`) asserts `print()` reproduces the original source verbatim for a trailing comment after the last parameter, after `)`, after a statement, and after an enum constant, plus a two-parameter case asserting the comment still appears exactly once.
- No comment loss across siblings. `issue4960EachTrailingCommentIsAttributedOnceToItsOwnParameter` parses two parameters each with its own trailing comment and asserts both survive as two distinct instances. Full `javaparser-core-testing` suite: 2840 tests, 0 failures.

**Non-blocking**

- "Nearest" wording. The comment now states the reasoning explicitly: candidates are the children whose end is on the comment's line; they are sorted by begin position and tried from the greatest begin (nearest to the comment) backwards, so the first one tried is the one the comment trails.
- Missing guard. Added `childEnd.isBefore(commentBegin)`.
- `areInOrder` / annotations. The sort now uses `configuration.isIgnoreAnnotationsWhenAttributingComments()`, consistent with the rest of `insertComments`.

**On point 7.** The implementation keeps the no-loss fallback: candidates on the comment's line are tried from nearest to earliest, and attribution stops at the first that accepts the comment. This removes the duplication (the old code iterated every sibling regardless) while preserving the old behavior where it matters.

One honest note: I could not construct a reachable input where the nearest candidate already carries a comment (two line comments cannot share a line, and a block comment on the same line does not go through this path), so the fallback is kept for behavioral parity rather than because a reachable case needs it. Happy to drop it if you would rather keep the method to a single nearest-attempt; I left it in since you listed no-loss as required.

Minor: the current loop sorts the whole candidate list and reverses it just to try the nearest first. Since siblings are disjoint, the nearest is simply the candidate with the greatest begin position, so a single pass tracking the max would do without the sort+reverse. I can fold that in if welcome.

jlerbsc:
Thanks for the rework — this addresses everything from the previous round, and the approach you landed on is better than what I suggested: collecting the candidates, sorting them descending by begin position and breaking on the first accept keeps the attribution unique and preserves the fallback to an earlier sibling that the old code had. The code comment now describes what the code actually does.

The global-uniqueness assertion cannot fail

attachedComments() dedups with an IdentityHashMap, so it counts distinct Comment instances, not attachment sites. The bug is one instance attached to several nodes — the helper collapses exactly what it should detect.

Two follow-ons:

issue4960EachTrailingCommentIsAttributedOnceToItsOwnParameter guards nothing today. Site counting makes it real (4 vs 2 expected).
In that test, attached.stream().map(System::identityHashCode).distinct().count() == 2 is tautological — the list is already identity-deduped. Drop it.

Swap the helper for site counting and check that all three CommentsInserterTest tests then fail on master. Once that holds, good to merge from my side.

fudianchn:
Updated in `f0e683d`.

`attachedComments` has been replaced with `commentAttachmentSites`. The helper now records every node whose `getComment()` is present and deliberately does not deduplicate by `Comment` identity. The tautological `identityHashCode().distinct()` assertion has been removed.

Mutation check against master `2ec0219`: the three `CommentsInserterTest` regressions fail on attachment-site counts `3 vs 1`, `3 vs 1`, and `4 vs 2`. With this branch's production change, the same tests pass with counts `1`, `1`, and `2`.

Verified locally on JDK 17: `CommentsInserterTest` and `Issue4960Test` focused runs pass (13 tests); `spotless:check` and `checkstyle:check` pass; the full `clean test --activate-profiles AlsoSlowTests` run passes with `5128 tests, 0 failures, 0 errors, 37 skipped`.

fudianchn:
The 3 failing checks share one cause outside this PR: a broken guava 33.7.0-jre artifact entered master with f3bbb7fc (#5094) and crashes the javac 9 / 10.0.1 compile of javaparser-symbol-solver-core.

I reported this to the guava maintainers and they have confirmed they will fix it with a patch release: google/guava#8614. Until it ships, #5096 reverts to 33.6.0-jre to keep the matrix green.

Update: guava shipped 33.7.1 (google/guava#8614), which removes the stray `Multi-Release` manifest header from the published jre and android artifacts. #5096 now updates to 33.7.1-jre instead of reverting.

jlerbsc:
Thank you for this PR.

PR #5096, opened 2026-08-18
fix(deps): update guava to 33.7.1-jre to fix JDK 9 compilation

### LINKED ISSUE #5095: guava 33.7.0-jre breaks JDK 9 / JDK 10.0.1 compilation of javaparser-symbol-solver-core (since #5094)
Since f3bbb7fc (#5094) the `maven_test (ubuntu-latest, 9)` and `maven_test (windows-latest, 9)` jobs and the AppVeyor build fail on every run, e.g. master run 32051029087:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.15.0:compile (default-compile) on project javaparser-symbol-solver-core: Compilation failure
[ERROR] bootstrap class path not set in conjunction with -source 1.8
```

That message is incidental. The actual failure is a compiler crash inside zipfs (from the AppVeyor log, JDK 10.0.1):

```
java.lang.NullPointerException
	at jdk.zipfs/jdk.nio.zipfs.JarFileSystem.getVersionMap(JarFileSystem.java:137)
	at jdk.nio.zipfs.JarFileSystem.createVersionedLinks(JarFileSystem.java:112)
	at jdk.nio.zipfs.JarFileSystem.<init>(JarFileSystem.java:85)
	at jdk.compiler/com.sun.tools.javac.file.JavacFileManager$ArchiveContainer.<init>(JavacFileManager.java:517)
```

Root cause chain:

1. guava 33.7.0-jre ships `Multi-Release: true` in its manifest but contains no `META-INF/versions/` entries, while 33.6.0-jre contains `META-INF/versions/9/module-info.class`.
2. javac opens every classpath jar through zipfs. On JDK 9 (all 9.0.x, including zulu 9.0.7 that the `9` matrix entry resolves to) and JDK 10.0.1, `JarFileSystem.createVersionedLinks` has no null guard for the missing versions directory and NPEs. JDK 10.0.2 added the `verdir != null` check (so the `10` matrix jobs pass), and JDK 8 has no multi-release support. That is why exactly the `9` jobs plus AppVeyor (preinstalled 10.0.1) fail.
3. guava is only on the compile classpath of javaparser-symbol-solver-core, so that is the only module failing.

Reproducer on zulu 9.0.7:

```java
import java.net.URI;
import java.nio.file.*;
import java.util.*;

public class Repro {
        Map<String, String> env = new HashMap<>();
        env.put("multi-release", "9");
        FileSystem fs = FileSystems.newFileSystem(URI.create("jar:" + Paths.get(args[0]).toUri()), env);
        System.out.println("OK: " + args[0]);
    }
}
```

- guava-33.6.0-jre.jar -> opens fine
- guava-33.7.0-jre.jar -> NPE at `JarFileSystem.getVersionMap(JarFileSystem.java:137)`, identical frames to the CI stack

Two-state build on zulu 9.0.7, `./mvnw -pl javaparser-symbol-solver-core -am clean compile`:
- pom at f3bbb7f (33.7.0-jre) -> compilation failure, only javaparser-symbol-solver-core fails
- same pom with 33.6.0-jre -> BUILD SUCCESS

Fix options: revert to 33.6.0-jre until guava publishes a corrected artifact (Maven Central currently lists 33.7.0-jre as latest), or upgrade once guava ships a fixed release. There is no build-side workaround: the crash happens inside javac while scanning the classpath.

Update: reported upstream as google/guava#8614; fixed in guava 33.7.1, which removes the stray `Multi-Release` manifest header from the published jre and android artifacts.

fudianchn:
Fixes #5095.

One-line update of guava to 33.7.1-jre, which replaces the broken 33.7.0-jre artifact that entered master with #5094.

Verified on zulu 9.0.7 (the JDK the failing `maven_test (ubuntu-latest, 9)` and `maven_test (windows-latest, 9)` matrix entries resolve to), `./mvnw -pl javaparser-symbol-solver-core -am clean compile`:
- 33.7.0-jre -> compilation failure (compiler NPE, root cause and reproducer in #5095)
- 33.7.1-jre -> BUILD SUCCESS

Regression check on JDK 17: full `javaparser-symbol-solver-testing` suite passes with the update (2090 tests, 0 failures).

guava 33.7.1 is the patch release confirmed in google/guava#8614; it removes the stray `Multi-Release` manifest header from the published jre and android jars, so future renovate bumps are safe again.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -188,7 +188,7 @@
             <dependency>
                 <groupId>com.google.guava</groupId>
                 <artifactId>guava</artifactId>
-                <version>33.7.0-jre</version>
+                <version>33.7.1-jre</version>
             </dependency>
             <dependency>
                 <groupId>org.checkerframework</groupId>


fudianchn:
Superseded by #5097, which is merged (master pom is already at 33.7.1-jre).

PR #5100, opened 2026-08-22
fix: match one-line Markdown comments in lexical preservation

### LINKED ISSUE #5099: Exception in setJavadocComment
calling setJavadocComment on the MethodDeclaration of a method with a markdown comment present results in an exception.

after removeJavaDocComment of the present javadoc comment, it results in an "IllegalArgumentException: I could not find child 'match node …"; 
when printing the class, the whole method is missing.

both do not happen if the method has no previous javadoc comment or a traditional one.
javaparser-core 3.28.2 .

```
java.lang.IllegalStateException: The matching comment to be replaced could not be found
    at com.github.javaparser.printer.lexicalpreservation.LexicalPreservingPrinter$Observer.concretePropertyChange (LexicalPreservingPrinter.java:245)
    at com.github.javaparser.ast.observer.PropagatingAstObserver.propertyChange (PropagatingAstObserver.java:64)
    at com.github.javaparser.ast.Node.lambda$notifyPropertyChange$7 (Node.java:675)
    at java.util.ArrayList.forEach (ArrayList.java:1604)
    at com.github.javaparser.ast.Node.notifyPropertyChange (Node.java:675)
    at com.github.javaparser.ast.Node.setComment (Node.java:308)
    at com.github.javaparser.ast.nodeTypes.NodeWithJavadoc.setJavadocComment (NodeWithJavadoc.java:78)
    at com.github.javaparser.ast.nodeTypes.NodeWithJavadoc.setJavadocComment (NodeWithJavadoc.java:67)
```

java source:
```
package test;
public class Error
{
  /// dummy.
  public void x()
  { }
}
```

kalayciburak:
Fixes #5099.

A one-line Markdown comment (`/// dummy.`) is a single token. Lexical preservation only completed a Markdown match when a later `SINGLE_LINE_COMMENT` token appeared, so that case stayed unmatched.

- `setJavadocComment` threw `IllegalStateException: The matching comment to be replaced could not be found`
- `removeJavaDocComment` dropped the method node and left the `///` line behind

The matcher now treats a start token that already equals the comment content as a complete match.

Tested locally (JDK 17):
- `Issue5099Test` 2/2
- `MethodDeclarationTransformationsTest` 32/32 (3 skipped)
- `CommentTest` 16/16
- `Issue3387Test` 1/1
- `com.github.javaparser.printer.lexicalpreservation.**` 684/684 (6 skipped)

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue5099Test.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/printer/lexicalpreservation/Issue5099Test.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2013-2026 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
+ * JavaParser is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ */
+
+package com.github.javaparser.printer.lexicalpreservation;
+
+import static com.github.javaparser.utils.TestUtils.assertEqualsStringIgnoringEol;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import com.github.javaparser.ast.body.MethodDeclaration;
+import org.junit.jupiter.api.Test;
+
+class Issue5099Test extends AbstractLexicalPreservingTest {
+
+    @Test
+    void setJavadocCommentReplacesSingleLineMarkdownComment() {
+        considerCode("package test;\n" + "public class Error\n"
+                + "{\n"
+                + "  /// dummy.\n"
+                + "  public void x()\n"
+                + "  { }\n"
+                + "}\n");
+
+        MethodDeclaration method = cu.findFirst(MethodDeclaration.class).get();
+        method.setJavadocComment(" replacement");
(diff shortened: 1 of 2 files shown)


jlerbsc:
Thanks for the fix.

**Root cause confirmed.**  a one-line `///` comment is a single `SINGLE_LINE_COMMENT` token and `getContent()` carries no trailing EOL (LF and CRLF alike), so the markdown matcher never completed the match. Worth noting the deletion path is worse than the reported exception: it falls back to `findChildTextElementForComment`, which resolves to the *commented node*, so `removeJavaDocComment()` silently deletes the member (reproduced on a method and on a field) instead of the comment.

**The fix is at the right layer.**. Beyond the two cases you test, it also repairs fields, type declarations, CRLF sources, empty `///`, trailing spaces, markdown→markdown replacement and duplicate-content comments.

**Two things before merge:**

1. CI will be red — `mvn spotless:check` fails on `Issue5099Test.java` (palantir formatting of the concatenated strings). `mvn spotless:apply` fixes it.
2. The `+ "\n"` / `+ "\r\n"` / `+ "\r"` clauses are unreachable: a parsed markdown comment's content never ends with an EOL, and a hand-built `MarkdownComment` doesn't reach that branch at all. Keeping only `oldContent.equals(startText)` gives identical results across the whole lexical-preservation package. Please drop them.

Nit: the new test's license header should match the other files:

```java
/*
 * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
 * Copyright (C) 2011, 2013-2026 The JavaParser Team.
 *
 * This file is part of JavaParser.
 *
 * JavaParser can be used either under the terms of
 * a) the GNU Lesser General Public License as published by
 *     the Free Software Foundation, either version 3 of the License, or
 *     (at your option) any later version.
 * b) the terms of the Apache License
 *
 * You should have received a copy of both licenses in LICENCE.LGPL and
 * LICENCE.APACHE. Please refer to those files for details.
 *
 * JavaParser is distributed in the hope that it will be useful,
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Lesser General Public License for more details.
 */
```

kalayciburak:
dropped the eol branches and ran spotless. header matches the other files now

jlerbsc:
Thanks for the quick turnaround — both points are addressed, and the license header now matches the other files.

Issue #5077, opened 2026-07-25
Test stability: tests mutate the global StaticJavaParser configuration


jlerbsc:
A large portion of the test suite configures the parser through StaticJavaParser by mutating its shared, process-global ParserConfiguration. This creates hidden coupling between tests, makes the suite order-dependent, and prevents parallel execution. This issue tracks reducing that technical debt.

Background

StaticJavaParser holds its configuration in a single ThreadLocal<ParserConfiguration>:

    private static final ThreadLocal<ParserConfiguration> localConfiguration =
        ThreadLocal.withInitial(ParserConfiguration::new);

There are ~115 occurrences of configuration mutation (setConfiguration(...), getParserConfiguration().setXxx(...)) across ~85 test files. Because the configuration is global mutable state, any test that changes it can affect subsequent tests.

Problem

Three inconsistent patterns coexist in the test code, two of which are unsafe:
- Reset on every test (correct, but duplicated everywhere) — e.g. UnaryExprTest resets with @BeforeEach → new ParserConfiguration().
- Fragile save/restore — e.g. PatternExprTest captures the "original" language level in a field initializer. If an earlier test already polluted the global config, the restored value is already corrupted.
- No reset at all (state leak) — e.g. ImportDeclarationTest, PrettyPrintVisitorTest, and others call getParserConfiguration().setLanguageLevel(...) without ever cleaning up, leaking state (language level, lexical preservation, symbol resolver, etc.) into every test that runs afterward on the same thread.

Consequences:

- Order dependency — a test that leaves BLEEDING_EDGE or lexicalPreservationEnabled=true behind can silently change the outcome of another test depending on execution order.
- Parallelization blocked — the ThreadLocal isolates per thread, but JUnit distributes methods of a class across a worker pool; a @BeforeAll set on one thread is not visible to methods running on another. The suite is effectively stuck running sequentially.
- Silent flakiness — nothing signals the leak as long as the execution order stays stable, so failures surface unpredictably (e.g. after adding/reordering tests or enabling parallelism).

sylviauwa748:
Hi @jlerbsc, I'd like to work on this issue if that's okay with you let me know if you're already deep into it or would rather pair on it together.

I have two ways I would like to go about it.

The quick fix to add a small piece of code that resets the config clean before and after every test, so tests stop leaking state into each other. 
[...]

PR #5102, opened 2026-08-24
Refs #5077: reset test parser config per test; SourceFileInfoExtractor no longer mutates the global configuration

sylviauwa748:
- AbstractSymbolResolutionTest now resets StaticJavaParser to a plain ParserConfiguration in both @BeforeEach and @AfterEach, instead of only resetting after each test with a resolver already attached.
- This exposed a hidden dependency in AnonymousClassesResolutionTest, which relied on a resolver set once in @BeforeAll on the shared config. Fixed by building an isolated TypeSolver/JavaParser inside the test itself via parseSample(sampleName, typeSolver).

Note: AnonymousClassesResolutionTest failed because it configures its resolver in @BeforeAll but parses inside the test method itself. By the time it parses, the per-test reset has already wiped the config. Classes that parse during @BeforeAll (e.g. MethodDescriptorTest) are unaffected, since the resolver is already captured in the AST before the reset ever runs.

Scope: this reset only covers test classes that extend AbstractSymbolResolutionTest. Standalone classes that configure StaticJavaParser directly in @BeforeAll, e.g. JavaParserTypePatternDeclarationTest and LeastUpperBoundTest, don't go through this base class, so they aren't covered and still leak state. Migrating those is part of the remaining ~231 call sites across ~103 files tracked in #5077.

Refs #5077.


jlerbsc:
Thanks @sylviauwa748 — this is the right shape for step 1, and the two things that mattered are both there: a blank configuration (no default resolver) and a reset on both sides. Three things before I merge.

mvn spotless:apply — AnonymousClassesResolutionTest.java lost its trailing newline, and the formatting workflow (spotless:apply + git diff --exit-code) will fail on it.
Please retitle to Refs #5077 rather than Fix #5077: 231 call sites across 103 files still mutate the shared configuration, so the issue must stay open.
Worth stating in the description that the reset only covers classes extending AbstractSymbolResolutionTest. Standalone classes such as JavaParserTypePatternDeclarationTest and LeastUpperBoundTest still configure the global parser in @BeforeAll and still leak.

One thing worth documenting in the base class while you're there: a per-test reset is incompatible with configuring in @BeforeAll. It happens to be harmless for MethodDescriptorTest, because it parses during @BeforeAll and the resolver is captured in the AST — AnonymousClassesResolutionTest failed precisely because it parsed inside the test. Two lines of comment will save the next contributor an hour.

Natural follow-up, if you want it: move the reset into an auto-detected JUnit extension so it covers the classes outside the hierarchy.

jlerbsc:
Thanks for pushing this — the CI failures are not a mistake on your side, they are a genuine consequence of your change, and a rather interesting one. Here is the full diagnosis so you can implement the fix yourself.

Why it fails. SourceFileInfoExtractor — production code in javaparser-symbol-solver-core — configures the global parser in its constructor:

public SourceFileInfoExtractor(TypeSolver typeSolver) {
    this.typeSolver = typeSolver;
    StaticJavaParser.getParserConfiguration().setSymbolResolver(new JavaSymbolSolver(typeSolver));
}

and then parses through the static import StaticJavaParser.parse(file). In AnalyseNewJavaParserTest the extractor is a static field, so that resolver is installed once, when the class is loaded. Your @BeforeEach reset then wipes it before every test. Resolution paths that go through Node.getSymbolResolver() — lambdas and stream chains, which is exactly what the two failing lines contain — fall back to ERROR.

So your reset is doing its job: it exposed production code that both pollutes and depends on the global configuration. Same defect as #5077, on the main side rather than the test side.

Why you did not see it locally. These classes are annotated @SlowTest and excluded by the default NonSlowTests profile. CI runs the other one:

mvn -B -pl :javaparser-symbol-solver-core,:javaparser-symbol-solver-testing test -P AlsoSlowTests


The fix. In SourceFileInfoExtractor, give the class its own parser instead of mutating the global one:

add a private final JavaParserAdapter parser field, built in the constructor from new JavaParser(new ParserConfiguration().setSymbolResolver(new JavaSymbolSolver(typeSolver)));
replace both CompilationUnit cu = parse(file); call sites with parser.parse(file);
drop import static com.github.javaparser.StaticJavaParser.parse; and the StaticJavaParser import;
expose public ParserConfiguration getParserConfiguration() returning parser.getParserConfiguration() — the next point needs it.

That surfaces one more failure, in AnalyseJavaSymbolSolver060Test: it steers the extractor by writing into the global config (StaticJavaParser.getParserConfiguration().setAttributeComments(false)). It must become sourceFileInfoExtractor.getParserConfiguration().setAttributeComments(false). Remember to remove the now-unused StaticJavaParser import there — palantir/spotless strips unused imports, so leaving it will fail the formatting check.

One trap worth knowing. Making the extractor field non-static instead would not work: JUnit runs instance field initializers before @BeforeEach, so your reset would still wipe the resolver. The fix has to be on the production side.

Also still pending: AnonymousClassesResolutionTest.java lost its trailing newline. mvn spotless:apply fixes it, and the formatting workflow (spotless:apply + git diff --exit-code) will stay red until it is.

If you would rather keep this PR focused, splitting it into a separate commit inside the same PR is perfectly fine.

sylviauwa748:
Hi @jlerbsc Thanks for the detailed feedback.

I have pushed a fix as a separate commit.

SourceFileInfoExtractor now owns its own JavaParserAdapter/config instead of mutating StaticJavaParser's global config, and exposes getParserConfiguration().

AnalyseJavaSymbolSolver060Test now configures the extractor directly (sourceFileInfoExtractor.getParserConfiguration().setAttributeComments(false)) instead of reaching into the global config. 
Removed the now-unused StaticJavaParser import.

Didn't need to touch AnalyseNewJavaParserTest directly since the fix is on the production side, it should be covered automatically.

Ran the full AlsoSlowTests profile locally. A few unrelated failures showed up, but they're because I'm on JDK 21. None of the failing tests are the ones from your diagnosis, and CI runs JDK 8–18 so it shouldn't see this at all. Happy to look into it separately if useful.

Also, the earlier spotless fix (AnonymousClassesResolutionTest's trailing newline) is already in from a previous push, in case that note in your last comment was against an older CI run.

jlerbsc:
Thanks @sylviauwa748, this turned into a genuinely useful contribution.

What lands here:

AbstractSymbolResolutionTest now resets StaticJavaParser to a blank ParserConfiguration before and after each test, instead of leaving a default symbol resolver behind. Tests that need a resolver must now configure their own.
AnonymousClassesResolutionTest builds its own TypeSolver inside the test rather than relying on shared state.
SourceFileInfoExtractor no longer configures the process-global parser in its constructor: it owns its parser and exposes its configuration. This one is a real behaviour change in javaparser-symbol-solver-core — the constructor side effect was undocumented and is the same defect as #5077 seen from the production side, so removing it is deliberate. The squash title reflects it so it shows up in the changelog.

Worth restating for anyone following: this is step one. The reset only covers classes extending AbstractSymbolResolutionTest  so #5077 stays open.

The natural follow-up, if you want to keep going: move the reset into an auto-detected JUnit extension so it also covers the classes outside that hierarchy. I measured it — that surfaces 19 errors in 3 classes, one of which this PR already fixed, leaving JavaParserTypePatternDeclarationTest (leaked language level) and LeastUpperBoundTest (resolver set in @BeforeAll).

sylviauwa748:
Thanks @jlerbsc, really appreciate the feedback. 
I’d be happy to take on the JUnit extension follow-up you mentioned.
Should I open a new issue to track that specifically, or would you rather keep it under #5077?

PR #5107, opened 2026-08-27
   Refs #5077: auto-detect StaticJavaParser reset via JUnit extension, covering classes outside AbstractSymbolResolutionTest

sylviauwa748:
Follow up to #5077. Per the earlier discussion, this moves the per-test StaticJavaParser reset out of AbstractSymbolResolutionTest and into a JUnit 5 extension that auto-applies to every test in this module,
including classes that don't extend AbstractSymbolResolutionTest.

What changed:
- Added StaticJavaParserConfigResetExtension (com.github.javaparser.symbolsolver.utils), implementing BeforeEachCallback/AfterEachCallback with the same blank-config reset AbstractSymbolResolutionTest used to do itself.
- Registered it for auto-detection via src/test/resources/META-INF/services/org.junit.jupiter.api.extension.Extension,
  and enabled JUnit's auto-detection for this module only via src/test/resources/junit-platform.properties (junit.jupiter.extensions.autodetection.enabled=true). This is scoped to javaparser-symbol-solver-testing and does not affect other modules.
- Removed the now redundant resetBefore()/resetAfter() from AbstractSymbolResolutionTest, since the extension now covers every test in the module, including subclasses of that base class.

As predicted, enabling the reset module wide surfaced the same class of issue AnonymousClassesResolutionTest hit in the original PR: classes that configure StaticJavaParser once in @BeforeAll but parse inside individual test methods. Two classes broke:
- JavaParserTypePatternDeclarationTest — set the language level once
  in @BeforeAll, parsed per-test in createValue().
- LeastUpperBoundTest — set a resolver once in @BeforeAll, parsed
  per-test via a private treeOf() helper.

Both are fixed the same way as AnonymousClassesResolutionTest and SourceFileInfoExtractor in the original PR: each now owns a private JavaParserAdapter/config instead of depending on the shared StaticJavaParser, so they're unaffected by the reset regardless of when JUnit decides to run them.

Verification:
- mvn spotless:check — clean.
- Full module test run (mvn test -pl javaparser-symbol-solver-testing) 2092 tests, only pre-existing JDK 21 failures remain (see notebelow), none touching the classes changed here.

Note: the full run surfaces 12 pre-existing failures unrelated to this change  8 are ancestor count assertions affected by JDK 21 standard library changes (e.g. java.util.SequencedCollection), and 4 are IllegalStateException: Unable to determine the current version of java running, from TestJdk.getCurrentHostJdk() only recognizing JDK 8–18. Both are pre existing environment mismatches unrelated to StaticJavaParser config handling, and don't affect CI, which only
tests JDK 8–18.

Refs #5077.


sylviauwa748:
Hi @jlerbsc , this is the JUnit extension follow up you suggested when merging #5077.

jlerbsc:
I checked that the extension is genuinely picked up rather than silently ignored: a throwaway test class that configures StaticJavaParser in @BeforeAll and parses inside the test method now fails with a ParseProblemException, exactly as intended. 

And your JDK 21 note is correct. That is SequencedCollection arriving in JDK 21 and adding an ancestor to the collection types; unrelated to this work. I'll open a separate issue for it.

Two things for a follow-up, not for this PR. autodetection.enabled=true turns on every ServiceLoader-registered extension on the test classpath. The properties file is module-scoped, so javaparser-core-serialization (1 remaining mutation site) and javaparser-core-testing are still uncovered; copying the two resource files there would lock in the gain everywhere.

Worth recording where this leaves #5077: 226 call sites across 99 files still mutate the shared configuration, but they no longer create order coupling, since every test now starts and ends clean. What remains is cleanup rather than a correctness risk — and the third step, turning this extension into a guard that fails on a dirty configuration, only becomes possible once those sites are migrated.

sylviauwa748:
Hi @jlerbsc, thanks for the detailed breakdown, both points make sense. I'll work through the resource file copy to javaparser-core-serialization and javaparser-core-testing, and keep #5077's remaining scope in mind as I go. Will follow up once I've made progress.

Issue #4991, opened 2026-03-24
Regression Resolving Method Declaration From Overloaded Function Call


LaFoster00:
When resolving this function call in 3.28.0
```java
package test;

public class Test{
  public static void main(String[] args){
    int[] arr = new int[10];
    arr = Arrays.copyOf(arr, 3); // This line fails
  }
}

class Arrays {
  public static native <T> T[] copyOf(T[] original, int newLength);

  public static native byte[] copyOf(byte[] original, int newLength);

  public static native short[] copyOf(short[] original, int newLength);

  public static native char[] copyOf(char[] original, int newLength);

  public static native int[] copyOf(int[] original, int newLength);

  public static native long[] copyOf(long[] original, int newLength);

  public static native float[] copyOf(float[] original, int newLength);

  public static native double[] copyOf(double[] original, int newLength);
}
```
 it fails with, while it does not in 2.27.1. 
```
Exception in thread "main" com.github.javaparser.resolution.MethodAmbiguityException: Ambiguous method call: cannot find a most applicable method: JavaParserMethodDeclaration{wrappedNode=public static native double[] copyOf(double[] original, int newLength);, typeSolver=com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver@11c20519}, JavaParserMethodDeclaration{wrappedNode=public static native float[] copyOf(float[] original, int newLength);, typeSolver=com.github.javaparser.symbolsolver.resolution.typesolvers.CombinedTypeSolver@11c20519}
	at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:1093)
	at com.github.javaparser.resolution.logic.MethodResolutionLogic.findMostApplicable(MethodResolutionLogic.java:992)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.JavaParserTypeDeclarationAdapter.solveMethod(JavaParserTypeDeclarationAdapter.java:316)
	at com.github.javaparser.symbolsolver.javaparsermodel.contexts.ClassOrInterfaceDeclarationContext.solveMethod(ClassOrInterfaceDeclarationContext.java:108)
	at com.github.javaparser.symbolsolver.javaparsermodel.declarations.JavaParserClassDeclaration.solveMethod(JavaParserClassDeclaration.java:334)
	at com.github.javaparser.resolution.logic.MethodResolutionLogic.solveMethodInType(MethodResolutionLogic.java:1329)
[...]

PR #5108, opened 2026-08-29
fix: overloads on primitive array parameters are no longer ambiguous (fixes #5065)

### LINKED ISSUE #5065: Method Resolution fails for Parameters of Primitive Array Types

the2015:
Fixes #5065

## Problem

Resolving a call to an overloaded method whose overloads differ only in the primitive component type of an array parameter fails with `MethodAmbiguityException`:

```java
int[] a = new int[1];
Arrays.copyOf(a, 2); // MethodAmbiguityException: copyOf(int[],int) / copyOf(float[],int) / copyOf(double[],int)
```

## Root cause

`MethodResolutionLogic.isBoxingCompatibleWithTypeSolver` recursed into the component types when both the expected and the actual type were arrays, and then accepted the pair through `isAssignableBy`. Because `int` is assignable to `float` and `double` by widening primitive conversion, `float[]` and `double[]` were reported as applicable for an `int[]` argument, so several overloads survived applicability filtering and the resolution ended up ambiguous.

Per JLS 5.2 there is neither a boxing nor a widening conversion between two distinct primitive array types, so this component-wise recursion is wrong for a real array parameter.

## Fix

`isBoxingCompatibleWithTypeSolver` now takes a flag that distinguishes a genuine array parameter from an array type that was synthesized for a variadic parameter:

- For a real array parameter, if either component type is primitive, the component types must be equal. This mirrors `ResolvedArrayType#isAssignableBy`.
- For a variadic parameter the previous component-wise behaviour is kept, because there the conversion applies to each individual argument rather than to the array as a whole (`groupTrailingArgumentsIntoArray` turns the trailing arguments into an array type, and boxing/widening per argument is legal there).

The flag is threaded through the wildcard, component and constraint recursions, and the call site passes `methodDeclaration.getParam(i).isVariadic()`.

## Tests

- New regression test `Issue5065Test#overloadsDifferingOnlyInPrimitiveArrayParameterAreNotAmbiguous` asserts that `Arrays.copyOf(intArray, n)` resolves to `Arrays.copyOf(int[], int)` and `Arrays.copyOf(floatArray, n)` to `Arrays.copyOf(float[], int)`. It fails with `MethodAmbiguityException` without the core change.
- `javaparser-symbol-solver-testing`: 2093 tests, 0 failures, 0 errors, 34 skipped.
- `javaparser-core-testing`: 2842 tests, 0 failures, 0 errors, 8 skipped.
- `javaparser-core-testing-bdd`: 625 tests, 0 failures, 0 errors.
- `spotless:check` passes for the touched sources.

Note: the varargs tests in `MethodsResolutionLogicTest` are exactly what motivated the flag; an unconditional strict-equality check in the array branch regressed five of them.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -224,8 +224,15 @@ public class MethodResolutionLogic {
                     expectedDeclaredType = replaceTypeParam(expectedDeclaredType, tp, typeSolver);
                 }
                 if (!expectedDeclaredType.isAssignableBy(actualArgumentType)) {
-                    // Check boxing/unboxing compatibility using TypeSolver
-                    if (isBoxingCompatibleWithTypeSolver(expectedDeclaredType, actualArgumentType, typeSolver)) {
+                    // Check boxing/unboxing compatibility using TypeSolver.
+                    // Boxing or widening of the component types of two array types is only
+                    // meaningful for a variadic parameter, where the conversion applies to each
+                    // individual argument instead of to the array as a whole.
+                    if (isBoxingCompatibleWithTypeSolver(
+                            expectedDeclaredType,
+                            actualArgumentType,
+                            typeSolver,
+                            methodDeclaration.getParam(i).isVariadic())) {
                         // This parameter is compatible via boxing/unboxing
                         continue;
                     }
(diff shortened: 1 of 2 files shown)


jlerbsc:
Thanks @the2015 for this work. Your analysis of the root cause in isBoxingCompatibleWithTypeSolver was spot on. The fix for #4991 takes the narrower approach discussed in the review: array compatibility is left to ResolvedArrayType, and component-wise conversion is only applied to the last variadic parameter. Since it also resolves #5065, I'm closing this PR.

Issue #5109, opened 2026-08-29
Test suite fails on JDK 21+ (symbol-solver-testing): SequencedCollection ancestors and TestJdk capped at JDK 18


jlerbsc:
javaparser-symbol-solver-testing cannot be built on any JDK newer than 18. On JDK 21 the module reports failures/errors in tests, from two independent and pre-existing causes. The other modules are unaffected.

Reproduction

JAVA_HOME=/path/to/jdk-21 mvn -B test

Cause 1 SequencedCollection (JEP 431, JDK 21) 

java.util.SequencedCollection was inserted into the List/Collection hierarchy in JDK 21, so every ancestor count and interface set computed by reflection gains one entry. 

Eight assertions hardcode the pre-21 values:

ReferenceTypeTest.testGetAllAncestorsOnRawType
ReferenceTypeTest.testGetAllAncestorsOnTypeWithSpecifiedTypeParametersForClassAbstractList
ReferenceTypeTest.testGetAllAncestorsOnTypeWithSpecifiedTypeParametersForClassArrayList
ReferenceTypeTest.testGetAllAncestorsOnTypeWithSpecifiedTypeParametersForInterface
ReflectionClassDeclarationTest.testAllAncestorsWithDepthFirstTraversalOrder
ReflectionClassDeclarationTest.testGetAllInterfaces
ReflectionInterfaceDeclarationTest.testAllAncestors
ReflectionInterfaceDeclarationTest.testAllAncestorsForAnInterfaceWithBreadthFirstFunc

These are assertions about the host JDK's class library, not about JavaParser's behaviour. The durable fix is to make them JDK-agnostic — assert that the expected ancestors are present rather than asserting an exact count or an exact set — so a future JDK adding another interface does not break them again.

Cause 2 —TestJdk stops at JDK 18

AbstractSymbolResolutionTest.TestJdk.getCurrentHostJdk() matches java.version against an enum listing JDK 8 through 18 and otherwise throws: java.lang.IllegalStateException: Unable to determine the current version of java running

Four tests call it and error out on any newer JDK:
JavaParserClassDeclarationTest.testGetAllMethods
JavaParserEnumDeclarationTest.testGetAllMethods
JavaParserInterfaceDeclarationTest.testGetAllMethods
ReferenceTypeTest.testDirectAncestorsOfClassWithInterfaces

[...]

PR #5110, opened 2026-08-29
fix: keep exact symbol-solver assertions under JDK 21 guards

kalayciburak:
Refs #5109.

`javaparser-symbol-solver-testing` fails on JDK 21+ for two independent reasons:

- Reflection against `List`/`ArrayList` now includes `java.util.SequencedCollection`, so exact ancestor counts and exact interface sets no longer match.
- JDK 21 split `Object.wait(long)` and `java.lang.Object.wait0(long)` shows up in method lists.
- `TestJdk` only enumerates through 18, so `getCurrentHostJdk()` throws on newer hosts.

This keeps the JDK 8-era exact counts / `containsInAnyOrder` / set equality, branching only where the library changed (same pattern as Constable on JDK 12). Host major version is parsed from `java.specification.version` (`1.8` -> 8, `21` -> 21). `Runtime.version()` is not used because the tests still compile at Java 8.

I did not add JDK 21 to the CI matrix: `javaparser-core` currently fails to compile on JDK 21+ (`NodeList.getLast()` vs `java.util.List.getLast()`). That is tracked in #5111.

Tested locally:

- JDK 17: `ReferenceTypeTest`, `ReflectionClassDeclarationTest`, `ReflectionInterfaceDeclarationTest`, `JavaParserClassDeclarationTest`, `JavaParserEnumDeclarationTest`, `JavaParserInterfaceDeclarationTest` — 247/247 (6 skipped)
- JDK 21 (classes compiled with JDK 17): same 6 classes, 247/247 (6 skipped)
- `spotless:apply`

### DIFF
--- a/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/AbstractSymbolResolutionTest.java
+++ b/javaparser-symbol-solver-testing/src/test/java/com/github/javaparser/symbolsolver/AbstractSymbolResolutionTest.java
@@ -46,85 +46,26 @@ public abstract class AbstractSymbolResolutionTest {
     }
 
     /**
-     * An initial attempt at allowing JDK-specific test cases. It is a work-in-progress, and subject to change.
-     * @deprecated <strong>Note that use of TestJdk should be a last-resort, preferably implementing JDK-agnostic tests.</strong>
+     * Host JDK major version parsed from {@code java.specification.version}
+     * ({@code 1.8} -&gt; 8, {@code 21} -&gt; 21). Prefer JDK-agnostic assertions; use this
+     * only when a test must branch on a known library change.
      */
-    @Deprecated
-    protected enum TestJdk {
-        JDK8(8),
-        JDK9(9),
-        JDK10(10),
-        JDK11(11),
-        JDK12(12),
-        JDK13(13),
-        JDK14(14),
-        JDK15(15),
-        JDK16(16),
-        JDK17(17),
-        JDK18(18);
-
-        private final Integer major;
-
-        /**
-         * @deprecated <strong>Note that use of TestJdk should be a last-resort, preferably implementing JDK-agnostic tests.</strong>
-         */
-        @Deprecated
-        TestJdk(Integer major) {
-            this.major = major;
+    protected static int currentHostJdkMajor() {
+        String spec = System.getProperty("java.specification.version");
+        if (spec == null || spec.isEmpty()) {
+            throw new IllegalStateException("java.specification.version is not set");
         }
-
-        /**
-         * @deprecated <strong>Note that use of TestJdk should be a last-resort, preferably implementing JDK-agnostic tests.</strong>
-         */
-        @Deprecated
-        public int getMajorVersion() {
-            return this.major;
+        if (spec.startsWith("1.")) {
+            spec = spec.substring(2);
         }
-
-        /**
-         * @deprecated <strong>Note that use of TestJdk should be a last-resort, preferably implementing JDK-agnostic tests.</strong>
-         */
-        @Deprecated
-        public static TestJdk getCurrentHostJdk() {
-            String javaVersion = System.getProperty("java.version");
-
-            // JavaParser explicitly requires a minimum of JDK8 to build.
-            if ("8".equals(javaVersion) || javaVersion.startsWith("1.8") || javaVersion.startsWith("8")) {
-                return JDK8;
-            } else if ("9".equals(javaVersion) || javaVersion.startsWith("9.")) {
-                return JDK9;
-            } else if ("10".equals(javaVersion) || javaVersion.startsWith("10.")) {
-                return JDK10;
-            } else if ("11".equals(javaVersion) || javaVersion.startsWith("11.")) {
-                return JDK11;
-            } else if ("12".equals(javaVersion) || javaVersion.startsWith("12.")) {
-                return JDK12;
-            } else if ("13".equals(javaVersion) || javaVersion.startsWith("13.")) {
-                return JDK13;
-            } else if ("14".equals(javaVersion) || javaVersion.startsWith("14.")) {
-                return JDK14;
-            } else if ("15".equals(javaVersion) || javaVersion.startsWith("15.")) {
-                return JDK15;
-            } else if ("16".equals(javaVersion) || javaVersion.startsWith("16.")) {
-                return JDK16;
-            } else if ("17".equals(javaVersion) || javaVersion.startsWith("17.")) {
-                return JDK17;
-            } else if ("18".equals(javaVersion) || javaVersion.startsWith("18.")) {
-                return JDK18;
-            }
-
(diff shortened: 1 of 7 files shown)


jlerbsc:
Thanks for tackling this. The `TestJdk` → `currentHostJdkMajor()` part is the right
call and I'd merge it as is. My concern is the second half: the exact-count and
exact-set assertions are dropped, and that costs coverage on JDK 8–18, which is what
CI actually runs, in exchange for green tests on a JDK the matrix never exercises.

I reproduced the failures locally: 11 failures, from exactly two causes.

- 8 failures: one extra ancestor, `java.util.SequencedCollection` (9→10, 5→6, 2→3,
  plus the `getAllInterfaces` set). Every delta is exactly +1.
- 3 failures: one new method, `java.lang.Object.wait0(long)` (JDK 21, Loom) —
  `expected: <106> but was: <107>` / `no match for: "java.lang.Object.wait0(long)"`.
  Note this is *not* `SequencedCollection`, as the PR description states.

Both are precise enough to encode, which keeps the assertions strong:

    // JDK 21 inserts java.util.SequencedCollection into the List/Collection hierarchy
    assertEquals(currentHostJdkMajor() >= 21 ? 10 : 9, ancestors.size());

    // JDK 21 split Object.wait(long) into wait0
    if (currentHostJdkMajor() >= 21) {
        expected.add("java.lang.Object.wait0(long)");
    }
    assertThat(signatures, containsInAnyOrder(expected.toArray()));

That is the same pattern this PR keeps in `ReferenceTypeTest.testDirectAncestorsOf
ClassWithInterfaces` for `Constable`/`ConstantDesc` (JDK 12), so the codebase would
stay consistent.

Specific requests:

1. Restore the exact counts / `containsInAnyOrder` / set equality under a JDK guard,
   as above. As written, `ReflectionInterfaceDeclarationTest.testAllAncestors` still
   carries the comment "Since List is an interface, Object cannot be an ancestor of
   List" while the only assertion that enforced it is gone.
2. `testAllAncestorsForAnInterfaceWithBreadthFirstFunc` no longer asserts any order,
   so it is now a duplicate of `testAllAncestors` and the breadth-first function is
   untested. A JDK-agnostic ordering check keeps the intent:

       assertTrue(names.indexOf("java.util.Collection") < names.indexOf("java.lang.Iterable"));

3. `Fixes #5109` → `Refs #5109`. The issue's reproduction still fails: with
   `JAVA_HOME=<jdk-21> mvn -B test`, `javaparser-core` fails to compile with four
   errors (`addFirst`, `addLast`, `getFirst`, `getLast` vs `java.util.List`), so the
   issue should stay open.

On that last point — the compile failure is not inherent to the source. It comes from
`<source>/<target>1.8` instead of `<release>8`; with `--release`, javac compiles
against the JDK 8 signatures and the clash disappears. Verified end to end on JDK 21
with `-Dmaven.compiler.release=8` alone: `javaparser-core` compiles (588 files) and
`javaparser-symbol-solver-testing` runs 2092 tests, 0 failures, BUILD SUCCESS. Filed
separately as #XXXX — that is what would keep this fix from silently regressing,
since the current matrix stops at 18.

Nits: `currentHostJdkMajor()` throws a bare `NumberFormatException` where the old code
threw an explicit `IllegalStateException`; and `assertTrue(...containsAll(...))` gives
no diagnostic on failure, `assertThat(..., hasItems(...))` names the missing element.

kalayciburak:
yes, restored the exact counts / containsInAnyOrder under a jdk 21 guard (SequencedCollection + wait0), plus the bfs Collection-before-Iterable check. also switched to Refs #5109 and wrapped the NumberFormatException.

the compile-on-21 part is #5111 like you said.

jlerbsc:
Thanks for the follow-up  — the second iteration addresses the points from my first pass: exact counts and equality checks are back, and the guards are narrow.

Two things I'd like resolved before this goes in.

1. The JDK version number is the wrong discriminator

All of these tests resolve through ReflectionTypeSolver (jreOnly), i.e. against the running JDK. The question that matters is not "is this JDK >= 21" but "does this JDK expose SequencedCollection". A capability probe is more accurate, self-documenting, and does not need revisiting for backports or unusual runtimes:

protected static boolean hasSequencedCollection() {
    try {
        Class.forName("java.util.SequencedCollection");
        return true;
    } catch (ClassNotFoundException e) {
        return false;
    }
}
Same idea for wait0, via Object.class.getDeclaredMethod("wait0", long.class). currentHostJdkMajor() is still worth keeping for the pre-existing guards (registerNatives >= 14, Constable < 12), but it shouldn't carry these two new cases.

2. The ordering assertion in testAllAncestorsForAnInterfaceWithBreadthFirstFunc is too weak

assertTrue(names.indexOf("java.util.Collection") < names.indexOf("java.lang.Iterable"));
If Collection ever disappeared from the result, indexOf returns -1 and the assertion still passes (-1 < 0). More importantly, on JDK 21 the expected breadth-first order is [SequencedCollection, Collection, Iterable], and nothing checks that SequencedCollection comes first — which is precisely the property this test is named after. Moving from ancestors.get(0) / get(1) to byName.get(...) dropped the positional check I asked to preserve.

Please build the expected ordered list under the guard and assert on it directly, e.g. assertEquals(expectedNames, names).


Nits

The wait0 block is duplicated verbatim in 3 classes and the SequencedCollection ternary in 6 places. Two helpers in AbstractSymbolResolutionTest would centralise the next JDK adjustment.

new HashSet<>(ImmutableSet.of(...)) in ReflectionClassDeclarationTest: the intermediate ImmutableSet no longer serves a purpose.

currentHostJdkMajor() re-reads System.getProperty on every call; a static final int would do.

JUnit 5 ships org.junit.jupiter.api.condition.JRE, which would avoid hand-rolled version parsing — worth checking against the junit-bom version we pin.

kalayciburak:
done, probes for SequencedCollection/wait0 instead of jdk >= 21, and the bfs list is asserted in order. also cached the major version and dropped the extra ImmutableSet.

jlerbsc:
Thanks @kalayciburak.

Verified locally on the full javaparser-symbol-solver-testing module:

One nit before I merge: currentHostJdkMajor() is now only used by the two pre-existing guards (registerNatives >= 14, Constable < 12). Turning those into probes as well — Object.class.getDeclaredMethod("registerNatives") and Class.forName("java.lang.constant.Constable") — would let you drop the method entirely and remove the last coupling to a version number.

Once that is in, I will merge. This lands before [#5111](https://github.com/jlerbsc/javaparser/issues/5111): the current matrix stops at 18, so the probes stay inactive and master stays green; #5111 then rebases on top, and its extended 19/20/21 matrix validates both.

kalayciburak:
done, registerNatives and Constable are probes now so currentHostJdkMajor is gone

jlerbsc:
Thanks for the follow-through. All coupling to JDK version numbers is gone: currentHostJdkMajor() is fully Temoved and the four capability probes are consistent with each other.
The inverted hasConstable() branch is correct: both sides are exercised across those runs, and containsInAnyOrder is an exact match, so a swapped list would have failed.

PR #5117, opened 2026-09-01
Refs #5077: copy StaticJavaParser reset extension to javaparser-core-testing and javaparser-core-serialization

sylviauwa748:
hi @jlerbsc ,
Follow-up to #5077 — extends the reset extension to the two remaining modules that still mutate StaticJavaParser.

What changed:
Copied StaticJavaParserConfigResetExtension into each module's own test sources (javaparser-core-testing's existing utils package; javaparser-core-serialization's single flat test package).
Added the matching junit-platform.properties and META-INF/services registration to each, scoped to that module only.

javaparser-core-serialization: 11/11 tests pass. Its one mutation site sets and uses a resolver within a single test method, so there's no @BeforeAll-then-parse-later gap for the reset to break.
javaparser-core-testing: 2 pre-existing failures in Issue5099Test (Javadoc/comment matching, unrelated to parsing config). Confirmed these exist on unmodified master too, by stashing these changes out and re-running same two failures either way.

Refs #5077.


jlerbsc:
Thanks for the follow-up — this is exactly the remaining scope.

## Requested changes (all minor)

1. Remove the now-redundant @AfterAll in JavaParserJsonDeserializerTest
   (lines 226-229). It does exactly what the extension now does for every test
   in the class, so keeping both leaves two competing reset mechanisms for a
   single test. This is the only file in either module that mutates the shared
   configuration, so it is worth cleaning up in the same commit.

2. Reword the commit message to match the project convention: a type prefix, a
   subject under 72 characters, a body explaining the why, and the issue
   reference in the commit itself rather than only in the PR title. For example:

       test: extend StaticJavaParser reset guard to two modules (refs #5077)

3. afterEach does not truly reset: localConfiguration is a
  ThreadLocal.withInitial(...), so setConfiguration(new ParserConfiguration())
  populates it rather than clearing it, and the worker thread keeps a
  configuration referenced after the last test. A StaticJavaParser
  .resetConfiguration() doing localConfiguration.remove() would be the honest
  reset, and would help library users too, not just our tests.

sylviauwa748:
Thanks for the feedback, @jlerbsc 
I’ll work through these changes and update the PR.

jlerbsc:
Thanks for the rework @sylviauwa748 — the redundant @AfterAll is gone, and resetConfiguration() is a better answer than the one I suggested: since the ThreadLocal is declared withInitial(ParserConfiguration::new), remove() gives the same fresh default on the next call while also releasing the entry from the thread's map, which matters when tests run in a pool. Nice catch.

Four small things and it can go in:

Title and description. Both still describe the previous version ("copy StaticJavaParser reset extension to…"). The PR now adds a public API method to javaparser-core, and since the changelog is generated per milestone from PR titles, that needs to show in the title itself, not only in the body.

Javadoc on resetConfiguration(). Two lines would do: that subsequent calls get a fresh default configuration, and that — like setConfiguration and getParserConfiguration — it is thread-scoped, so it resets the calling thread only. That second point is the one users will get wrong otherwise.

Align the third copy. javaparser-symbol-solver-testing's copy of StaticJavaParserConfigResetExtension (merged in #5107) still calls setConfiguration(new ParserConfiguration()). Three copies with two implementations will drift; it's a one-line change here.

Eight hand-written resets in javaparser-symbol-solver-testing are now redundant thanks to the extension (Issue2684Test:71, JavaParserInterfaceDeclarationTest:1236, ConstructorsResolutionTest:57, EnumResolutionTest:106 and :143, FieldsResolutionTest:57, MethodsResolutionTest:57, MethodsResolutionWithJavassistTest:49). And javaparser-core-testing-bdd is still uncovered — worth either adding it for consistency or saying why not.

Thanks again for your patience through the back and forth on this one — the end result is better than what I originally asked for.

sylviauwa748:
Hi @jlerbsc, thanks for the review on resetConfiguration. Glad to hear it looks good overall! I'll push an update soon addressing all four points you raised.

sylviauwa748:
Hi @jlerbsc, I’ve pushed the follow-up changes covering all four points you raised, including the symbol-solver testing updates and redundant reset cleanup.

I also checked javaparser-core-testing-bdd. It does use StaticJavaParser parsing methods, but its configuration-related tests operate on locally created ParserConfiguration/JavaParser instances and do not mutate StaticJavaParser’s global/thread-scoped configuration. Therefore, the reset extension doesn’t appear necessary there, so I left that module unchanged.

I ran the full javaparser-symbol-solver-testing suite after the cleanup. It still has 12 failures and 5 errors, and I reproduced the exact same failures on clean master, so they appear to be pre-existing rather than related to these changes.

jlerbsc:
Thanks for the quick turnaround. One blocker left, then this is good to go.

## Blocker

spotless check fails on this branch. Removing the try/finally in EnumResolutionTest left the two
method bodies at the old indentation.

## Nits

1. The Javadoc now differs between the three copies. You updated the
   symbol-solver one to "Resets StaticJavaParser to its default configuration",
   but the two new copies still say "to a blank ParserConfiguration" — which is
   also no longer accurate, since the call no longer builds a ParserConfiguration.
   Same sentence in all three, please.

2. Commit type. The project merges branches rather than squashing, so both
   subjects land on master verbatim. The first commit adds a public API method to
   javaparser-core/src/main, so `feat:` fits it better than `test:`; the second
   one is genuinely `test:`.

3. The first commit's body says remove() is used "to truly reset ... instead of
   replacing it with a new configuration". I checked: swapping
   localConfiguration.remove() back to set(new ParserConfiguration()) leaves
   StaticJavaParserTest green. That is fine — the test covers the public contract, which is the
   right level — but it means the previous reset was not wrong. remove() is still
   the better choice, for non-retention rather than for correctness. Worth
   softening the wording so the history does not claim a bug that was not there.

4. MethodsResolutionTest and MethodsResolutionWithJavassistTest keep an
   @AfterEach named resetConfiguration() that now only calls
   Log.setAdapter(...). Good call not deleting them wholesale — but the name is
   now misleading, and it collides with the new StaticJavaParser API. resetLog()
   would say what they do.

5. The description still carries the original "2 pre-existing failures in
   Issue5099Test" line. That test is 2/2 on current master and on this branch, so
   the line reads as merging over known failures when nothing is failing. Please
   drop it.

Issue #2597, opened 2020-04-08
Documentation -- APIs for triggering a resolve


MysterAitch:
There are many variations on how we can resolve things, and I'm not sure of there being a place where these are collected together in one place. Here are some of my notes towards trying to put this together.

Personally I have been tripped up by this in the past when blindly stumbling through the IDE's autocomplete rather than trying to understand what is what. Given this, I'm sure it is all fine/logical, but for now I just don't see it at the moment.

Any input in how best to categorise these would be helpful!

**Parsing:**
```java
StaticJavaParser.parse(..) //CompilationUnit
StaticJavaParser.parseBlock(..) //BlockStmt
StaticJavaParser.parseStatement(..) //Statement
StaticJavaParser.parseImport(..) //ImportDeclaration
... and so forth


new JavaParser().parse(..) //ParseResult<CompilationUnit>
new JavaParser().parse(T, ..) //ParseResult<T>
new JavaParser().parseBlock(..) //ParseResult<BlockStmt>
new JavaParser().parseStatement(..) //ParseResult<Statement>
new JavaParser().parseImport(..) //ParseResult<ImportDeclaration>
... and so forth
```

**Resolution**
This is where it gets a bit fuzzier about the "correct" / intended way to do things.

Skimming through, some techniques are labelled as being intended for private/internal use.  

In the case of the `JavaParserFacade`, it is explicitly documented as being a helper class. 

Additionally, it seems that some implementations are just wrappers for others or take the others as parameters (e.g. some require the context to be passed as a parameter, and then the method just delegates to the context's resolution options). 

```java
// Helper class..
JavaParserFacade#solve(X) //ResolvedX


// interface SymbolResolver
SymbolResolver.resolveDeclaration(Node, T) // T
SymbolResolver.toResolvedType(Type, T) // T
SymbolResolver.calculateType(Expression) //ResolvedType

// This implementation of the SymbolResolver wraps the functionality of the library[..]
// An instance of this class should be created once and then injected in all the CompilationUnit for which we want to enable symbol resolution
// Can use the method inject, or call ParserConfiguration#setSymbolResolver(SymbolResolver) and the parser will do the injection for you.
JavaSymbolSolver.resolveDeclaration(Node, T) //T
JavaSymbolSolver.toResolvedType(Type, T) // T
JavaSymbolSolver.calculateType(Expression) //ResolvedType


// SymbolSolver
[...]

PR #5118, opened 2026-09-02
Document parse and resolve APIs #2597

AzazelSensei:
Refs #2597.

I put the parse vs resolve entry points in one page, with the layering (node → JavaSymbolSolver → JavaParserFacade → SymbolSolver → Context) rather than a ranking of which to call first.

### DIFF
--- a/doc/symbol-resolution.md
+++ b/doc/symbol-resolution.md
@@ -1,0 +1,167 @@
+# Parsing and symbol resolution APIs
+
+There are two steps: parse source into an AST, then (optionally) resolve names and types on that AST. This page is a short guide, then a per-class reference.
+
+Related: [About the Symbol Solver](https://github.com/javaparser/javaparser/wiki/About-the-Symbol-Solver) (lookup strategy and history), [JavaParser vs StaticJavaParser](https://javaparser.org/the-quick-and-the-full-api-of-javaparser/).
+
+## Parsing
+
+Use `JavaParser` with a `ParserConfiguration`. The configuration is what carries the `SymbolResolver`; the parser injects it into every successfully parsed compilation unit.
+
+```java
+ParserConfiguration config = new ParserConfiguration()
+        .setSymbolResolver(new JavaSymbolSolver(typeSolver));
+JavaParser parser = new JavaParser(config);
+
+ParseResult<CompilationUnit> result = parser.parse(source);
+if (!result.isSuccessful()) {
+    result.getProblems().forEach(System.err::println);
+    return;
+}
+CompilationUnit cu = result.getResult().get();
+```
+
+The same instance also has the fragment parsers, each returning `ParseResult<T>`:
+
+```java
+ParseResult<BlockStmt> block = parser.parseBlock("{ return 1; }");
+ParseResult<Statement> stmt = parser.parseStatement("int x = 1;");
+ParseResult<ImportDeclaration> imp = parser.parseImport("import java.util.List;");
+```
+
+Language level and the symbol resolver both live on `ParserConfiguration`. Build one `JavaParser` per configuration and reuse it.
+
+### `StaticJavaParser`
+
+`StaticJavaParser` is a convenience shortcut — its Javadoc calls it "a simpler, static API than `JavaParser`". It is fine for scripts, tests, and single-threaded one-offs. It is not what to use for real resolution work:
+
+- Parse errors are thrown instead of returned as `ParseResult`.
+- Configuration is a `ThreadLocal`. `setConfiguration` on one thread does not apply on another, so the `SymbolResolver` is silently missing and `resolve()` fails with the "not configured" exception below.
+
+## Configure and resolve
+
+```java
+TypeSolver typeSolver = new TypeSolverBuilder()
+        .withCurrentJRE()
+        .withSourceCode("src/main/java")
+        .build();
+
+ParserConfiguration config = new ParserConfiguration()
+        .setSymbolResolver(new JavaSymbolSolver(typeSolver));
+JavaParser parser = new JavaParser(config);
+
+ParseResult<CompilationUnit> result = parser.parse(source);
+CompilationUnit cu = result.getResult().get();
+MethodCallExpr call = cu.findFirst(MethodCallExpr.class).get();
+ResolvedMethodDeclaration resolved = call.resolve();
+ResolvedType type = call.calculateResolvedType();
+```
+
+These go through `SymbolResolver` and throw `UnsolvedSymbolException` when the name cannot be found.
+
+| Call | Typical result |
+| --- | --- |
+| `Resolvable.resolve()` | Declaration for that node |
+| `Expression.calculateResolvedType()` | `ResolvedType` of the expression |
+| `Type.resolve()` | `ResolvedType` for that type node |
+
+Nodes that implement `Resolvable` include method/constructor/field/parameter/type declarations, `MethodCallExpr`, `ObjectCreationExpr`, `MethodReferenceExpr`, `FieldAccessExpr`, `NameExpr`, `ThisExpr`, `AnnotationExpr`, and `ExplicitConstructorInvocationStmt`.
+
+## What goes wrong
+
+Without a resolver, or if the node is not in a compilation unit, `getSymbolResolver()` throws before resolution even starts:
+
+```
+IllegalStateException: Symbol resolution not configured: to configure consider setting a SymbolResolver in the ParserConfiguration
+IllegalStateException: The node is not inserted in a CompilationUnit
+```
+
+Once a resolver is present, a name that cannot be found throws `UnsolvedSymbolException`.
+
+Two styles exist next to each other:
+
+- **Throwing:** `TypeSolver.solveType(String)`, `node.resolve()`, `calculateResolvedType()`. Failure is `UnsolvedSymbolException` (or the `IllegalStateException`s above).
+- **`SymbolReference`:** `TypeSolver.tryToSolveType(String)`, `Context.solveSymbol` / `solveType` / `solveMethod`. Check `isSolved()` before `getCorrespondingDeclaration()`.
+
+`tryToSolveType` vs `solveType` on `TypeSolver` is the same split: one returns an unsolved reference, the other throws.
+
+Resolved types are cached on the node through `DataKey`s (`TYPE_WITH_LAMBDAS_RESOLVED` / `TYPE_WITHOUT_LAMBDAS_RESOLVED`). `calculateResolvedType()` therefore goes stale if the AST is mutated afterwards. `JavaParserFacade` also caches instances per `typeSolver.getRoot()` — see below.
+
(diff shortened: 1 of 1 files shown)


jlerbsc:
Thanks for picking this up. I checked the page against `master` before commenting, and the
technical content mostly holds up: the API listings, the `Resolvable` implementors, the
`TypeSolverBuilder` methods and the `Context.solveMethod` / `solveMethodAsUsage`
distinction are all correct. What needs work is where this lands, and the hierarchy the
page puts on top of the APIs.

### 1. This should go on the wiki rather than into `doc/`

`readme.md` already points at javaparser.org and the wiki as the information sites, and
`doc/` is not a documentation home — it currently holds a stale copy of the readme and a
`.puml`. The wiki also already has [About the Symbol
Solver](https://github.com/javaparser/javaparser/wiki/About-the-Symbol-Solver), which your
page links to and largely continues. Adding `doc/symbol-resolution.md` would create a
fourth place telling the same story, and the least discoverable one.

Could you move the guide to the wiki, either as a new page alongside *About the Symbol
Solver* or as a section inside it? Ping me if you need wiki access.

### 2. Please drop the `changelog.md` hunk

`changelog.md` is generated at release time — `scripts/run_generate_changelog_by_milestone_id.sh`
runs the changelog generator over the milestone, and the sections come from the `PR: *`
labels (`PR: Added`, `PR: Fixed`, …), with `excludeIssues: true` and the PR author added
automatically. Hand-written entries get duplicated or overwritten. Nothing for you to do
here: I'll set the milestone and the label.

### 3. Blocking: lead with `JavaParser`, not `StaticJavaParser`

`StaticJavaParser` is a convenience shortcut — its own Javadoc calls it "a simpler, static
API than `JavaParser`" — and it is not what we recommend for real use. It hides parse
problems behind exceptions instead of `ParseResult`, and its configuration lives in a
`ThreadLocal` (`StaticJavaParser.java:52-54`).

That second point matters here specifically, because the `ParserConfiguration` is what
carries the `SymbolResolver`. Configure on one thread, parse on another, and resolution
silently isn't enabled. So the page's flagship example currently teaches the one pattern
most likely to produce "why doesn't `resolve()` work?".

Please build both the parsing section and the resolution walkthrough on
`new JavaParser(config)` / `ParseResult`, and demote `StaticJavaParser` to a short note
covering when it is fine (scripts, tests, single-threaded one-offs) and what it costs.

### 4. Blocking: `JavaParserFacade` is a layer, not a rejected alternative

The page files `JavaParserFacade` under "Other entry points (usually not first)" and says
"prefer `node.resolve()`". That misdescribes the architecture rather than simplifying it:
`JavaSymbolSolver` delegates to `JavaParserFacade.get(typeSolver)` for essentially every
resolution (`JavaSymbolSolver.java:195-476`), and `calculateType` is literally
`JavaParserFacade.get(typeSolver).getType(expression)` (`JavaSymbolSolver.java:476`).
`node.calculateResolvedType()` *is* the facade, one hop further out.

So "prefer one over the other" isn't a meaningful recommendation, and it isn't a position
the project has taken — `JavaParserFacade` is public, supported, and its Javadoc
("Class to be used by final users to solve symbols for JavaParser ASTs") is accurate.

Please describe the actual difference, which is about configuration rather than
endorsement: the node API needs a `SymbolResolver` injected into the compilation unit and
then reads naturally from the AST; `JavaParserFacade` needs nothing but a `TypeSolver` in
hand and is what you reach for when you have an AST that was parsed without a resolver.
Showing the layering (node → `JavaSymbolSolver` → `JavaParserFacade` → `SymbolSolver` →
`Context`) would answer #2597 far better than a ranking does — that layering is exactly
what the issue could not work out from the outside.

If we do want to declare an officially preferred entry point, that's a maintainer decision
and a separate discussion; a documentation PR shouldn't be where it first appears.

### 5. Blocking: `Path.of` doesn't compile on our baseline

`Path.of` is Java 9+; we compile with `<java.release>8</java.release>` (`pom.xml:173`).
Use `Paths.get(...)`, or the `withSourceCode(String)` overload that already exists on
`TypeSolverBuilder`.

### 6. Missing — these are the questions people actually turn up with

* Quote the real failure messages verbatim — people search for them:
  `IllegalStateException("Symbol resolution not configured: ...")`, and the second case
  the page doesn't mention at all, `"The node is not inserted in a CompilationUnit"`
  (`Node.java:845-855`).
* Caching. `JavaParserFacade` holds a static `WeakHashMap<TypeSolver, JavaParserFacade>`
  (`JavaParserFacade.java:69`) with `clearInstances()`, and resolved types are cached on
  the node through `DataKey`s — so `calculateResolvedType()` goes stale if the AST is
  mutated afterwards.
* Thread safety. `JavaParserFacade.get` carries an explicit warning pointing at #2668 and
  #2671; the page is silent on it.
* The `solveMethod` collision is only half covered. You explain `Context.solveMethod` vs
  `solveMethodAsUsage`, but `Solver.solveMethod(...)` returns a `MethodUsage` while
  `Context.solveMethod(...)` returns a `SymbolReference<ResolvedMethodDeclaration>` — same
  name, swapped semantics. #2597 names this case specifically, so it should be spelled out.
* `Context.solveType(String)` is deprecated in favour of `solveType(String, List)`
  (`Context.java:85-90`); the list doesn't mention it.
* Nothing on the general failure-handling pattern (`SymbolReference.isSolved()` vs
  throwing, `tryToSolveType` vs `solveType`), nor on the `CombinedTypeSolver` /
  `getRoot()` pitfall.

### 7. `Fixes #2597`

#2597 is really two requests: document the entry points, and address the fact that the
relationships between them are opaque. A guide covers the first half well. Could you
switch to `Refs #2597`? I'd rather keep the issue open until the layering above is written
down somewhere durable.


Thanks again — the research behind this is solid, and the layering question is the one
piece that would make it genuinely answer the issue.

AzazelSensei:
I don't have wiki write access — could you grant it? I'll move the page once I can.

Dropped the changelog.md hunk and switched this to Refs #2597. Rewrote the guide around `JavaParser`/`ParseResult`, the actual layering instead of a ranking, Java 8 (`withSourceCode(String)`), and the failure/cache/thread-safety/`solveMethod` notes.

jlerbsc:
Thanks for the rewrite — this is a substantial improvement.

Three things left.

### 1. Placement

You asked about wiki write access — I can't grant that, and it's worth explaining why so it
doesn't read as a brush-off. GitHub has no wiki-only permission: wiki write is repository
push access, so the only ways to give it are to make you a repo collaborator or to open the
wiki to every GitHub account. Neither is proportionate to a documentation change, and
neither is about you specifically.

Simplest path: leave the file where it is, I'll approve the text here, and once we're done
I'll copy the page onto the wiki myself alongside *About the Symbol Solver*, crediting you.
The review then still happens in a PR, which is what caught the `Path.of` issue and the
facade-vs-node framing last round.

### 2. Blocking: the review conversation leaked into the page

A few passages now address me rather than the reader:

- *"Do not use `Path.of` — the project compiles with `<java.release>8</java.release>`"*.
  That's a constraint on **contributors to JavaParser**, not on users of the library. A
  reader on JDK 21 can use `Path.of` perfectly well, so as user-facing advice it's simply
  wrong. Just use `withSourceCode("src/main/java")` in the example and say nothing.
- *"People search for them verbatim"* — that justifies including the exception messages; it
  isn't information for the reader. Drop the sentence, keep the messages.
- *"its Javadoc (\"Class to be used by final users…\") is accurate"* and *"`JavaParserFacade`
  is not a rejected alternative to `node.resolve()`"* — the page is defending itself against
  a criticism its reader never saw. The layering diagram and the "the difference is
  configuration" bullet already make the point; these two lines only raise a question that
  wasn't there.

Worth a general pass for the same pattern: the page should read as if written from scratch
for someone who has never seen this PR.

### 3. Structure: the page doubled and started repeating itself

`JavaParserFacade` now appears three times (layering, caching, and a dedicated section
reduced to a single line of example calls); `TypeSolver` twice. The running order also
alternates between tutorial and reference — Parsing, Layering, Walkthrough, Node methods,
Failure handling, Caching, Thread safety, then six per-class sections.

Two clear halves would fix it: a guide (parse → configure → resolve → what goes wrong), then
a per-class reference. Fold the one-line `JavaParserFacade` section into the layering
section, and move the `Solver.solveMethod` row out of the table headed `Context` or retitle
that table.

### Commits

Please squash to one commit. The repo convention is a Conventional Commits subject of 72
characters or less with the issue reference as a suffix, plus a body when the change isn't
trivial — for example:

```
docs: describe the parse and resolve entry points (refs #2597)
```

AzazelSensei:
Left the file in doc/. Thanks for moving it to the wiki after this.

Dropped the Path.of note and the other lines that were talking to the review. Split into a guide then a per-class reference, folded JavaParserFacade into layering, and took Solver.solveMethod out of the Context table. Squashed to one commit.

jlerbsc:
Merging this as a wiki page rather than into `doc/`, and closing the PR here — that is not
a rejection, so let me explain where it landed and why.

**Where it is now:** [Parsing and Symbol Resolution
APIs](https://github.com/javaparser/javaparser/wiki/Parsing-and-Symbol-Resolution-APIs),
next to *About the Symbol Solver*, with links between the two. That pairing is the right
one: *About* covers what the solver is and how lookup walks outward, yours covers which
entry point to call and how the pieces sit on top of each other.

`doc/` was the wrong home for it. It holds a stale copy of the readme and a `.puml`, and
nothing links to it; the readme points readers at javaparser.org and the wiki instead. On
the wiki the page gets a sidebar entry and shows up in wiki search, which is what a page
like this needs.

**On wiki write access**, since you asked: GitHub has no wiki-only permission — wiki write
is repository push access, so the only ways to grant it are to add you as a repo
collaborator or to open the wiki to every GitHub account. Neither is proportionate to a
documentation change, and neither is a judgement about you. So I copied the page across
myself. If you want to revise it later, open an issue or ping me and I'll apply it.

**What I adjusted while copying**, so nothing is a surprise:

- Moved `Layering` ahead of `What goes wrong`. The failure section referred to
  `JavaParserFacade` (and had a "see below") before the reader met the class.
- Merged the two halves of the caching material into a `Caching` subsection under
  `Layering`, and pointed `CombinedTypeSolver and getRoot()` at it instead of restating it.
- Split `Thread safety` into its own section — it is cross-cutting rather than a resolution
  failure mode.
- Named the calls in "`resolve()` and `calculateResolvedType()` are the node-level entry
  points" instead of the bare "These".
- Added the trailing period to `This TypeSolver already has a parent.`

Everything else is your text unchanged, and the page credits you at the bottom.

**On #2597:** I'm closing it against this page. Worth being explicit that the page
*documents* the naming inconsistencies rather than removing them — `Solver.solveMethod`
returning a `MethodUsage` while `Context.solveMethod` returns a declaration reference is
still there, it is now just written down. If we ever want to act on that, it should be its
own issue with an API-change label.

Thanks for three rounds of this. You checked the source rather than guessing, you took the
`Path.of` and facade-versus-node corrections on board without arguing them, and the
`CombinedTypeSolver` / `getRoot()` section — the parent link, and the fact that the facade
cache keys on the root rather than the instance you passed — is genuinely new material that
existed nowhere in our documentation. That part alone answers a question we get regularly.

Since the PR closes unmerged, it won't show in your contribution graph. If you'd like, say
the word and I'll add you to the contributors list for the next release notes.

PR #5121, opened 2026-09-02
Make identifier and keyword case conversions locale-independent

dylanpulver:
Java keywords and identifiers are normalised with the default-locale `String.toLowerCase()`/`toUpperCase()` overloads. In a Turkish locale `I` folds to dotless `ı` and `i` to dotted `İ`, so these conversions corrupt the strings they exist to normalise.

The worst site is `Utils.screamingToCamelCase`, which `ObservableProperty.camelCaseName()` uses to build a reflective getter name. **43 of the 104 `ObservableProperty` constants contain an `I`**, so their getters are not found and everything routed through the concrete syntax model throws, `LexicalPreservingPrinter` included:

```
RuntimeException: Unable to get value for ANNOTATIONS from class A { } (ClassOrInterfaceDeclaration)
```

**Measurement.** 23 probes over the public API, run in two separate JVMs (`-Duser.language=en` vs `tr`) rather than `Locale.setDefault`, because some values are captured in static initialisers and a single JVM cannot see those:

| build | divergences |
|---|---|
| master | **20 / 23** |
| this PR | 5 / 23 |
| this PR + #5086 | **0 / 23** |

**Relationship to #5086.** @tarann26's open PR fixes `PrimitiveType`, which is exactly the 5 that remain here. This PR deliberately does not touch that file. The two share no file, and cherry-picking this commit onto `pull/5086/head` applies with zero conflicts. Happy to fold them together or rebase if you would rather have one PR.

**Tests.** No test in the suite touches `Locale`, which is why 13 sites survived since 2016-17. The new `LocaleIndependenceTest` asserts values under a Turkish default; all 6 fail on master. Reverting only `Utils.java` still fails 2, so the other sites are independently load-bearing.

Three of the sites I pinned (`NoBinaryIntegerLiteralsValidator`, `TypeExtractor`, `JavaParserTypeSolver`) are latent. I could not make them misbehave in a Turkish locale since their inputs have no `i`/`I`. They are changed for consistency, not because I measured a failure.

Full suite green: 2092 run, 0 failures, 34 skipped. Part of #3018.

---
Drafted with the assistance of Claude Opus 5 (`claude-opus-5`); I have reviewed and verified every change and the measurements above.

### DIFF
--- a/javaparser-core-testing/src/test/java/com/github/javaparser/utils/LocaleIndependenceTest.java
+++ b/javaparser-core-testing/src/test/java/com/github/javaparser/utils/LocaleIndependenceTest.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright (C) 2007-2010 Júlio Vilmar Gesser.
+ * Copyright (C) 2011, 2013-2026 The JavaParser Team.
+ *
+ * This file is part of JavaParser.
+ *
+ * JavaParser can be used either under the terms of
+ * a) the GNU Lesser General Public License as published by
+ *     the Free Software Foundation, either version 3 of the License, or
+ *     (at your option) any later version.
+ * b) the terms of the Apache License
+ *
+ * You should have received a copy of both licenses in LICENCE.LGPL and
+ * LICENCE.APACHE. Please refer to those files for details.
+ *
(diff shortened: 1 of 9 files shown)


jlerbsc:

PR #5135, opened 2026-09-13
Close qualified method resolution through the receiver type's static imports

### LINKED ISSUE #5105: Qualified method calls resolve through the receiver type's own static imports
## Summary

`MethodCallExpr#resolve()` resolves a qualified call `X.m()` / `expr.m()` to a method that is **not a
member of `X`**, when the file that declares `X` happens to statically import a method of that name from
somewhere else. `javac` rejects such code, because static imports are not transitive: `import static Y.*`
in `X.java` is a property of the *file*, not of the type `X`, and does not add members to `X`
(JLS 7.5.3 / 7.5.4).

The unqualified form of this was fixed for methods in #4358. The qualified form was not, because the
mechanism that fix relies on (`staticOnly`) is not set on the qualified path.

The same escape exists on the field side, addressed separately in #4450 / #4982 by a different mechanism
again. Four cases, three ad-hoc enforcements, one gap — see *The same defect elsewhere* below.

## Reproducer

Three files, package `qq`:

```java
// Sink.java
package qq;
public class Sink {
    public static String ping() { return "p"; }
    public String inst() { return "i"; }
}
```

```java
// Mid.java   -- declares NOTHING; only imports
package qq;
import static qq.Sink.*;
public class Mid { }
```

```java
// Outer.java
package qq;
public class Outer {
    public String a() { return Mid.ping(); }        // javac: cannot find symbol
    public String b() { return new Mid().ping(); }  // javac: cannot find symbol
    public String c() { return new Sink().inst(); } // legal -- control
}
```

`javac` output:

```
Outer.java:3: error: cannot find symbol
  symbol:   method ping()
  location: class Mid
Outer.java:4: error: cannot find symbol
  symbol:   method ping()
  location: class Mid
2 errors
```

Resolution code:

```java
CombinedTypeSolver ts = new CombinedTypeSolver();
ts.add(new ReflectionTypeSolver());
ts.add(new JavaParserTypeSolver(pathTo("qq's parent directory")));

CompilationUnit cu = /* parse Outer.java with the solver configured */;
for (MethodCallExpr call : cu.findAll(MethodCallExpr.class)) {
    System.out.println(call + " -> " + call.resolve().getQualifiedSignature());
}
```

## Expected vs actual

| call | expected | actual |
|---|---|---|
| `Mid.ping()` | unsolved (`UnsolvedSymbolException`) | **`qq.Sink.ping()`** |
| `new Mid().ping()` | unsolved (`UnsolvedSymbolException`) | **`qq.Sink.ping()`** |
| `new Sink().inst()` | `qq.Sink.inst()` | `qq.Sink.inst()` (correct) |

`Mid` declares no members at all, yet a call qualified by `Mid` resolves to a method of `Sink`.

## Control experiment

Deleting the single import line from `Mid.java` — a file neither call mentions — flips both results:

| call | `Mid.java` **has** `import static qq.Sink.*` | import removed |
|---|---|---|
| `Mid.ping()` | `qq.Sink.ping()` | UNSOLVED |
| `new Mid().ping()` | `qq.Sink.ping()` | UNSOLVED |
| `new Sink().inst()` | `qq.Sink.inst()` | `qq.Sink.inst()` |

This isolates the cause: resolution is consulting `Mid.java`'s import list while answering "does `Mid`
have a member `ping`?". The control call is unaffected in both runs, so the type solver and the
instance-call machinery are working normally.

## Root cause

`JavaParserTypeDeclarationAdapter#solveMethod` searches the type's declared methods, then its ancestors,
and then falls through to the **enclosing lexical context** — which for a top-level type is its
compilation unit, where the imports live:

```java
// JavaParserTypeDeclarationAdapter.java:317
if (candidateMethods.isEmpty() && !staticOnly) {
    SymbolReference<ResolvedMethodDeclaration> parentSolution = context.getParent()
            .orElseThrow(...)
            .solveMethod(name, argumentsTypes, staticOnly);
    ...
}
```

That fall-through is correct and necessary for **lexical** lookups — an unqualified name written inside
the class body must be able to reach the file's imports (e.g. #1726). It is wrong for **member**
lookups, which must stop at the type's declared and inherited members.

`staticOnly` currently distinguishes the two cases, and it is only ever `true` from one place:

- `CompilationUnitContext:369,380` — resolving a static import — passes `true`, so the gate closes.
- `MethodCallExprContext:172` — resolving a qualified call — passes `false`, so the gate stays open.

Hence the qualified path escapes into the receiver type's file.

## Why simply passing `staticOnly = true` on the qualified path does not fix it

`staticOnly` does two independent jobs:

1. filters candidates to static members only (`JavaParserTypeDeclarationAdapter:283`)
2. closes the parent-context fall-through (`:317`)

Passing `staticOnly = true` at `MethodCallExprContext:172` does close the gate and fix both `ping` calls —
but job 1 then filters out every instance method, for all qualified calls. Measured on this reproducer:

| call | `staticOnly = false` (current) | `staticOnly = true` |
|---|---|---|
| `Mid.ping()` | `qq.Sink.ping()` (wrong) | unsolved (correct) |
| `new Mid().ping()` | `qq.Sink.ping()` (wrong) | unsolved (correct) |
| `new Sink().inst()` (legal) | `qq.Sink.inst()` (correct) | **unsolved (regression)** |

Across `javaparser-symbol-solver-testing` that change adds **+36 failures and +125 errors**. The two jobs
genuinely cannot share one boolean.

(`new Mid().ping()` is valid Java *syntax* — invoking a static method through an instance receiver is
legal, if discouraged. `javac` reports `cannot find symbol`, a semantic error, not a parse error, and the
call would compile if `ping` were actually a member of `Mid`.)

## The same defect elsewhere

Non-transitivity of static imports applies to fields and methods alike (same sentence of JLS 7.5.3/7.5.4),
so there are four cases — and this is the last one unenforced:

| | unqualified — via another file's `import static T.*` | qualified — `T.x` / `expr.x` |
|---|---|---|
| **methods** | fixed in #4358 — `staticOnly` closes the gate at `:317` | **BROKEN — this report** |
| **fields** | fixed in #4982 — thread-local "name in flight" guard in `CompilationUnitContext` | correct, but incidentally |

- **methods / unqualified** — #4358 is the same escape from the other direction. Reverting `&& !staticOnly`
  at `:317` makes `Issue4358Test` overflow again; the loop is
  `adapter(A) -> CU_A -> import static B.d -> adapter(B) -> ancestors -> adapter(A) -> ...`. This report
  surfaces as a wrong resolution rather than an overflow only because the qualified path forms no cycle.
- **fields / unqualified** — before #4982, `Outer` with `import static Holder.*` resolved a bare `ID` to
  `Sink.ID` where `Holder` merely imports it; in a cyclic import graph this overflowed (#4450).
- **fields / qualified** — `Mid.FIELD` where `Mid.java` statically imports `Sink.FIELD` is correctly
  unsolved, but nothing guards it: `FieldAccessContext#solveField` resolves via `rrtd.getField(...)`, a
  lookup **on the declaration**, so it never enters the type's `Context`. Member-only by construction
  rather than by intent.

One language rule, three unrelated enforcements — a boolean parameter on the method chain, a thread-local
name set on the value chain, and one call site that bypasses `Context` — and one cell enforced by nothing.
That asymmetry is the underlying problem; this report is where it becomes visible.

## Proposed symmetric fix

Give both chains one explicit concept — "this is a member lookup, do not escape to the enclosing
context" — and set it at every member-lookup entry point. Concretely:

1. Add a `memberOnly` parameter, as `default` methods so no external implementer breaks:
   - values: `Solver#solveSymbolInType`, `SymbolResolutionCapability#solveSymbol`, `Context#solveSymbol`
   - methods: separate the two jobs `staticOnly` currently does, leaving it to mean only "static members
     are eligible"
2. Gate both fall-throughs on it:
   - `JavaParserTypeDeclarationAdapter:317` — `&& !memberOnly` instead of `&& !staticOnly`
   - `ClassOrInterfaceDeclarationContext:71` (and the `Enum`/`Record`/`Annotation` equivalents) — return
     unsolved instead of calling `solveSymbolInParentContext`
3. Set it at the member-lookup entry points: `MethodCallExprContext`, `MethodReferenceExprContext`,
   `FieldAccessContext`, and both the static-import loop and the absolute-reference branch of
   `CompilationUnitContext`.
4. Delete the thread-local guard in `CompilationUnitContext`, which then has nothing to do.

This closes the broken cell, replaces the two workarounds with the mechanism they are approximating, and
leaves the inherited-member and lexical paths untouched — the fall-through remains available for
unqualified names written inside a class body, which genuinely need it (#1726).

A prototype of the **value half** (steps 1–2 for `solveSymbol`, plus 3 for `CompilationUnitContext`,
plus 4) has been built and run: 14 files changed, and both `javaparser-symbol-solver-testing` and
`javaparser-core-testing` come out with results identical to the unmodified baseline. The method half is
untried.

One implementation note: `StatementContext` already declares a private `solveSymbol(String, boolean)` with
an unrelated meaning, which collides with the new interface default and breaks its eight subclasses until
renamed. A distinctly named entry point (e.g. `solveSymbolAsMember`) may be preferable to a bare boolean
for that reason, and reads better at the call site.

## Scope

- **The wrong resolution is on one path: methods, qualified.** The proposed fix spans both fields and
  methods, because it replaces the two existing workarounds rather than patching only the gap.
- Only reproduces when the receiver type is resolved **from source** (`JavaParserTypeSolver`).
  Types resolved reflectively or from a jar have no compilation unit to escape into, which is likely why
  this has gone unnoticed.
- The `solveType` path (nested type lookups, also non-transitive) has **not** been checked and may be a
  fifth case.

## Environment

- JavaParser `3.29.0-SNAPSHOT`, `javaparser/javaparser` master at `fea481092`
- JDK 21.0.10 (Temurin)

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

Fixes #5105.

`Mid.ping()` and `new Mid().ping()` no longer resolve to a method that `Mid` only statically imports. They are unsolved when `Mid` has no such member, matching javac. `new Sink().inst()` is unchanged.

Root cause: `JavaParserTypeDeclarationAdapter` falls through to the enclosing lexical context, which holds the imports of the file declaring the type. Correct for an unqualified name written inside that class. Wrong for a member lookup from outside, because static imports are not transitive (JLS 7.5.3/7.5.4). `staticOnly` closes that fall-through but also filters out instance methods, so it cannot carry both jobs.

Fix: a new `memberOnly` flag runs through `Context`, `MethodResolutionCapability`, `MethodUsageResolutionCapability` and their JavaParser implementations, including the ancestor recursion of `JavaParserTypeDeclarationAdapter` and `AnonymousClassDeclarationContext`. It is set when a method call has a receiver.

What stays as before: the fall-through stays open for unqualified names. A lookup on a subclass no longer escapes into the file of its superclass. Inherited static members still resolve. A receiver that is a type name also passes `staticOnly`, matching JLS 15.12 for `TypeName` receivers.

What changes: enum synthetic methods (`values`, `valueOf`) keep their dispatch on the new paths. A `this`-qualified call to a method that only the file's own static imports provide is now unsolved, matching javac.

Not covered: fields and method references keep their existing mechanisms (#4982 thread local, `FieldAccessContext`). Unifying them is left as follow-up.

Verification: `Issue5105Test` covers `qq/Sink`, `qq/Mid` (imports only), `qq/MidSub` (extends Mid), `qq/Deep` (extends Sink) and `qq/Outer`.

1. Both qualified `ping` calls fail on main and are unsolved with the fix.
2. The subclass-qualified call is unsolved as well.
3. The inherited-static call and the control call resolve.
4. `javaparser-symbol-solver-testing` 2105 and `javaparser-core-testing` 2853 tests pass.

### DIFF
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/Context.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/Context.java
@@ -371,6 +371,18 @@ public interface Context {
         return solveMethodInParentContext(name, argumentsTypes, staticOnly);
     }
 
+    /**
+     * Similar to {@link #solveMethod(String, List, boolean)}, but a member lookup
+     * stops at the declared and inherited members of the type it started from:
+     * it must not escape into the enclosing lexical context (e.g. the imports of
+     * the file declaring that type, which are not visible from outside, see JLS
+     * 7.5.3/7.5.4 on the non-transitivity of static imports).
+     */
+    default SymbolReference<ResolvedMethodDeclaration> solveMethod(
+            String name, List<ResolvedType> argumentsTypes, boolean staticOnly, boolean memberOnly) {
+        return solveMethod(name, argumentsTypes, staticOnly);
+    }
+
     default SymbolReference<ResolvedMethodDeclaration> solveMethodInParentContext(
             String name, List<ResolvedType> argumentsTypes, boolean staticOnly) {
         Optional<Context> optionalParentContext = getParent();
@@ -386,4 +398,14 @@ public interface Context {
      * A MethodUsage corresponds to a MethodDeclaration plus the resolved type variables.
      */
     Optional<MethodUsage> solveMethodAsUsage(String name, List<ResolvedType> argumentsTypes);
+
+    /**
+     * Similar to {@link #solveMethodAsUsage(String, List)}, but the lookup is a
+     * member lookup and must not escape into the enclosing lexical context of
+     * the receiver type declaration.
+     */
+    default Optional<MethodUsage> solveMethodAsUsage(
+            String name, List<ResolvedType> argumentsTypes, boolean memberOnly) {
+        return solveMethodAsUsage(name, argumentsTypes);
+    }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionCapability.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionCapability.java
@@ -29,4 +29,14 @@ public interface MethodResolutionCapability {
 
     SymbolReference<ResolvedMethodDeclaration> solveMethod(
             String name, List<ResolvedType> argumentsTypes, boolean staticOnly);
+
+    /**
+     * Similar to {@link #solveMethod(String, List, boolean)}, but a member lookup
+     * must not escape into the enclosing lexical context of this declaration
+     * (e.g. the imports of the file declaring the type).
+     */
+    default SymbolReference<ResolvedMethodDeclaration> solveMethod(
+            String name, List<ResolvedType> argumentsTypes, boolean staticOnly, boolean memberOnly) {
+        return solveMethod(name, argumentsTypes, staticOnly);
+    }
 }
--- a/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
+++ b/javaparser-core/src/main/java/com/github/javaparser/resolution/logic/MethodResolutionLogic.java
@@ -1206,6 +1206,19 @@ public class MethodResolutionLogic {
         throw new UnsupportedOperationException(typeDeclaration.getClass().getCanonicalName());
     }
 
+    public static SymbolReference<ResolvedMethodDeclaration> solveMethodInType(
+            ResolvedTypeDeclaration typeDeclaration,
+            String name,
+            List<ResolvedType> argumentsTypes,
+            boolean staticOnly,
+            boolean memberOnly) {
+        if (typeDeclaration instanceof MethodResolutionCapability) {
+            return ((MethodResolutionCapability) typeDeclaration)
+                    .solveMethod(name, argumentsTypes, staticOnly, memberOnly);
+        }
+        throw new UnsupportedOperationException(typeDeclaration.getClass().getCanonicalName());
+    }
+
     public static void inferTypes(
             ResolvedType source, ResolvedType target, Map<ResolvedTypeParameterDeclaration, ResolvedType> mappings) {
         if (source.equals(target)) {
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/MethodUsageResolutionCapability.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/core/resolution/MethodUsageResolutionCapability.java
@@ -33,4 +33,18 @@ public interface MethodUsageResolutionCapability {
             List<ResolvedType> argumentTypes,
             Context invocationContext,
             List<ResolvedType> typeParameters);
+
+    /**
+     * Similar to {@link #solveMethodAsUsage(String, List, Context, List)}, but the
+     * lookup is a member lookup on this declaration and must not escape into the
+     * enclosing lexical context (e.g. the imports of the file declaring the type).
+     */
+    default Optional<MethodUsage> solveMethodAsUsage(
+            String name,
+            List<ResolvedType> argumentTypes,
+            Context invocationContext,
+            List<ResolvedType> typeParameters,
+            boolean memberOnly) {
+        return solveMethodAsUsage(name, argumentTypes, invocationContext, typeParameters);
+    }
 }
--- a/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
+++ b/javaparser-symbol-solver-core/src/main/java/com/github/javaparser/symbolsolver/javaparsermodel/contexts/AbstractJavaParserContext.java
@@ -343,7 +343,13 @@ public abstract class AbstractJavaParserContext<N extends Node> implements Conte
      */
     @Override
     public Optional<MethodUsage> solveMethodAsUsage(String name, List<ResolvedType> argumentsTypes) {
-        SymbolReference<ResolvedMethodDeclaration> methodSolved = solveMethod(name, argumentsTypes, false);
+        return solveMethodAsUsage(name, argumentsTypes, false);
+    }
+
+    @Override
+    public Optional<MethodUsage> solveMethodAsUsage(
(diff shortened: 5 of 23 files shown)


jlerbsc:
Thanks — the diagnosis is right (JLS 7.5.3/7.5.4: static imports are not
transitive) and the added tests pass. Three points block this as it stands.

**1. Regression.** In `MethodCallExprContext`, `solveMethodInType(ref, name,
argumentsTypes, true)` passes `staticOnly`, not `memberOnly` — that overload's
4th parameter is `staticOnly`. The comment above it describes member-only
semantics; the escape is blocked only as a side effect of the `!staticOnly`
guard in the adapter. It silently makes `Name.method()` static-only, which
breaks obscuring (JLS 6.4.2):

```java
public class Obscure {
    static Sink Sink = new Sink();
    public String m() { return Sink.inst(); }   // legal Java
}
```

`calculateResolvedType()` now throws `UnsolvedSymbolException` (master returns
`java.lang.String`) while `resolve()` still succeeds — the two disagree on the
same node.

**2. The rule is not enforced, only patched per call site.**
`MethodReferenceExprContext` goes through the same `solveMethodInType` and is
untouched, so `Mid::ping` still resolves to `Sink.ping()` while `Mid.ping()` no
longer does.

**3. `memberOnly` has only one correct value, so it should not be a parameter.**
`ReflectionClassDeclaration` and `JavassistClassDeclaration` already do a
members-only lookup for the same `MethodResolutionCapability` contract;
`JavaParserClassDeclaration.solveMethod` is the outlier, delegating to
`getContext()` and therefore to the declaring file's imports. The same call
resolves differently depending on whether `Mid` comes from source or from a jar
— that is the real bug, and `memberOnly=true` just means "honour the contract".
Cutting the escape inside the five `JavaParser*Declaration` makes ancestor
recursion correct by construction: no threading, no new API on `Context`,
`MethodResolutionCapability`, `MethodUsageResolutionCapability`,
`MethodResolutionLogic` or `ContextHelper`, and `Mid::ping` fixed for free.

I'll take this one over and rework it along those lines — thanks for the
report and the test cases, they're useful and I'll keep them.

PR #5142, opened 2026-09-18
Fixes #3990

### LINKED ISSUE #3990: Local Enum and Interface (Java 16)
From Java 16's [JEP395](https://openjdk.org/jeps/395):

> The ability to declare local record classes, local enum classes, and local interfaces was introduced

JavaParser does allow for local records, but does not allow local enums or interfaces.

Enums:
```java
void b() {
    enum Z { Y, X }
    System.out.println("Z: " + Z.Y);
}
```

Error: 

> Parse error. Found "{", expected one of  "," ";" "=" "@" "["

Interfaces:
```java
void a() {
    interface B { void b(); }
    B b = new B() {
        @Override 
        public void b() { 
            System.out.println("B");
        }
    };
    b.b();
 }
```

Error:

> There is no such thing as a local interface.

mbarjak:
Fixes #3990.

PR #4626 added support for local interface declarations but explicitly left local enum declarations unsupported. This PR implements the remaining local-enum portion of the issue.

Changes:

- parses local enum declarations as block statements
- introduces `LocalEnumDeclarationStmt`
- integrates the node with the metamodel, visitors, printers, validators, symbol solver, and control-flow handling
- rejects local enums before Java 16
- validates the permitted local-enum modifiers
- adds AST, language-level, modifier, cloning, printing, and visitor tests

PR passes checks:

- `./mvnw test`
- `./mvnw spotless:check checkstyle:check`
-  core metamodel and source generators produce no additional changes


javaparser-core/src/main/javacc/java.jj
```diff
@@ -4561,14 +4561,20 @@ Statement BlockStatement():
...
     ClassOrInterfaceDeclaration typeDecl;
     RecordDeclaration recordDeclaration;
+    EnumDeclaration enumDeclaration;
     ModifierHolder modifier;
 }
 {
     try {
         (
-            LOOKAHEAD( Modifiers() ("class" | "interface") ) // JLS specifies local class. Since java 16 local enum or interface are allowed.
+            LOOKAHEAD( Modifiers() ("class" | "interface") ) // JLS specifies local class. Since java 16 local interface is also allowed.
```
mbarjak, line 4570: - The LOOKAHEAD( Modifiers() ("class" | "interface") ) line only covers the class/interface alternative — it has nothing to do with enum, which is a completely separate alternative I added right below it (line 4574–4577) with its own comment: // Since java 16 local enum is also allowed (JEP 395).
- The old comment text: "...Since java 16 local enum or interface are allowed.", was inaccurate even before my change touched local enums at all: it lumped "enum" into a comment that sits on the class/interface lookahead line, where enum was never handled (that's exactly the bug this whole branch fixes). Once local enum got its own alternative with its own comment, leaving "enum" in the class/interface line's comment would be actively misleading, it would look like enum support belonged to that branch when it's handled two lines down.
- So narrowing it to "...Since java 16 local interface is also allowed." on that line, and stating enum support separately on its own line, makes each comment accurately describe the branch it's attached to.

jlerbsc:
Blocking — the new BlockStatement() alternative breaks parsing of enum used as an identifier

In java.jj, <ENUM> is accepted by Identifier() (comment: "Make sure older Java versions parse"); rejecting it as an identifier is the job of ReservedKeywordValidator("enum"), registered from Java5Validator onwards. Since Modifiers() can match empty, LOOKAHEAD( Modifiers() "enum" ) commits any statement that starts with a variable named enum to EnumDeclaration().

I built javaparser-core from this branch and from its base (fdc458d) and ran the same probe. Through ParseStart.STATEMENT at JAVA_1_4, enum = 3;, enum.foo();, enum x = null;, enum[0] = 1; and enum++; all yield an ExpressionStmt on the base and a parse error on this PR; same results inside a COMPILATION_UNIT. This is the Enumeration enum = ...; enum.hasMoreElements(); idiom of pre-Java-5 sources.

Diagnostics regress at every language level, not just the old ones: from JAVA_5 up the parse now fails before validation, so

'enum' cannot be used as an identifier as it is a keyword.
is replaced by a raw JavaCC trace:

Parse error. Found "=", expected one of  "_" "assert" "enum" "exports" ... <IDENTIFIER>
  com.github.javaparser.GeneratedJavaParser.EnumDeclaration(GeneratedJavaParser.java:941)
  com.github.javaparser.GeneratedJavaParser.BlockStatement(GeneratedJavaParser.java:6373)
CI cannot catch this: the only existing coverage for enum-as-identifier is int enum; (enumAllowedAsIdentifier in Java1_4ValidatorTest and Java5ValidatorTest), which starts with int and never reaches the new lookahead.

A statement beginning with enum has to stay parsable, and the fix needs a regression test covering that form.

PR #5144, opened 2026-09-18
Deduplicate CombinedTypeSolver lookup and caching logic

piyushjoshi01:
Ordinary and module-qualified lookups in `CombinedTypeSolver` duplicate cache access, ordered solver traversal, exception filtering, and result caching. Extract the shared policy into a private helper accepting the lookup operation so changes to that policy apply consistently to both paths. Public APIs, cache keys, solver ordering, and exception behavior remain unchanged.

Add regression coverage for both lookup modes: first-match ordering, cached successes and misses, cache invalidation when adding a solver, exception propagation and fallback, and separate cache entries for ordinary lookups and different modules.

Validation on current upstream master:
- Focused `CombinedTypeSolverTest`: 30 tests passed.
- Full reactor with `./mvnw -PAlsoSlowTests --fail-at-end test`: 5673 passed, 41 skipped, no failures or errors.
- Targeted Spotless check and `git diff --check` passed.
- Both contribution generator scripts completed with `MAVEN_ARGS=-Dmaven.clean.skip=true`; plain runs hit local output-directory cleanup errors. Generation and formatting left only the two intended files changed.


piyushjoshi01:
@jlerbsc Could you please review this PR and merge it if everything looks good? Thank you!

jlerbsc:
Thanks for this cleanup. The two lookup paths had duplicated the same loop, and having one place for the cache and exception policy lowers the risk of fixing one path and forgetting the other.

A few requests before merging:

1. **Rename the private helper.** It overloads the public `tryToSolveType(String)`, but its first `String` means something else (a cache key, not a type name). Two same-named methods whose `String` argument means different things are easy to misread. A distinct name such as `solveWithCache(String cacheKey, Function<...> lookup)` would make that clear.

2. **Keep the diff focused.**
   - Please revert the `ts` → `solver` and `res` → `result` renames. They add lines to the diff without helping the refactoring.
   - Please restore the comment `// When unable to solve, cache the value with unsolved symbol` in the helper. It documents the non-obvious fact that unsolved lookups are cached too.

3. **Tests (minor):**
   - Please import `org.mockito.InOrder` instead of writing the fully qualified name.
   - `lookup(verify(x, never()), mode)` works but takes a moment to read. A short comment on the `lookup` helper explaining that it's reused for stubbing and verification would help.
   - `verify(first, times(1))` / `verify(second, times(1))` partly repeat the `InOrder` checks just above.
   - `lookupAppliesExceptionPolicy` overlaps with the existing `testExceptionFilter` for the ordinary path. The part that's new is the module path, so the test could focus on that.

4. **Commit message:** please use the project convention, e.g. `refactor: deduplicate CombinedTypeSolver lookup and caching logic`.

piyushjoshi01:
@jlerbsc All requested changes are addressed, including the helper rename, focused diff, test cleanup, and conventional commit message.
The full Maven test suite passes across all modules: 5,610 tests reported, 41 skipped, and no failures or errors. Formatting checks also pass.
Could you please review the updates and merge if everything looks good? Thank you!

jlerbsc:

